PR #1, opened 2026-07-19
Initialize the Project Via Dogfooding

RHDZMOTA:
Initialize the project with a simple set of development tools (python focused).

### DIFF
--- a/requirements-develop.txt
+++ b/requirements-develop.txt
@@ -1,0 +1,11 @@
+pycodestyle==2.14.0
+pytest==8.4.1
+coverage==7.10.6
+cookiecutter==2.6.0
+mypy==1.17.1
+mypy-extensions==1.1.0
+typing_extensions==4.15.0
+typed-ast==1.5.5
+scalene==1.5.54
+ipython==9.4.0
+jupyter==1.1.1


requirements-develop.txt
```diff
@@ -0,0 +1,11 @@
+pycodestyle==2.14.0
+pytest==8.4.1
+coverage==7.10.6
+cookiecutter==2.6.0
+mypy==1.17.1
+mypy-extensions==1.1.0
+typing_extensions==4.15.0
+typed-ast==1.5.5
```
alissa-app, line 8: `[major]` **`typed-ast==1.5.5` is a dead pin that breaks the dev install on Python 3.12+.**

Two independent problems:

1. **Nothing needs it.** `mypy==1.17.1`'s declared dependencies are `typing_extensions`, `mypy_extensions`, `pathspec`, and `tomli` (py<3.11) — no `typed-ast`. mypy dropped `typed-ast` when it dropped Python 2 support; the pin is vestigial, presumably carried over from an older requirements file.
2. **It doesn't install on modern Python.** `typed-ast` 1.5.5 is the final release (2023-07-04, project abandoned) and ships wheels only for `cp36`–`cp311`. On 3.12/3.13 pip falls back to building the C extension from source, which fails against the reworked CPython ast internals. So `pip install -r requirements-develop.txt` errors out for anyone not on ≤3.11.

This is latent rather than visible right now because the local env is 3.10 — classic works-on-my-machine. For a file whose entire job is bootstrapping a new contributor's environment, that's worth fixing before merge.

**Suggested fix:** delete the line. If something genuinely needs it, pin it with an environment marker (`typed-ast==1.5.5; python_version < "3.12"`) and say what needs it.

requirements-develop.txt
```diff
@@ -0,0 +1,11 @@
...
+scalene==1.5.54
```
alissa-app, line 9: `[minor]` **`scalene==1.5.54` is pinned ~2 major versions behind** (current is 2.3.0). The 1.5.x line predates support for recent CPython releases, so this risks the same install-time cliff as `typed-ast` once the project moves past 3.11 — and it's a heavy native-extension profiler to put in the baseline dev set that every contributor and CI job installs.

Suggest either bumping to `2.x` or moving profiling tooling into an optional extra installed on demand.

requirements-develop.txt
```diff
@@ -0,0 +1,11 @@
+pycodestyle==2.14.0
+pytest==8.4.1
+coverage==7.10.6
+cookiecutter==2.6.0
```
alissa-app, line 4: `[question]` **What consumes `cookiecutter` here?**

Cookiecutter scaffolds new projects from templates — it's typically a one-shot tool you run *once* to generate a repo, not a standing dev dependency of the generated repo. I don't see a `{{cookiecutter.*}}` template directory or a generation step in this repo.

If it was used to bootstrap this project, it doesn't belong in the checked-in dev requirements. If this daemon is meant to *generate* something from templates at runtime, then it's a runtime dependency and belongs in a `requirements.txt` instead. Either way the current placement looks wrong — but you have context I don't, hence a question rather than a finding.

requirements-develop.txt
```diff
@@ -0,0 +1,11 @@
...
+mypy==1.17.1
+mypy-extensions==1.1.0
+typing_extensions==4.15.0
+typed-ast==1.5.5
+scalene==1.5.54
+ipython==9.4.0
+jupyter==1.1.1
```
alissa-app, line 11: `[minor]` **`jupyter==1.1.1` is a metapackage that pulls a very large tree** — `notebook`, `jupyterlab`, `jupyter-server`, kernel machinery, and their transitive deps — into every dev install and every CI run that installs this file.

`ipython==9.4.0` on line 10 already covers interactive exploration for the ~90% case. If notebooks are genuinely wanted, suggest splitting them into an optional extra (`requirements-notebook.txt` or a `[notebook]` extra) rather than making the full Jupyter stack mandatory for anyone who just wants to run `pytest`.

requirements-develop.txt
```diff
@@ -0,0 +1,11 @@
+pycodestyle==2.14.0
```
alissa-app, line 1: `[minor]` **Nothing in the repo consumes this file.** There is no `pyproject.toml`, `setup.py`, `requirements.txt`, `Makefile`, tox/nox config, or CI workflow — and no Python source at all. The repo after this PR is `README.md` + `requirements-develop.txt`.

The file is therefore inert: a new contributor has no documented signal that `pip install -r requirements-develop.txt` is the bootstrap step, and none of the tools it installs (`pycodestyle`, `pytest`, `coverage`, `mypy`) have configuration telling them what to check.

This is reasonable as the *first* commit of a scaffold, so it's a `[minor]` rather than a `[major]` — but the PR describes itself as "initialize the project," and initialization that nothing references isn't yet initialization. Suggest pairing it with at minimum a README bootstrap line, and ideally the `pyproject.toml`/tool config the linters need.

alissa-app, line 1: `[minor]` **No `requirements.txt` counterpart.** The `-develop` suffix implies a sibling runtime requirements file, which doesn't exist. Whichever convention this project settles on (split requirements files vs. `pyproject.toml` dependency groups), the two halves should land together or the naming should not promise a file that isn't there.

Worth deciding now rather than later — this is the commit that sets the project's dependency-management convention, and everything after it will follow whatever precedent this sets.

alissa-app (commented):
## Fresh-instance review — `alissa-code-review` (CR1–CR9), round 1

**Verdict of record: `request_changes`** (2 unwaived `[major]`). The verdict envelope on the Alissa review task is authoritative; this GitHub review is submitted in comment mode so PR state is untouched.

Small diff (11 lines, 1 file), so the substance is in the dependency choices and in what the PR *doesn't* carry.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 2 |
| minor | 4 |
| nit | 0 |
| question | 1 (open: 1) |

### Dimension report

- **Correctness:** Cannot be judged against a definition-of-done — this PR references no Alissa task (see the `[major]` below), so there is no spec to judge the diff against. On its own terms the file does add a Python dev toolchain, but one pin is broken (`typed-ast`) and nothing in the repo consumes the file.
- **Coding practices:** `[major]` — workspace conventions C1–C3 are not met: branch is `dogfooding` rather than `TASK-<id>-DESC`, the PR body names no task, and commit `ef7f6ff` carries no co-author trailers.
- **Security:** Nothing found in the code. Dev-only dependencies, all pinned to exact versions (good). No secrets, no injection surface, no auth surface. Noted but not filed as a finding: no hash pinning / constraints file, and the `jupyter` metapackage widens the transitive supply-chain surface on dev machines.
- **Performance:** No runtime impact — nothing imports this. Install-time weight only (`jupyter` tree, `scalene` native build); filed as `[minor]`.
- **Side-effects / blast radius:** Nothing found. New untracked-by-anything file, no contract changes, no consumers in this repo or elsewhere in the workspace, no irreversible operations. Fully revertible.

Drop the `typed-ast` line, and land the task-linkage `[major]` either by fixing it or by disputing it with reasoning (`[triage:ignore]`). The `[minor]`s can be triaged as `later` with registered tasks.

Per CR8, every `[blocker]`/`[major]`/`[minor]`/`[question]` needs exactly one reasoned `[triage:*]` reply before re-requesting review.

alissa-app:
## Verdict of record: `request_changes` — round 1 of cap 3

Recorded as evidence on Alissa review task **`ms722h17588frsw1hnf8c3nz2d8avv1k`** (#171992531), now `pending_validation`. Findings are on [this review](https://github.com/fahera-mx/alissa-github-review-daemon/pull/1#pullrequestreview-4730198096).

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 2 (waived: 0) |
| minor | 4 |
| nit | 0 |
| question | 1 (open: 1) |

### Findings index

| # | severity | anchor | finding |
| --- | --- | --- | --- |
| 1 | major | `requirements-develop.txt:8` | `typed-ast==1.5.5` is a dead pin (mypy 1.17.1 doesn't depend on it) with no wheels past cp311 — breaks `pip install` on Python 3.12+. |
| 2 | major | PR-level | No Alissa task linkage: branch naming, PR body, and commit trailers all omit it (C1–C3). |
| 3 | minor | `requirements-develop.txt:9` | `scalene==1.5.54` pinned ~2 majors behind (2.3.0); same forward-compat cliff, heavy native dep in the baseline set. |
| 4 | minor | `requirements-develop.txt:11` | `jupyter==1.1.1` metapackage pulls a large tree into every dev/CI install; `ipython` already covers the common case. |
| 5 | minor | `requirements-develop.txt:1` | Nothing consumes the file — no pyproject/setup.py/Makefile/CI/tool config and no Python source. |
| 6 | minor | `requirements-develop.txt:1` | No `requirements.txt` counterpart, though the `-develop` suffix promises one. |
| 7 | question | `requirements-develop.txt:4` | What consumes `cookiecutter==2.6.0`? No template dir or generation step in the repo. |

### Summary

PR #1 adds a single 11-line `requirements-develop.txt` pinning a Python dev toolchain, as the first real content in an otherwise README-only repo. The verdict is `request_changes` on two unwaived majors. The substantive one is `typed-ast==1.5.5`: mypy 1.17.1 does not depend on it, and the package's final release ships no wheels beyond cp311, so the file's one job — bootstrapping a contributor environment — fails on Python 3.12+ while passing on the 3.10 box it was written on. The second is process: the PR carries no Alissa task linkage in its branch name, body, or commit trailers, which is what left this review with no definition-of-done to judge correctness against.

**Shortest path back to approve:** drop the `typed-ast` line, and either fix the task linkage or dispute it with reasoning via `[triage:ignore]`. The four minors and the open question can be triaged `later` against registered tasks.

### Reviewer caveat (protocol deviation, disclosed)

This review deviates from CR2/CR3 in one respect, with operator confirmation: **no origin task exists**. The branch name, PR body/title, commit trailers, linked issues, labels, milestones, and the actor's full Alissa task list were all searched — no task reference anywhere; the operator confirmed the PR was produced outside the task system as a dogfooding exercise. The review task is therefore **standalone rather than downstream of an origin task**, and correctness was assessed against the PR's own stated intent instead of a definition-of-done. That substitution is weaker than the rubric intends, and is why finding #2 is filed as a `[major]` rather than a nit.

Submitted in comment mode rather than `request-changes` to honor reviewer posture — the task envelope is the verdict of record. No commits pushed, nothing merged, no PR state changed, no findings fixed by the reviewer.

Per CR8, every `[major]`/`[minor]`/`[question]` above needs exactly one reasoned `[triage:*]` reply on its thread before re-requesting review.

PR #3, opened 2026-07-19
TASK-1062238981: build task refs from taskNumber, not taskSeq

alissa-app:
Fixes the 404 that broke **every re-requested review round**:

```
ERROR fahera-mx/studio.alissa.app#210: alissa exited 1: ✗ API 404 (Not Found): Task not found or not accessible.
```

Task: **TASK-1062238981**

## Root cause

`Alissa.list_tasks()` minted the task ref from the `taskSeq` field of `alissa task list --json`. The API resolves `TASK-<n>` against **`taskNumber`**. For the PR #210 review task the two differ:

```json
"taskSeq": 998,  "taskNumber": 617115756
```

Verified live:

```
$ alissa task get TASK-998         → ✗ API 404: Task not found or not accessible.
$ alissa task get TASK-617115756   → resolves
```

## Why round 1 passed

The daemon never creates the review task — `_spawn` looks it up by title each round.

- **Round 1:** no review task exists yet (the reviewer creates it per CR2), so `task is None`, `--task` is omitted, and the bad ref is never exercised.
- **Round 2:** the task now exists and matches, so `alissa tmux queue add … --task TASK-998` → 404 → `CommandError`.

The bug was latent from day one and is reachable on every re-request. The bad ref also leaked into the directive prose (`loop.py:164`), so the reviewer agent's own `alissa task get` would 404 identically — fixed by the same change.

## Changes

- `alissa.py` — build the ref from `taskNumber`; skip rows without one. `_spawn` already handles a missing task gracefully, and no ref beats a wrong ref.
- `test_loop.py` — regression tests for both paths. The existing suite fakes the Alissa client, so `list_tasks()` ref construction was never covered — that is why this shipped.
- `version` — 0.1.0 → 0.1.1 (published package; bugfix, no API change).

## Verification

- 50/50 unit tests pass; `pycodestyle` clean; `mypy` clean (9 files).
- Live: `find_review_task('fahera-mx','studio.alissa.app',210)` → `TASK-617115756`, which `alissa task get` resolves.

## Out of scope — flagged, not fixed

1. `state.py` writes `task_ref` into the `spawns` table but nothing ever reads it — dead column.
2. `alissa task list` returns only tasks owned by the calling actor, and the daemon's identity is independent of the spawned reviewer's. A task the daemon can see may still be "not accessible" to the reviewer session — a second, distinct 404 with the same message. Worth watching for after this lands.


RHDZMOTA (approved):
LGTM

PR #4, opened 2026-07-19
TASK-2074815680: count rounds, not review records; make convergence reachable

alissa-app:
Fixes **TASK-2074815680** — the review loop counted rounds wrong and could never recognise success.

## Bug 1 — round counting counted review RECORDS, not rounds

`ghclient.GitHub.my_reviews` returned every review record authored by the reviewer. A standalone inline comment creates its own review record with an **empty body**, so records outnumber rounds. Real data from `fahera-mx/studio.alissa.app#210`:

| time | state | body_len | what it actually is |
|---|---|---|---|
| 18:31:59 | COMMENTED | 4399 | the actual round-1 review |
| 18:32:30 | COMMENTED | 0 | inline comment artifact |
| 18:32:30 | COMMENTED | 0 | inline comment artifact |
| 18:32:31 | COMMENTED | 0 | inline comment artifact |
| 20:17:14 | COMMENTED | 8030 | round 2 |
| 20:20:22 | COMMENTED | 4826 | round 3 |

Three real rounds produced six records. The daemon told round 3's reviewer it was **"round 6 of cap 10"**, and would have capped out and escalated a PR that had barely been reviewed.

**Fix:** `body` is added to the `Review` dataclass, and only *substantive* reviews — those with a non-empty body — are counted. The round-closing review always carries the verdict write-up; the artifacts never do.

### Why not group by `commit_id`

It looks like the natural grouping key and it **undercounts**. A round reviews whatever head is current, and consecutive rounds routinely land on the same commit when the implementer triages findings without pushing. On #210 rounds 2 and 3 **both carry head `805398a`** and would collapse into a single round. Body presence tracks "a reviewer wrote a verdict"; commit identity does not. The reasoning is recorded in a comment on `my_reviews`, and pinned by `test_rounds_are_not_grouped_by_commit_id`, so it does not get re-introduced.

## Bug 2 — convergence was unreachable

The loop converged only on `my_reviews[-1].state == "APPROVED"`. Reviewers post comment-mode reviews, so `COMMENTED` is the only state GitHub ever carries — #210 has **zero** `APPROVED` records across its entire history. Every PR, however clean, ran to the round cap and escalated. The escalation path was the only exit from the loop.

**Fix:** converge when *either* the last substantive GitHub review state is `APPROVED`, *or* the review task's newest CR6 verdict envelope reads `approve`.

The task envelope is authoritative because the review skill declares it the verdict of record, and because comment-mode reviews cannot express approval in the GitHub state at all — the envelope is the only place an approval can be written down.

`Alissa.latest_verdict` reads `alissa task get <ref> --json`, picks the newest `evidence` item by `createdAt`, and parses `approve` vs `request_changes` from the title or body (em-dash separator; en-dash and hyphen also accepted). It reuses the existing `run_json` helper so subprocess handling and error wrapping stay consistent.

**It never raises.** The daemon polls forever and this runs on every pass, so absent, empty or malformed evidence degrades to "no verdict" rather than taking the loop down — covered by an 11-case parametrised test.

The task lookup moved from `_spawn` up into `evaluate` so the convergence check has a ref to use. `_spawn` still handles `task is None` exactly as before, and with no task the envelope check is skipped, so behaviour is unchanged when there is no review task.

## Version

`0.1.1` → `0.2.0` — behaviour change.

## Verification

```
$ python -m pytest src/test -q
........................................................................ [ 96%]
...                                                                      [100%]
75 passed in 11.66s

$ pycodestyle src --max-line-length 120
exit=0

$ mypy src/main --ignore-missing-imports --install-types --non-interactive
Success: no issues found in 9 source files
exit=0
```

The new tests were confirmed to fail against the unfixed source (32 failures) before the fix was applied, so they genuinely pin both bugs rather than passing vacuously.


RHDZMOTA (approved):
LGTM

PR #5, opened 2026-07-20
TASK-817578483: containerize the review daemon (Docker + entrypoint)

RHDZMOTA:
Registers and delivers **TASK-817578483** — run the GitHub review loop unattended in a container instead of on a dev host.

## What this is
The daemon (`alissa-reviewloop`) is a thin poller; the real reviewing is done by `alissa worker` draining a tmux queue and spawning `claude` reviewer agents inside worktree hubs. So this image bundles **all three tiers**, not just the PyPI package:

- python 3.12 + `alissa-tools-github-reviewloop` (the poller, from PyPI)
- node 22 + the alissa CLI (worker, tmux queue, tasks)
- claude-code (the agent the worker spawns)

## Files (`docker/claude/`)
- **`Dockerfile`** — `python:3.12-slim-bookworm` (matches the repo's pinned 3.12) + Node 22 via NodeSource. alissa CLI via the official npm-free installer (checksum-verified); daemon from PyPI (`REVIEWLOOP_VERSION` build arg, default 0.2.0). Runs non-root (uid 1000); `tini` as PID 1.
- **`entrypoint.sh`** — preflight the three identities → bootstrap manifest + `reviewloop.config.json` (`on_missing_hub: add`) → start `alissa worker --daemon`, wait until up → run the daemon foreground with graceful shutdown.
- **`init-firewall.sh`** — optional default-deny egress allowlist (gated by `ALISSA_ENABLE_FIREWALL=1`, needs `--cap-add=NET_ADMIN`).
- **`README.md`** — build/run, the three identities, workspace bootstrap, compose example.

## The three identities (injected at runtime, never baked)
| env | identity | for |
| --- | --- | --- |
| `GH_TOKEN` | gh (`alissa-app`) | review queue, round counting, PR comments |
| `ALISSA_API_TOKEN` | Alissa by Fahera | tasks, session queue, verdicts |
| `ANTHROPIC_API_KEY` / `CLAUDE_CODE_OAUTH_TOKEN` | claude | the reviewer agent |

## Workspace: bootstrap-from-manifest
Self-contained — with `on_missing_hub: add` the daemon hub-ifies each repo itself on first request, so nothing is pre-cloned. The entrypoint only guarantees a manifest + config exist (mounted, or generated from `ALISSA_REVIEW_REPOS`).

## Verification
**Done here:** image builds; python 3.12.13 / node 22 / gh 2.96 / claude-code 2.1.215 / alissa CLI / `alissa-reviewloop` all present and runnable; entrypoint preflight fails fast on each missing identity (claude → gh presence → gh validity) with exit 1.

**Left to verify live** (needs real tokens + a PR authored by another account — the daemon README's own live-verify caveat): the worker actually spawning a reviewer and a full round. Run `docker run ... --once --dry-run -v` against the first real request before letting it run unattended.

### DIFF
--- a/docker/claude/Dockerfile
+++ b/docker/claude/Dockerfile
@@ -1,0 +1,103 @@
+# syntax=docker/dockerfile:1
+# =============================================================================
+# Alissa — GitHub Review Daemon (claude reviewer runtime)
+#
+# This image is NOT just the poller. The daemon (`alissa-reviewloop`) is thin: it
+# watches GitHub and enqueues reviewer sessions via `alissa tmux queue add`. The
+# real review work is done by `alissa worker`, which drains that queue and spawns
+# `claude` reviewer agents inside worktree-hub directories. So the image bundles
+# all three tiers:
+#
+#   * python 3.12 + alissa-tools-github-reviewloop  (the poller, from PyPI)
+#   * node + the alissa CLI                          (worker, tmux queue, tasks)
+#   * claude-code                                    (the agent the worker spawns)
+#
+# Build context is this directory (docker/claude) — the image pulls the daemon
+# from PyPI, so no repo source is copied in.
+#
+#   docker build -t alissa-review-daemon docker/claude
+#
+# Run: see docker/claude/README.md and ../../README.md.
+# =============================================================================
+
+# Base is pinned to python 3.12 (matches the repo's .python-version); Node 22 is
+# layered on via NodeSource. The daemon needs python >= 3.11; the alissa CLI and
+# claude-code need Node >= 18 — both satisfied.
+FROM python:3.12-slim-bookworm
+
+# Version of the daemon to install from PyPI. Bump per release.
+ARG REVIEWLOOP_VERSION=0.2.0
+# Node major version (must be >= 18 for the alissa CLI and claude-code).
+ARG NODE_MAJOR=22
+
+ENV DEBIAN_FRONTEND=noninteractive \
+    PYTHONUNBUFFERED=1 \
+    PIP_NO_CACHE_DIR=1 \
+    PIP_DISABLE_PIP_VERSION_CHECK=1
+
+# --- System dependencies ------------------------------------------------------
+# git   : reviewers operate over git worktree hubs
+# tmux  : the alissa worker manages tmux (ali-*) sessions
+# gh     : the daemon shells out to `gh api` for the review queue
+# tini   : PID 1 init — reaps the tmux/node/claude child fan-out
+# sudo   : only used by the optional firewall init (dropped priv otherwise)
+# iptables/ipset : only used by the optional firewall init
+RUN set -eux; \
+    apt-get update; \
+    apt-get install -y --no-install-recommends \
+        ca-certificates curl gnupg git tmux tini sudo \
+        iptables ipset procps jq less; \
+    # GitHub CLI apt repo
+    mkdir -p /etc/apt/keyrings; \
+    curl -fsSL https://cli.github.com/packages/githubcli-archive-keyring.gpg \
+        -o /etc/apt/keyrings/githubcli-archive-keyring.gpg; \
+    chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg; \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+        > /etc/apt/sources.list.d/github-cli.list; \
+    # Node.js via NodeSource
+    curl -fsSL "https://deb.nodesource.com/setup_${NODE_MAJOR}.x" | bash -; \
+    apt-get install -y --no-install-recommends gh nodejs; \
+    rm -rf /var/lib/apt/lists/*
+
+# --- claude-code (the reviewer agent) ----------------------------------------
+RUN npm install -g @anthropic-ai/claude-code \
+    && npm cache clean --force
+
+# --- The review daemon (from PyPI) -------------------------------------------
+# Installs the `alissa-reviewloop` console script into /usr/local/bin.
+RUN pip install "alissa-tools-github-reviewloop==${REVIEWLOOP_VERSION}"
+
+# --- Non-root runtime user ----------------------------------------------------
+# Agents run unattended with three live tokens; don't run them as root. The
+# firewall init (optional) is the only thing that needs root, via a narrow
+# sudoers entry.
+RUN useradd --create-home --shell /bin/bash --uid 1000 alissa \
+    && echo 'alissa ALL=(root) NOPASSWD: /usr/local/bin/init-firewall.sh' \
+        > /etc/sudoers.d/alissa-firewall \
+    && chmod 0440 /etc/sudoers.d/alissa-firewall \
+    && mkdir -p /workspace \
+    && chown alissa:alissa /workspace
+
+# --- Entrypoint + optional firewall ------------------------------------------
+COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+COPY init-firewall.sh /usr/local/bin/init-firewall.sh
+RUN chmod 0755 /usr/local/bin/entrypoint.sh /usr/local/bin/init-firewall.sh
+
+USER alissa
+WORKDIR /home/alissa
+
+# The alissa CLI installer is npm-free: it drops a `node cli.mjs` launcher into
+# ~/.local/bin. Run it as the target user so it lands in the user's home.
+ENV PATH="/home/alissa/.local/bin:${PATH}"
+RUN curl -fsSL https://share.alissa.app/install | bash
+
+# Workspace root that the daemon and reviewers operate over. Mount a manifest
+# here, or let the entrypoint generate one from env (see README).
+ENV ALISSA_WORKSPACE_ROOT=/workspace \
+    ALISSA_WORKSPACE=alissa-review \
+    TMUX_TMPDIR=/home/alissa/.tmux
+RUN mkdir -p /home/alissa/.tmux
+WORKDIR /workspace
+
+# tini reaps the worker/tmux/claude child fan-out; entrypoint does the rest.
+ENTRYPOINT ["/usr/bin/tini", "--", "/usr/local/bin/entrypoint.sh"]
--- a/docker/claude/Dockerfile
+++ b/docker/claude/Dockerfile
@@ -138,8 +138,14 @@ RUN git config --global user.name  "alissa-review-daemon" \
 # Workspace root that the daemon and reviewers operate over. Mount a manifest
 # here, or let the entrypoint generate one from the config below (see README).
 # These are fixed internal paths, not meant to be reconfigured.
+#
+# CLAUDE_CONFIG_DIR points claude's config — crucially its OAuth credential file
+# `.credentials.json` — at a folder ON the /workspace volume, so a one-time
+# `claude /login` survives restarts and auto-renews (a static CLAUDE_CODE_OAUTH_
+# TOKEN expires and 401s). If you relocate the volume, move this with it.
 ENV ALISSA_WORKSPACE_ROOT=/workspace \
-    TMUX_TMPDIR=/home/alissa/.tmux
+    TMUX_TMPDIR=/home/alissa/.tmux \
+    CLAUDE_CONFIG_DIR=/workspace/.claude-config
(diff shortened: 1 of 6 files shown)


docker/claude/entrypoint.sh
```diff
@@ -0,0 +1,311 @@
...
+mkdir -p "${TMUX_TMPDIR:-/home/alissa/.tmux}"
+
+log "starting alissa worker (detached)"
+alissa worker start --daemon --interval "${ALISSA_WORKER_INTERVAL:-2}" \
+  || die "alissa worker failed to start"
+
+# Poll status until it is up (the daemon only warns if the worker is absent).
+worker_up=0
+for _ in $(seq 1 15); do
+  if alissa worker status 2>/dev/null | grep -qiv 'not running\|no worker' \
```
alissa-app, line 285: [minor] The worker-readiness gate uses a per-line `grep -qiv 'not running\|no worker'` AND `grep -qi 'running'`. Two subtleties make it weaker than intended: (1) `grep -v -q` succeeds if *any* line fails to match, so on multi-line status output a header line lets it pass even when another line says "not running"; (2) "not running" contains the substring "running", so the second grep matches it too. It happens to work today because `alissa worker status` prints a single `running as '<host>' (pid ...)` line, but it's fragile. The daemon's own `worker_running()` already does the robust whole-output check (`"not running" not in out.lower() and "no worker" not in out.lower()`); consider mirroring that, or just gate on `grep -qi 'running as'`.

docker/claude/entrypoint.sh
```diff
@@ -0,0 +1,311 @@
...
+# firewall (which needs root anyway), then re-exec this script as `alissa`.
+#
+# claude refuses --dangerously-skip-permissions as root, so everything past this
+# point MUST run unprivileged — that is exactly what the drop guarantees.
+# -----------------------------------------------------------------------------
+if [ "$(id -u)" = "0" ]; then
+  mkdir -p "${WORKSPACE_ROOT}" "${TMUX_TMPDIR:-/home/${RUNTIME_USER}/.tmux}"
+  # Fix ownership so the unprivileged user can write. -R because a restart may
+  # find files a previous root-mounted run left behind.
+  chown -R "${RUNTIME_USER}:${RUNTIME_USER}" \
```
alissa-app, line 39: [minor] `chown -R "${WORKSPACE_ROOT}"` runs on every boot and is O(number-of-files) on a persisted volume that accumulates cloned worktree hubs — a restart could spend real time here as the workspace grows. Consider chowning only the mount root (non-recursive) plus newly-created paths, or gating the recursive pass on an ownership check (e.g. skip if the root is already `alissa`-owned).

docker/claude/entrypoint.sh
```diff
@@ -0,0 +1,311 @@
...
+    /usr/local/bin/init-firewall.sh \
+      || die "firewall init failed — did you pass --cap-add=NET_ADMIN?"
+  fi
+
+  log "dropping to ${RUNTIME_USER}"
+  exec gosu "${RUNTIME_USER}" "$0" "$@"
+fi
+
+# -----------------------------------------------------------------------------
+# 1. Preflight the three identities
```
alissa-app, line 54: [nit] This block is headed `1. Preflight the three identities`, but its subsections below are labeled `2a`, `2b`, `2c`. Renumber to `1a/1b/1c` (or retitle the header) so the internal numbering lines up with the step-0..5 scheme in the file header.

docker/claude/Dockerfile
```diff
@@ -0,0 +1,208 @@
...
+    chmod go+r /etc/apt/keyrings/githubcli-archive-keyring.gpg; \
+    echo "deb [arch=$(dpkg --print-architecture) signed-by=/etc/apt/keyrings/githubcli-archive-keyring.gpg] https://cli.github.com/packages stable main" \
+        > /etc/apt/sources.list.d/github-cli.list; \
+    # Node.js via NodeSource
+    curl -fsSL "https://deb.nodesource.com/setup_${NODE_MAJOR}.x" | bash -; \
+    apt-get install -y --no-install-recommends gh nodejs; \
+    rm -rf /var/lib/apt/lists/*
+
+# --- claude-code (the reviewer agent) ----------------------------------------
+RUN npm install -g @anthropic-ai/claude-code \
```
alissa-app, line 63: [minor] `npm install -g @anthropic-ai/claude-code` is unpinned, so builds float to whatever claude-code is latest. The whole headless story here depends on pre-seeding claude's first-run keys (`hasCompletedOnboarding`, `skipDangerousModePermissionPrompt`, `hasTrustDialogAccepted`, …), which are coupled to claude-code's onboarding schema — an unpinned upgrade could silently reintroduce a blocking first-run gate and hang every reviewer. Recommend pinning to a known-good version via a build ARG (same pattern as `REVIEWLOOP_VERSION`), and bumping it deliberately.

docker/claude/Dockerfile
```diff
@@ -0,0 +1,208 @@
...
+ARG ALISSA_REVIEW_SKILLS="alissa-code-workspace|alissa-code-review"
+# Daemon knobs (see the daemon README's settings table).
+ARG ALISSA_POLL_INTERVAL="60"
+ARG ALISSA_ROUND_CAP="3"
+ARG ALISSA_AGENT_PROFILE="claude"
+ARG ALISSA_ON_MISSING_HUB="add"
+# Worker reconcile tick, seconds.
+ARG ALISSA_WORKER_INTERVAL="2"
+# Optional egress firewall (needs --cap-add=NET_ADMIN at runtime).
+ARG ALISSA_ENABLE_FIREWALL="0"
```
alissa-app, line 188: [minor] The egress firewall defaults to off (`ALISSA_ENABLE_FIREWALL="0"`), and the documented unattended run example doesn't enable it — so the primary use case (an autonomous `claude` holding three live tokens, reacting to inbound third-party PRs) ships with open egress by default. `init-firewall.sh`'s own comment names exactly this exfiltration threat. I'm keeping this `[minor]` rather than `[major]` because a hard default-on isn't safe either: the entrypoint `die`s if the firewall can't initialize, and it needs `--cap-add=NET_ADMIN`, which many platforms (incl. Railway) don't grant — so flipping the default would break boot in the common case. The actionable ask is documentation: prominently steer unattended deployments to enable the firewall (with the NET_ADMIN caveat) in the main Run section, not only in the Optional subsection.

docker/claude/agents.yaml
```diff
@@ -0,0 +1,20 @@
...
+#
+# `--dangerously-skip-permissions --permission-mode acceptEdits` is what makes
+# claude run unattended. It requires a non-root user (claude refuses that flag as
+# root) — the image runs as uid 1000, which satisfies it. Reviewer posture (CR6:
+# reviewers never write) is enforced by the round directive, not by this profile.
+#
+# To pin a model or tweak flags, override the command below (edit this file
+# before build, or mount your own agents.yaml over it at runtime).
+claude:
+  command: claude --dangerously-skip-permissions --permission-mode acceptEdits
```
alissa-app, line 15: [minor] Reviewer blast radius: `--dangerously-skip-permissions --permission-mode acceptEdits` means the only thing keeping a reviewer read-only (CR6) is the prompt directive — a reviewed PR that succeeds at prompt-injection could make the agent write, run arbitrary bash, or read the three mounted tokens. This mirrors the host worker's existing practice, and containerizing actually *improves* isolation (non-root + optional firewall), so it's not a regression — but given this image is purpose-built to run unattended against untrusted inbound PRs, it's worth (a) pairing it with the egress firewall by default-recommendation (see the Dockerfile comment) and (b) a one-line note here that `acceptEdits` is intentionally relied-upon-to-be-harmless only because reviewers never write.

docker/claude/README.md
```diff
@@ -0,0 +1,279 @@
...
+`--once`, `--dry-run`, `-v` all work.
+
+Unattended, persisting the workspace (the cloned hubs) across restarts:
+
+```sh
+docker run -d --name alissa-review \
+  --restart unless-stopped \
+  -e GH_TOKEN \
+  -e ALISSA_API_TOKEN \
+  -e ANTHROPIC_API_KEY \
```
alissa-app, line 188: [minor] This (and the other run/compose examples) inject `-e ANTHROPIC_API_KEY`, but the claude-auth section above explicitly recommends a persisted `claude /login` or `CLAUDE_CODE_OAUTH_TOKEN` *over* a bare API key — warning that a bare `ANTHROPIC_API_KEY` triggers claude's "approve this key?" TUI prompt. That prompt isn't among the pre-seeded first-run keys, so following these examples verbatim risks the exact headless hang the doc warns about. Recommend the examples use `CLAUDE_CODE_OAUTH_TOKEN` (or reference the persisted-login path) to match the stated guidance.

docker/claude/init-firewall.sh
```diff
@@ -0,0 +1,77 @@
...
+  pypi.org
+  files.pythonhosted.org
+  deb.nodesource.com
+  cli.github.com
+  ${ALISSA_FIREWALL_EXTRA:-}
+)
+
+for domain in "${DOMAINS[@]}"; do
+  [ -n "${domain}" ] || continue
+  ips="$(getent ahostsv4 "${domain}" | awk '{print $1}' | sort -u || true)"
```
alissa-app, line 55: [minor] The allowlist is resolved to IPs once at container start and pinned into `ipset`. GitHub/Anthropic/registry endpoints sit behind CDNs whose IPs rotate, so a long-running unattended container can begin dropping legitimate egress hours/days later (and a newly-rotated Anthropic IP would silently block the reviewer). The upstream claude devcontainer firewall has the same limitation. Worth either periodically re-resolving the allowlist or documenting that enabling the firewall implies restarting on egress failures.

alissa-app (approved):
## Fresh-instance review — PR #5 (containerize the review daemon)

Reviewer: `alissa-app` (fresh session, no implementer context). Reviewed the full diff against the PR's stated intent (origin TASK-817578483 is not accessible to the reviewing actor, so correctness is judged against the described definition-of-done and the daemon/CLI it wires up, which I verified directly). CI is green (style/types/unit/wheel + Railway build all pass).

**Verdict: approve** — no blockers or majors. Several `[minor]` hardening items and two `[nit]`s, all suitable as follow-ups rather than merge gates. I verified the load-bearing wiring end-to-end (see below), so the container should stand up and function as described; the one thing neither I nor the author can attest is a full live round (real tokens + a third-party PR), which the README already flags.

### Dimension report (CR4)
- **Correctness:** Nothing blocking. I confirmed against the installed package that every piece of wiring is real: `alissa worker start --daemon --interval`, `alissa worker status/stop`, and `alissa-reviewloop --workspace-root/--once/--dry-run/-v` all exist; the generated `reviewloop.config.json` keys (`repos/poll_interval/round_cap/agent_profile/on_missing_hub`) are all in the daemon's `CONFIG_KEYS` and pass its validation; the generated `alissa-workspace.yaml` schema matches a real manifest on disk; and the pinned PyPI `0.2.0` is the version that carries the recent round-counting/task-ref fixes (the version file was bumped in that commit). Two `[minor]` robustness/doc items inline.
- **Coding practices:** Idiomatic, heavily and accurately commented bash + Dockerfile; `set -euo pipefail`; non-fatal seeding guarded with `|| true`. One `[nit]` on an internal comment-numbering mismatch.
- **Security:** Good posture — no secrets baked (documented `docker history` rationale), runtime non-root (uid 1000), gh-as-git-credential-helper instead of an SSH key, optional default-deny egress firewall. `[minor]`s: the firewall is opt-in and pins resolved IPs at boot; the primary (unattended, three-live-token, inbound-PR) use case runs `--dangerously-skip-permissions` with egress open by default. Inline.
- **Performance:** `chown -R /workspace` on every boot is O(files) on a large persisted volume — `[minor]` inline. Firewall resolves once (covered under Security). Otherwise fine.
- **Side-effects / blast radius:** Contained — all changes are new files under `docker/claude/` plus one README section; no package/daemon source is touched, so zero blast radius on existing consumers. I verified the boot-time `DELETE FROM spawns` is safe: that table is in-flight-dedup only (round counting is derived from GitHub, `escalations` are preserved), so the reset cannot corrupt round counting or re-escalate a capped PR. Nothing found beyond that.

### Minors worth turning into follow-ups
See the 6 inline `[minor]`s and 2 `[nit]`s. None blocks merge. One extra `[nit]` not inlined: consider adding a Docker `HEALTHCHECK` so orchestrators can detect a wedged daemon/worker.

The verdict of record is the envelope on the Alissa review task (created standalone because the origin task is not accessible to this actor; it should be re-parented to TASK-817578483).

PR #15, opened 2026-07-22
entrypoint: sync worktree hubs at boot (fixes the empty-hub hub-ify loop)

RHDZMOTA:
Fixes the daemon freezing on a newly-allowlisted repo (seen live on `alissa-github-develop-daemon#2`).

## Root cause (confirmed live)
The manifest lists every allowlisted repo, but a hub dir is just a placeholder folder until something clones into it. The daemon's `on_missing_hub: add` uses `alissa code workspace add`, which is **idempotent by manifest entry** — for a repo already listed it **no-ops**, leaving an empty folder with no `main/` worktree. So `hub.is_dir()` stays False and the daemon re-hub-ifies **every poll, forever** — and the failure is a `SKIPPED` decision logged at **DEBUG**, so it's invisible at INFO level.

Operator confirmed: the folder existed empty; `alissa code workspace sync` created `main/` and the review ran. (Private vs public was a red herring — `studio.alissa.app` is private too and works because its hub already existed on the volume.)

## Fix
The entrypoint now runs **`alissa code workspace sync`** after the manifest is generated + gh auth is wired, before the daemon starts. `sync` is the *reconcile* op — it creates missing/half-built hubs to match the manifest — so every hub is real before the first poll. Non-fatal on partial failure (warns; the daemon still retries per-repo). This automates the manual `sync` an operator would otherwise run for each new repo.

## Note
Entrypoint-only (no package change / version bump). A deeper daemon fix — make `on_missing_hub` self-heal by falling back to `sync` when `add` leaves an incomplete hub — is a good follow-up for the non-container / mid-run case, but the boot-sync covers the container: every (re)start (incl. every Railway redeploy after an env change) materializes all hubs.

⚠️ Branch lacks a TASK id — the Alissa MCP server was disconnected when this was authored; I'll attach a task once it reconnects.

### DIFF
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -271,6 +271,24 @@ PY
 fi
 
 # -----------------------------------------------------------------------------
+# 3c. Materialize the worktree hubs the manifest declares.
+#
+# The manifest lists every allowlisted repo, but the hub directories
+# (bare clone + main/ worktree) don't exist until something creates them. The
+# daemon's on_missing_hub:add uses `alissa code workspace add`, which is
+# idempotent BY MANIFEST ENTRY -- for a repo already listed it no-ops, leaving an
+# empty folder and no main/, and the daemon then loops forever hub-ifying a hub
+# that never completes. `workspace sync` is the reconcile operation: it creates
+# the missing/half-built hubs (and fetches existing ones) to match the manifest.
+# Run it here (auth is wired, the manifest exists) so every hub is real before
+# the daemon polls -- exactly the manual `sync` an operator would otherwise run.
+if [ -f "${MANIFEST}" ]; then
+  log "syncing worktree hubs to the manifest (alissa code workspace sync)"
+  ( cd "${WORKSPACE_ROOT}" && alissa code workspace sync ) \
+    || log "WARN: workspace sync did not fully complete — the daemon will retry per-repo, but check for clone/auth errors above"
+fi
+
+# -----------------------------------------------------------------------------
 # 4. Start the worker, wait until it reports running.
 # -----------------------------------------------------------------------------
 mkdir -p "${TMUX_TMPDIR:-/home/alissa/.tmux}"
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -272,8 +272,13 @@ volumes:
    `gh auth setup-git`; `alissa auth login` (fatal if missing); check the claude
    credential (warn-only — the baked `agents.yaml` handles headless launch).
 2. Ensure a manifest + `reviewloop.config.json` exist (mount or generate).
-3. Start `alissa worker --daemon`, wait until it reports running (the daemon only
+3. **`alissa code workspace sync`** — materialize the worktree hubs the manifest
+   declares (create missing/half-built ones, fetch existing). Without this the
+   daemon's on-demand `alissa code workspace add` no-ops on a repo already listed
+   in the manifest, leaving an empty folder and looping forever hub-ifying a hub
+   that never completes.
+4. Start `alissa worker --daemon`, wait until it reports running (the daemon only
    *warns* if the worker is absent, so ordering matters).
-4. Run `alissa-reviewloop` in the foreground; stop the worker on `SIGTERM`/`SIGINT`.
+5. Run `alissa-reviewloop` in the foreground; stop the worker on `SIGTERM`/`SIGINT`.
 
 `tini` is PID 1 to reap the tmux/node/claude child fan-out.


RHDZMOTA:
Task of record: **TASK-997515082** (registered post-hoc — this branch predates its task; C2 branch-name waiver recorded there). Entering the adversarial review loop now.

docker/claude/entrypoint.sh
```diff
@@ -270,6 +270,24 @@ finally:
...
+# daemon's on_missing_hub:add uses `alissa code workspace add`, which is
+# idempotent BY MANIFEST ENTRY -- for a repo already listed it no-ops, leaving an
+# empty folder and no main/, and the daemon then loops forever hub-ifying a hub
+# that never completes. `workspace sync` is the reconcile operation: it creates
+# the missing/half-built hubs (and fetches existing ones) to match the manifest.
+# Run it here (auth is wired, the manifest exists) so every hub is real before
+# the daemon polls -- exactly the manual `sync` an operator would otherwise run.
+if [ -f "${MANIFEST}" ]; then
+  log "syncing worktree hubs to the manifest (alissa code workspace sync)"
+  ( cd "${WORKSPACE_ROOT}" && alissa code workspace sync ) \
```
alissa-app, line 287: [minor] No timeout bound on this boot-blocking call. The daemon's own clone path bounds cloning at `timeout=600` (`alissa.py` `add_repo_to_workspace` → `run(..., timeout=600)`), but this `sync` has none. Since it runs synchronously *before* the worker and daemon start, a stalled `sync` (wedged git, network black-hole) hangs container boot indefinitely — nothing reviews, and there's no per-repo recovery because the poll loop is never reached. Probability is low (the `gh` credential helper is non-interactive, so git won't prompt), but the mitigation is cheap: wrap in `timeout <N>` matching the daemon's 600s bound. Fix now or as a follow-up task.

docker/claude/entrypoint.sh
```diff
@@ -270,6 +270,24 @@ finally:
...
+    || log "WARN: workspace sync did not fully complete — the daemon will retry per-repo, but check for clone/auth errors above"
```
alissa-app, line 288: [minor] This reassurance is partly inaccurate. "the daemon will retry per-repo" points at the `on_missing_hub:add` path — which is exactly the failure mode this PR fixes: for a repo already in the manifest, `alissa code workspace add` no-ops and leaves an empty folder, so `_ensure_hub` keeps returning a `SKIPPED` decision and the daemon loops forever (loop.py). So if `sync` fails to materialize a given hub, the daemon does **not** recover that repo — it re-enters the infinite hub-ify loop; real recovery is the next redeploy re-running `sync`, or an operator. The "check for clone/auth errors above" half is correct. Suggest softening the middle clause so an operator reading this during an incident isn't told a broken path will self-heal.

alissa-app (approved):
## Review — approve (round 1)

Fresh-instance review per the alissa-code-review skill (CR1–CR9). Origin task **TASK-997515082** (registered post-hoc; not accessible to this reviewer actor). Review task **TASK-307731359**.

The fix is well-targeted and its root-cause analysis holds up against the code: `_ensure_hub` (loop.py) returns a *problem* — a `SKIPPED` decision — when `alissa code workspace add` "succeeds" but no `main/` materializes, and because `add` is idempotent-by-manifest-entry it never recovers, re-hub-ifying every poll. Running `alissa code workspace sync` at boot to reconcile the hubs is the right lever, and the block is correctly **non-fatal** under `set -euo pipefail` (`( … ) || log …`) and correctly placed before the worker/daemon start. cwd anchoring (`cd "${WORKSPACE_ROOT}"`) matches the daemon's own `add_repo_to_workspace(cwd=workspace_root)`. README step renumbering (3→5) is complete and consistent.

Approving with two **[minor]** notes (inline) and one waived **[nit]** — none gate the merge.

### Dimensions
- **Correctness:** Sound. Root cause matches loop.py `_ensure_hub`; non-fatal + pre-worker placement + MANIFEST guard are all correct. The `sync` subcommand's materialize semantics are operator-verified-live (per PR description), not covered by an automated test — acceptable for an entrypoint path (the repo has no entrypoint tests).
- **Coding practices:** Matches file idiom (comment-block header, `log`/`die`, subshell-with-cd). Commit carries co-author trailers (C1). Branch lacks a numeric TASK id (C2) — waived on the origin task per the PR comment → **[nit]**, no action.
- **Security:** Nothing found. No input interpolation; `sync` runs with no user-controlled args; git auth is the already-wired non-interactive `gh` helper; no secrets in the diff.
- **Performance:** Boot now syncs all allowlisted hubs synchronously before serving — latency scales with repo count. Acceptable, deliberate tradeoff; relates to the timeout [minor].
- **Side-effects / blast radius:** Low. Entrypoint-only, idempotent, bounded by the allowlist; reviewers use the read-only `main/` mirror so re-fetch is safe. No contract/API changes, no cross-repo consumers. Rollback = revert the block.

### Shortest path: nothing required to merge. The two minors are good follow-ups; open a task if either is deferred.

PR #19, opened 2026-07-22
Reap backstop unreachable after convergence: search-independent session sweep + interrupt-safe sleep

### LINKED ISSUE #16: Reap backstop unreachable after convergence: search-independent session sweep + interrupt-safe sleep
## Problem / Scope

Reviewer sessions accumulate in the deployed container (12 idle observed 2026-07-22): the reap backstop `_reap_finished` runs inside `evaluate()`, which is fed by the `review-requested:@me` search — submitting a review CLEARS the request, so a finished round's PR vanishes from the search and the backstop is unreachable exactly when it's needed; terminal (approved) rounds are never reaped. The self-kill directive fast path is empirically unreliable (claude idles at its prompt). Fix it structurally, plus one same-file item folded in.

## Acceptance criteria

- [ ] The reaper becomes a **search-independent sweep**: each poll (or every Nth, documented), list `review-*` sessions via `alissa tmux ls`, cross-reference the spawn ledger (round ≤ completed reviews) and/or GitHub state (review submitted for that round / PR closed or merged), kill idle finished sessions (`alissa tmux kill`, never kill-server), spare in-flight rounds, honor dry-run, keep the ledger's tolerate-killed-sessions contract
- [ ] Tests (mirroring this repo's fakes): the sweep reaps a converged PR's session with NO pending review request (the exact starved case — this test must fail if the reap moves back inside the search-fed path); terminal approved rounds reaped; in-flight spared; dry-run logs only
- [ ] Same-file fold: `run_forever`'s `time.sleep(backoff)` moves inside the KeyboardInterrupt guard (Ctrl-C lands in the sleep, the dominant wall-clock state) + a test scripting the interrupt from the sleep — reference implementation: the devloop repo's fix (`a5de06d`)
- [ ] All four repo scripts green; version bumped (0.7.0 → 0.8.0) so the fix is releasable — but do NOT touch docker/claude/Dockerfile (the re-pin is a separate operator-sequenced step; note it in the PR body)

## Out of scope

`.github/workflows/` (a parallel branch owns them), docker/.

---
Alissa-Task: TASK-1609166327

RHDZMOTA:
Closes #16

## The bug

Reviewer sessions accumulate in the deployed container (12 idle observed 2026-07-22). The reap backstop `_reap_finished` ran inside `evaluate()`, which is fed by the `review-requested:@me` search — and **submitting a review clears the request**, so a finished round's PR vanishes from the search at exactly the moment its session becomes reapable. Terminal (approved) rounds were never reaped, and the reviewer self-kill fast path is empirically unreliable (claude idles at its prompt).

## The fix: a search-independent sweep

`sweep_sessions()` runs at the top of every poll (documented in its docstring: one `alissa tmux ls` when no `review-*` session is live, plus one PR fetch per distinct PR otherwise), **before** the search-fed evaluation:

- Lists live `review-*` managed sessions via `alissa tmux ls --json --live`.
- Maps each back to its (repo, PR, round) through the spawn ledger (`State.find_spawn_by_session`); sessions with no ledger row (another workspace's daemon, a human) are spared.
- A round is finished when the PR is **closed or merged**, or when `round ≤ completed` (verdict envelopes on the review task, falling back to the substantive-review count without a task).
- Kills only individual idle finished sessions with `alissa tmux kill <name>` — never the server, never a busy session, never an in-flight round. Dry-run logs only. Undecidable state (a failed fetch) spares the session until the next poll.
- Idempotency now comes from the live list itself: a reaped session drops off `tmux ls`, and the ledger keeps tolerating sessions killed behind its back (self-kill, operator).

Tests pin the starved case directly: `test_sweep_reaps_converged_pr_absent_from_the_search` runs `poll_once()` with the search returning **nothing** and asserts the reap still happens — it fails if the reap ever moves back inside the search-fed path. Also covered: terminal approved rounds, closed/merged PRs, in-flight spared, busy spared, foreign sessions spared, dry-run, idempotency, list-failure and undecidable-GitHub degradation.

## Same-file fold: interrupt-safe sleep

`run_forever`'s `time.sleep(backoff)` moved inside the `KeyboardInterrupt` guard — with a 60s poll interval (up to 900s backing off) Ctrl-C almost always lands in the sleep and must hit the same clean-exit path. Mirrors the devloop repo's fix (`a5de06d`), including the interrupt-from-sleep test.

## Version

`0.7.0 → 0.8.0` so the fix is releasable. **docker/claude/Dockerfile is deliberately untouched** — the `REVIEWLOOP_VERSION` re-pin to 0.8.0 is a separate operator-sequenced step after this publishes (same pattern as PR #12).

## Validation

pycodestyle clean · mypy clean · **104 tests pass** · coverage report green (loop.py 85%).

Alissa-Task: TASK-1609166327 (implementation task: TASK-351929130)


alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
```diff
@@ -244,30 +239,94 @@ def _convergence_reason(
...
+        # one PR can be live at once). None = undecidable this pass.
+        completed_cache: dict[tuple[str, int], float | None] = {}
+
+        for ses in sessions:
+            if not ses.is_idle:
+                # A busy session is still doing something (reviewing, or
+                # closing out its round) -- never yank the slot from under it.
+                continue
+            row = self.state.find_spawn_by_session(ses.name)
+            if row is None:
```
alissa-app, line 302: [major] "No ledger row ⇒ not ours" is not a sound inference, because the ledger can lose the row for a session we spawned ourselves — and when it does, this branch spares that session **permanently**.

`spawns` is `PRIMARY KEY (repo, number, round)` (`state.py:26`) and `record_spawn` is `INSERT OR REPLACE` (`state.py:88`). So the row is keyed by *round*, not by session. But `evaluate()` deliberately re-enqueues a round whose reviewer is presumed stalled:

```python
if age is not None:
    log.warning("%s round %d has been in flight %.0f min with no submitted "
                "review — re-enqueuing (reviewer session presumed stalled)", ...)
return self._spawn(pr, round_, task)
```

`_spawn` mints a fresh nonce'd name via `session_name(pr, round_)` and records it for the *same* `(repo, number, round)` — overwriting the row. The original session is still live and still holding a worker slot, but `find_spawn_by_session(<old name>)` now returns `None`, so it lands here and is spared as "another workspace's daemon (or a human)."

I confirmed the mechanism against the real schema rather than by reading:

```
find_spawn_by_session(review-r-pr19-r1-AAAAAA) -> None  => sweep SPARES it as foreign
find_spawn_by_session(review-r-pr19-r1-BBBBBB) -> row round=1
rows in spawns: 1
```

Why this is worth blocking on rather than filing: it is the same accumulation this PR exists to stop, and it fires on precisely the population the upstream task singles out. A reviewer that stalls and idles at its prompt without submitting is *the* case `STALE_ROUND_SECONDS` exists to detect — so the sessions most likely to be orphaned from the ledger are the ones most likely to sit idle forever. `_reap_finished` is gone, and the self-kill fast path is unreliable by hypothesis, so nothing else will ever collect them.

The docstring above says the live session list "cannot lose a finished session." True of tmux — but the guarantee has to survive the hop through the ledger, and right now it doesn't.

Suggested direction: give sessions their own identity instead of inheriting the round's. A `sessions(session TEXT PRIMARY KEY, repo, number, round, spawned_at)` table appended on every spawn keeps superseded rounds addressable while leaving the existing `(repo, number, round)` in-flight guard exactly as it is. Worth a test that records the same round twice under two session names, marks both live and idle, and asserts the sweep reaps **both**.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
```diff
@@ -244,30 +239,94 @@ def _convergence_reason(
...
+        the review-requested:@me search -- and submitting a review CLEARS the
+        request, so a finished round's PR vanished from the search at exactly
+        the moment its session became reapable; terminal (approved) rounds
+        were never reaped and idle reviewer sessions accumulated in the
+        worker. The sweep instead starts from the live session list, which
+        cannot lose a finished session, and works back to the round via the
+        spawn ledger. It must stay search-independent: never move it (back)
+        into the evaluate() path.
+
+        Every-poll is affordable: one `alissa tmux ls` when no review-*
```
alissa-app: [minor] This cost claim is the stated justification for running the sweep on every poll, so it should be accurate — and it currently undercounts by roughly 4×.

Per distinct PR with a live idle session, `_completed_rounds` can issue four external calls, not one:

1. `gh api repos/<o>/<r>/pulls/<n>` — the PR fetch named here;
2. `alissa task list --json` inside `find_review_task` (`alissa.py:70`, 90s timeout) — and this one fetches the acting actor's **entire** task list, then regex-matches titles; on this workspace that list is already well over a hundred tasks;
3. then either `alissa task get <ref> --json` (`count_verdicts`, another 90s timeout) or `gh api .../pulls/<n>/reviews`.

At a 60s poll interval with several live sessions that is a materially different budget from the one the docstring describes, and the whole-task-list fetch is the expensive part rather than the GitHub call.

Either correct the docstring to describe what actually runs, or cut the work: PRs that are also in the review-requested search get all of this fetched again moments later in `evaluate()`, so a per-poll memo shared between the sweep and the evaluation pass would remove the duplication rather than just re-document it.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
```diff
@@ -244,30 +239,94 @@ def _convergence_reason(
...
+        """
+        owner, _, repo = repo_slug.partition("/")
+        try:
+            pr = self.github.pull_request(owner, repo, number)
+            if pr.is_terminal:
+                return float("inf")
+            task = self.alissa.find_review_task(owner, repo, number)
+            if task is not None:
+                return self.alissa.count_verdicts(task.ref)
+            return len(self.github.my_reviews(owner, repo, number))
```
alissa-app: [minor] This fallback can mark a round "complete" while the reviewer is still mid-close-out, and `is_idle` is the only thing standing between that and a kill.

`len(my_reviews)` increments the moment the GitHub review is submitted — but per `procedures/review-a-pr.md` the reviewer still has to record the CR6 verdict envelope and move its task to `pending_validation` after that. So there is a real window where `completed >= round` while the session still has work to do. Kill it there and the round's verdict never exists; the loop sees no envelope and re-spawns the whole round from scratch.

The envelope path (`count_verdicts`, the branch above) does not have this problem — it increments only once the round is genuinely closed. But the fallback is reached more often than "round 1 before the task exists" suggests: `find_review_task` filters on `t.is_open`, and `OPEN_STATUSES` excludes `validated`, so a review task that a human has validated drops the daemon back onto this branch for every later sweep of that PR.

`is_idle` is a weaker guard than it looks, because a claude session parked at a prompt — waiting on a permission gate, or just between turns — reports `idle` while very much mid-round. `alissa tmux ls --json` already returns `lastActivity` (I checked the live output); carrying it onto `ManagedSession` and requiring a short quiet period before reaping would close the window cheaply, and is strictly more conservative than what is here now.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
```diff
@@ -244,30 +239,94 @@ def _convergence_reason(
...
             try:
-                self.alissa.kill_session(session)
+                self.alissa.kill_session(ses.name)
             except Exception:  # pragma: no cover - defence in depth
-                log.exception("failed to reap session %s", session)
+                log.exception("failed to reap session %s", ses.name)
                 continue
-            self.state.record_reap(session)
-            log.info("reaped finished reviewer session %s (round %d done)", session, round_)
+            self.state.record_reap(ses.name)
```
alissa-app, line 332: [minor] After this change the `reaps` table is write-only: `record_reap` is still called here, but `State.is_reaped` (`state.py:104`) now has no caller anywhere in `src/main` — the old `_reap_finished` was its only reader, and the sweep gets idempotency from the live list instead (which is the better design — `test_sweep_is_idempotent_across_polls` shows why).

The new `state.py` docstring says as much ("a reap record is bookkeeping, never a precondition"), so this is deliberate, and I am not asking you to reverse it. But leaving a public read API with zero readers is how the next person concludes the table is authoritative and starts gating on it — which would reintroduce exactly the failure this PR removes, since a session killed behind the ledger's back would then be un-reapable.

Either drop `is_reaped` and let the docstring's "bookkeeping" stand alone (the audit trail keeps working), or keep it and say in one line at the call site why it is intentionally not consulted before killing.

alissa-app (changes requested):
## Review — round 1 of cap 3 (CR4 rubric, all five dimensions)

Review task: **TASK-1242175580** · Origin task: **TASK-351929130** · Reviewer: `alissa-app`

The core of this is right, and the framing is the best part of it: the sweep starts from live tmux state rather than from the search that a submitted review empties, and `test_sweep_reaps_converged_pr_absent_from_the_search` pins exactly that by running `poll_once()` with `review_requests()` returning nothing — a test that fails if the reap ever drifts back inside `evaluate()`. I verified the external contract the sweep depends on: `alissa tmux ls --json --live` really does emit `name` without the `ali-` prefix and `status` in `idle`/`busy`, so `name.startswith("review-")` and `status == "idle"` match reality and line up with the names in the spawn ledger.

Requesting changes on **one [major]** — a still-live session can lose its ledger row, and the sweep then spares it forever — plus **one load-bearing [question]** about whether this fix actually reaches the deployed container.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 |
| minor | 3 |
| nit | 1 |
| question | 1 (open: 1) |

### Dimension report

- **Correctness** — the origin task's DoD and all four validation criteria are met: PR open and `Closes #16` (vc-1), the sweep is genuinely search-independent (vc-2), `time.sleep` moved inside the `KeyboardInterrupt` guard (vc-3), `version` 0.7.0 → 0.8.0 (vc-4). One [major] on an unsound inference in the sweep, one [minor] on a reap window — both inline.
- **Coding practices** — matches this repo's idiom closely, including the house style of docstrings that record *why* a previous shape failed. CI is green on all four checks. One [minor] (a now-dead read API) and one [nit], below.
- **Security** — nothing found. Every call is `subprocess.run` with a list argv and no `shell=True`; the only externally-sourced string that reaches a command is the session name from `alissa tmux ls`, and it is doubly constrained (must start with `review-`, must resolve to a row in our own ledger) before it is passed as a single argument to `alissa tmux kill`. Only individual sessions are killed — never `kill-server`. No secrets, no widened access, no new dependencies.
- **Performance** — one [minor]: the docstring's cost claim, which is the justification for running this every poll, understates the actual per-poll work. Inline.
- **Side-effects / blast radius** — killing sessions is irreversible and this is the code that decides to do it, so the guards matter: sparing busy sessions, sparing in-flight rounds, sparing foreign sessions, and sparing on any undecidable GitHub/CLI failure are all present and all tested. Adding `state`/`merged` to `PullRequest` with `"open"`/`False` defaults is safe — every other construction site keeps `is_terminal` false, which fails toward sparing. The remaining blast-radius concern is the [question] below.

### [question] — does this fix actually reach the box with the problem?

`docker/claude/Dockerfile:29` currently reads `ARG REVIEWLOOP_VERSION=0.6.0`. The PR body says the re-pin to 0.8.0 is "a separate operator-sequenced step after this publishes (same pattern as PR #12)" — but the pin sitting at **0.6.0**, not 0.7.0, means the PR #12 precedent being cited is itself the case where that follow-up step never happened, and the deployed container is now two minor versions behind.

That matters here more than usual: the symptom in the upstream task (12 idle `review-*` sessions, observed 2026-07-22) was observed *on the deployed container*, and the upstream DoD explicitly reads "a version bump + Dockerfile re-pin follow so the deployed container gets it." Which Alissa task carries the 0.8.0 re-pin? If the answer is "none yet," that is a `[triage:later]` needing a registered task, not an assumption — otherwise this merges, publishes, and the container keeps accumulating sessions exactly as before.

I am deliberately **not** asking you to touch the Dockerfile in this PR — keeping it out is the right call. I only need the follow-up to exist somewhere durable.

### [nit] `kill_session`'s `dry_run` parameter no longer has a caller

`alissa.py` — the sweep now gates dry-run itself (`loop.py:289`) and calls `self.alissa.kill_session(ses.name)` with no `dry_run` argument, and it is the only call site. The parameter and its `if dry_run:` branch are unreachable in production. Harmless, but it is the kind of leftover that makes a future reader think dry-run is handled one layer deeper than it is.

### Also noted, not a finding

The branch and commit are named for `TASK-1609166327` (the upstream task) while the implementation task is `TASK-351929130`. Both are disclosed in the PR body, so nothing is ambiguous — flagging only because this repo has already merged one fix for task-ref confusion (`afd08d6`), so the convention is worth pinning down eventually.

Fix the [major] (make the ledger able to name every session it spawned, and add a test that records the same round twice and asserts both sessions are reaped), and answer the [question] with a task reference. The three [minor]s and the [nit] are yours to pursue or reasonably decline — per CR8 each needs one `[triage:*]` reply with reasoning, and a reasoned `[triage:ignore]` on any of them will not be re-raised.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
RHDZMOTA: [triage:pursue] — Confirmed, and it is the worst kind of miss: the orphan population is exactly the stalled-and-idle sessions this PR exists to reap. Fix in the next push: `spawns` re-keys on `session` (the nonce makes it unique per spawn) with a secondary index on `(repo, number, round)`; `record_spawn` stops destroying history, and `get_spawn`/`spawn_age` return the *newest* spawn for a round — preserving the staleness-aging semantics evaluate() relies on. Existing state.db files get a copy-migration on open (SQLite cannot alter a PK in place), detected via `PRAGMA table_info`. Tests added per your ask: two recorded spawns of the same round, both live, both reaped by one sweep; plus a migration test that opens an old-schema DB and asserts the rows survive and a second same-round spawn can now be recorded.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
RHDZMOTA: [triage:pursue] — Rather than re-documenting the 4× cost, the next push removes the expensive call entirely: the sweep stops calling `find_review_task` (the whole-task-list fetch) and reads `task_ref` straight off the ledger row it already holds — the daemon recorded it at spawn time. Per distinct PR the sweep now issues the PR fetch plus exactly one of `alissa task get <ref>` (row has a task ref) or the reviews fetch (it does not). The docstring's cost claim is corrected to enumerate those calls. Side benefit relevant to your next finding: the ledger ref keeps the envelope-counting path even after a human validates the review task, where `find_review_task` (which filters to open statuses) would have dropped the sweep back onto the racier GitHub-count fallback.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
RHDZMOTA: [triage:pursue] — Both halves taken. (1) The fallback's reach shrinks as described in the thread above: with `task_ref` read from the ledger row, the GitHub-count path only runs for spawns recorded before any review task existed, and the validated-task regression you traced through `OPEN_STATUSES` disappears. (2) The close-out window itself: `ManagedSession` gains `last_activity` from the `lastActivity` field you verified in the live `tmux ls --json` output, and the sweep requires a quiet period (`REAP_QUIET_SECONDS = 300`) of idleness before killing — a session that just submitted its review and is parked mid-close-out reports recent activity and is spared until it has been genuinely quiet. A missing `lastActivity` defaults to 0 (= long quiet) so the field's absence cannot indefinitely immunize a session; `is_idle` and round-finished remain the primary gates. Test added: finished round, idle session, recent activity → spared; same session past the quiet period → reaped.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
RHDZMOTA: [triage:pursue] — Taking the second of your two options: `is_reaped` stays (the sweep tests also use it to assert bookkeeping), and the `record_reap` call site gains the one-line warning you asked for — the live list is the authority, and gating a kill on the reaps table would spare any session killed behind the ledger's back, reintroducing the accumulation this PR removes. That puts the do-not-gate-on-this contract at the exact spot a future reader would be tempted to add the gate.

RHDZMOTA:
## CR8 triage — round 1

All four inline findings are replied on their threads (all **[triage:pursue]**). The two body-level items:

**[question] Does this fix reach the deployed container?** — [triage:answer] You are right that the PR #12 precedent is the cautionary tale, not the reassurance: the pin sits at 0.6.0 because the 0.7.0 re-pin after PR #13 never happened. The follow-up now exists durably: **TASK-107268632** ("Re-pin reviewloop Dockerfile REVIEWLOOP_VERSION 0.6.0 → 0.8.0 and redeploy the reviewer container"), linked downstream of origin TASK-1609166327, with a DoD requiring the deployed container to report 0.8.0 — so the deploy carries both the convergence-head fix and this sweep. Dockerfile remains untouched in this PR.

**[nit] `kill_session`'s dead `dry_run` parameter** — [triage:pursue] Removed in the next push; the sweep is the only caller and gates dry-run itself, and the parameter's presence implies a second layer of dry-run handling that does not exist.

**On the task-ref naming note** — acknowledged; branch/commit carry the origin task ref per this repo's existing convention (PRs #11/#13), implementation task disclosed in the PR body. Happy to follow a pinned convention once one is declared.

Fixes and re-request follow this comment.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/state.py
```diff
@@ -45,9 +49,35 @@ def __init__(self, path: Path):
         path.parent.mkdir(parents=True, exist_ok=True)
         self._db = sqlite3.connect(str(path))
         self._db.row_factory = sqlite3.Row
+        stale = self._spawns_keyed_by_round()
+        if stale:
+            self._db.execute("ALTER TABLE spawns RENAME TO spawns_v0")
```
alissa-app: [minor] The migration is destructive but not atomic, so a crash inside it wipes the ledger permanently and silently.

`sqlite3.Connection.executescript` issues an implicit `COMMIT` before it runs its script. So line 55 commits the `ALTER TABLE ... RENAME` **and** the empty new `spawns` — while the `INSERT ... SELECT` that carries the rows across, and the `DROP`, are still uncommitted below. Die in that window and the next open finds a `spawns` whose pk is already `['session']`, so `_spawns_keyed_by_round()` returns `False`, the migration never runs again, and the rows sit in `spawns_v0` forever.

I confirmed it by replaying `__init__` and stopping right after `executescript`:

```
tables      : ['spawns', 'spawns_v0']
spawns rows : 0
v0 rows     : 1
pk          : ['session'] -> next open sees stale = False
```

An empty ledger is not a degraded state here, it is this PR's own bug restored: `find_spawn_by_session` returns `None` for every live reviewer, so the sweep spares them all as "another workspace's daemon" — permanently, and with no log line saying anything went wrong. The in-flight guard goes with it, so rounds already in flight get re-enqueued.

I went back and forth on `[major]` and landed on `[minor]` on probability, not consequence: it needs a crash in a genuinely narrow window. Two things kept it from being lower. The window is at **startup**, and per TASK-107268632 this code's first-ever run is a container redeploy — startup is exactly when a process gets OOM-killed or rolled back. And the failure is silent and permanent, so nobody learns about it until sessions are piling up again for a reason that looks like the bug you just fixed. Happy to be argued down to a `[nit]` if you think the window is smaller than I do.

The fix is small — drive the DDL through `execute()` inside one explicit transaction instead of letting `executescript` commit mid-migration, so the rename, the create, the copy and the drop land or roll back together.

(For what it is worth I also tested the two-processes-open-at-once interleaving, expecting it to be worse: it is benign. Process B just sees an empty ledger for one poll and spares everything, then A's copy commits. Only the crash path loses data.)

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
```diff
@@ -244,30 +247,113 @@ def _convergence_reason(
...
+                # Idle but recently active: likely mid-close-out (the review
+                # is submitted before the envelope and task move land). Wait
+                # for a real quiet period; see REAP_QUIET_SECONDS.
+                continue
+            row = self.state.find_spawn_by_session(ses.name)
+            if row is None:
+                # Not in our ledger: another workspace's daemon (or a human)
+                # owns it. Not ours to judge.
+                continue
+            key = (row["repo"], row["number"], row["task_ref"])
```
alissa-app, line 312: [nit] The cost docstring you just corrected can still undercount, because the cache key is finer than "per distinct PR".

The docstring above promises one PR fetch "per distinct PR with a live idle quiet session", but `completed_cache` is keyed `(repo, number, task_ref)`. Two live sessions on the same PR with different `task_ref`s — the realistic case being a round-1 spawn recorded before the review task existed (`task_ref = None`) alongside a later round that has one — miss each other in the cache and fetch the same PR twice.

Genuinely trivial in cost, and I would not raise it at all except that the whole point of the round-1 finding was that this docstring should be believable. Either key the PR fetch on `(repo, number)` and the round count on the full triple, or soften the wording to "per distinct PR and task ref".

alissa-tools-github-reviewloop/src/test/test_alissa/test_tools/test_github/test_reviewloop/test_loop.py
```diff
@@ -951,65 +963,304 @@ def test_task_without_a_number_is_skipped(monkeypatch):
...
+    pr = make_pr()
+    w, _, al = watcher(config, pr, [review()])
+    s1 = _record(w, pr, 1)
+    _live(al, s1, last_activity=time.time())  # just did something
+
+    w.sweep_sessions()
+    assert al.killed == []
+
+    al.sessions = []
+    _live(al, s1, last_activity=time.time() - STALE_ROUND_SECONDS)  # long quiet
```
alissa-app: [nit] This expresses "long quiet" with `STALE_ROUND_SECONDS`, a constant about something else.

The assertion is about `REAP_QUIET_SECONDS` (300s), but the offset borrows the stalled-round constant (5400s). It passes only because of the accidental relationship between two unrelated numbers — raise `REAP_QUIET_SECONDS` above 90 minutes and this test starts failing for a reason that has nothing to do with what it is testing.

`REAP_QUIET_SECONDS * 2` (or `+ 1`) would say what is meant and stay true under any future retuning. The `last_activity=0.0` default in `_live` already has the right comment for the other half of this.

alissa-app (approved):
## Review — round 2 of cap 3 (CR4 rubric, all five dimensions)

Review task: **TASK-1242175580** · Origin task: **TASK-351929130** · Reviewer: `alissa-app`

**Verdict: approve.** All six round-1 findings are resolved, and the delta introduces no blocker or major. What is left is one [minor] and two [nit]s, all in the new commit, none of which gate the merge.

### Prior-finding reconciliation (CR9 step 2)

Triage is complete: four inline threads carry `[triage:pursue]`, and the body-level `[nit]` and `[question]` are triaged in the CR8 round-1 comment. Nothing bounced. I verified each claimed fix against `e05db98` rather than against the reply:

| --- | --- | --- |
| [major] ledger loses the row of a re-enqueued round, sparing a live session forever | pursue | **fixed** (`e05db98`) |
| [minor] sweep cost docstring undercounts ~4x | pursue | **fixed** (`e05db98`) |
| [minor] reap window during round close-out | pursue | **fixed** (`e05db98`) |
| [minor] `State.is_reaped` is a reader-less public API | pursue | **fixed** (`e05db98`) |
| [nit] `kill_session`'s unreachable `dry_run` parameter | pursue | **fixed** (`e05db98`) |
| [question] does the fix reach the deployed container? | answer | **answered** — TASK-107268632 verified |

The [major] is genuinely fixed, not just re-described. `spawns` is now keyed `session TEXT NOT NULL PRIMARY KEY` with a `spawns_by_round` index, and `get_spawn` orders `spawned_at DESC, rowid DESC` — the `rowid` tiebreak matters, because `spawned_at` is whole seconds and two spawns of one round can share one. I re-ran my round-1 repro against the new schema:

```
old session row: FOUND      (was: None -> ORPHANED)
new session row: FOUND
rows in spawns : 2          (was: 1)
get_spawn newest: ...BBBBBB
```

The migration also works: an old-schema DB with rows in `spawns` and `reaps` comes through with every row preserved, `spawns_v0` dropped, pk `['session']`, a second same-round spawn recordable, and a re-open is a no-op. `test_old_round_keyed_ledger_is_migrated_on_open` pins it, and `test_sweep_reaps_both_sessions_of_a_twice_spawned_round` is exactly the two-sessions-one-round test I asked for.

On the [question], I checked the cited task rather than taking the reply's word: **TASK-107268632** exists, is open (`committed`), and its DoD names the 0.6.0 → 0.8.0 re-pin, the skipped 0.7.0, and a redeploy — including the reason the pin sits at 0.6.0. The Dockerfile is correctly untouched here. That satisfies the question; the deploy risk is now carried somewhere durable instead of being an assumption.

### Dimension report — the new delta

- **Correctness** — the round-1 [major] and all three [minor]s are fixed, and the origin task's DoD and four validation criteria still hold (sweep still search-independent, interrupt-safe sleep untouched, `version` 0.7.0 → 0.8.0, `Closes #16`). One new [minor] on the migration's atomicity, inline. 109 tests pass locally on the head commit; CI is green on all four checks.
- **Coding practices** — still matches repo idiom, including the house habit of docstrings that record why a prior shape failed (`_spawns_keyed_by_round` explains the orphaning it exists to undo, which is exactly right). Two [nit]s inline.
- **Security** — nothing found. The delta adds no new external input reaching a command: `last_activity` is parsed with an `isinstance` guard and only ever compared numerically, and `kill_session` lost a parameter rather than gaining one. Session names are still doubly constrained (`review-` prefix, then a lookup in our own ledger) before reaching `alissa tmux kill` as a single argv element. No secrets, no widened access, no new dependencies.
- **Performance** — improved, and this is the round-1 [minor] properly discharged rather than re-documented: dropping `find_review_task` from the sweep removes a whole-task-list fetch per PR per poll, which was the expensive call. Reading `task_ref` off the ledger row also removes the `OPEN_STATUSES` regression, so a human-validated review task no longer demotes the sweep to the racier GitHub-count path. Cost is now one PR fetch plus one of `alissa task get` / the reviews fetch — modulo the [nit] below.
- **Side-effects / blast radius** — this round's blast radius is the schema migration, which mutates a deployed `state.db` in place and irreversibly. That is the one place the delta adds real risk, and it is the subject of the [minor]. The reap guards themselves are unchanged and still fail toward sparing: `count_verdicts` degrades to `0` on an unreadable task ref, which makes `row["round"] > completed` true and spares the session, and the new quiet gate spares on negative skew.

### On the quiet-period fix — I checked the contract, not just the code

`REAP_QUIET_SECONDS` is only sound if `lastActivity` is (a) in the same unit as `time.time()` and (b) actually goes stale for a parked session. Both are worth verifying because the Alissa API elsewhere reports epoch **milliseconds** (`task get --json` returns `createdAt: 1784755407316`), and a millisecond field here would make `time.time() - last_activity` hugely negative for every session — sparing all of them forever and silently reverting this PR's entire purpose.

It is fine. Live `alissa tmux ls --json` emits `"lastActivity": 1784753083` — epoch **seconds**. And idle sessions do go quiet:

```
idle  quiet=    3113s  review-...-pr18-r3-f9aa9d
idle  quiet=    1073s  review-...-pr18-r1-b8e61d
busy  quiet=      78s  review-...-pr21-r2-943a42
```

So a parked claude session's activity really does freeze — the TUI does not keep bumping it — and the 300s threshold separates "mid-close-out" from "done" as intended. The `0.0` default for a missing field is a deliberate, documented trade (fail toward reaping rather than let an absent field immunize a session), and I am not second-guessing it.

### Shortest path from here

The [minor] is a ~3-line change (one explicit transaction around the migration). Fix it before merge, or carry it on a registered task per CR5 — a `[minor]` that evaporates was a mislabeled `[nit]`, and this one has a permanent consequence, so I would rather it not evaporate. The two [nit]s are yours to take or drop. None of the three blocks the merge, and no further review round is owed on my account.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/state.py
RHDZMOTA: [triage:pursue] — Agreed, and thank you for replaying the crash window rather than eyeballing it; `executescript`'s implicit COMMIT is exactly the kind of footgun that deserves a repro. Next push makes the migration a single explicit transaction: `BEGIN IMMEDIATE` → rename → create new table → copy → drop → `COMMIT`, with a `ROLLBACK` on any failure — so an interrupted migration leaves the round-keyed table fully intact and the next open simply retries. The new-table DDL is extracted to a shared constant so SCHEMA and the migration cannot drift. Test added: an obstructing `spawns_v0` table makes the migration's first statement fail mid-transaction; the old ledger must come through untouched and still be detected as stale, and a retry after clearing the obstruction must migrate every row.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/loop.py
RHDZMOTA: [triage:pursue] — Taking the first option so the docstring stays true without softening: the PR fetch is memoized per `(repo, number)` in a separate per-sweep map, and only the round count keys on the full `(repo, number, task_ref)` triple. `_completed_rounds` now takes the already-fetched PR, so the mixed-task-ref case (round-1 spawn pre-task alongside a later ref-carrying row) costs one PR fetch, and a test pins the fetch count at 1 for exactly that shape.

alissa-tools-github-reviewloop/src/test/test_alissa/test_tools/test_github/test_reviewloop/test_loop.py
RHDZMOTA: [triage:pursue] — Right, the assertion should not depend on an accidental inequality between two unrelated constants. The offset becomes `REAP_QUIET_SECONDS * 2` (imported alongside the other loop constants), which stays correct under any retuning of either knob.

alissa-tools-github-reviewloop/src/test/test_alissa/test_tools/test_github/test_reviewloop/test_loop.py
```diff
@@ -951,67 +971,367 @@ def test_task_without_a_number_is_skipped(monkeypatch):
...
+    db = tmp_path / "state.db"
+    conn = sqlite3.connect(str(db))
+    conn.executescript(
+        """
+        CREATE TABLE spawns (
+            repo TEXT NOT NULL, number INTEGER NOT NULL, round INTEGER NOT NULL,
+            head_sha TEXT NOT NULL, session TEXT NOT NULL, task_ref TEXT,
+            spawned_at INTEGER NOT NULL, PRIMARY KEY (repo, number, round)
+        );
+        CREATE TABLE spawns_v0 (blocker INTEGER);  -- makes the RENAME fail
```
alissa-app, line 1273: [minor] This test passes against the exact bug it is named for, so it does not pin the fix.

Making `spawns_v0` pre-exist fails the **first** statement of the migration — the `ALTER TABLE ... RENAME`. At that point nothing has happened yet, so the old code and the new code are indistinguishable: the pre-fix `executescript` version also raises `OperationalError` here and also leaves `spawns` intact with its round-keyed PK. I checked rather than assumed — reverting `state.py` to its `e05db98` (non-atomic) form and running this test: `1 passed`.

The window the round-2 finding was actually about is a failure **after** the rename, during the copy. That one does discriminate. Injecting a crash on the `INSERT ... SELECT`:

- new code → rolls back, `spawns` has 2 rows, PK still `(repo, number, round)` so it is still detected as stale, retry migrates both sessions;
- pre-fix code → `spawns` committed **empty**, both rows stranded in `spawns_v0`, PK now `session` so it no longer looks stale, and the retry finds **neither** session — the permanent ledger loss that makes the sweep spare every live session forever as "not ours".

Suggest pointing the test at that window instead (or additionally): patch the connection so the copy raises, then assert the rollback — the two assertions already here (`COUNT(*) == 1`, PK is round-keyed) become load-bearing. Worth noting the sibling test added in this push, `test_sweep_fetches_each_pr_once_even_across_task_refs`, *does* fail against pre-fix `loop.py` (`assert 2 == 1`) — so this is a one-off gap, not a pattern.

Not blocking: the behaviour itself is correct, verified directly.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/state.py
```diff
@@ -45,9 +53,52 @@ def __init__(self, path: Path):
...
+            self._db.execute(_SPAWNS_TABLE)
+            self._db.execute(
+                "INSERT OR REPLACE INTO spawns "
+                "(repo, number, round, head_sha, session, task_ref, spawned_at) "
+                "SELECT repo, number, round, head_sha, session, task_ref, spawned_at "
+                "FROM spawns_v0"
+            )
+            self._db.execute("DROP TABLE spawns_v0")
+        except BaseException:
+            self._db.execute("ROLLBACK")
```
alissa-app, line 83: [nit] This `ROLLBACK` can replace the exception that caused it, hiding the real failure.

When SQLite itself aborts the transaction — `SQLITE_FULL`, `SQLITE_IOERR`, `SQLITE_NOMEM` — the transaction is already gone by the time this line runs, so the explicit rollback raises `OperationalError: cannot rollback - no transaction is active` (confirmed: a second `ROLLBACK` on a rolled-back connection raises exactly that). That new error propagates in place of the disk-full error the operator actually needs to see. The original survives as `__context__`, so it is a readability loss rather than a lost cause, and **data safety is unaffected** — SQLite already rolled back, which is why the rollback failed.

`self._db.rollback()` (the DB-API method, a no-op when no transaction is active) instead of `execute("ROLLBACK")` sidesteps it, or wrap the line in a narrow `try/except sqlite3.OperationalError: pass`. Purely a diagnostics nit — the atomicity property this method exists for holds either way.

alissa-app (approved):
## Review round 3 — `approve`

Delta reviewed: `e05db98..8690f05` (round-2 fixes). All three round-2 findings carry exactly one `[triage:pursue]` reply, and all three are **verified fixed against the diff**, not accepted on the reply. No blocker, no major — approving with one `[minor]` and one `[nit]`, both about the *test* and *diagnostics* around an otherwise correct fix.

| round-2 finding | triage | disposition |
| --- | --- | --- |
| `[minor]` migration destructive but not atomic (`executescript`'s implicit COMMIT) | pursue | **fixed** — `_migrate_spawns` runs `BEGIN IMMEDIATE` → rename → create → copy → drop → `COMMIT`, with `ROLLBACK` on any failure. Verified by injecting a crash **mid-copy** (the actual dangerous window): the new code rolls back to 2 rows, still round-keyed, and the retry migrates both sessions. Against the pre-fix code the same injection commits an **empty** `spawns`, strands both rows in `spawns_v0`, and the retry finds neither session — exactly the permanent-ledger-loss the finding described. Bonus: `_SPAWNS_TABLE` shared between `SCHEMA` and the migration removes the drift risk that shape usually carries. |
| `[nit]` cache key finer than the docstring's "one PR fetch per distinct PR" | pursue | **fixed** — PR fetch memoized per `(repo, number)` in `prs` via the new `_sweep_pr`; the round count still keys on the full triple. Docstring corrected to match. Guard ordering still puts the fetch *after* busy/quiet/ledger filters, so the corrected claim is accurate. |
| `[nit]` test borrowed `STALE_ROUND_SECONDS` to mean "long quiet" | pursue | **fixed** — now `REAP_QUIET_SECONDS * 2`, imported alongside the other loop constants. |

I also re-checked the two refactor assumptions the delta newly leans on, since both would be silent breakage: `PullRequest` really does expose `owner`/`repo`/`number`/`slug` (`ghclient.py:22-46`), and `count_verdicts` really never raises (`alissa.py:176-197` catches `CommandError` *and* bare `Exception` in both stages), so moving it outside the `try` in `_completed_rounds` is behaviour-preserving rather than a swallowed-error regression.

### Dimensions

- **Correctness:** Origin DoD and all four validation criteria still hold (search-independent sweep, interrupt-safe sleep, version file at `0.8.0`, `Closes #16`). 111 tests pass on `8690f05` (109 → 111, the two new tests). Round-1's `[question]` stays discharged: TASK-107268632 re-confirmed by `get_task` — exists, open (`committed`), DoD names the 0.6.0 → 0.8.0 re-pin. Nothing found beyond the `[minor]` below.
- **Coding practices:** Matches repo idiom, including the house style of docstrings recording why a prior shape failed (`_migrate_spawns` explains the `executescript` footgun it avoids). One `[minor]` (test does not pin its fix), one `[nit]` (rollback diagnostics).
- **Security:** Nothing found — the delta adds no new external input reaching a command. `_sweep_pr` partitions a slug that came from our own ledger, session names remain doubly constrained (`review-` prefix + ledger lookup) before reaching `alissa tmux kill` as a single argv element, no `shell=True`, no secrets, no widened access, no new dependencies.
- **Performance:** Improved, and this is the delta's main win — the PR fetch is now memoized per distinct PR instead of per `(PR, task_ref)`, so the realistic mixed-`task_ref` case (a pre-task round-1 row beside a later ref-carrying one) costs one fetch rather than two.
- **Side-effects / blast radius:** Reduced. The migration is the only irreversible operation in the delta and it is now all-or-nothing; every failure path still fails toward *sparing* a session (`_sweep_pr` → `None` spares every session on that PR, `count_verdicts` degrades to `0` so `round > completed` spares, negative clock skew spares).

### Verdict

`approve`. Reconciliation is clean and the delta carries no blocker and no major. The `[minor]` is a test-coverage gap over code I verified correct by direct repro — worth fixing so the property stays pinned, but not worth another round; it is registered as a follow-up task.

alissa-tools-github-reviewloop/src/test/test_alissa/test_tools/test_github/test_reviewloop/test_loop.py
RHDZMOTA: [triage:later] → **TASK-2132422030** — You are right, and the revert-and-run check settles it: a test that passes against the pre-fix code pins nothing. The discriminating shape is exactly the one you injected (crash on the `INSERT ... SELECT`, after the rename), and the registered task's DoD requires the test to FAIL against the `e05db98` non-atomic implementation and pass against the atomic one. Deferred rather than pushed now only to keep your approve bound to `8690f05` — a post-approve push at round 3 of cap 3 would re-open a loop that owes no round.

alissa-tools-github-reviewloop/src/main/alissa/tools/github/reviewloop/state.py
RHDZMOTA: [triage:later] → **TASK-2132422030** (bundled with the test fix above, same file) — Confirmed mechanism: SQLite's auto-rollback on FULL/IOERR/NOMEM ends the transaction, so the explicit `execute("ROLLBACK")` raises and shadows the operator-relevant error, surviving only as `__context__`. The DB-API `self._db.rollback()` (no-op when no transaction is active) is the fix the task names. Data safety unaffected, as you note — diagnostics only, hence deferred with the same keep-the-approve-head-bound reasoning.

RHDZMOTA:
## Converged — round 3 approve on head `8690f05`

Both round-3 findings are triaged **[triage:later]** on their threads → **TASK-2132422030** (discriminating migration test + `rollback()` diagnostics), linked downstream of origin TASK-1609166327. No push follows, so the approve stays head-bound.

Two more follow-ups registered while closing out:
- **TASK-107268632** — Dockerfile `REVIEWLOOP_VERSION` 0.6.0 → 0.8.0 re-pin + redeploy (the round-1 [question]).
- **TASK-1564968556** — `latest_verdict` treats numeric `createdAt` as undated, so the FIRST envelope wins forever: observed twice this loop (`alissa-pr-review` exited `request_changes` while the task already held a newer approve envelope). It also blunts the daemon's envelope-based convergence signal. Filed with repro and blast radius.

Ready for operator merge. After merge, 0.8.0 auto-publishes; TASK-107268632 sequences the container re-pin.

PR #28, opened 2026-07-23
Container: pin the reviewer model via ALISSA_AGENT_MODEL (default opus)

### LINKED ISSUE #26: Container: pin the reviewer model via ALISSA_AGENT_MODEL (default opus)
# Container: pin the reviewer model via `ALISSA_AGENT_MODEL` (default `opus`)

**Repo:** fahera-mx/alissa-github-review-daemon — `docker/claude/` tier only.

## Problem

`docker/claude/agents.yaml` is a static `COPY` whose claude profile runs:

```yaml
claude:
  command: claude --dangerously-skip-permissions --permission-mode acceptEdits
```

No `--model` pin. Reviewer sessions therefore inherit whatever the persisted
`/login` credential's account default is — and on plan-based accounts that
default can silently fall back to a smaller model when usage thresholds are
hit. The reviewer is the pipeline's quality gate: if review recall degrades,
nothing downstream notices. Model choice is currently a rebuild-time decision
("edit this file before build"), contradicting the image's env-driven runtime
configuration.

## Change

Make the model a boot-time env var:

1. **New env var `ALISSA_AGENT_MODEL`, default `opus`.** At container boot the
   entrypoint produces the effective `agents.yaml` claude profile with
   `--model "$ALISSA_AGENT_MODEL"` appended to the command. Implementation
   choice is yours (generate from a heredoc, or targeted substitution on the
   baked copy), but log the effective command at startup like other resolved
   config.
2. **Opt-out:** `ALISSA_AGENT_MODEL=default` (or explicitly empty) omits the
   `--model` flag entirely, restoring today's account-default behavior. Any
   other value passes through verbatim (aliases `opus`/`sonnet` and full ids
   like `claude-opus-4-8` are both valid — no hardcoded allowlist).
3. **Docs:** `docker/claude/README.md` — add the variable to the runtime
   variables table with default, accepted values, opt-out semantics, and the
   silent-fallback rationale. Update the stale "edit this file before build"
   comment in `agents.yaml` to point at the env var.
4. **No behavior change elsewhere.** Flags, mode, quietSeconds, promptPatterns
   stay as-is; reviewer posture (CR6: reviewers never write) remains enforced
   by the round directive, unaffected by this change. A runtime-mounted custom
   agents.yaml must still win (document the precedence you implement).

## Acceptance

- Container boot (dry-run acceptable) with no `ALISSA_AGENT_MODEL` set shows the
  claude command carrying `--model opus`.
- Boot with `ALISSA_AGENT_MODEL=claude-opus-4-8` shows that exact value.
- Boot with `ALISSA_AGENT_MODEL=default` shows no `--model` flag.
- README updated; entrypoint tests only if the repo already tests entrypoint
  behavior (match existing patterns).

Keep the diff small and container-scoped: no Python package changes, no
reviewloop version bump. This mirrors an identical issue in
fahera-mx/alissa-github-develop-daemon — keep the variable name, default, and
opt-out semantics byte-consistent with that implementation, but do not touch
the devloop repo from this branch.

Alissa-Task: TASK-1843563761

RHDZMOTA:
## Summary

Makes the reviewer's claude model a **boot-time env var** instead of a rebuild-time edit of `agents.yaml`. The baked profile pinned no `--model`, so reviewers inherited the persisted `/login` account default — which on plan-based accounts can silently fall back to a smaller model under usage throttling. The reviewer is the pipeline's quality gate, so that degrades review recall unnoticed.

`docker/claude/` tier only — **no Python package changes, no reviewloop version bump**.

## Change

- **`ALISSA_AGENT_MODEL`, default `opus`.** At boot the entrypoint (new step `2d`) rewrites the baked claude profile's `command:` with `--model <value>` and logs the effective command (`effective reviewer command: …`).
- **Opt-out:** `ALISSA_AGENT_MODEL=default` (or empty) omits the `--model` flag, restoring account-default behavior. Any other value passes through **verbatim** — aliases (`opus`/`sonnet`) and full ids (`claude-opus-4-8`) both valid, no allowlist.
- **Precedence:** only the baked default (identified by an `alissa-managed:` marker comment) is rewritten. A runtime-mounted custom `agents.yaml` carries no marker and is used **verbatim** — the mounted command always wins.
- **No behavior change elsewhere:** flags, `mode`, `quietSeconds`, `promptPatterns` untouched; CR6 reviewer posture stays enforced by the round directive.
- Dockerfile: `ALISSA_AGENT_MODEL` added as an ARG→ENV (Railway-friendly, default `opus`). README: runtime-variables table row + a "Pinning the reviewer model" section (default, accepted values, opt-out, precedence, silent-fallback rationale). `agents.yaml`: stale "edit this file before build" comment now points at the env var.

## Acceptance (verified locally against the extracted entrypoint block)

| `ALISSA_AGENT_MODEL` | resulting `command:` |
| --- | --- |
| *(unset)* | `claude … --model opus` ✅ |
| `claude-opus-4-8` | `claude … --model claude-opus-4-8` ✅ |
| `default` / empty | `claude …` (no `--model`) ✅ |
| mounted custom (no marker) | left verbatim, ignored ✅ |

`bash -n` clean; baked + rewritten `agents.yaml` both parse (PyYAML) with all other profile fields intact. No entrypoint test file added — the repo has no existing entrypoint tests (its CI covers only the Python package, unchanged here).

Mirrors fahera-mx/alissa-github-develop-daemon#28 (variable name / default / opt-out kept byte-consistent); the devloop repo is not touched from this branch.

Closes #26

---

**Task refs**
- Origin task: TASK-1843563761
- Implementation task: TASK-1721233512

### DIFF
--- a/docker/claude/agents.yaml
+++ b/docker/claude/agents.yaml
@@ -9,8 +9,16 @@
 # root) — the image runs as uid 1000, which satisfies it. Reviewer posture (CR6:
 # reviewers never write) is enforced by the round directive, not by this profile.
 #
-# To pin a model or tweak flags, override the command below (edit this file
-# before build, or mount your own agents.yaml over it at runtime).
+# To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default: opus) —
+# the entrypoint rewrites the `command:` below with `--model "$ALISSA_AGENT_MODEL"`
+# at container boot (set it to `default` or empty to omit the flag and inherit the
+# account default). To change flags instead, mount your own agents.yaml over this
+# path at runtime; a mounted file lacks the `alissa-managed:` marker below and is
+# used verbatim, so ALISSA_AGENT_MODEL does not touch it (your command wins).
+#
+# alissa-managed: the entrypoint regenerates the `command:` line of this default
+# profile at boot from ALISSA_AGENT_MODEL — do not remove this marker line, it is
+# what distinguishes the baked default from a runtime-mounted custom agents.yaml.
 claude:
   command: claude --dangerously-skip-permissions --permission-mode acceptEdits
   mode: interactive
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -101,6 +101,61 @@ alissa auth login --token "${ALISSA_API_TOKEN}" >/dev/null 2>&1 \
 log "alissa authenticated"
 
 # -----------------------------------------------------------------------------
+# 2d. Resolve the reviewer model into the baked agents.yaml.
+#
+# The reviewer is the pipeline's quality gate, but the baked claude profile pins
+# no model — so it inherits the persisted /login account's default, which can
+# silently fall back to a smaller model when a plan hits its usage threshold. We
+# pin it explicitly at boot: ALISSA_AGENT_MODEL (default `opus`) is appended to
+# the profile's `command:` as `--model <value>`. The value passes through
+# verbatim — aliases (`opus`, `sonnet`) and full ids (`claude-opus-4-8`) are both
+# valid, no allowlist. `default` or an empty value omits the flag entirely,
+# restoring the account-default behavior.
+#
+# Precedence: we only rewrite the BAKED default (identified by its `alissa-managed:`
+# marker). A custom agents.yaml mounted over this path carries no marker, so it is
+# left verbatim and ALISSA_AGENT_MODEL is ignored for it — the mounted command
+# always wins. The baked file lives in the ephemeral home (not on the volume), so
+# it is pristine on every boot and this rewrite is idempotent.
+# -----------------------------------------------------------------------------
+AGENTS_YAML="${HOME}/.config/alissa/agents.yaml"
+# Default `opus` only when UNSET (use `-`, not `:-`): an explicitly empty value is
+# a valid opt-out and must NOT be re-defaulted back to opus.
+AGENT_MODEL="${ALISSA_AGENT_MODEL-opus}"
+if [ ! -f "${AGENTS_YAML}" ]; then
+  log "WARN: ${AGENTS_YAML} not found — skipping model pin (worker will fall back to a bare claude)"
+elif ! grep -q 'alissa-managed:' "${AGENTS_YAML}"; then
+  log "custom agents.yaml in effect (no alissa-managed marker) — using it verbatim, ALISSA_AGENT_MODEL ignored"
+else
+  BASE_CMD="claude --dangerously-skip-permissions --permission-mode acceptEdits"
+  if [ -z "${AGENT_MODEL}" ] || [ "${AGENT_MODEL}" = "default" ]; then
+    CLAUDE_CMD="${BASE_CMD}"
+    log "reviewer model: account default (ALISSA_AGENT_MODEL='${AGENT_MODEL}') — no --model flag"
+  else
+    CLAUDE_CMD="${BASE_CMD} --model ${AGENT_MODEL}"
+    log "reviewer model: ${AGENT_MODEL} (ALISSA_AGENT_MODEL)"
+  fi
+  # Rewrite the profile's `command:` line. python (not sed) so an arbitrary
+  # passed-through model value can't collide with a substitution metacharacter.
+  python3 - "${AGENTS_YAML}" "${CLAUDE_CMD}" <<'PY' || die "failed to render agents.yaml"
+import re, sys
+path, cmd = sys.argv[1], sys.argv[2]
+out, seen = [], False
+for ln in open(path).read().splitlines(keepends=True):
+    m = re.match(r'^(\s*)command:\s', ln)
+    if m and not seen:
+        out.append(f"{m.group(1)}command: {cmd}\n")
+        seen = True
+    else:
+        out.append(ln)
+if not seen:
+    sys.exit("no `command:` line found in agents.yaml")
+open(path, "w").writelines(out)
+PY
+  log "effective reviewer command: ${CLAUDE_CMD}"
+fi
+
+# -----------------------------------------------------------------------------
 # 3. Bootstrap the workspace (bootstrap-from-manifest model)
 #
 # Reviewers cd into {root}/{repo}/main worktree hubs. With on_missing_hub:add
--- a/docker/claude/Dockerfile
+++ b/docker/claude/Dockerfile
@@ -181,6 +181,11 @@ ARG ALISSA_REVIEW_SKILLS="alissa-code-workspace|alissa-code-review"
 ARG ALISSA_POLL_INTERVAL="60"
 ARG ALISSA_ROUND_CAP="3"
 ARG ALISSA_AGENT_PROFILE="claude"
+# Reviewer model pinned into the agents.yaml claude command at boot (the reviewer
+# is the quality gate, so it must not silently fall back to a smaller model on a
+# usage-throttled account). Alias (`opus`/`sonnet`) or full id (`claude-opus-4-8`);
+# `default` or empty omits the --model flag. See the entrypoint (step 2d) + README.
+ARG ALISSA_AGENT_MODEL="opus"
 ARG ALISSA_ON_MISSING_HUB="add"
 # Worker reconcile tick, seconds.
 ARG ALISSA_WORKER_INTERVAL="2"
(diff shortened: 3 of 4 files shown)


docker/claude/entrypoint.sh
```diff
@@ -100,6 +100,61 @@ alissa auth login --token "${ALISSA_API_TOKEN}" >/dev/null 2>&1 \
...
+AGENTS_YAML="${HOME}/.config/alissa/agents.yaml"
+# Default `opus` only when UNSET (use `-`, not `:-`): an explicitly empty value is
+# a valid opt-out and must NOT be re-defaulted back to opus.
+AGENT_MODEL="${ALISSA_AGENT_MODEL-opus}"
+if [ ! -f "${AGENTS_YAML}" ]; then
+  log "WARN: ${AGENTS_YAML} not found — skipping model pin (worker will fall back to a bare claude)"
+elif ! grep -q 'alissa-managed:' "${AGENTS_YAML}"; then
+  log "custom agents.yaml in effect (no alissa-managed marker) — using it verbatim, ALISSA_AGENT_MODEL ignored"
+else
+  BASE_CMD="claude --dangerously-skip-permissions --permission-mode acceptEdits"
```
alissa-app, line 130: [minor] `BASE_CMD` hard-codes a second copy of the baked profile's command (`claude --dangerously-skip-permissions --permission-mode acceptEdits`). The rewrite reconstructs the whole `command:` from this constant instead of appending `--model` to the line already in `agents.yaml`, so the two copies must be kept in sync by hand. If a future change edits the flags in the baked `agents.yaml` but not this constant, boot will silently revert them — the `alissa-managed:` marker is still present, so the line is regenerated from `BASE_CMD`. The hard-coding is what makes the rewrite idempotent, so keep it, but add a comment here cross-referencing `agents.yaml` (or derive the base by stripping any existing `--model …` from the file's own command line) so the coupling is visible. Fix now or track as a follow-up task.

docker/claude/entrypoint.sh
```diff
@@ -100,6 +100,61 @@ alissa auth login --token "${ALISSA_API_TOKEN}" >/dev/null 2>&1 \
...
+for ln in open(path).read().splitlines(keepends=True):
+    m = re.match(r'^(\s*)command:\s', ln)
+    if m and not seen:
+        out.append(f"{m.group(1)}command: {cmd}\n")
+        seen = True
+    else:
+        out.append(ln)
+if not seen:
+    sys.exit("no `command:` line found in agents.yaml")
+open(path, "w").writelines(out)
```
alissa-app, line 153: [minor] The rewrite writes the new `command:` line but never re-parses the result, and the `|| die` only fires if python raises. Since the model value is passed through verbatim with no allowlist (by design), a value with YAML-significant or whitespace characters (e.g. a newline, or `opus --some-flag`) yields a malformed or flag-injected command that this step still reports as success — the breakage surfaces only later when the worker loads `agents.yaml`. The env is operator-controlled so this is not a privilege issue, but a cheap `yaml.safe_load(...)` round-trip check before `die` would convert a late, confusing worker failure into a clear boot-time one.

docker/claude/agents.yaml
```diff
@@ -9,8 +9,16 @@
...
-# To pin a model or tweak flags, override the command below (edit this file
-# before build, or mount your own agents.yaml over it at runtime).
+# To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default: opus) —
+# the entrypoint rewrites the `command:` below with `--model "$ALISSA_AGENT_MODEL"`
+# at container boot (set it to `default` or empty to omit the flag and inherit the
+# account default). To change flags instead, mount your own agents.yaml over this
+# path at runtime; a mounted file lacks the `alissa-managed:` marker below and is
+# used verbatim, so ALISSA_AGENT_MODEL does not touch it (your command wins).
+#
+# alissa-managed: the entrypoint regenerates the `command:` line of this default
```
alissa-app, line 19: [nit] Precedence keys off the literal substring `alissa-managed:` appearing anywhere in the file. A user who copies this baked file (marker included) and mounts the copy would have it rewritten by step 2d — contradicting the README's "a mounted file … always wins." Worth either a one-line caveat in the precedence docs ("remove this marker if you copy the baked file") or matching the marker more specifically.

alissa-app (approved):
## Review — TASK-18873090 (origin TASK-1843563761) · round 1

**Verdict: approve.** The change delivers the origin task's definition-of-done cleanly: `ALISSA_AGENT_MODEL` (default `opus`) is resolved at boot (entrypoint step 2d) into the baked `agents.yaml` claude `command:` as `--model <value>`, `default`/empty opts out, and a runtime-mounted custom file (no `alissa-managed:` marker) is left verbatim so the mounted command wins. Step 2d correctly runs before the worker starts, the rewrite is idempotent (base is reconstructed from a constant, not appended), and I reproduced the acceptance table locally against the real `agents.yaml` (default → `--model opus`; `claude-opus-4-8` → verbatim; `default`/empty → no flag; double-run → single flag). No `[blocker]`/`[major]` findings.

Only two `[minor]` and one `[nit]` — all non-gating, safe as follow-ups. Scope is container-tier only (4 files under `docker/claude/`); no Python/reviewloop changes, matching the PR's claim.

### Dimension report
- **Correctness:** meets origin DoD; ordering, opt-out, precedence, idempotency all verified. No entrypoint tests added, but the repo has none and CI covers only the Python package — a pre-existing gap, not introduced here.
- **Coding practices:** matches repo idiom (numbered step comments, `log`/`die`, python-not-sed rationale); commit trailers, branch, task-link all correct. One `[minor]`: `BASE_CMD` duplicates the baked command.
- **Security:** no shell injection (model passed as argv into a single-quoted heredoc). Verbatim/no-allowlist passthrough is an operator-trust-boundary choice, not a privilege escalation. One `[minor]`: rewritten YAML isn't re-validated.
- **Performance:** one-time boot cost, single-file rewrite. Nothing found.
- **Side-effects / blast radius:** container-tier only, no contract/consumer changes. One `[nit]` on the marker-substring precedence edge.

The minors are worth folding in or tracking; none block the merge.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Agreed — the hard-coded `BASE_CMD` was a second copy of the baked flags that could silently revert them on boot. Fixed by **deriving** the base from the file's own `command:` line: the rewrite strips any existing `--model …` and re-appends the resolved pin, so the base flags now live in exactly one place (`agents.yaml`) and a future flag change there is honored for free. The strip also keeps it idempotent (re-running with a new model replaces the pin rather than stacking). Pushing this fix next.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Good call on converting a late worker failure into a clear boot-time one. I did **not** use `yaml.safe_load` for the round-trip, though: the entrypoint runs the system `python3` (python:3.12-slim), where PyYAML is not installed — importing it would either crash the step or add a dependency just for validation. Instead the injection vector is closed directly at the source: `ALISSA_AGENT_MODEL` is rejected at boot (`die`, clear message) if it contains whitespace, so a value like `opus --dangerous` or one with a newline can no longer produce a malformed/flag-injected `command:`. A real alias/id is a single whitespace-free token, so this is a sanity guard, not an allowlist. Pushing next.

docker/claude/agents.yaml
RHDZMOTA: [triage:pursue] Fair — keying on the substring means a copied-then-mounted file (marker included) would still be rewritten, contradicting "mounted always wins". Took the documentation route from your two options: added an explicit caveat in both `agents.yaml` (next to the marker) and the README precedence section — "if you build your mounted file by copying the baked default, delete the `alissa-managed:` marker line". Kept the substring match (rather than a more specific regex) so the marker stays greppable and human-obvious; the caveat makes the one sharp edge explicit. Pushing next.

RHDZMOTA:
**Note on the round-1 improvement commit (`ea2d3f2`).** After the round-1 **approve** landed on `42d205d`, I posted CR8 triage replies and pushed `ea2d3f2` folding all three non-blocking comments (derive base command from `agents.yaml` instead of a hard-coded `BASE_CMD`; reject a whitespace-bearing `ALISSA_AGENT_MODEL` at boot; document deleting the `alissa-managed:` marker when copying the baked file). Convergence merged the approved head (`42d205d`) ~29s before that push arrived, so `ea2d3f2` is **orphaned on the branch and not in `main`**.

The merged code fully satisfies issue #26's acceptance criteria (verified: unset→`--model opus`, `claude-opus-4-8`→exact, `default`/empty→no flag, mounted custom→verbatim). The orphaned improvements are non-blocking (`[minor]`/`[nit]`, reviewer sanctioned a follow-up) and are tracked in **TASK-1476573014** with the ready-made diff at `ea2d3f2`.

PR #21, opened 2026-07-22
TASK-430061608: re-pin REVIEWLOOP_VERSION → 0.8.0 (session sweep to the deployed container)

RHDZMOTA:
Re-pins the reviewer container to reviewloop **0.8.0** so the deployed daemon picks up the fixes that merged to main in PR #19. One-line change to `docker/claude/Dockerfile` (`ARG REVIEWLOOP_VERSION` 0.6.0 → 0.8.0), same release pattern as the prior re-pin PRs. Verified `alissa-tools-github-reviewloop==0.8.0` is published on PyPI (latest) before pinning.

Supersedes draft PR #14 (which pinned 0.7.0 and should be closed, not merged).

## Release content since 0.7.0

- **Search-independent session sweep** — the daemon now enumerates and reaps idle `review-*` sessions directly instead of relying on search, fixing the live session-accumulation in the deployed container.
- **Interrupt-safe `run_forever` sleep** — the loop sleep now wakes promptly on interrupt instead of blocking shutdown for the full interval.

## Operator follow-through

1. Rebuild and redeploy the reviewer container from this Dockerfile.
2. Observe the sweep reaping the accumulated idle `review-*` sessions after the redeployed daemon's first cycles.

## Verification

All four repo scripts green locally: `tests-unit.sh` (111 passed), `tests-coverage.sh`, `check-style.sh`, `check-types.sh`. Nothing outside the Dockerfile pin changed (entrypoint hub-sync is PR #15, separate branch).

Closes #20

Alissa-Task: TASK-430061608

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #31, opened 2026-07-23
entrypoint: stop shadowing library config defaults (pass-through-when-unset)

### LINKED ISSUE #30: entrypoint: stop shadowing library config defaults (pass-through-when-unset)
## Problem / Scope

The env-driven entrypoint injects EVERY config key into the generated config with its own hardcoded fallback, e.g.:

```sh
--argjson cap "${ALISSA_ROUND_CAP:-3}"
```

This **shadows library defaults**: reviewloop 0.11.0 raised the `round_cap` default to 10, but every env-driven container boot still writes `round_cap: 3` unless the operator happens to set `ALISSA_ROUND_CAP`. Library defaults are dead code in the container. (Same defect class just bit the devloop's `attempt_cap`, silently keeping a capped issue parked after an upgrade.)

## Change (preferred: pass-through-when-unset)

Build the generated config so a key is **included only when its env var is set**; when unset, omit the key entirely and let the daemon's library default apply. Precedence contract: env var > library default — no hidden entrypoint layer. If pass-through is genuinely unsafe for a specific key, argue it in the PR and byte-sync that fallback to the library default with a pinning test instead.

## Acceptance criteria

- [ ] Boot with NO optional config env vars set → effective daemon config equals library defaults (specifically: `round_cap` resolves to 10 without `ALISSA_ROUND_CAP`), verified/tested
- [ ] Each env var still overrides when set (spot-tested)
- [ ] Required inputs (repos allowlist etc.) keep their current fatal-if-missing behavior — this change is about OPTIONAL keys only
- [ ] docker/claude/README documents the precedence and drops any stale fallback-value claims
- [ ] Repo checks green

## Out of scope

Python package changes, version bumps, agents.yaml/model handling (just landed in #28 — don't touch), any config *semantics*.

Note: a mirror issue exists in fahera-mx/alissa-github-develop-daemon (held behind its in-flight entrypoint lane). Keep the precedence contract and README language consistent, but do not touch that repo from this branch.

---
Alissa-Task: TASK-1799365990

RHDZMOTA:
## Summary

The env-driven container entrypoint injected **every** reviewloop config key with its own hardcoded fallback (`--argjson cap "${ALISSA_ROUND_CAP:-3}"`, …), which **shadowed the daemon library's own defaults**: a library that raised a default (e.g. `round_cap` → 10) could never take effect in a container unless the operator happened to set the matching env var. Library defaults were dead code in the container.

This implements the issue's preferred **pass-through-when-unset** model, with the precedence contract **env var > library default, no hidden entrypoint layer**.

## What changed

- **`docker/claude/reviewloop-config.sh`** (new) — the `reviewloop.config.json` renderer, extracted so it is unit-testable standalone.
  - **Pass-through** (optional tuning knobs): `poll_interval`, `round_cap` are emitted **only when their env var is set**. Unset ⇒ the key is omitted ⇒ the daemon library applies its own current default.
  - **Structural** (container constants, always emitted, argued exceptions): `on_missing_hub=add` and `agent_profile=claude`. Pass-through is unsafe for these — the library default `on_missing_hub=skip` would make a fresh volume review nothing, and `agent_profile` must name a profile the baked [`agents.yaml`](docker/claude/agents.yaml) ships (`claude`). Both stay explicit and pinned by tests, and remain overridable via their env var.
- **`docker/claude/entrypoint.sh`** — sources the renderer and calls it in place of the inline `jq` block.
- **`docker/claude/Dockerfile`** — empties the `ARG` defaults for the pass-through knobs (`ALISSA_POLL_INTERVAL`, `ALISSA_ROUND_CAP`) so an unset knob is genuinely unset (a baked concrete value would re-introduce the shadow); COPYs the new script. Structural ARGs (`add`, `claude`) kept.
- **`docker/claude/tests-entrypoint-config.sh`** (new) — asserts pass-through / override / structural behavior, and cross-checks that omitted keys resolve to the **real library default** through `Config.build` — dynamically, so it tracks the pinned release (`round_cap` → **3**, `poll_interval` → **60** at the pinned `REVIEWLOOP_VERSION` 0.8.0).
- **`.github/workflows/check-entrypoint.yaml`** (new) — runs the test in CI.
- **`docker/claude/README.md`** — documents the `env var > library default` precedence and drops the stale baked fallback values from the config table.

## Acceptance criteria

- [x] Boot with NO optional config env vars set → effective config equals library defaults (`round_cap` resolves to the pinned library default — **3** at `REVIEWLOOP_VERSION` 0.8.0 — without `ALISSA_ROUND_CAP`) — cross-checked dynamically against the pinned release in `tests-entrypoint-config.sh`.
- [x] Each env var still overrides when set — spot-tested (`round_cap`, `poll_interval`, and the structural keys).
- [x] Required inputs (repos allowlist) keep their fatal-if-missing behavior — untouched; change is OPTIONAL keys only.
- [x] `docker/claude/README` documents precedence and drops stale fallback-value claims.
- [x] Repo checks green — the entrypoint config test passes locally (12/12, incl. the dynamic library-default cross-check) and, after the round-1 rebase, `check-entrypoint.yaml` and the rest of CI now build the merge-ref and run on the PR; no Python package changes (unit/coverage/style/types unaffected).

## Out of scope (untouched)

Python package, version bumps, `agents.yaml`/model handling (#28), config semantics. The mirror issue in `alissa-github-develop-daemon` is not touched from this branch; the precedence contract and README language are kept consistent with it.

## Round-1 review fixes (request_changes → this update)

- **[blocker] stale/conflicting branch** — rebased onto `main`, keeping the newer `REVIEWLOOP_VERSION=0.8.0` pin (0.8.0 also defaults `round_cap=3`) and resolving the Dockerfile/README/entrypoint conflicts. The PR is now mergeable so GitHub can build the merge-ref and CI runs.
- **[major] wrong README round_cap default** — `README.md` config table corrected "currently 10" → **"currently 3"** to match the pinned 0.8.0 default and this PR's own `Dockerfile` comment (10 lands only in 0.11.0, out of scope). The same drift was corrected in this PR body and the commit message.
- **[minor] CI cross-check installed the source tree** — `check-entrypoint.yaml` now installs the pinned PyPI release `alissa-tools-github-reviewloop==${REVIEWLOOP_VERSION}` (parsed from the Dockerfile), so the library-default cross-check validates the library the image actually ships, not the repo source tree.
- **[nit] non-numeric override aborts with a raw jq error** — `[triage:ignore]`: not a regression (same failure class as the `--argjson` code it replaced) and the knobs are operator-set build-args, not untrusted input; a `die` would pull validation into the deliberately dependency-free, standalone-testable renderer. Reasoned on the thread; happy to file a follow-up.

---

Closes #30

**Alissa tasks** (canonical cross-actor record):
- Origin: `TASK-1799365990`
- Implementation: `TASK-664459042`

### DIFF
--- a/.github/workflows/check-entrypoint.yaml
+++ b/.github/workflows/check-entrypoint.yaml
@@ -1,0 +1,36 @@
+name: Check Container Entrypoint Config Rendering
+on: [pull_request]
+
+# The env-driven entrypoint renders reviewloop.config.json with a
+# pass-through-when-unset contract (optional tuning keys are omitted so the
+# daemon library default applies; structural keys stay pinned). This check
+# exercises that renderer directly — see docker/claude/tests-entrypoint-config.sh.
+jobs:
+  entrypoint-config-check:
+    name: Entrypoint config renderer
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Repo
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: '3.12'
+      - name: Install jq
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
+      - name: Install daemon package (for the library-default cross-check)
+        # Install the SAME PyPI release the image ships (REVIEWLOOP_VERSION pinned
+        # in the Dockerfile), not the repo source tree, so the cross-check
+        # validates the library the container actually runs and would catch a
+        # pin/default divergence (e.g. a README default that drifts from the pin).
+        run: |
+          python -m pip install --upgrade pip
+          version="$(grep -oP '^ARG REVIEWLOOP_VERSION=\K[0-9.]+' docker/claude/Dockerfile)"
+          echo "Installing alissa-tools-github-reviewloop==${version} (pinned in Dockerfile)"
+          python -m pip install "alissa-tools-github-reviewloop==${version}"
+      - name: Execute Entrypoint Config Tests
+        run: bash docker/claude/tests-entrypoint-config.sh
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -155,14 +155,33 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_REVIEW_REPOS` | *(required if no manifest mounted)* | allowlist as one `\|`-separated string (see below) |
 | `ALISSA_WORKSPACE` | `alissa-review` | workspace name in the generated manifest |
 | `ALISSA_REVIEW_SKILLS` | `alissa-code-workspace\|alissa-code-review` | skills installed into every reviewer session (manifest `skills:`), `\|`-separated |
-| `ALISSA_POLL_INTERVAL` | `60` | seconds between polls (≥10) |
-| `ALISSA_ROUND_CAP` | `3` | CR9 round cap |
-| `ALISSA_AGENT_PROFILE` | `claude` | agent the worker launches |
+| `ALISSA_POLL_INTERVAL` | *daemon default* (currently 60) | seconds between polls (≥10); **pass-through** — unset ⇒ library default |
+| `ALISSA_ROUND_CAP` | *daemon default* (currently 3) | CR9 round cap; **pass-through** — unset ⇒ library default |
+| `ALISSA_AGENT_PROFILE` | `claude` | agent the worker launches (must name a profile in `agents.yaml`) |
 | `ALISSA_AGENT_MODEL` | `opus` | model pinned into the reviewer's claude command (see [Pinning the reviewer model](#pinning-the-reviewer-model)); `default` or empty omits the pin |
 | `ALISSA_ON_MISSING_HUB` | `add` | `add` hub-ifies on demand; `skip` to require a mounted workspace |
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
+
+#### Config precedence: env var > daemon library default
+
+The optional tuning knobs `ALISSA_POLL_INTERVAL` and `ALISSA_ROUND_CAP` are
+**pass-through**: their build `ARG` default is empty, and when they are unset the
+entrypoint **omits the key entirely** from the generated `reviewloop.config.json`
+so the daemon library applies its own current default. There is no hidden
+entrypoint fallback layer that would shadow it — set the env var to override,
+leave it unset to inherit the library default (which is why the "default" column
+above says *daemon default* rather than a baked number, and why upgrading the
+daemon can change these without a Dockerfile edit). The parenthetical values are
+the library defaults at the pinned `REVIEWLOOP_VERSION`, informational only.
+
+`ALISSA_AGENT_PROFILE` and `ALISSA_ON_MISSING_HUB` are **not** pass-through: they
+are container constants the image requires. `agent_profile` must name a profile
+that the baked [`agents.yaml`](./agents.yaml) ships (`claude`), and
+`on_missing_hub` must be `add` for the self-contained hub-ify-on-demand model —
+the library default `skip` would make a fresh volume review nothing. Both are
+still overridable via their env var.
 
 ### The repos allowlist string
 
--- a/docker/claude/reviewloop-config.sh
+++ b/docker/claude/reviewloop-config.sh
@@ -1,0 +1,60 @@
+#!/usr/bin/env bash
+# =============================================================================
+# Render reviewloop.config.json from the environment.
+#
+# Precedence contract:  env var  >  daemon library default.  There is NO hidden
+# entrypoint layer in between for optional tuning knobs — the entrypoint used to
+# inject EVERY key with its own hardcoded fallback (e.g. `round_cap: 3`), which
+# SHADOWED the library's own default: a library that raised its default (say to
+# 10) could never take effect in the container, because the entrypoint always
+# wrote the old value unless the operator happened to set ALISSA_ROUND_CAP.
+#
+# So keys fall into two classes:
+#
+#   * PASS-THROUGH (optional tuning knobs) — emitted ONLY when the env var is
+#     set. When unset the key is omitted entirely and the daemon library applies
+#     its own current default. These are pure tuning values where the library is
+#     the authority: poll_interval, round_cap.
+#
+#   * STRUCTURAL (container constants) — always emitted with an explicit value
+#     the container requires, INDEPENDENT of the library default. Pass-through is
+#     unsafe here (see the per-key rationale below), so the value is byte-pinned
+#     and covered by tests-entrypoint-config.sh:
+#       - on_missing_hub = add     the container's whole model is self-contained
+#                                  hub-ify on demand; the library default is
+#                                  `skip`, which would make a fresh volume review
+#                                  nothing. (Bounded: `add` requires a non-empty
+#                                  repos allowlist, which env-driven mode always
+#                                  has.)
+#       - agent_profile  = claude  must name a profile that exists in the baked
+#                                  agents.yaml, which defines exactly `claude`;
+#                                  drifting to some future library default would
+#                                  select a profile the image does not ship.
+#
+# `repos` is required (env-driven mode only calls this with a non-empty
+# allowlist) and always emitted.
+#
+# Usage:  reviewloop-config.sh '<repos-json-array>'   # prints config JSON
+# Or source it and call render_reviewloop_config '<repos-json-array>'.
+# =============================================================================
+set -euo pipefail
+
+render_reviewloop_config() {
+  local repos_json="$1"
+  # --arg (string) + tonumber for the numeric pass-through keys: an unset/empty
+  # env var yields "" and the key is dropped, so the library default wins.
+  jq -n \
+    --argjson repos "${repos_json}" \
+    --arg     hub    "${ALISSA_ON_MISSING_HUB:-add}" \
+    --arg     agent  "${ALISSA_AGENT_PROFILE:-claude}" \
+    --arg     poll   "${ALISSA_POLL_INTERVAL:-}" \
+    --arg     cap    "${ALISSA_ROUND_CAP:-}" \
+    '{ repos: $repos, on_missing_hub: $hub, agent_profile: $agent }
+     + (if $poll == "" then {} else { poll_interval: ($poll | tonumber) } end)
+     + (if $cap  == "" then {} else { round_cap:     ($cap  | tonumber) } end)'
+}
+
+# Direct execution renders to stdout; sourcing just defines the function.
+if [ "${BASH_SOURCE[0]}" = "${0}" ]; then
+  render_reviewloop_config "${1:?usage: reviewloop-config.sh <repos-json-array>}"
+fi
--- a/docker/claude/Dockerfile
+++ b/docker/claude/Dockerfile
@@ -93,8 +93,9 @@ RUN git config --system url."https://github.com/".insteadOf "git@github.com:" \
 
 # --- Entrypoint + optional firewall ------------------------------------------
 COPY entrypoint.sh /usr/local/bin/entrypoint.sh
+COPY reviewloop-config.sh /usr/local/bin/reviewloop-config.sh
 COPY init-firewall.sh /usr/local/bin/init-firewall.sh
-RUN chmod 0755 /usr/local/bin/entrypoint.sh /usr/local/bin/init-firewall.sh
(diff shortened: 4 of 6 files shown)


alissa-app:
- 2026-07-23 04:08:07 UTC — `review-alissa-github-review-daemon-pr31-r1-06078c` — round 1 of 10 — spawned
- 2026-07-23 04:31:50 UTC — `review-alissa-github-review-daemon-pr31-r2-ee42b2` — round 2 of 10 — spawned

docker/claude/README.md
```diff
@@ -123,14 +123,33 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_REVIEW_REPOS` | *(required if no manifest mounted)* | allowlist as one `\|`-separated string (see below) |
 | `ALISSA_WORKSPACE` | `alissa-review` | workspace name in the generated manifest |
 | `ALISSA_REVIEW_SKILLS` | `alissa-code-workspace\|alissa-code-review` | skills installed into every reviewer session (manifest `skills:`), `\|`-separated |
-| `ALISSA_POLL_INTERVAL` | `60` | seconds between polls (≥10) |
-| `ALISSA_ROUND_CAP` | `3` | CR9 round cap |
-| `ALISSA_AGENT_PROFILE` | `claude` | agent the worker launches |
+| `ALISSA_POLL_INTERVAL` | *daemon default* (currently 60) | seconds between polls (≥10); **pass-through** — unset ⇒ library default |
+| `ALISSA_ROUND_CAP` | *daemon default* (currently 10) | CR9 round cap; **pass-through** — unset ⇒ library default |
```
alissa-app: [major] round_cap's library default is not 10 for this image. The container pins `REVIEWLOOP_VERSION=0.6.0` (config `round_cap: int = 3`), `main` is already on 0.8.0 (also 3), and the raise to 10 landed only in 0.11.0 — which this PR does not adopt (version bumps are out of scope per the description). The prose at line 144 says this parenthetical is "the library defaults at the pinned `REVIEWLOOP_VERSION`", and `Dockerfile:188` in this same PR says "round_cap=3", so "currently 10" contradicts both the shipped default and your own Dockerfile comment. Operational risk: an operator who trusts "10" may remove their interim `ALISSA_ROUND_CAP=10` mitigation and silently regress the cap to 3. Fix: state "currently 3" to match the pinned version, or bump the pin to ≥0.11.0 (separate scope). (poll_interval "(currently 60)" is correct — 60 in both 0.6.0 and 0.8.0.)

.github/workflows/check-entrypoint.yaml
```diff
@@ -0,0 +1,30 @@
...
+        with:
+          python-version: '3.12'
+      - name: Install jq
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y jq
+      - name: Install daemon package (for the library-default cross-check)
+        run: |
+          python -m pip install --upgrade pip
+          python -m pip install -e ./alissa-tools-github-reviewloop
```
alissa-app: [minor] This installs the local source tree (`-e ./alissa-tools-github-reviewloop`), but the container ships the PyPI-pinned `alissa-tools-github-reviewloop==${REVIEWLOOP_VERSION}` (0.6.0 on this branch). So the "effective config equals library defaults" cross-check validates a *different* library than the image actually runs. Harmless today (both default round_cap=3), but it won't catch a pin/README divergence like the README:127 one, and would assert the wrong number if the source tree's default ever drifts from the pinned release. Consider installing `alissa-tools-github-reviewloop==${REVIEWLOOP_VERSION}` for the cross-check, or explicitly note that it validates the source tree, not the shipped image.

docker/claude/reviewloop-config.sh
```diff
@@ -0,0 +1,60 @@
...
+  # env var yields "" and the key is dropped, so the library default wins.
+  jq -n \
+    --argjson repos "${repos_json}" \
+    --arg     hub    "${ALISSA_ON_MISSING_HUB:-add}" \
+    --arg     agent  "${ALISSA_AGENT_PROFILE:-claude}" \
+    --arg     poll   "${ALISSA_POLL_INTERVAL:-}" \
+    --arg     cap    "${ALISSA_ROUND_CAP:-}" \
+    '{ repos: $repos, on_missing_hub: $hub, agent_profile: $agent }
+     + (if $poll == "" then {} else { poll_interval: ($poll | tonumber) } end)
+     + (if $cap  == "" then {} else { round_cap:     ($cap  | tonumber) } end)'
```
alissa-app, line 54: [nit] A non-numeric override (e.g. `ALISSA_ROUND_CAP=foo`) makes `tonumber` throw and, under `set -euo pipefail`, aborts the entrypoint with a raw jq error instead of the friendly `die()` used elsewhere in the entrypoint. Same failure class as the old `--argjson` code, so not a regression — optional: validate and `die` with a clear message before rendering.

alissa-app (changes requested):
## Reviewer verdict: REQUEST_CHANGES — round 1

Fresh-instance review (alissa-code-review CR1–CR9) of the pass-through-when-unset entrypoint change, judged against origin TASK-1799365990 / implementation TASK-664459042.

**The core mechanism is correct and well-tested.** I ran `docker/claude/tests-entrypoint-config.sh`: all 12 assertions pass, including the dynamic library-default cross-check. Verified end-to-end: empty build `ARG` → empty `ENV` → `${VAR:-}` yields `""` → jq drops the key → the library default applies (the test's empty-string case covers this). The extraction into `reviewloop-config.sh` is clean and the structural-vs-pass-through split is well argued.

Two items block approval.

### [blocker] Branch is stale and conflicting; naive merge would regress the version pin
`main` has already re-pinned `REVIEWLOOP_VERSION 0.6.0 → 0.8.0` (merged PR #21), but this branch (`docker/claude/Dockerfile:29`) still pins **0.6.0**, and the PR is `mergeable: CONFLICTING` / `mergeStateStatus: DIRTY` — `git merge-tree` shows real conflict markers in `Dockerfile`, `README.md`, and `entrypoint.sh`. Consequences:
- The PR cannot merge as-is.
- **No CI has run on it** (the reviews/status rollup is empty) — GitHub can't build the merge-ref for a conflicting PR, so the DoD's "repo checks green" and "the new CI check runs" are currently **unverifiable on GitHub**. The new test passes locally, but `check-entrypoint.yaml` has not executed in CI.
- Resolving the conflict carelessly could regress the pin 0.8.0 → 0.6.0.

Fix: rebase onto `main`, keep the newer `0.8.0` pin, resolve the three conflicts, and let CI go green before merge.

### [major] README documents the wrong round_cap library default (see inline on README:127)
`README.md:127` says round_cap "*daemon default* (currently 10)" and the prose at :144 ties that parenthetical to "the library defaults at the pinned `REVIEWLOOP_VERSION`". But the pinned version is 0.6.0 (default **3**), main is on 0.8.0 (default **3**), and round_cap became 10 only in **0.11.0**, which this PR explicitly does not adopt (version bumps out of scope). This PR's own `Dockerfile:188` comment correctly says "round_cap=3", so the README self-contradicts and re-introduces exactly the doc-vs-reality drift #30 set out to remove.

### Dimension report
- **Correctness:** Pass-through mechanism correct and tested (12/12, dynamic lib-default cross-check). Gated by the stale/conflicting branch [blocker] and the wrong README default [major]; CI unverified on the PR.
- **Coding practices:** Clean sibling-script extraction, good comments; commit carries co-author + Alissa-Task trailers; branch/PR name the task. One internal doc contradiction (README:127 vs Dockerfile:188) [major].
- **Security:** Nothing found — `jq --arg/--argjson` escape inputs; no injection, secrets, or auth surface.
- **Performance:** Nothing found — trivial boot-time shell rendering.
- **Side-effects / blast radius:** Version-pin regression risk on conflict resolution [blocker]; the runtime behavior change is intended and correct; ARG→ENV empty-string bridge verified; structural keys (`on_missing_hub=add`, `agent_profile=claude`) correctly pinned and still overridable.

**Shortest path back to approve:** rebase onto `main` (keep the 0.8.0 pin), fix the README round_cap line (state "currently 3", or bump the pin to ≥0.11.0 as separate scope), and let CI go green. The [minor]/[nit] inline are non-gating.

_Verdict of record: TASK-1999828717 (review task)._

docker/claude/README.md
RHDZMOTA: [triage:pursue] Correct. The pinned `REVIEWLOOP_VERSION` is 0.6.0 (→ 0.8.0 after the rebase below), and both default `round_cap=3`; the raise to 10 only lands in 0.11.0, which this PR does not adopt. So "currently 10" contradicts the shipped default and this PR's own `Dockerfile:188` ("round_cap=3"). Fixing README:127 to "currently 3". (`poll_interval` "currently 60" is correct in both 0.6.0 and 0.8.0 and stays.)

.github/workflows/check-entrypoint.yaml
RHDZMOTA: [triage:pursue] Fair. The `-e ./alissa-tools-github-reviewloop` install cross-checks against the repo source tree, not the PyPI release the image actually ships (`==${REVIEWLOOP_VERSION}`). Pinning the CI cross-check install to `alissa-tools-github-reviewloop==${REVIEWLOOP_VERSION}` (parsed from the Dockerfile) so it validates the shipped library and would catch a pin/default divergence, not just the source-tree default.

docker/claude/reviewloop-config.sh
RHDZMOTA: [triage:ignore] Acknowledged, leaving as-is. It is not a regression (same failure class as the `--argjson` code it replaced), and `ALISSA_ROUND_CAP`/`ALISSA_POLL_INTERVAL` are operator-set build-args/env, not untrusted input. A friendly `die` would require pulling validation/`die` into `reviewloop-config.sh`, which is deliberately dependency-free so it stays standalone unit-testable — out of scope for this pass-through change. Happy to file a follow-up if desired.

RHDZMOTA:
[triage:pursue] re: **[blocker]** stale/conflicting branch (review body, no inline thread). Rebasing `TASK-664459042-CONFIG-PASSTHROUGH` onto `main`, keeping the newer `REVIEWLOOP_VERSION=0.8.0` pin (0.8.0 also defaults `round_cap=3`, so no extra README change beyond the [major] fix) and resolving the Dockerfile/README/entrypoint conflicts. This lets GitHub build the merge-ref so `check-entrypoint.yaml` and the rest of CI actually run on the PR.

alissa-app (approved):
## Reviewer verdict: APPROVE — round 2

Re-review (alissa-code-review CR9) of the pass-through-when-unset entrypoint change, against origin TASK-1799365990 / implementation TASK-664459042. Round 1 was `request_changes` on 1 blocker + 1 major (2 non-gating items). All three `[triage:pursue]` fixes hold, the one `[triage:ignore]` is accepted, and the new delta is clean across all five dimensions.

- **[blocker] stale/conflicting branch, pin-regression risk** — `[triage:pursue]` → **fixed**. Branch rebased onto `main`; `docker/claude/Dockerfile:29` now pins `REVIEWLOOP_VERSION=0.8.0` (matches `main`, no 0.6.0 regression). PR is `mergeable: MERGEABLE` / `mergeStateStatus: CLEAN`, and **all five CI checks pass on the merge-ref**, including the new `Entrypoint config renderer` (check-entrypoint.yaml) — the DoD's "repo checks green / new CI check runs", previously unverifiable, is now satisfied on GitHub.
- **[major] README round_cap default "currently 10"** — `[triage:pursue]` → **fixed**. `README.md:159` now reads *daemon default* (currently 3), consistent with the 0.8.0 library default and this PR's own `Dockerfile:188` comment. `poll_interval` "currently 60" is correct for 0.8.0 and retained.
- **[minor] CI cross-check installs the local source tree** — `[triage:pursue]` → **fixed**. `check-entrypoint.yaml` now parses the pin from the Dockerfile (`grep -oP '^ARG REVIEWLOOP_VERSION=\K[0-9.]+'` → `0.8.0`, verified against the branch file) and installs `alissa-tools-github-reviewloop==${version}` from PyPI — the library the image actually ships, so a future pin/default divergence would be caught.
- **[nit] `tonumber` aborts on a non-numeric override** — `[triage:ignore]` → **accepted, closed**. Reasoning holds: not a regression (same failure class as the replaced `--argjson`), the knobs are operator-set build-args/env rather than untrusted input, and `reviewloop-config.sh` is deliberately dependency-free so it stays standalone unit-testable. No new argument to re-raise.

### Dimension report (new delta)
- **Correctness:** Nothing found. Renderer logic unchanged and still correct (empty ENV → key dropped → library default; structural keys always emitted). The CI `Config.build` cross-check dynamically confirms omitted `round_cap`/`poll_interval` resolve to the pinned-0.8.0 library defaults, and it passed on the merge-ref.
- **Coding practices:** Nothing found. README precedence section and Dockerfile comments are clear and internally consistent; the CI install comment documents why it pins the shipped release.
- **Security:** Nothing found. `jq --arg` escapes inputs; no injection, secrets, or auth surface. CI grep reads a repo-controlled Dockerfile line.
- **Performance:** Nothing found. Boot-time shell rendering only.
- **Side-effects / blast radius:** Nothing found. Rebase kept the newer pin; structural keys (`on_missing_hub=add`, `agent_profile=claude`) remain pinned and overridable; the new CI job's only added coupling (Dockerfile line format, PyPI availability of the pin) holds today and is the intended, correct behavior.

No new blocker/major/minor/nit/question findings this round. Clean reconciliation + clean delta → **approve**.

_Verdict of record: TASK-1999828717 (review task)._

PR #37, opened 2026-07-26
TASK-1700707384: reviewer console sidecar `alissa-revloop-ui` (0.14.0)

RHDZMOTA:
Closes #36

**Alissa-Task (origin):** TASK-688130685
**Alissa-Task (implementation):** TASK-1700707384 — *Revloop console sidecar: `alissa-revloop-ui`* (downstream of the origin, soft dependency)

Ports the devloop worker console (`alissa.tools.github.devloop.webui`, its PR #38) to the reviewer daemon as `alissa.tools.github.revloop.webui` + the `alissa-revloop-ui` console script — stdlib-only (`ThreadingHTTPServer`, no third-party deps), copy-adapted to reviewer semantics per family precedent (module shapes copied; a shared-webui package stays a deferred lane).

## What's here

| area | shape |
| --- | --- |
| `webui/auth.py` | fail-closed `ALISSA_UI_PASSCODE`, constant-time compare, HMAC session cookie (passcode + per-boot nonce), sliding-window login throttle, CSRF token bound to the cookie |
| `webui/sources.py` | the read-only data layer: `poll_snapshots` (UI-1 reader from #35), spawn ledger, escalations, pings; `alissa tmux ls`; `/proc` pane-tree walk; `gh api rate_limit` (60s cache) + PyPI JSON (10m cache); the one retry-now `UPDATE` |
| `webui/sysinfo.py` | sample-free CPU% + summed RSS off `/proc`, vanished-PID tolerant; workspace disk usage |
| `webui/page.py` | one static page, studio tokens, parchment/ink light + glass-dark, ONE gold accent (drift chip), mono overlines, tabular-nums |
| `webui/server.py` | routing, security headers, body cap, auth/CSRF gating, the two audit-logged actions |
| `webui/__main__.py` | `alissa-revloop-ui`; resolves the *daemon's* config (no preflight, no `gh api user`) |
| `state.py` | console bridge: `read_spawns` / `read_escalations` / `read_pings` + `age_out_spawn` (the retry-now UPDATE) |

[...]

PR #33, opened 2026-07-23
TASK-1955630946: rename reviewloop -> revloop repo-wide; first revloop release (0.12.0)

RHDZMOTA:
## Rename `reviewloop` → `revloop` (three-letter loop family); first `revloop` release

Closes #32

**Alissa tasks**
- Origin: TASK-375600773
- Implementation: TASK-1955630946

### What changed (name-only; no behaviour altered)

| Was | Now |
| --- | --- |
| module `alissa.tools.github.reviewloop` | `alissa.tools.github.revloop` |
| dist `alissa-tools-github-reviewloop` | `alissa-tools-github-revloop` |
| CLI `alissa-reviewloop` | `alissa-revloop` |
| Dockerfile ARG `REVIEWLOOP_VERSION` | `REVLOOP_VERSION` (installs new dist, default `0.12.0`) |
| config file `reviewloop.config.json` | `revloop.config.json` |
| state dir `<ws>/.reviewloop/` | `<ws>/.revloop/` |

- `alissa-pr-review` **keeps its entry-point name** (it names the action, not the loop); only its home module path moved to `…github.revloop`.
- All five workflows updated (matrix path, publish `paths` filter + `--skip-existing`, wheel assert against the new package + `alissa-revloop --help`).
- `docker/claude` swept: ARG `REVLOOP_VERSION`, pip installs `alissa-tools-github-revloop`, entrypoint config/state/CLI references (merged pass-through-when-unset structure preserved).
- Version → **0.12.0** (continues the line monotonically; publishes on merge).
- README migration note added (frozen old dist, operator + cross-repo follow-ups).

### Acceptance criteria

- [x] Directory/module/imports renamed; **96 tests green** under `alissa.tools.github.revloop`; PEP 420 namespace intact
- [x] pyproject/dist metadata → `alissa-tools-github-revloop`; console scripts `alissa-revloop` + `alissa-pr-review`
- [x] All five workflows updated; publish targets new dist (`--skip-existing`); wheel assert against new package
- [x] docker/claude updated: ARG `REVLOOP_VERSION`, installs new dist, entrypoint swept; pass-through-when-unset structure not regressed
- [x] Repo-wide sweep: no stale `reviewloop` outside the migration note (grep proof below)
- [x] README migration note: old dist frozen w/ deprecation pointer; cross-repo devloop consumer flagged as a **separate follow-up in the other repo** (not touched here)
- [x] Version 0.12.0; publishes on merge
- [x] Repo checks green (table below)

### Results

[...]

PR #29, opened 2026-07-23
TASK-586326564: raise reviewloop round_cap default 3→10; release 0.11.0

RHDZMOTA:
## Summary

Operator decision (2026-07-23): both daemons should allow up to **10 loop rounds by default**. This raises the reviewloop's CR9 per-PR `round_cap` default from `3` to `10`.

## Changes

- **config.py** — dataclass default `round_cap: int = 3` → `10`; the from-raw fallback (`int(raw.get("round_cap", 3))`) is deduplicated to reference the class default (`cls.round_cap`), matching the existing `hub_template` pattern, so there is now a single source of truth. The `>= 1` validation and CLI/file overrides are unchanged.
- **README.md** — config table default `3` → `10`.
- **reviewloop.config.example.json** — `round_cap` example `3` → `10`.
- **tests** — the default assertion in `test_config_file_is_optional` is `3` → `10`; added `test_round_cap_default_is_ten` pinning both the dataclass default and the from-raw fallback (and confirming an explicit override still wins) so a silent revert fails.
- **version** — `0.10.0` → `0.11.0`.

## Verification

All repo checks green against the worktree source (`PYTHONPATH=src/main`):
- `tests-unit.sh` — 133 passed
- `tests-coverage.sh` — ran clean (config.py 97%)
- `check-style.sh` — clean
- `check-types.sh` — Success: no issues found in 14 source files

## Out of scope

Per the issue: no change to round-counting semantics, `loop.py`, or `docker/`. Note `docker/claude/README.md` still documents the `ALISSA_ROUND_CAP` build-arg default as `3`; that is a separate deployment default and is left untouched per the explicit out-of-scope — flagging it as a possible follow-up so the "both daemons, 10 rounds" intent isn't silently dropped at the container layer.

Closes #27

---
Origin task: TASK-708656813
Implementation task: TASK-586326564

PR #39, opened 2026-07-26
TASK-1578377135: wire the revloop console into the reviewer container (ALISSA_UI_ENABLED / ${PORT} / /healthz, 0.15.0)

### LINKED ISSUE #38: Revloop console: container wiring (ALISSA_UI_ENABLED, PORT, /healthz) + 0.15.0
Wire the revloop console sidecar (parent task's deliverable) into the reviewer container. **Reference implementation: the devloop's UI-3 lane — docker/claude/entrypoint.sh + README in fahera-mx/alissa-github-develop-daemon (merged PR #46); copy-adapt the same contract with ALISSA_UI_ENABLED / ALISSA_UI_PASSCODE / ${PORT:-8080} / /healthz.** Respect the reviewer entrypoint's pass-through-when-unset structure (merged PR #31) — do not regress it; add the UI vars to its env mapping. Reviewer-specific: the sidecar reads the REVLOOP state.db and review-* sessions; no other daemon behavior changes. Out of scope: webui module changes (parent lane owns it), any Python beyond the version file, the devloop repo.

## Definition of Done

The reviewer container serves the console when enabled: entrypoint starts alissa-revloop-ui under tini alongside worker+daemon when ALISSA_UI_ENABLED=1 (default 0) on 0.0.0.0:${PORT:-8080}; enabled-without-passcode dies at boot; /healthz liveness endpoint documented as the Railway healthcheck path; fail-visible sidecar supervision; Dockerfile EXPOSE + README env rows and Railway section (public networking, healthcheck, the passcode-is-the-only-gate warning); verified container boots with UI enabled (login + /healthz 200) and disabled (no listener), transcripts in the PR body; version 0.14.0 → 0.15.0 (publishes on merge). Merged through the adversarial loop.

## Acceptance criteria

- [ ] UI off by default; enabled+no-passcode dies at boot; enabled+passcode serves on ${PORT} with /healthz 200 — verified in container runs, transcripts in the PR body
- [ ] README env rows + Railway section cover enablement, passcode, healthcheck path, public-URL warning; pass-through-when-unset entrypoint structure not regressed
- [ ] Adversarial review loop converged on approve, or a recorded operator decision closed it

Alissa-Task: TASK-120677211

RHDZMOTA:
Wires the reviewer console sidecar (`alissa-revloop-ui`, shipped by the parent lane in #37) into the reviewer container, copy-adapting the devloop's UI-3 contract: `ALISSA_UI_ENABLED` / `ALISSA_UI_PASSCODE` / `${PORT:-8080}` / `/healthz`. Reviewer-side only — the sidecar reads this container's `.revloop/state.db` and the `review-*` sessions; no daemon behaviour changes, no webui module changes, no Python beyond the version file.

Closes #38

- Origin task: **TASK-120677211** (`alissa-github-review-daemon` develop lane)
- Implementation task: **TASK-1578377135** (this PR's work; soft-linked downstream of the origin — `create_downstream_task` 403s across actor partitions)

## What changed

| file | change |
| --- | --- |
| `docker/claude/entrypoint.sh` | **2d** — resolve the console gate *before* any bootstrap work; enabled-without-a-passcode **dies at boot** (fail-closed, fail-fast, same posture as the identity gates). Truthiness matches the sidecar's own `_env_flag` (`1/true/yes/on`). **4b** — start `alissa-revloop-ui --host 0.0.0.0 --port ${PORT:-8080} --workspace-root ${WORKSPACE_ROOT}` *after* the worker, backgrounded under tini, with a monitor subshell that logs a death **loudly** without tearing the container down. The shutdown trap silences the monitor first, then the sidecar. Steps renumbered: the model pin is now 2e. |
| `docker/claude/Dockerfile` | `EXPOSE 8080` (metadata only), the runtime-only rationale for the console knobs (passcode is a secret, `PORT` is platform-injected, the flag already defaults off), and `REVLOOP_VERSION` **0.12.0 → 0.15.0**. |
| `docker/claude/README.md` | Console env rows (`ALISSA_UI_ENABLED`, `ALISSA_UI_PASSCODE`, `PORT`, `ALISSA_UI_SECURE_COOKIE`), a `docker run` example, **The console on Railway** (public networking, `/healthz` as the Healthcheck Path, `Secure` cookie) and the passcode-is-the-only-gate warning; entrypoint step list now carries the gate + 4b. |
| `README.md` | The "container wiring is out of scope / this release ships the script only" note is now false — replaced with the enablement contract and a pointer; Tests section lists the two entrypoint suites. |
| `docker/claude/tests-entrypoint-ui.sh` | New suite (below), wired into `check-entrypoint.yaml`. |
| `.../revloop/version` | **0.14.0 → 0.15.0** (publishes on merge). |

**Pass-through-when-unset (#31) is not regressed**: the gate is a new preflight block and the sidecar launch is a new step — nothing in the config renderer or its callers changed, and the new suite additionally asserts end-to-end through a real boot that `poll_interval` / `round_cap` stay **omitted** while `on_missing_hub` stays pinned.

**Why the version pin moves in this PR**: the entrypoint now starts a console script that does not exist at the old pin (`0.12.0`); it landed in `0.14.0`. Pinning the version this PR publishes matches the precedent from #33 (rename + first release pinned in one PR); `check-entrypoint.yaml` already falls back to the source tree while the release is unpublished.

## Verification

This runner has **no docker daemon** (`docker: command not found`), so the wiring is verified by booting the **real `entrypoint.sh`** with the three CLIs it shells out to (`gh` / `alissa` / `alissa-revloop`) stubbed and the **console sidecar left real**, in a hermetic `env -i` sandbox. That covers every line this PR adds — the container adds image layers, not different entrypoint behaviour. The same suite runs in CI (`docker/claude/tests-entrypoint-ui.sh`), and it passes both against the installed console script and against the repo source tree.

<details><summary><b>A. <code>ALISSA_UI_ENABLED=1</code> with no passcode → dies at boot, before any bootstrap work</b></summary>

```
[entrypoint] gh authenticated as: alissa-app
[entrypoint] git credential helper configured (gh)
[entrypoint] alissa authenticated
[entrypoint] FATAL: ALISSA_UI_ENABLED is set but ALISSA_UI_PASSCODE is empty — the console is fail-closed on the passcode (it is the ONLY gate, and it rides the public URL once you enable networking). Set ALISSA_UI_PASSCODE, or unset ALISSA_UI_ENABLED.
exit=1
worker started: NO — died before any bootstrap work
```
</details>

<details><summary><b>B. Console disabled (default) → no listener</b></summary>

```
[entrypoint] reviewer console disabled (ALISSA_UI_ENABLED unset/off) — no listener
[entrypoint] alissa worker is running
-- listener check on :8099 --
(no listener on :8099)
-- curl --
(connection refused)
[entrypoint] shutting down
```
</details>

<details><summary><b>C. Console enabled with a passcode → serves on <code>0.0.0.0:${PORT}</code>, <code>/healthz</code> 200, login works</b></summary>

```
[entrypoint] reviewer console ENABLED (ALISSA_UI_ENABLED) — will serve on 0.0.0.0:8099 (passcode required)
[entrypoint] alissa worker is running
[entrypoint] starting reviewer console (alissa-revloop-ui) on 0.0.0.0:8099
alissa-revloop-ui 0.15.0 -- serving on http://0.0.0.0:8099 (passcode required); watching /tmp/tmp.EJJ7O9m0dt/workspace
-- GET /healthz --
HTTP/1.1 200 OK
{"ok": true, "version": "0.15.0"}
-- GET /api/state (no session) --      HTTP 401
-- POST /login wrong passcode --       HTTP 401
-- POST /login correct passcode --     HTTP 303
-- GET /api/state (with session) --    HTTP 200
   state.json keys: config, generated_at, header, inbox, log, pipeline, sessions, sparklines, tiles
-- off-loopback GET http://10.250.25.171:8099/healthz --  HTTP 200
-- SIGTERM --
10.250.25.171 - "GET /healthz HTTP/1.1" 200 -
[entrypoint] shutting down
-- listener after shutdown --
(connection refused)
```
The off-loopback 200 is the `0.0.0.0` bind (not the sidecar's own `127.0.0.1` default) — what lets a platform router reach it.
</details>

<details><summary><b>D. <code>bash docker/claude/tests-entrypoint-ui.sh</code> — the committed suite (4 boots, all green)</b></summary>

```
reviewer console entrypoint wiring — sidecar from: installed console script (/home/alissa/.local/bin/alissa-revloop-ui)
using port 37125

1. console OFF by default (ALISSA_UI_ENABLED unset)
  ok   logs the console as disabled
  ok   does not start the sidecar
  ok   no listener on ${PORT} when disabled
  ok   generated config omits unset poll_interval (pass-through intact)
  ok   generated config omits unset round_cap (pass-through intact)
  ok   generated config keeps the structural on_missing_hub
  ok   shuts down on SIGTERM

2. console ENABLED without a passcode -> dies at boot
  ok   exits non-zero (1)
  ok   fails with the fail-closed passcode message
  ok   dies before any worker/bootstrap work (fail-fast)

3. console ENABLED with a passcode -> serves on 0.0.0.0:${PORT}
  ok   logs the console as enabled
  ok   binds the platform ${PORT} on 0.0.0.0
  ok   console answers on ${PORT}
  ok   GET /healthz -> 200 (no passcode needed)
  ok   /healthz payload is {"ok": true, …}
  ok   /healthz reports the running version
  ok   GET /api/state without a session -> 401
  ok   POST /login with the wrong passcode -> 401 (no session)
  ok   POST /login with the right passcode -> 303 + session cookie
  ok   GET /api/state with the session -> 200
  ok   reachable off-loopback (10.250.25.171) -> 0.0.0.0 bind
  ok   shuts down on SIGTERM
  ok   orderly shutdown does not trip the sidecar alarm
  ok   console listener is gone after shutdown

4. sidecar killed under the container -> fail-VISIBLE, not fail-fatal
  ok   monitor logs the sidecar death loudly
  ok   container did NOT tear down with the sidecar
  ok   entrypoint (daemon + worker) still running
  ok   worker was left running

All reviewer-console entrypoint checks passed.
```
</details>

<details><summary><b>E. Repo suites (unchanged, still green at 0.15.0)</b></summary>

```
bash tests-unit.sh alissa-tools-github-revloop        -> 295 passed in 10.00s
bash tests-coverage.sh alissa-tools-github-revloop    -> TOTAL 1661 stmts, 88%
bash check-style.sh alissa-tools-github-revloop       -> clean
bash check-types.sh alissa-tools-github-revloop       -> Success: no issues found in 28 source files
bash docker/claude/tests-entrypoint-config.sh         -> ALL PASS (incl. library-default cross-check: round_cap=10 poll_interval=60)
```
</details>

## Acceptance criteria

- [x] UI off by default; enabled+no-passcode dies at boot; enabled+passcode serves on `${PORT}` with `/healthz` 200 — transcripts A/B/C/D above
- [x] README env rows + Railway section cover enablement, passcode, healthcheck path, public-URL warning; pass-through-when-unset entrypoint structure not regressed (asserted in suite 1)
- [ ] Adversarial review loop converged on approve

## Notes for the reviewer

- Two stale facts were corrected while touching their lines: the `ALISSA_ROUND_CAP` "currently 3" default (it has been **10** since #29) in both the README table and the Dockerfile comment, and the `--build-arg REVLOOP_VERSION=0.2.0` build example. Say the word if you'd rather they moved to their own lane.
- The fail-visible monitor polls the sidecar pid every 30s, so suite case 4 waits out two windows (~40s); the CI step carries `timeout-minutes: 5`.


alissa-app:
- 2026-07-26 23:23:11 UTC — `review-alissa-github-review-daemon-pr39-r1-8a46bb` — round 1 of 10 — spawned

docker/claude/Dockerfile
```diff
@@ -25,8 +25,11 @@
 # claude-code need Node >= 18 — both satisfied.
 FROM python:3.12-slim-bookworm
 
-# Version of the daemon to install from PyPI. Bump per release.
-ARG REVLOOP_VERSION=0.12.0
+# Version of the daemon to install from PyPI. Bump per release. 0.15.0 is the
+# release this repo publishes on merge; it is the first version whose console
+# script `alissa-revloop-ui` the entrypoint starts (0.14.0 added the script,
+# this release wires it into the container), so the pin must not go below it.
+ARG REVLOOP_VERSION=0.15.0
```
alissa-app, line 32: [major] **The definition-of-done's "verified in container runs" clause is unmet, and nothing in CI builds this image — so this pin is the one line in the PR that no test ever executes.**

Two gaps stack here:

1. `0.15.0` does not exist on PyPI while this PR is open — it publishes *on merge*, via `package-publish.yaml`'s `pull_request: closed` trigger. So `docker build` from this commit fails outright at `pip install alissa-tools-github-revloop==0.15.0`. The PR cites #33 as precedent and `check-entrypoint.yaml` has an explicit source-tree fallback for exactly this, so the window is short and understood.
2. But this repo has **no docker-build workflow at all** (`.github/workflows/` is style / types / tests / wheel / entrypoint / publish). Combined with (1), nothing ever proves the image builds — not before merge, not after publish. A pin that stays broken (publish fails, or a later bump goes below the floor this comment warns about) surfaces on a Railway deploy, not in CI.

**I am waiving this in the verdict** — the runner has no docker daemon, the PR says so plainly, the committed suite exercises every added line (I re-ran it green), and the wheel check plus the `alissa-revloop-ui` console-script entry point in `setup.py` cover the assumption that actually matters. Criterion `c1` is manual, so the container-run call stays with the human validator.

Flagging it rather than dropping it so `c1`'s validator sees that the clause was substituted, not satisfied. The durable fix is a docker-build check — the orcloop repo already carries a task for one (TASK-287205231); the same shape here would make the next pin bump self-verifying. Worth registering as a follow-up rather than widening this PR.

docker/claude/entrypoint.sh
```diff
@@ -108,7 +109,35 @@ alissa auth login --token "${ALISSA_API_TOKEN}" >/dev/null 2>&1 \
...
+# so `ALISSA_UI_ENABLED=true` behaves like `=1`; anything else (incl. unset) is off.
+# -----------------------------------------------------------------------------
+UI_ENABLED=0
+case "$(printf '%s' "${ALISSA_UI_ENABLED:-0}" | tr '[:upper:]' '[:lower:]')" in
+  1|true|yes|on) UI_ENABLED=1 ;;
+esac
+UI_PORT="${PORT:-8080}"
+if [ "${UI_ENABLED}" = "1" ]; then
+  [ -n "${ALISSA_UI_PASSCODE:-}" ] \
+    || die "ALISSA_UI_ENABLED is set but ALISSA_UI_PASSCODE is empty — the console is fail-closed on the passcode (it is the ONLY gate, and it rides the public URL once you enable networking). Set ALISSA_UI_PASSCODE, or unset ALISSA_UI_ENABLED."
```
alissa-app, line 133: [minor] **The gate fail-fasts on the passcode but not on the sidecar existing — so the misconfiguration the Dockerfile explicitly warns about is the one this gate can't catch.**

This block's own comment sets the bar: *"validate its one hard requirement now — before any bootstrap work — so a misconfigured deploy fails FAST."* The passcode is one hard requirement; `alissa-revloop-ui` being installed is the other. `Dockerfile:31` says *"the pin must not go below it"* — a below-floor `REVLOOP_VERSION` is precisely the deploy that should die here and doesn't.

I booted this path to see what actually happens (`ALISSA_UI_ENABLED=1`, real entrypoint, no `alissa-revloop-ui` on `PATH`):

```
[entrypoint] reviewer console ENABLED (ALISSA_UI_ENABLED) — will serve on 0.0.0.0:39411 (passcode required)
[entrypoint] alissa worker is running
[entrypoint] starting reviewer console (alissa-revloop-ui) on 0.0.0.0:39411
entrypoint.sh: line 420: alissa-revloop-ui: command not found
[entrypoint] starting alissa-revloop over /tmp/.../workspace
[entrypoint] WARN: reviewer console (alissa-revloop-ui, pid 3721) EXITED — the console URL is dead until the next redeploy
```

Credit where due: the container **survives** and the warning does fire, so the fail-visible design holds even for a binary that was never installed — that's why this is `[minor]` and not worse. But the operator gets two confident "console ENABLED / starting" lines, a bare `command not found` that doesn't look like an entrypoint diagnostic, and then a ~32 s gap before the WARN — all after the worker is already up. Boot-time detection is one line next to the passcode check:

```sh
command -v alissa-revloop-ui >/dev/null 2>&1 \
  || die "ALISSA_UI_ENABLED is set but alissa-revloop-ui is not installed — REVLOOP_VERSION must be >= 0.15.0 (currently pinned: see Dockerfile)."
```

That makes 2d's fail-fast claim complete, and turns a below-floor pin into a clear boot failure instead of a healthy-looking container with a dead URL.

docker/claude/README.md
```diff
@@ -331,6 +419,12 @@ volumes:
    that never completes.
 4. Start `alissa worker --daemon`, wait until it reports running (the daemon only
    *warns* if the worker is absent, so ordering matters).
-5. Run `alissa-revloop` in the foreground; stop the worker on `SIGTERM`/`SIGINT`.
-
-`tini` is PID 1 to reap the tmux/node/claude child fan-out.
+4b. When `ALISSA_UI_ENABLED` is set, start the reviewer console
```
alissa-app, line 422: [minor] **`4b.` isn't a list marker, so the step this PR exists to document doesn't render as a step.**

An ordered-list marker is digits followed by `.` or `)`; `4b.` is neither, so it becomes a lazy-continuation line of item 4's paragraph. Confirmed against GitHub's own renderer (`POST /markdown`) on this exact text:

```html
<ol start="4">
<li>Start the worker, wait until it reports running (the daemon only
<em>warns</em> if the worker is absent, so ordering matters).
4b. When ALISSA_UI_ENABLED is set, start the reviewer console
backgrounded on 0.0.0.0.</li>
<li>Run alissa-revloop in the foreground; stop the worker.</li>
</ol>
```

Step 5 survives as its own `<li>`; **4b is swallowed into step 4** and reads as run-on prose. Since the whole point of this section is that the console is a distinct step between the worker and the daemon, that's the one line readers most need to see as a step.

Renumbering to `5.`/`6.` (matching `entrypoint.sh`, which already calls the daemon step 5) is the clean fix; a nested sub-bullet under 4 also works if you want to keep the a/b framing. Same applies to the `4b` in the entrypoint header comment only insofar as you want the two lists to agree — that one is a shell comment and renders fine as-is.

docker/claude/README.md
```diff
@@ -183,6 +189,85 @@ that the baked [`agents.yaml`](./agents.yaml) ships (`claude`), and
...
+#### The console on Railway
+
+1. Set `ALISSA_UI_ENABLED=1` and a **long, random** `ALISSA_UI_PASSCODE` as
+   service variables (both runtime-only — the passcode is a secret, so it must
+   not be an ARG). Enabling without a passcode makes the container **die at
+   boot** with a clear message, before the worker or daemon start.
+2. Under the service's **Settings → Networking**, enable a public domain (or a
+   TCP proxy). Railway injects `PORT` and routes the public URL to it; the
+   entrypoint already binds `0.0.0.0:${PORT:-8080}`.
+3. Set the **Healthcheck Path** to **`/healthz`** — the console's unauthenticated
```
alissa-app, line 256: [minor] **This step puts an unauthenticated exact-version endpoint on the public internet — liveness doesn't need the version.**

`/healthz` returns `{"ok": true, "version": "0.15.0"}` before any auth check (`webui/server.py`, first branch of `do_GET`). Correct for a healthcheck: Railway's prober has no session. But this section is also the one telling operators to enable a **public domain**, so the version ships to anyone who can reach the URL — free fingerprinting against a surface whose sole gate is one shared passcode, and whose dashboard carries kill and retry-now.

`{"ok": true}` satisfies every healthcheck; the version is already on the dashboard for an authenticated operator, and the deployed pin is knowable from the image anyway. So this is low-severity — it's the gratuitousness that makes it worth a line, not the secrecy of the number.

Two things keep this out of blocker territory: the exposure is opt-in and this PR warns about it twice, prominently; and the fix is **not in this diff** — `webui/server.py` belongs to the parent lane, which TASK-1578377135 lists as out of scope. So please **don't widen this PR for it**. `[triage:later]` with a follow-up task against the console module is the right disposition; if the version is deliberate (deploy-verification after redeploy, say), `[triage:ignore]` with that reasoning closes it just as well.

docker/claude/tests-entrypoint-ui.sh
```diff
@@ -0,0 +1,338 @@
...
+# up/down, a foreground daemon); everything about the console is the real thing.
+# -----------------------------------------------------------------------------
+TMPROOT="$(mktemp -d)"
+BIN="${TMPROOT}/bin"; mkdir -p "${BIN}"
+FAKE_HOME="${TMPROOT}/home"; mkdir -p "${FAKE_HOME}/.config/alissa"
+WORKSPACE="${TMPROOT}/workspace"; mkdir -p "${WORKSPACE}"
+MARKERS="${TMPROOT}/markers"; mkdir -p "${MARKERS}"
+cp "${HERE}/agents.yaml" "${FAKE_HOME}/.config/alissa/agents.yaml"
+
+cleanup() { rm -rf "${TMPROOT}"; }
```
alissa-app, line 63: [nit] `cleanup()` removes the temp tree but doesn't stop what the suite started.

Under `set -euo pipefail`, an unexpected failure outside an assertion (the `cp` on line 60, a `jq`/`curl` absence, a `pgrep` surprise) aborts before the matching `stop_entrypoint`. `rm -rf "${TMPROOT}"` then deletes the sandbox out from under a still-running entrypoint, its stubbed worker, and a **real** console sidecar holding a socket — all reparented and left behind. CI doesn't care (the runner is torn down), so this is purely local-dev ergonomics: `PORT_FREE` is re-randomised per run, so it won't even wedge the next invocation.

Tracking the pids makes it self-cleaning:

```sh
EP_PIDS=()
cleanup() { for p in "${EP_PIDS[@]:-}"; do kill -TERM "$p" 2>/dev/null || true; done; rm -rf "${TMPROOT}"; }
```
with `EP_PIDS+=("${EP_PID}")` at the end of `run_entrypoint`.

Genuinely discretionary — the suite is well-built and this is the only thing I'd change in 338 lines.

alissa-app (approved):
## Review verdict: **approve** (round 1 of cap 3)

Reviewer: fresh instance, alissa-code-review CR1–CR9. Reviewed against **TASK-1578377135**'s definition-of-done (the origin task named in the PR body, TASK-120677211, is not reachable from the reviewer actor partition). Review task: **TASK-1010923661**.

**1 major (waived, justified below) · 4 minor · 1 nit · 0 blockers.** The wiring is correct and, unusually for a container PR, independently checkable — I re-ran the committed suite from a clean worktree of `24e2e9d` and got **28/28 green**, and separately probed a path the suite does not cover.

### What I verified myself (not taken from the PR body)

- `bash docker/claude/tests-entrypoint-ui.sh` from a throwaway worktree at the PR head — all four boots pass, sidecar resolved from the repo source tree.
- **Missing-sidecar path (untested by the suite):** booted the real entrypoint with `ALISSA_UI_ENABLED=1` and no `alissa-revloop-ui` on `PATH`. Result: `command not found`, the container **survives**, and the monitor fires `WARN: … EXITED` after ~32 s. The fail-visible/not-fail-fatal claim holds even for a binary that was never installed.
- Every env var the READMEs document is real and read where claimed: `ALISSA_UI_PASSCODE` and `ALISSA_UI_SECURE_COOKIE` in `webui/__main__.py` / `webui/auth.py`, `/healthz` in `webui/server.py`, and the console-script entry point `alissa-revloop-ui=…webui.__main__:main` in `setup.py`.
- Both stale-fact corrections check out: `round_cap: int = 10` in `config.py:66`, and `0.15.0` matches the version file this PR ships.
- Anchors in the new cross-links resolve (`#reviewer-console-alissa-revloop-ui`, `#the-console-on-railway`, and the long console anchor).
- The `check-entrypoint` job **rename** does not orphan a required status check — `mergeStateStatus` is `CLEAN` with no expected/pending checks. Flagging that I checked, since renaming a job's `name:` is the usual way this bites.
- CI is green on all five checks, including the new console-wiring step.

### Dimension report (CR4 — every dimension, "nothing found" written out)

- **Correctness:** the gate (2d), launch (4b), fail-visible supervision and shutdown ordering all behave as documented; `pass-through-when-unset` is asserted end-to-end through a real boot, so #31 is demonstrably not regressed. One **[major]** on the definition-of-done's *container-run* verification clause (waived), one **[minor]** on the gate not preflighting the sidecar binary.
- **Coding practices:** matches the entrypoint's established idiom — numbered step blocks with rationale, `log`/`die`, `${VAR:-default}`, `|| true` guards under `set -euo pipefail`. Branch `TASK-1578377135-UI-CONTAINER` and the PR title name the task; the PR is attached to the task as evidence. The two drive-by doc corrections are disclosed in the PR body rather than smuggled. One **[nit]** on test-harness cleanup.
- **Security:** the passcode never reaches `argv` or the logs (inherited env only; the log says "passcode required", not the value); the fail-closed gate is verified by a real boot; the README's "login throttle and CSRF" claim is backed by `LoginThrottle` and session-bound CSRF tokens in `webui/auth.py`; no secrets in the diff; the `0.0.0.0` + public-URL exposure is warned about twice, in both READMEs, including the kill/retry-now blast radius. One **[minor]** on unauthenticated version disclosure at `/healthz`.
- **Performance:** nothing found. The supervision costs one subshell waking every 30 s; the console is opt-in, reads only this container's local `state.db` plus the tmux session list, and adds no GitHub API budget.
- **Side-effects / blast radius:** nothing found. No contract, schema or exported-type changes; `EXPOSE` is metadata only; the version bump publishes through the existing paths-filtered `package-publish.yaml` on merge; the renamed CI job does not orphan a required check (verified above); no other hub in the workspace consumes any of this.

### Waived major

**[major] The definition-of-done's "verified in container runs" was not performed, and no CI job ever builds the image** (inline on `Dockerfile:32`). **Waived**, with reasoning: the runner genuinely has no docker daemon and the PR discloses this rather than papering over it; the substitute exercises every line the diff adds and I reproduced it independently; the wheel check plus the declared console-script entry point cover the one image-level assumption that matters (`alissa-revloop-ui` lands on `PATH` at the new pin); and criterion `c1` is **manual**, so the human validator keeps the container-run decision rather than inheriting an approval that quietly redefined it. Recording it as a waived major rather than dropping it, so it stays visible to whoever validates `c1`.

### Minors and the nit

Four **[minor]** findings and one **[nit]**, all inline — none gates the merge:

1. `entrypoint.sh:133` — the gate doesn't preflight that `alissa-revloop-ui` exists.
2. `docker/claude/README.md:422` — `4b.` is not a list marker, so the newly documented step renders as trailing prose inside step 4.
3. `docker/claude/README.md:256` — `/healthz` discloses the exact running version, unauthenticated, on the public URL this section tells operators to enable.
4. `Dockerfile:32` — the pin/CI gap that accompanies the waived major.
5. `tests-entrypoint-ui.sh:63` — `cleanup()` leaks backgrounded entrypoints and real sidecars on an early abort.

Per CR5 a deferred `[minor]` should become a registered task rather than evaporate; #3 in particular belongs to the **parent lane** (`webui/server.py` is out of scope for this task), so it wants a follow-up task, not a widening of this PR.

### Summary

This PR wires the `alissa-revloop-ui` sidecar into the reviewer container behind an opt-in, fail-closed gate: off by default, dead at boot when enabled without a passcode, bound to `0.0.0.0:${PORT:-8080}` after the worker, and supervised fail-visibly so its death warns loudly without taking the daemon down. It satisfies the definition-of-done's behavioural clauses, and — the part that earns the approve — it ships a 338-line suite that boots the **real** entrypoint four times with only the CLIs stubbed, which I re-ran green and then pushed past its own coverage. The one substantive gap is that "verified in container runs" became "verified via the real entrypoint with stubs", forced by a runner with no docker; I'm waiving it with the reasoning above rather than bouncing a round over a constraint the implementer cannot resolve in this lane. **Approved** — the four minors and the nit are the implementer's to triage, and the deferred ones should land as tasks.

docker/claude/Dockerfile
RHDZMOTA: [triage:later] **Agreed, and thank you for waiving it with the reasoning intact rather than silently.** The clause really was *substituted*, not satisfied: no docker daemon on this runner, and — as you found — no image-build workflow anywhere in this repo, so the pin is the one line no test executes. I am not widening this PR for it (the approve is terminal; a post-approve commit gets orphaned by convergence, which is a lesson this lane has already paid for once).

Registered as **TASK-1456046886** — "CI: add a docker image build check to the reviewer daemon repo", shaped like the orcloop repo's TASK-287205231 you point at, and explicitly required to tolerate the release-PR window where the pinned version is not yet on PyPI (otherwise the check false-fails on every release PR, including the next one). That makes the next pin bump self-verifying, which is the durable fix rather than a one-off container run.

Leaving `c1`'s container-run decision with the human validator, as you intended.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:later] **Correct, and the sharpest of the five — the gate's own comment claims a fail-fast that it only half delivers.** Your boot trace is the argument: two confident "console ENABLED / starting" lines, a bare `command not found` that reads like shell noise rather than an entrypoint diagnostic, and then ~32s of silence before the WARN — all *after* the worker is up, i.e. after exactly the bootstrap work the 2d block exists to run in front of. And it is the failure mode `Dockerfile:31` names out loud (a pin below the console-script floor), so the gate not catching it is a real gap, not a hypothetical.

Deferred only because of ordering, not disagreement: the approve is terminal here, so pushing your one-liner now would orphan it on the branch instead of landing it in `main`. Registered as **TASK-443470187** with your `command -v` check verbatim as the acceptance criterion, alongside the other two implementer-side findings on this PR.

One refinement I'll carry into that task: the message should name the floor *and* the currently pinned value, since the operator reading it on Railway cannot see the Dockerfile.

docker/claude/README.md
RHDZMOTA: [triage:later] **Confirmed — thank you for rendering it through GitHub's own `POST /markdown` rather than eyeballing it.** `4b.` is not an ordered-list marker, so it becomes lazy continuation of item 4, and the one step this PR exists to document is the one that stops looking like a step. I copied the `4b` framing from the devloop README (same defect there, inherited rather than invented).

Going with your clean fix — renumber to `5.`/`6.` so the rendered list agrees with `entrypoint.sh`, which already calls the daemon step 5. The a/b framing buys nothing once the console is a permanent step. The shell-comment `4b` in the entrypoint header will move with it so the two lists stay in sync.

Folded into **TASK-443470187** (same reason as the sibling finding: post-approve pushes are orphaned by convergence).

docker/claude/README.md
RHDZMOTA: [triage:later] **Agreed on the disposition you proposed — this is the parent lane's, and I am not widening the PR for it.** The version in `/healthz` is not deliberate on my side: I documented the endpoint's existing payload, I did not choose it. `{"ok": true}` satisfies every healthcheck, the authenticated dashboard already shows the version, so the disclosure is gratuitous — exactly your framing, and the reason it is worth a line at all.

Registered against the console module as **TASK-2023500360**, scoped to `webui/server.py` (out of scope for TASK-1578377135, which is why it is a task and not a commit here).

Not touching this PR's README for it either: the section is honest about what the endpoint returns today, and it would go stale the moment the payload changes. The follow-up owns both sides.

docker/claude/tests-entrypoint-ui.sh
RHDZMOTA: [triage:later] **Fair, and worth doing — `rm -rf` on a sandbox that a live sidecar still has open is the kind of thing that only ever bites the person debugging the harness at 2am.** Precisely: an abort outside an assertion (the `cp`, a missing `jq`/`curl`, a `pgrep` surprise) skips `stop_entrypoint` and reparents a stubbed worker plus a *real* console holding a socket. CI does not care, which is why I am comfortable deferring rather than treating it as a correctness issue.

Taking your `EP_PIDS` array approach as-is into **TASK-443470187** — it is the smallest change that makes the suite self-cleaning, and `stop_entrypoint` already does the graceful-then-KILL dance the trap can reuse.

(Deferred for the same terminal-approve reason as the other implementer-side findings, not because I disagree.)

RHDZMOTA:
## Round 1 triage — approve accepted, nothing pushed

Verdict: **approve** (round 1 of 3), 1 waived major · 4 minors · 1 nit, all triaged on their threads above (one reply per finding).

**Deliberately no commits after the approve.** Convergence merges the *approved* head, so a post-approve push is orphaned on the branch rather than landing in `main` — this lane has paid for that lesson once already (#28). So every actionable finding became a registered task instead of a late commit; the merged tree is exactly the reviewed one, `24e2e9d`.

| finding | disposition | task |
| --- | --- | --- |
| [major] no docker-build check; the pin is untested (`Dockerfile:32`) | `[triage:later]` — waiver accepted, `c1` stays with the human validator | **TASK-1456046886** — docker image build check in CI, tolerant of the release-PR window |
| [minor] gate doesn't preflight `alissa-revloop-ui` (`entrypoint.sh:133`) | `[triage:later]` — agreed, the reviewer's `command -v` check is the acceptance criterion | **TASK-443470187** |
| [minor] `4b.` isn't a list marker (`README.md:422`) | `[triage:later]` — renumbering to 5./6. so the rendered list matches `entrypoint.sh` | **TASK-443470187** |
| [minor] `/healthz` discloses the exact version (`README.md:256`) | `[triage:later]` — parent lane's, `webui/server.py` is out of scope here | **TASK-2023500360** |
| [nit] harness `cleanup()` leaks processes (`tests-entrypoint-ui.sh:63`) | `[triage:later]` — taking the `EP_PIDS` approach as-is | **TASK-443470187** |

All three follow-ups are soft-linked downstream of **TASK-1578377135** and carry the finding ids, so nothing evaporates (CR5).

Thanks for re-running the suite from a clean worktree and then pushing *past* its coverage — the missing-sidecar boot is a case I did not write, and it is now the first line of TASK-443470187.

PR #41, opened 2026-07-28
TASK-1058266947: CI guard — fail revloop src PRs without a version bump (and bumps to a published version)

### LINKED ISSUE #40: CI guard: fail PRs that touch alissa-tools-github-revloop/src/** without bumping the version file
Repo: fahera-mx/alissa-github-review-daemon. Sibling of the same guard being added to the develop- and orchestrate-daemon repos, after the 2026-07-28 develop-daemon incident: source-touching PRs merged without a version bump make the merge-triggered "Python Package Publish" workflow a silent no-op (it reports success while publishing nothing), leaving main ahead of PyPI with no signal. This repo's history is clean (0.15.0 published on its merge), but the failure mode is identical.

## Required change

Add a `.github/workflows/` check (own workflow file, named in the repo's existing "Check ..." convention, same pull_request trigger pattern as the current check workflows):

1. Compute the PR's changed files (per the repo's existing workflow idiom).
2. If ANY changed file matches `alissa-tools-github-revloop/src/**` AND the version file `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version` is NOT among the changed files → FAIL with an actionable message: name the rule, the version-file path, and the silent-no-op consequence (cite the develop-daemon #62/#64 incident, 2026-07-28). Note this repo also carries a `version.py` beside the `version` file — the guard keys on whichever file is the source of truth for the published version (state which in the PR; if version.py is derived, keying on `version` alone is correct).
3. If the version file IS changed → verify the NEW version does not already exist on PyPI (GET https://pypi.org/pypi/alissa-tools-github-revloop/<version>/json; 404 = good). Confirmed-exists → FAIL; network error → pass with a visible warning (never block merges on a PyPI outage).
4. PRs touching only docs, docker/, workflows, or other non-src paths are unaffected (this PR itself must pass its own check).

## Verification expectations
- Dry-run demonstrations in the PR body: src-without-bump (fails), src-with-new-version (passes), docs-only (skips), bump-to-existing-version (fails).
- Repo check scripts green; no version bump needed for this PR (workflows only — consistent with the rule).

## Definition of Done

A pull_request check fails any PR that modifies alissa-tools-github-revloop/src/** without also changing the revloop version source-of-truth file, fails bumps to a version already on PyPI, warns-but-passes on PyPI network errors, and ignores non-src PRs — merged via a reviewed PR whose body demonstrates the dry-run cases and states the version/version.py source-of-truth decision.

## Acceptance criteria

- [ ] Workflow fails src-without-bump and bump-to-existing-PyPI-version; passes src-with-new-version and non-src PRs; PyPI network errors warn but pass — dry-run demonstrations in the PR body.
- [ ] Workflow follows the repo's existing check naming/trigger conventions and passes on its own PR; the version vs version.py source-of-truth decision is stated.
- [ ] The failure message names the version-file path and the silent-no-op consequence with the incident reference.

Autonomous-Scope: .github/workflows/**
Alissa-Task: TASK-1625720421

RHDZMOTA:
Closes #40

- Origin task: **TASK-1625720421**
- Implementation task: **TASK-1058266947**

## What this adds

`.github/workflows/check-version-bump.yaml` — a `pull_request` check (`Check Version Bump for Source Changes`) that closes the silent-no-op publish gap on the PR side:

1. Computes the PR's changed files from git (`merge-base(base.sha, head.sha)..head.sha`, `--no-renames` so both sides of a move appear). No pull-request API call, so the job keeps `permissions: contents: read` like the other check workflows.
2. **Fails** if anything *published* under `alissa-tools-github-revloop/**` changed and `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version` did not. `src/test/**` is exempt (nothing under it is packaged); repo-level paths — `README.md`, `docker/`, `.github/`, the `*.sh` check scripts — are outside the dist directory and never require a bump.
3. When the version file *did* change, **fails** if that version already exists on PyPI (`GET https://pypi.org/pypi/alissa-tools-github-revloop/<version>/json` → 200). 404 passes.
4. Anything the check cannot confirm — connection failure, timeout, 5xx, rate limit — emits a `::warning::` and **passes**. A PyPI outage never blocks a merge.
5. PRs that touch nothing published (docs, `docker/`, workflows — including this one — and test-only changes) skip both rules with a `::notice::`.
6. A deleted or empty version file fails with the same RULE + step-summary + annotation shape as every other failure, rather than a bare `head: cannot open`.

### Scope: the guarded subtree is the publish trigger, not `src/**`

Round 1 raised a `[major]` that this PR takes: the property worth enforcing is **"if merging this PR runs a publish, it must publish something."** `package-publish.yaml` triggers on all of `alissa-tools-github-revloop/**`, and `requirements.txt` (`install_requires`), `setup.py` (dist name, `entry_points`, `python_requires`), `README.md` (`long_description`) and `MANIFEST.in` (sdist contents) each change the artifact *and* start a publish — but a `src/**`-only prefix exempted all four, reaching exactly the failure this workflow exists to prevent.

**This is a deliberate, recorded deviation from the issue's literal `src/**`.** Every repo-level path the issue's acceptance criteria call out as unaffected — docs, `docker/`, workflows, the check scripts — stays exempt, and this PR still passes its own check. The only files that move from exempt to guarded are the four inside the distribution directory, and for each of them "unaffected" would be wrong on the issue's own terms. In the other direction, `src/test/**` is now exempt: it is not packaged, so a test-only change yields a byte-identical wheel and must not cost a release number.

Failure messages name the rule, the version-file path, the `twine upload --skip-existing` → "reports SUCCESS while publishing nothing" consequence, and the 2026-07-28 `alissa-github-develop-daemon` #62/#64 incident. They go to the job log, to `$GITHUB_STEP_SUMMARY`, and to a `::error file=<version file>::` annotation.

## Source-of-truth decision: `version`, not `version.py`

**The guard keys on the plain-text `version` file alone.** `version.py` is a *derived reader*, not a second source of truth:

- `setup.py` opens `src/main/alissa/tools/github/revloop/version` and passes its contents as `setup(version=...)` — that string is what lands on PyPI.
- `version.py`'s `Version.from_path()` opens the same file at import time (falling back to `0.0.0` if it is missing); it holds no version literal of its own.
- `MANIFEST.in` ships `version`, and `check-wheel.yaml` asserts the installed `version.value != '0.0.0'` — i.e. it asserts the *file* was packaged.

So editing `version.py` cannot change what gets published, and requiring it to change would be a false gate. Keying on `version` alone is correct. (`__init__.py` also carries a stale `__version__ = "0.1.0"` literal that nothing reads — untouched here, out of this PR's `.github/workflows/**` scope, and worth a separate cleanup.)

## Dry-run demonstrations

Executed against the **shipped** workflow file: a harness parses `check-version-bump.yaml`, extracts the two guard steps' actual `run:` blocks, and runs them under a GitHub-Actions-shaped environment (`RUNNER_TEMP` / `GITHUB_OUTPUT` / `GITHUB_STEP_SUMMARY`, the step `env:` maps with `${{ matrix.* }}` and `${{ steps.*.outputs.* }}` resolved, workflow-level `env`, and the `if:` condition evaluated). Only the "Compute Changed Files" step is stubbed with a fixture file list; everything after it is the real code path. PyPI calls are real (`0.15.0` is published → 200; `0.16.0` is not → 404).

| # | Case | Changed files | `version` | Result |
|---|------|---------------|-----------|--------|
| 1 | src without bump | `…/revloop/loop.py` | 0.15.0 | **FAIL** ✅ |
| 2 | src with new version | `…/loop.py`, `…/version` | 0.16.0 | PASS ✅ |
| 3 | **this PR itself** | `.github/workflows/check-version-bump.yaml` | 0.15.0 | PASS (skipped) ✅ |
| 4 | repo docs / docker only | `README.md`, `docker/claude/Dockerfile` | 0.15.0 | PASS (skipped) ✅ |
| 5 | bump to a published version | `…/loop.py`, `…/version` | 0.15.0 | **FAIL** ✅ |
| 6 | PyPI unreachable | `…/loop.py`, `…/version` | 0.16.0 | PASS + warning ✅ |
| 7 | test-only change, no bump | `…/src/test/…/test_directives.py` | 0.15.0 | PASS (exempt) ✅ |
| 8 | dist `requirements.txt`, no bump | `alissa-tools-github-revloop/requirements.txt` | 0.15.0 | **FAIL** ✅ |
| 9 | dist `setup.py` + dist `README.md`, no bump | `…/setup.py`, `…/README.md` | 0.15.0 | **FAIL** ✅ |
| 10 | dist `MANIFEST.in` with new version | `…/MANIFEST.in`, `…/version` | 0.16.0 | PASS ✅ |
| 11 | src file moved *out* of the dist tree, no bump | `…/loop.py`, `docs/loop.py` | 0.15.0 | **FAIL** ✅ |
| 12 | version file deleted | `…/loop.py`, `…/version` | *(deleted)* | **FAIL** ✅ |
| 13 | version file emptied | `…/loop.py`, `…/version` | *(empty)* | **FAIL** ✅ |

`13/13 cases matched expectations`

Cases 7–13 are the round-1 review's findings; case 7 **inverted** in round 2 (a test-only change no longer requires a release).

<details>
<summary>Case 1 — src change without a version bump (FAIL)</summary>

```
changed files: ['alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py']
version file : 0.15.0
--- Enforce Version Bump for Published Changes (exit 1) ---
RULE: a pull request that modifies alissa-tools-github-revloop/** must also change alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version
      (alissa-tools-github-revloop/src/test/** is exempt — nothing under it is packaged)

This pull request changes published files under alissa-tools-github-revloop/** but leaves
alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version at 0.15.0.

Why that blocks the merge: on merge, "Python Package Publish" uploads with
`twine upload --skip-existing`. Re-uploading a version PyPI already has is skipped
silently, so the workflow reports SUCCESS while publishing nothing. main ends up
ahead of PyPI with no signal, and everything pinned to 0.15.0 keeps
running the old code. That is the 2026-07-28 alissa-github-develop-daemon incident
(its PRs #62 and #64).

Fix: bump alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version to the next unpublished version
(the version.py beside it is a derived reader — editing that file alone changes nothing),
or move the change out of alissa-tools-github-revloop/** if this pull request must not release.
::error file=alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version::Published files under alissa-tools-github-revloop/ changed without bumping alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version — the merge-triggered publish would be a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump the version file.
```
</details>

<details>
<summary>Cases 8 &amp; 9 — dist metadata without a bump (FAIL) — the round-1 [major]</summary>

```
changed files: ['alissa-tools-github-revloop/requirements.txt']
version file : 0.15.0
--- Enforce Version Bump for Published Changes (exit 1) ---
RULE: a pull request that modifies alissa-tools-github-revloop/** must also change alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version
      (alissa-tools-github-revloop/src/test/** is exempt — nothing under it is packaged)

This pull request changes published files under alissa-tools-github-revloop/** but leaves
alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version at 0.15.0.

Why that blocks the merge: on merge, "Python Package Publish" uploads with
`twine upload --skip-existing`. Re-uploading a version PyPI already has is skipped
silently, so the workflow reports SUCCESS while publishing nothing. main ends up
ahead of PyPI with no signal, and everything pinned to 0.15.0 keeps
running the old code. That is the 2026-07-28 alissa-github-develop-daemon incident
(its PRs #62 and #64).

Fix: bump alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version to the next unpublished version
(the version.py beside it is a derived reader — editing that file alone changes nothing),
or move the change out of alissa-tools-github-revloop/** if this pull request must not release.
::error file=alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version::Published files under alissa-tools-github-revloop/ changed without bumping alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version — the merge-triggered publish would be a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump the version file.
changed files: ['alissa-tools-github-revloop/setup.py', 'alissa-tools-github-revloop/README.md']
version file : 0.15.0
--- Enforce Version Bump for Published Changes (exit 1) ---
RULE: a pull request that modifies alissa-tools-github-revloop/** must also change alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version
      (alissa-tools-github-revloop/src/test/** is exempt — nothing under it is packaged)

This pull request changes published files under alissa-tools-github-revloop/** but leaves
alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version at 0.15.0.

Why that blocks the merge: on merge, "Python Package Publish" uploads with
`twine upload --skip-existing`. Re-uploading a version PyPI already has is skipped
silently, so the workflow reports SUCCESS while publishing nothing. main ends up
ahead of PyPI with no signal, and everything pinned to 0.15.0 keeps
running the old code. That is the 2026-07-28 alissa-github-develop-daemon incident
(its PRs #62 and #64).

Fix: bump alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version to the next unpublished version
(the version.py beside it is a derived reader — editing that file alone changes nothing),
or move the change out of alissa-tools-github-revloop/** if this pull request must not release.
::error file=alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version::Published files under alissa-tools-github-revloop/ changed without bumping alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version — the merge-triggered publish would be a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump the version file.
```
</details>

<details>
<summary>Cases 4 &amp; 7 — repo docs / test-only (PASS, rule does not apply)</summary>

```
changed files: ['README.md', 'docker/claude/Dockerfile']
version file : 0.15.0
--- Enforce Version Bump for Published Changes (exit 0) ---
::notice::No published files changed under alissa-tools-github-revloop/ (alissa-tools-github-revloop/src/test/ is exempt — nothing under it is packaged) — the version-bump rule does not apply to this pull request.
[step skipped by `if:`] Verify New Version Is Unpublished
changed files: ['alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_directives.py']
version file : 0.15.0
--- Enforce Version Bump for Published Changes (exit 0) ---
::notice::No published files changed under alissa-tools-github-revloop/ (alissa-tools-github-revloop/src/test/ is exempt — nothing under it is packaged) — the version-bump rule does not apply to this pull request.
[step skipped by `if:`] Verify New Version Is Unpublished
```
</details>

<details>
<summary>Case 5 — bump to a version already on PyPI (FAIL)</summary>

```
changed files: ['alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py', 'alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version']
version file : 0.15.0
--- Enforce Version Bump for Published Changes (exit 0) ---
::notice::alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version is part of this pull request — the merge will publish a new release.
--- Verify New Version Is Unpublished (exit 1) ---
Asking PyPI whether alissa-tools-github-revloop 0.15.0 already exists: https://pypi.org/pypi/alissa-tools-github-revloop/0.15.0/json
RULE: the version in alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version must not already exist on PyPI.

alissa-tools-github-revloop 0.15.0 is ALREADY published (https://pypi.org/pypi/alissa-tools-github-revloop/0.15.0/json responded HTTP 200).

Why that blocks the merge: on merge, "Python Package Publish" uploads with
`twine upload --skip-existing`, which skips the duplicate and reports SUCCESS
while publishing nothing. main ends up ahead of PyPI with no signal — the
2026-07-28 alissa-github-develop-daemon incident (its PRs #62 and #64).

Fix: bump alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version to a version PyPI does not have yet.
::error file=alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version::alissa-tools-github-revloop 0.15.0 is already on PyPI — merging would make the publish a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version again.
```
</details>

<details>
<summary>Case 6 — PyPI unreachable (PASS with a warning)</summary>

```
changed files: ['alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py', 'alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version']
version file : 0.16.0
--- Enforce Version Bump for Published Changes (exit 0) ---
::notice::alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version is part of this pull request — the merge will publish a new release.
--- Verify New Version Is Unpublished (exit 0) ---
Asking PyPI whether alissa-tools-github-revloop 0.16.0 already exists: https://pypi.invalid.localhost.example/pypi/alissa-tools-github-revloop/0.16.0/json
::warning::Could not reach PyPI (curl exited 6) at https://pypi.invalid.localhost.example/pypi/alissa-tools-github-revloop/0.16.0/json; could not confirm whether alissa-tools-github-revloop 0.16.0 is already published. Passing — a PyPI outage must not block a merge.
curl: (6) Could not resolve host: pypi.invalid.localhost.example
curl: (6) Could not resolve host: pypi.invalid.localhost.example
curl: (6) Could not resolve host: pypi.invalid.localhost.example
```
</details>

<details>
<summary>Cases 12 &amp; 13 — version file deleted / emptied (FAIL with the RULE shape)</summary>

```
changed files: ['alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py', 'alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version']
version file : <deleted>
--- Enforce Version Bump for Published Changes (exit 0) ---
::notice::alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version is part of this pull request — the merge will publish a new release.
--- Verify New Version Is Unpublished (exit 1) ---
RULE: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version must exist and hold the version this pull request publishes.

This pull request deletes (or moves away) alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version. setup.py opens that
exact path to set the distribution version, so the merge-triggered publish would
fail to build — and the version.py beside it would fall back to 0.0.0.

Fix: restore alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version holding the version this pull request publishes.
::error file=alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version::alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version is missing — setup.py reads it to set the published version. Restore it.
changed files: ['alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py', 'alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version']
version file : 
--- Enforce Version Bump for Published Changes (exit 0) ---
::notice::alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version is part of this pull request — the merge will publish a new release.
--- Verify New Version Is Unpublished (exit 1) ---
RULE: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version must exist and hold the version this pull request publishes.

The file is empty. setup.py reads it to set the distribution version, so the
merge-triggered publish would fail to build.

Fix: write the version this pull request publishes into alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version.
::error file=alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version::alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version is empty — it must hold the version this pull request publishes.

13/13 cases matched expectations
```
</details>

## Verification

- **The check ran for real on this PR and passed in 6s** ([run 30372698439](https://github.com/fahera-mx/alissa-github-review-daemon/actions/runs/30372698439), head `629df04`) — case 3 confirmed on GitHub Actions, not just in the harness:
  ```
  Changed files in this pull request (6348b5a..629df04):
    .github/workflows/check-version-bump.yaml
  ::notice::No published files changed under alissa-tools-github-revloop/ (alissa-tools-github-revloop/src/test/ is exempt — nothing under it is packaged) — the version-bump rule does not apply to this pull request.
  ```
  The job's `GITHUB_TOKEN` permissions resolved to `Contents: read` / `Metadata: read` only. (Round 1 ran the same way on `9335e42`: [run 30371283769](https://github.com/fahera-mx/alissa-github-review-daemon/actions/runs/30371283769).)
- `actionlint 1.7.7` clean on all six workflow files (`rc=0`).
- Repo check scripts green on this branch (unchanged Python, run for completeness): `tests-unit.sh` **295 passed**, `tests-coverage.sh` **88%**, `check-style.sh` clean, `check-types.sh` **"Success: no issues found in 28 source files"**.
- No version bump in this PR — it touches `.github/workflows/**` only, which is exactly what the new rule exempts (case 3 above).

## Round 1 → round 2

Reviewer `alissa-app` returned `request_changes` on 1 major, 2 minors, 2 nits. Triage replies are on each thread; all four code findings are taken:

| Finding | Triage | Change |
|---|---|---|
| `[major]` guarded prefix narrower than the publish trigger | `pursue` | `SRC_PREFIX` → `DIST_PREFIX: alissa-tools-github-revloop/` (cases 8–10) |
| `[minor]` `src/test/**` should not force a release | `pursue` | `TEST_PREFIX` exemption ordered before the dist match (case 7) |
| `[minor]` required-status-check / up-to-date branch protection | `answer` | Repo-admin, outside `Autonomous-Scope`; tracked as **TASK-568945951** |
| `[nit]` rename detection hides a file moved out of the dist tree | `pursue` | `git diff --no-renames` (case 11) |
| `[nit]` deleted/empty version file fails opaquely | `pursue` | RULE + summary + `::error file=` shape for both (cases 12–13) |
| `[nit]` no monotonicity check | `later` | Outside issue #40's rule and not the silent-no-op mode; tracked as **TASK-1492477** |

## Remaining judgement calls

- **Scope stays inside `.github/workflows/**`** per the issue's `Autonomous-Scope`, so the rule is documented in the workflow's header comment rather than in the README.
- **`PYPI_API_BASE` is a workflow-level env var** rather than a hardcoded URL, so the network-error branch is exercisable (case 6). It defaults to `https://pypi.org/pypi`.
- **This guard cannot see `main` move underneath an open PR** (a `pull_request` job only re-runs on `synchronize`), so two PRs bumping to the same version both pass. Closing that needs "require branches to be up to date" — TASK-568945951.

### DIFF
--- a/.github/workflows/check-version-bump.yaml
+++ b/.github/workflows/check-version-bump.yaml
@@ -1,0 +1,183 @@
+name: Check Version Bump for Source Changes
+on: [pull_request]
+
+# "Python Package Publish" runs on merge and uploads with `twine upload
+# --skip-existing`. That flag is what keeps a duplicate release from erroring —
+# but it also means a merged source change that forgot to bump the version is
+# published as NOTHING while the workflow reports success. main silently ends up
+# ahead of PyPI and every pinned consumer keeps running the old code. That is the
+# 2026-07-28 alissa-github-develop-daemon incident (its PRs #62 and #64).
+#
+# This check closes the gap on the pull_request side: touch the dist source, bump
+# the version; bump the version, make it one PyPI does not already have.
+#
+# SOURCE OF TRUTH: the plain-text `version` file, NOT the `version.py` beside it.
+# setup.py reads `version` to set the distribution version, and version.py is a
+# derived reader (Version.from_path() opens the same file at runtime, falling back
+# to 0.0.0). Editing version.py cannot change what gets published, so the guard
+# keys on `version` alone.
+
+# Least privilege: the changed-file set is computed from git history rather than
+# the pull-request API, so a read-only checkout is all this job ever needs (same
+# posture as the other check workflows).
+permissions:
+  contents: read
+
+env:
+  # PyPI JSON API root. Overridable so the guard can be dry-run against an
+  # unreachable host to exercise the network-error path.
+  PYPI_API_BASE: https://pypi.org/pypi
+
+jobs:
+  version-bump-check:
+    strategy:
+      matrix:
+        # The distribution name on PyPI is the directory name, so `path` doubles
+        # as the dist name. `version_file` is spelled out because the module path
+        # under src/main is not derivable from the dist name.
+        include:
+          - path: alissa-tools-github-revloop
+            version_file: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version
+    name: Version Bump Check for ${{ matrix.path }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Clone Repo
+        uses: actions/checkout@v4
+        with:
+          submodules: true
+          # Full history: the diff below needs the merge base of the pull request.
+          fetch-depth: 0
+      - name: Compute Changed Files
+        id: changes
+        env:
+          BASE_SHA: ${{ github.event.pull_request.base.sha }}
+          HEAD_SHA: ${{ github.event.pull_request.head.sha }}
+        run: |
+          set -euo pipefail
+
+          if ! git cat-file -e "${BASE_SHA}^{commit}" 2>/dev/null; then
+            echo "Base commit ${BASE_SHA} is missing from the local clone; fetching it."
+            git fetch --no-tags origin "${BASE_SHA}"
+          fi
+
+          # Diff against the merge base (equivalent to BASE...HEAD), which is the
+          # file set GitHub shows in the pull request's "Files changed" tab —
+          # unrelated commits that landed on main meanwhile are not attributed
+          # to this pull request.
+          merge_base="$(git merge-base "${BASE_SHA}" "${HEAD_SHA}")"
+          changed="${RUNNER_TEMP}/changed-files.txt"
+          git diff --name-only "${merge_base}" "${HEAD_SHA}" > "${changed}"
+
+          echo "changed_files=${changed}" >> "${GITHUB_OUTPUT}"
+          echo "Changed files in this pull request (${merge_base}..${HEAD_SHA}):"
+          sed 's/^/  /' "${changed}"
+      - name: Enforce Version Bump for Source Changes
+        id: guard
+        env:
+          CHANGED_FILES: ${{ steps.changes.outputs.changed_files }}
+          SRC_PREFIX: ${{ matrix.path }}/src/
+          VERSION_FILE: ${{ matrix.version_file }}
+        run: |
+          set -euo pipefail
+
+          src_changed=false
+          version_changed=false
+          while IFS= read -r file; do
+            if [ -z "${file}" ]; then
+              continue
+            fi
+            case "${file}" in
+              "${SRC_PREFIX}"*) src_changed=true ;;
+            esac
+            if [ "${file}" = "${VERSION_FILE}" ]; then
+              version_changed=true
+            fi
+          done < "${CHANGED_FILES}"
+
+          if [ "${src_changed}" = "false" ]; then
+            echo "src_changed=false" >> "${GITHUB_OUTPUT}"
+            echo "version_changed=false" >> "${GITHUB_OUTPUT}"
+            echo "::notice::No changes under ${SRC_PREFIX} — the version-bump rule does not apply to this pull request."
+            exit 0
+          fi
+          echo "src_changed=true" >> "${GITHUB_OUTPUT}"
+
+          if [ "${version_changed}" = "false" ]; then
+            current="$(head -n 1 "${VERSION_FILE}" 2>/dev/null | tr -d '[:space:]' || true)"
+            {
+              echo "RULE: a pull request that modifies ${SRC_PREFIX}** must also change ${VERSION_FILE}"
+              echo ""
+              echo "This pull request changes ${SRC_PREFIX}** but leaves ${VERSION_FILE} at ${current:-<unreadable>}."
+              echo ""
+              echo "Why that blocks the merge: on merge, \"Python Package Publish\" uploads with"
+              echo "\`twine upload --skip-existing\`. Re-uploading a version PyPI already has is skipped"
+              echo "silently, so the workflow reports SUCCESS while publishing nothing. main ends up"
+              echo "ahead of PyPI with no signal, and everything pinned to ${current:-that version} keeps"
+              echo "running the old code. That is the 2026-07-28 alissa-github-develop-daemon incident"
+              echo "(its PRs #62 and #64)."
+              echo ""
+              echo "Fix: bump ${VERSION_FILE} to the next unpublished version"
+              echo "(the version.py beside it is a derived reader — editing that file alone changes nothing),"
+              echo "or move the change out of ${SRC_PREFIX}** if this pull request must not release."
+            } | tee -a "${GITHUB_STEP_SUMMARY}"
+            echo "::error file=${VERSION_FILE}::Source under ${SRC_PREFIX} changed without bumping ${VERSION_FILE} — the merge-triggered publish would be a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump the version file."
+            exit 1
+          fi
+
+          echo "version_changed=true" >> "${GITHUB_OUTPUT}"
+          echo "::notice::${VERSION_FILE} is part of this pull request — the merge will publish a new release."
+      - name: Verify New Version Is Unpublished
+        if: steps.guard.outputs.version_changed == 'true'
+        env:
+          DIST_NAME: ${{ matrix.path }}
+          VERSION_FILE: ${{ matrix.version_file }}
+        run: |
+          set -euo pipefail
+
+          version="$(head -n 1 "${VERSION_FILE}" | tr -d '[:space:]')"
+          if [ -z "${version}" ]; then
+            echo "::error file=${VERSION_FILE}::${VERSION_FILE} is empty — it must hold the version this pull request publishes."
+            exit 1
+          fi
+          case "${version}" in
+            *[!A-Za-z0-9.+!_-]*)
+              echo "::error file=${VERSION_FILE}::Version '${version}' contains characters that cannot appear in a PEP 440 version."
+              exit 1
+              ;;
+          esac
+
+          url="${PYPI_API_BASE}/${DIST_NAME}/${version}/json"
+          echo "Asking PyPI whether ${DIST_NAME} ${version} already exists: ${url}"
+
+          # A confirmed 200 is the only outcome that fails this step. Anything the
+          # check cannot confirm — connection failure, timeout, 5xx, rate limit —
+          # warns and passes: a PyPI outage must never block a merge.
+          if http_code="$(curl -sS -o /dev/null -w '%{http_code}' --max-time 30 --retry 2 --retry-delay 3 "${url}")"; then
+            case "${http_code}" in
+              404)
+                echo "::notice::${DIST_NAME} ${version} is not on PyPI (HTTP 404) — this pull request can publish it."
+                ;;
+              200)
+                {
+                  echo "RULE: the version in ${VERSION_FILE} must not already exist on PyPI."
+                  echo ""
+                  echo "${DIST_NAME} ${version} is ALREADY published (${url} responded HTTP 200)."
+                  echo ""
+                  echo "Why that blocks the merge: on merge, \"Python Package Publish\" uploads with"
+                  echo "\`twine upload --skip-existing\`, which skips the duplicate and reports SUCCESS"
+                  echo "while publishing nothing. main ends up ahead of PyPI with no signal — the"
+                  echo "2026-07-28 alissa-github-develop-daemon incident (its PRs #62 and #64)."
+                  echo ""
+                  echo "Fix: bump ${VERSION_FILE} to a version PyPI does not have yet."
+                } | tee -a "${GITHUB_STEP_SUMMARY}"
+                echo "::error file=${VERSION_FILE}::${DIST_NAME} ${version} is already on PyPI — merging would make the publish a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump ${VERSION_FILE} again."
+                exit 1
+                ;;
+              *)
+                echo "::warning::PyPI answered HTTP ${http_code} for ${url}; could not confirm whether ${DIST_NAME} ${version} is already published. Passing — a PyPI outage must not block a merge."
+                ;;
+            esac
+          else
+            rc=$?
+            echo "::warning::Could not reach PyPI (curl exited ${rc}) at ${url}; could not confirm whether ${DIST_NAME} ${version} is already published. Passing — a PyPI outage must not block a merge."
+          fi
--- a/.github/workflows/check-version-bump.yaml
+++ b/.github/workflows/check-version-bump.yaml
@@ -8,8 +8,21 @@ on: [pull_request]
 # ahead of PyPI and every pinned consumer keeps running the old code. That is the
 # 2026-07-28 alissa-github-develop-daemon incident (its PRs #62 and #64).
 #
-# This check closes the gap on the pull_request side: touch the dist source, bump
-# the version; bump the version, make it one PyPI does not already have.
+# This check closes the gap on the pull_request side: touch anything that gets
+# published, bump the version; bump the version, make it one PyPI does not already
+# have.
+#
+# SCOPE: the guarded prefix is the whole distribution directory — the same subtree
+# package-publish.yaml triggers on — minus src/test/. The property being enforced
+# is "if merging this pull request runs a publish, it must publish something", so
+# the guard has to cover every path that starts a publish AND changes the artifact:
+# requirements.txt (install_requires), setup.py (name, entry_points,
+# python_requires), README.md (long_description) and MANIFEST.in (sdist contents),
+# not just src/. src/test/ is exempt in the other direction: nothing under it is
+# packaged (setup.py packages src/main only, MANIFEST.in adds requirements.txt and
+# version), so a test-only change yields a byte-identical wheel and must not cost a
+# release number. Repo-level paths — README.md, docker/, .github/, the *.sh check
+# scripts — are outside the dist directory and never require a bump.
 #
 # SOURCE OF TRUTH: the plain-text `version` file, NOT the `version.py` beside it.
 # setup.py reads `version` to set the distribution version, and version.py is a
@@ -64,50 +77,61 @@ jobs:
           # file set GitHub shows in the pull request's "Files changed" tab —
           # unrelated commits that landed on main meanwhile are not attributed
           # to this pull request.
+          #
+          # --no-renames: with rename detection on, a file moved OUT of the dist
+          # directory is reported only at its destination, so the wheel changes
+          # while the guard never sees a dist path. Without it, both sides of a
+          # rename appear and the set is exactly "paths this pull request touches".
           merge_base="$(git merge-base "${BASE_SHA}" "${HEAD_SHA}")"
           changed="${RUNNER_TEMP}/changed-files.txt"
-          git diff --name-only "${merge_base}" "${HEAD_SHA}" > "${changed}"
+          git diff --no-renames --name-only "${merge_base}" "${HEAD_SHA}" > "${changed}"
 
           echo "changed_files=${changed}" >> "${GITHUB_OUTPUT}"
           echo "Changed files in this pull request (${merge_base}..${HEAD_SHA}):"
           sed 's/^/  /' "${changed}"
-      - name: Enforce Version Bump for Source Changes
+      - name: Enforce Version Bump for Published Changes
         id: guard
         env:
           CHANGED_FILES: ${{ steps.changes.outputs.changed_files }}
-          SRC_PREFIX: ${{ matrix.path }}/src/
+          # The whole dist directory (what package-publish.yaml triggers on)…
+          DIST_PREFIX: ${{ matrix.path }}/
+          # …except the tests, which are not packaged. See the SCOPE note above.
+          TEST_PREFIX: ${{ matrix.path }}/src/test/
           VERSION_FILE: ${{ matrix.version_file }}
         run: |
           set -euo pipefail
 
-          src_changed=false
+          dist_changed=false
           version_changed=false
           while IFS= read -r file; do
             if [ -z "${file}" ]; then
               continue
             fi
             case "${file}" in
-              "${SRC_PREFIX}"*) src_changed=true ;;
+              "${TEST_PREFIX}"*) ;;
+              "${DIST_PREFIX}"*) dist_changed=true ;;
             esac
             if [ "${file}" = "${VERSION_FILE}" ]; then
               version_changed=true
             fi
           done < "${CHANGED_FILES}"
 
-          if [ "${src_changed}" = "false" ]; then
-            echo "src_changed=false" >> "${GITHUB_OUTPUT}"
+          if [ "${dist_changed}" = "false" ]; then
+            echo "dist_changed=false" >> "${GITHUB_OUTPUT}"
             echo "version_changed=false" >> "${GITHUB_OUTPUT}"
-            echo "::notice::No changes under ${SRC_PREFIX} — the version-bump rule does not apply to this pull request."
+            echo "::notice::No published files changed under ${DIST_PREFIX} (${TEST_PREFIX} is exempt — nothing under it is packaged) — the version-bump rule does not apply to this pull request."
             exit 0
           fi
-          echo "src_changed=true" >> "${GITHUB_OUTPUT}"
+          echo "dist_changed=true" >> "${GITHUB_OUTPUT}"
 
           if [ "${version_changed}" = "false" ]; then
             current="$(head -n 1 "${VERSION_FILE}" 2>/dev/null | tr -d '[:space:]' || true)"
             {
-              echo "RULE: a pull request that modifies ${SRC_PREFIX}** must also change ${VERSION_FILE}"
+              echo "RULE: a pull request that modifies ${DIST_PREFIX}** must also change ${VERSION_FILE}"
+              echo "      (${TEST_PREFIX}** is exempt — nothing under it is packaged)"
               echo ""
-              echo "This pull request changes ${SRC_PREFIX}** but leaves ${VERSION_FILE} at ${current:-<unreadable>}."
+              echo "This pull request changes published files under ${DIST_PREFIX}** but leaves"
+              echo "${VERSION_FILE} at ${current:-<unreadable>}."
               echo ""
               echo "Why that blocks the merge: on merge, \"Python Package Publish\" uploads with"
               echo "\`twine upload --skip-existing\`. Re-uploading a version PyPI already has is skipped"
@@ -118,9 +142,9 @@ jobs:
               echo ""
               echo "Fix: bump ${VERSION_FILE} to the next unpublished version"
               echo "(the version.py beside it is a derived reader — editing that file alone changes nothing),"
-              echo "or move the change out of ${SRC_PREFIX}** if this pull request must not release."
+              echo "or move the change out of ${DIST_PREFIX}** if this pull request must not release."
             } | tee -a "${GITHUB_STEP_SUMMARY}"
-            echo "::error file=${VERSION_FILE}::Source under ${SRC_PREFIX} changed without bumping ${VERSION_FILE} — the merge-triggered publish would be a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump the version file."
+            echo "::error file=${VERSION_FILE}::Published files under ${DIST_PREFIX} changed without bumping ${VERSION_FILE} — the merge-triggered publish would be a silent no-op (see the 2026-07-28 develop-daemon #62/#64 incident). Bump the version file."
             exit 1
           fi
 
@@ -134,8 +158,33 @@ jobs:
         run: |
           set -euo pipefail
 
+          # A pull request can DELETE the version file, which reaches this step as
+          # version_changed=true. Say so in the same shape as every other failure
+          # here rather than dying on a bare `head: cannot open …`.
+          if [ ! -f "${VERSION_FILE}" ]; then
+            {
+              echo "RULE: ${VERSION_FILE} must exist and hold the version this pull request publishes."
+              echo ""
+              echo "This pull request deletes (or moves away) ${VERSION_FILE}. setup.py opens that"
+              echo "exact path to set the distribution version, so the merge-triggered publish would"
+              echo "fail to build — and the version.py beside it would fall back to 0.0.0."
+              echo ""
+              echo "Fix: restore ${VERSION_FILE} holding the version this pull request publishes."
+            } | tee -a "${GITHUB_STEP_SUMMARY}"
+            echo "::error file=${VERSION_FILE}::${VERSION_FILE} is missing — setup.py reads it to set the published version. Restore it."
+            exit 1
+          fi
+
           version="$(head -n 1 "${VERSION_FILE}" | tr -d '[:space:]')"
           if [ -z "${version}" ]; then
+            {
+              echo "RULE: ${VERSION_FILE} must exist and hold the version this pull request publishes."
+              echo ""
+              echo "The file is empty. setup.py reads it to set the distribution version, so the"
+              echo "merge-triggered publish would fail to build."
+              echo ""
+              echo "Fix: write the version this pull request publishes into ${VERSION_FILE}."
+            } | tee -a "${GITHUB_STEP_SUMMARY}"
             echo "::error file=${VERSION_FILE}::${VERSION_FILE} is empty — it must hold the version this pull request publishes."
             exit 1
           fi


alissa-app:
- 2026-07-28 15:02:52 UTC — `review-alissa-github-review-daemon-pr41-r1-e03274` — round 1 of 10 — spawned
- 2026-07-28 15:20:20 UTC — `review-alissa-github-review-daemon-pr41-r2-5b9f17` — round 2 of 10 — spawned

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,183 @@
...
+          git diff --name-only "${merge_base}" "${HEAD_SHA}" > "${changed}"
+
+          echo "changed_files=${changed}" >> "${GITHUB_OUTPUT}"
+          echo "Changed files in this pull request (${merge_base}..${HEAD_SHA}):"
+          sed 's/^/  /' "${changed}"
+      - name: Enforce Version Bump for Source Changes
+        id: guard
+        env:
+          CHANGED_FILES: ${{ steps.changes.outputs.changed_files }}
+          SRC_PREFIX: ${{ matrix.path }}/src/
```
alissa-app: `[major]` **The guarded prefix is narrower than the publish trigger, so a whole class of publishing changes skips the rule.**

`package-publish.yaml` fires on `paths: [.github/workflows/package-publish.yaml, alissa-tools-github-revloop/**]`. This guard fires on `alissa-tools-github-revloop/src/`. Everything in between changes what gets published *and* triggers the publish job on merge, while being exempt from the version-bump requirement:

- `alissa-tools-github-revloop/requirements.txt` → `install_requires` (setup.py reads it)
- `alissa-tools-github-revloop/setup.py` → dist name, `entry_points`, `python_requires`
- `alissa-tools-github-revloop/README.md` → `long_description` (the PyPI page)
- `alissa-tools-github-revloop/MANIFEST.in` → what lands in the sdist

Merging any of those without a bump produces exactly the failure this workflow exists to prevent: the publish job runs, `twine upload --skip-existing` skips the duplicate, the job reports SUCCESS, and nothing is published.

Verified against the shipped `run:` blocks (fixture changed-file list, real step `env:`, `version` = 0.15.0):

```
changed = [alissa-tools-github-revloop/requirements.txt]
  ::notice::No changes under alissa-tools-github-revloop/src/ — the version-bump rule does not apply to this pull request.
  guard exit 0 · verify step skipped by its if: · OVERALL PASS

changed = [alissa-tools-github-revloop/setup.py, alissa-tools-github-revloop/README.md]
  (same notice) · OVERALL PASS
```

Fix: key this on the same subtree the publish workflow triggers on — `SRC_PREFIX: ${{ matrix.path }}/` (rename to `DIST_PREFIX`, and the messages read fine as `alissa-tools-github-revloop/**`). Coverage is then defined by the property that actually matters — "would merging this run a publish?" — instead of by a path prefix that happens to be narrower. Pair it with the `src/test/` exclusion in the `[minor]` on line 90.

Issue #40 does say `src/**` literally, so `[triage:ignore]` with that reasoning, or `[triage:later]` with a task, is a legitimate answer under CR8 — the point is that this gap should be a recorded decision rather than a side effect of the prefix.

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,183 @@
...
+          set -euo pipefail
+
+          src_changed=false
+          version_changed=false
+          while IFS= read -r file; do
+            if [ -z "${file}" ]; then
+              continue
+            fi
+            case "${file}" in
+              "${SRC_PREFIX}"*) src_changed=true ;;
```
alissa-app: `[minor]` **Answering the judgement call raised in the PR description:** `src/**` covering `src/test/**` is the one part of the current prefix I would narrow rather than keep.

Nothing under `src/test/` ships. `setup.py` packages `find_namespace_packages(where=src/main, include=[alissa.tools.github.revloop, alissa.tools.github.revloop.*])`, and `MANIFEST.in` adds only `requirements.txt` and `version`. A test-only change therefore produces a byte-identical wheel, and requiring a bump for it spends a release number on a release that changes nothing shipped.

The consistency argument in the PR body doesn't quite carry: `package-publish.yaml` triggers on the whole dist subtree, so a test-only merge runs a build-and-upload that publishes nothing *whether or not* the version moved. That is an argument for narrowing that workflow's own `paths:` (a separate change, out of this PR's scope) — not for forcing a release.

Confirmed with the shipped guard step that a test-only change fails today:

```
changed = [alissa-tools-github-revloop/src/test/test_loop.py], version 0.15.0
  → guard exit 1 (RULE: … must also change …/version)
```

Concretely, alongside the `[major]`: match on `${{ matrix.path }}/` and add a `case` arm that skips `${{ matrix.path }}/src/test/`*.

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,183 @@
+name: Check Version Bump for Source Changes
+on: [pull_request]
```
alissa-app, line 2: `[minor]` Two things outside this file decide whether this check *blocks* anything, and neither is in place yet:

1. Unless "Version Bump Check for alissa-tools-github-revloop" is a **required** status check in `main`'s branch protection, a red result is advisory — a merge can still happen, which is how the develop-daemon incident got merged in the first place.
2. Without **"Require branches to be up to date before merging"**, the check is decided against a stale base: two open PRs that both bump to `0.16.0` each pass independently, and the second merge is the silent no-op again (the job only re-runs on `synchronize`, not when `main` moves).

Both are repo-admin work outside this PR's `Autonomous-Scope: .github/workflows/**`, so no change is needed here — registered as **TASK-568945951** so it doesn't evaporate.

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,183 @@
...
+            git fetch --no-tags origin "${BASE_SHA}"
+          fi
+
+          # Diff against the merge base (equivalent to BASE...HEAD), which is the
+          # file set GitHub shows in the pull request's "Files changed" tab —
+          # unrelated commits that landed on main meanwhile are not attributed
+          # to this pull request.
+          merge_base="$(git merge-base "${BASE_SHA}" "${HEAD_SHA}")"
+          changed="${RUNNER_TEMP}/changed-files.txt"
+          git diff --name-only "${merge_base}" "${HEAD_SHA}" > "${changed}"
```
alissa-app: `[nit]` `git diff --name-only` applies rename detection, so a file *moved out of* the dist source tree is reported only at its destination — e.g. `…/src/main/alissa/tools/github/revloop/loop.py` → `docs/loop.py` lists just `docs/loop.py`, and the guard never sees a `src/` path even though the wheel changed. Verified:

```
$ git diff --name-only HEAD~1 HEAD          → docs/mod.py
$ git diff --no-renames --name-only HEAD~1 HEAD → docs/mod.py
                                                  pkg/src/main/mod.py
```

`--no-renames` makes both sides appear. One flag, and the changed-file set becomes exactly "paths this PR touches on either side".

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,183 @@
...
+          echo "::notice::${VERSION_FILE} is part of this pull request — the merge will publish a new release."
+      - name: Verify New Version Is Unpublished
+        if: steps.guard.outputs.version_changed == 'true'
+        env:
+          DIST_NAME: ${{ matrix.path }}
+          VERSION_FILE: ${{ matrix.version_file }}
+        run: |
+          set -euo pipefail
+
+          version="$(head -n 1 "${VERSION_FILE}" | tr -d '[:space:]')"
```
alissa-app, line 178: `[nit]` Two small ones on this line:

1. **A deleted version file fails opaquely.** If a PR deletes `…/revloop/version`, the guard step sees it in the changed set, sets `version_changed=true`, and this step dies with a bare `head: cannot open '…/revloop/version': No such file or directory` — no RULE text, no `::error file=` annotation, unlike every other failure path in the workflow. Verified (guard exit 0 → verify exit 1 with only the `head` message). Line 106 already has the pattern to mirror: `2>/dev/null || true` plus an explicit message, here something like "the version file was deleted; it must exist and hold the version this pull request publishes".
2. **No monotonicity check.** Only "not already on PyPI" is enforced, so a bump *down* to an unpublished lower number (`0.15.0` → `0.14.1`) passes and publishes, leaving PyPI's "latest" behind `main`. Outside the rule as specified in issue #40 — noting it, not asking for it.

alissa-app (changes requested):

Fresh-instance review per the `alissa-code-review` skill (CR1–CR9). Alissa review task: **TASK-1826183148** (origin **TASK-1625720421** / issue #40, implementation **TASK-1058266947**). Round 1 of cap 3.

**Findings:** blocker 0 · major 1 · minor 2 · nit 2 · question 0 → `request_changes` on the single unwaived `[major]`.

This is a careful piece of work. The changed-file set is computed from the merge base (so it matches the "Files changed" tab), the job keeps `contents: read`, every `${{ }}` value reaches the shell through an `env:` map rather than being interpolated into the script, the version string is charset-validated before it is put in a URL, and only a *confirmed* HTTP 200 fails the PyPI step — an outage warns and passes. Failure messages are genuinely actionable and land in the log, the step summary, and a file annotation. The `version` vs `version.py` source-of-truth analysis is correct: `setup.py` reads the plain `version` file and `version.py`'s `Version.from_path()` is a derived reader, so keying on `version` alone is right.

I re-ran the shipped `run:` blocks myself rather than reading the PR body's table (same harness shape: fixture changed-file list, real step `env:`, real PyPI calls). All seven of the PR's cases reproduce, plus two the table doesn't cover — those are the `[major]` and one `[nit]` below.

### Dimension report

- **Correctness** — the rule as literally specified in issue #40 works: src-without-bump fails, src+new-version passes, bump-to-a-published-version fails, non-src PRs skip, PyPI-unreachable warns and passes. The check ran green on this PR itself (run 30371283769), and all six checks on `9335e42` are green. One gap against the *purpose*: `[major]` below — the guarded prefix is narrower than the set of paths that actually trigger a publish.
- **Coding practices** — matches repo idiom closely: `on: [pull_request]`, `permissions: contents: read`, `strategy.matrix` with `name: … for ${{ matrix.path }}`, "Check …" workflow name, `Clone Repo` step naming. The `matrix.include` form is a justified extension for the second key. Branch `TASK-1058266947-VERSION-BUMP-GUARD` (C2), PR title names the task (C3), PR is on the task as evidence (C4), commit carries the `Co-Authored-By: alissa-app` trailer the repo's other commits use (C1). No dead code, no debug leftovers, no untracked TODOs. Two `[nit]`s.
- **Security** — nothing found. Least privilege (`contents: read`); no secrets touched; no `pull_request_target`; no untrusted value interpolated into a script (`${{ }}` → `env:` throughout, and the branch-controlled `version` string is charset-validated before URL construction and always quoted). `PYPI_API_BASE` is a workflow-level constant, not an input.
- **Performance** — nothing found. One extra ~5s job per PR; `fetch-depth: 0` is required for the merge base and is the only real cost. The PyPI call is a single request with `--max-time 30 --retry 2`.
- **Side-effects / blast radius** — additive, no existing workflow or code path changed; the job can only fail a PR, never publish or write. Sibling repos carry their own copies of this guard, so nothing cross-repo depends on this file. One `[minor]`: whether it *blocks* a merge is branch-protection configuration outside this PR's scope — registered as **TASK-568945951**.

Resolve the `[major]` — either widen the matched prefix to the subtree `package-publish.yaml` triggers on, or triage it `[triage:ignore]`/`[triage:later]` with the "issue #40 says `src/**` literally" reasoning per CR8. The two minors and two nits do not gate.

*Reviewer note:* this hub's local `main/` mirror is ~13 releases stale (it predates the `reviewloop` → `revloop` rename), so everything above was verified against `origin/main` at `6348b5a`, the PR's actual base.

RHDZMOTA:
- 2026-07-28 15:14:01 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr41-r1-a1` — fix edge — spawned

.github/workflows/check-version-bump.yaml
RHDZMOTA: `[triage:pursue]` — widening in this round.

Your framing is the right one, and it is a better rule than the one I wrote: the property worth enforcing is *"if merging this pull request runs a publish, it must publish something."* `package-publish.yaml` triggers on the whole `alissa-tools-github-revloop/**` subtree, and each of the four files you name changes the artifact — `requirements.txt` → `install_requires`, `setup.py` → dist name / `entry_points` / `python_requires`, `README.md` → `long_description`, `MANIFEST.in` → sdist contents. Merging any of them without a bump reaches the exact end state this workflow exists to prevent, through a door my prefix left open.

Changed: `SRC_PREFIX` → `DIST_PREFIX: ${{ matrix.path }}/`, with a `src/test/` exemption per your `[minor]` on line 90. Messages and the skip notice now read `alissa-tools-github-revloop/**`, and the header carries a SCOPE note stating the property and both boundaries.

New dry-run cases (same harness, real step `env:`, `version` = 0.15.0):

```
changed = [alissa-tools-github-revloop/requirements.txt]                    → guard exit 1  FAIL
changed = [alissa-tools-github-revloop/setup.py, …/README.md]               → guard exit 1  FAIL
changed = [alissa-tools-github-revloop/MANIFEST.in, …/version] (0.16.0)     → PASS
```

**Recording the deviation rather than making it silently:** issue #40 says `src/**` literally, and its acceptance criteria say PRs touching "docs, docker/, workflows, or other non-src paths" are unaffected. Every repo-level path in that list stays exempt — `README.md`, `docker/`, `.github/`, the `*.sh` check scripts, `*.config.example.json` — and this PR still passes its own check. The only files that move from exempt to guarded are the four inside the distribution directory, and for each of them "unaffected" is wrong on the issue's own terms, since each one both starts a publish and changes what would be published. Called out explicitly in the PR body so validation reads a decision, not a drift.

.github/workflows/check-version-bump.yaml
RHDZMOTA: `[triage:pursue]`

Agreed, and my consistency argument was backwards. Nothing under `src/test/` is packaged — `setup.py` calls `find_namespace_packages(where=src/main, …)` and `MANIFEST.in` adds only `requirements.txt` and `version` — so a test-only change produces a byte-identical wheel, and requiring a bump spends a release number on a release that ships nothing new. And "the publish job runs anyway" is an argument about that workflow's `paths:`, not a reason to force a release; I'm not touching `package-publish.yaml` here.

Implemented as a `case` arm ordered before the dist-prefix match:

```sh
case "${file}" in
  "${TEST_PREFIX}"*) ;;
  "${DIST_PREFIX}"*) dist_changed=true ;;
esac
```

Case 7 in the PR body inverts accordingly — a test-only change with no bump now **passes**.

.github/workflows/check-version-bump.yaml
RHDZMOTA: `[triage:answer]` — no change here; agreed on both points.

Required-status-check and "require branches to be up to date" are repo-admin settings outside `Autonomous-Scope: .github/workflows/**`, so this PR cannot close either. Your second point is the sharper one and worth restating for whoever configures protection: without up-to-date-before-merging, two open PRs can each bump to `0.16.0`, both pass against a stale base, and the second merge is the silent no-op again — a `pull_request` job only re-runs on `synchronize`, so it cannot see `main` move underneath it. That is a genuine residual hole in this guard, not a hypothetical.

TASK-568945951 linked from the PR body so the gap stays visible.

.github/workflows/check-version-bump.yaml
RHDZMOTA: `[triage:pursue]` — one flag, real hole. Added `--no-renames`, with a comment explaining why.

Reproduced from scratch before taking it:

```
$ git mv pkg/src/main/mod.py docs/mod.py && git commit
$ git diff --name-only HEAD~1 HEAD          → docs/mod.py
$ git diff --no-renames --name-only HEAD~1 HEAD → docs/mod.py
                                                  pkg/src/main/mod.py
```

Without the flag a file moved out of the dist tree changes the wheel while the guard sees no dist path at all — exactly the silent-no-op shape. Covered as case 11.

.github/workflows/check-version-bump.yaml
RHDZMOTA: **1.** `[triage:pursue]` — that was the one failure path in the workflow that didn't explain itself, which undercuts the whole point of the actionable-message criterion. Fixed: the verify step now checks `[ ! -f "${VERSION_FILE}" ]` first and fails with the same RULE text + `$GITHUB_STEP_SUMMARY` + `::error file=` shape as every other failure, naming that `setup.py` opens that exact path and that `version.py` would fall back to `0.0.0`. I gave the empty-file branch the same treatment while I was there — it previously had the annotation but no RULE block. Cases 12 (deleted) and 13 (emptied) in the PR body.

**2.** `[triage:later]` — real, but out of scope for this round on two counts: it is outside the rule issue #40 specifies, and a downgrade to an unpublished number does publish something, so it is not the silent-no-op failure mode this guard targets. It is still a hole — PyPI's "latest" drifting behind `main` — so it is registered as **TASK-1492477** (soft-linked to this work) rather than dropped on the floor.

.github/workflows/check-version-bump.yaml
```diff
@@ -0,0 +1,232 @@
...
+          set -euo pipefail
+
+          dist_changed=false
+          version_changed=false
+          while IFS= read -r file; do
+            if [ -z "${file}" ]; then
+              continue
+            fi
+            case "${file}" in
+              "${TEST_PREFIX}"*) ;;
```
alissa-app, line 111: `[nit]` The exemption list is one arm keyed on `src/test/`, which is exactly right for the dist tree as it stands — I enumerated it on `origin/main` and every path except `src/test/**` reaches the artifact. It turns into a false gate the first time a non-packaged file lands at the dist *root*: `pytest.ini`, `tox.ini`, `.flake8`, a `Makefile`. Verified with a fixture — `alissa-tools-github-revloop/tox.ini` alone, no bump, guard exits 1 and demands a release number for a byte-identical wheel.

That is arguably correct under the property you state (such a change *does* start a publish that publishes nothing), so I'd not pre-empt it with speculative arms. One clause in the SCOPE note — that the exemption list is expected to grow alongside non-packaged files, and the fix is another `case` arm here — would point the next person at this line instead of at `package-publish.yaml`'s `paths:`. Implementer's discretion; nothing gates on it.

alissa-app (approved):
## Round 2 — `approve`

Re-reviewed `9335e42..629df04` against the round-1 findings, then swept the whole file again with the full rubric. **All five round-1 findings reconcile clean and the delta introduces nothing that gates.** One new `[nit]` inline, non-blocking.

Everything below was re-verified by **executing the shipped `run:` blocks**, not by reading the PR body: a harness parses `check-version-bump.yaml` at `629df04`, extracts the two guard steps verbatim, resolves their `env:` maps and the `if:` condition from the matrix, and runs them under a GitHub-Actions-shaped environment against a fixture changed-file list. 20 cases — the PR body's 13 plus 7 of my own. **20/20 matched.**

| round-1 finding | triage | round-2 disposition |
| --- | --- | --- |
| `[major]` guarded prefix narrower than the publish trigger (line 78) | `pursue` | **fixed** in `629df04` — `SRC_PREFIX` → `DIST_PREFIX: ${{ matrix.path }}/`. Independently re-run: `alissa-tools-github-revloop/requirements.txt` alone, no bump → exit 1; `setup.py` + dist `README.md`, no bump → exit 1; `MANIFEST.in` + a new version → pass. I also enumerated the dist tree on `origin/main` — `MANIFEST.in`, `README.md`, `requirements.txt`, `setup.py`, `src/main/**`, `src/test/**`, nothing else — so with `src/test/` exempt the guarded set is *exactly* the packaged set today. No over-reach in the other direction. |
| `[minor]` `src/test/**` should not force a release (line 90) | `pursue` | **fixed** — the `"${TEST_PREFIX}"*) ;;` arm is ordered before the dist arm, which is what makes the exemption win. Verified on three shapes, the last two not claimed in the PR body: test-only + no bump → pass; test + `src/main` + no bump → **exit 1** (the exemption does not leak); test-only + a bump → pass and the PyPI check still runs. |
| `[minor]` required status check / "require branches to be up to date" (line 2) | `answer` | **accepted, closed.** Both are repo-admin settings genuinely outside `Autonomous-Scope: .github/workflows/**`, and the reply states the residual race correctly — a `pull_request` job only re-runs on `synchronize`, so it cannot see `main` move underneath it. TASK-568945951 verified: exists, open, and its title names both settings. |
| `[nit]` rename detection hides a file moved *out* of the dist tree (line 69) | `pursue` | **fixed** — `git diff --no-renames`. Reproduced from scratch rather than taking the reply's word: `git mv pkg/src/main/mod.py docs/mod.py` → `--name-only` reports only `docs/mod.py`; `--no-renames --name-only` reports both. Case: `…/loop.py` + `docs/loop.py`, no bump → exit 1. |
| `[nit]` deleted/empty version file fails opaquely; no monotonicity check (line 178) | `pursue` + `later` | **fixed / deferred.** Deleted → the `[ ! -f ]` branch now emits RULE text + `$GITHUB_STEP_SUMMARY` + `::error file=` and exits 1; emptied → same shape. Both re-run here. Monotonicity: TASK-1492477 verified — exists, open, and scoped to exactly that gap. |

No `still-open` rows. Nothing from round 1 was re-raised, and no reasoned `[triage:ignore]` was disputed (there were none).

### Delta sweep — all five dimensions

- **Correctness:** The rule holds on every shape I could construct, including four the PR body does not claim: test+src without a bump fails, test-only with a bump passes, a hypothetical non-packaged file at the dist root (`tox.ini`) fails, and a sibling directory sharing the prefix stem (`alissa-tools-github-revloop-extras/x.py`) correctly does *not* match — the trailing slash on `DIST_PREFIX` is doing real work there. Version-file-only-with-a-bump passes; a trailing newline in the version file is stripped before the PyPI URL is built. All six checks are green on `629df04`, including this workflow's own run ([30372698439](https://github.com/fahera-mx/alissa-github-review-daemon/actions/runs/30372698439)). One thing for validation to *ratify*, not a defect: the guard now deviates from issue #40's literal `src/**` in both directions — wider (the four dist-root files) and narrower (`src/test/**`). Both deviations are ones round 1 asked for, both are recorded in the PR body and in the workflow's SCOPE header, and the origin DoD's spirit ("if merging runs a publish, it must publish something") is better served by the new prefix than by the literal one. I'm calling it satisfied.
- **Coding practices:** Nothing found. The SCOPE header block earns its length — it states the property, both boundaries, and why each of the four dist-root files changes the artifact, so the next person editing the prefix has the reasoning rather than a bare glob. The `--no-renames` comment explains the failure mode it prevents instead of restating the flag. C1–C4 still hold on the round-2 commit: branch `TASK-1058266947-VERSION-BUMP-GUARD`, `Co-Authored-By: alissa-app` trailer, the commit message names the round-1 findings it answers, PR titles the task. `dist_changed` is written to `$GITHUB_OUTPUT` and never consumed — same as `src_changed` before it, pre-existing, and it is legitimately useful in the job log. Not worth a finding.
- **Security:** Nothing found. Posture is unchanged and still correct: `permissions: contents: read`, no `pull_request_target`, no secrets, every `${{ }}` reaching the shell through an `env:` map rather than script interpolation. The new `TEST_PREFIX`/`DIST_PREFIX` values are matrix constants, and `case` patterns built from them contain no glob metacharacters. The branch-controlled `version` string is still charset-validated before URL construction — I re-ran it with `0.16.0; rm -rf /` in the version file: rejected at the charset gate, exit 1, nothing interpolated.
- **Performance:** Nothing found. Unchanged shape — one ~6s job per PR, one `git diff`, one PyPI request with `--max-time 30 --retry 2`. `--no-renames` makes the diff marginally *cheaper*.
- **Side-effects / blast radius:** Nothing found. Still a single added file that can only fail a PR — no existing workflow or code path touched, nothing published or written. The widened prefix's blast radius is confined to this repo's own PRs, and this PR still passes its own check. The one behavioural change a human should expect: PRs touching `alissa-tools-github-revloop/{requirements.txt,setup.py,README.md,MANIFEST.in}` now require a release, which is the point.

### Verdict

`approve`. The round-1 `[major]` is fixed with a better rule than the one I asked for — keying on the publish trigger rather than on `src/`, with the test exemption as its mirror — and both nits produced real behavioural fixes rather than comment changes. The two deferrals cite tasks that exist and are open. Remaining non-gating items: the new `[nit]` below, and TASK-568945951 / TASK-1492477, which are already registered and outside this PR's scope.

.github/workflows/check-version-bump.yaml
RHDZMOTA: `[triage:later]` — agreed on both halves, and deliberately not landing it in this round.

Your point stands: the exemption is one arm keyed on `src/test/`, which is exactly the packaged/non-packaged boundary of the dist tree *today*, and the first `tox.ini`/`pytest.ini`/`Makefile` at the dist root will demand a release number for a byte-identical wheel. As you say that is defensible under the stated property — such a change does start a publish that publishes nothing — so the right answer is a pointer, not speculative arms.

Not taken here because this round's verdict is an approve, and an approve is terminal in this loop: convergence merges the **approved** commit, so a post-approve doc commit would be orphaned on the branch rather than reaching `main`. Trading a one-clause comment for that risk is not worth it.

Registered as **TASK-2026645875** (soft-linked to TASK-1058266947), scoped to exactly what you describe: one clause in the SCOPE note saying the exemption list is expected to grow and that the fix is another `case` arm on this line — not `package-publish.yaml`'s `paths:`. Guard behaviour unchanged.

PR #43, opened 2026-07-28
revloop: operator re-entry ack after cap-out (bounded fresh cap + verification round)

### LINKED ISSUE #42: revloop: operator re-entry after cap-out — an explicit ack grants a bounded fresh cap, plus a verification round for post-cap fixes
Repo: fahera-mx/alissa-github-review-daemon. Live incident (2026-07-28, studio PR #277 — a ~1300-line design-doc PR): the loop ran the full ALISSA_ROUND_CAP=10 rounds, every round's findings were fixed (repeatedly acknowledged by the reviewer as past the ask), and the round-10 verdict even enumerated a four-item "shortest path back to approve" — which the fix session then implemented and pushed ten minutes later. The cap-out escalation fired, and the fixed head sat unreviewable: the CR9 escalation message offers "direct specific fixes and re-enter with a fresh cap" as an operator option, but NO such mechanism exists — the round count is verdict-envelope-derived against a global config.round_cap, so the only lever is raising the cap for every PR and restarting the daemon. The operator had to hand-verify the final commit against the review text and merge on their own authority.

## Required change

1. **Operator re-entry ack.** After a cap-out escalation, an explicit operator signal on the PR grants a bounded number of additional rounds. Design the signal to be trivially auditable and impossible to fire accidentally — e.g. a marker comment with a fixed grammar (`alissa-review: re-enter +N` with a sane N ceiling, authored by an allowlisted operator identity — reuse the daemon's existing operator/identity notions; NOT any-commenter). The decision loop treats the ack as raising this PR's effective cap by N (per (PR, ack) — a second ack requires a second comment; acks are counted, never inferred). Log the grant loudly; append to the activity comment.
2. **The cheap high-value special case — a verification round.** When the head has MOVED past the last verdict's head after cap-out (the PR #277 shape: fixes pushed post-cap), one re-entry round is exactly what is needed: the reviewer verifies the fix against its own final findings and flips to approve (or re-requests, consuming the grant). Make the escalation comment say this explicitly — tell the operator the ack grammar and that one round is usually enough — so the lever is discoverable at the moment it is needed.
3. **Escalation stays once-only**: after a granted re-entry is consumed without approval, a fresh cap-out escalation fires once (naming the ack that granted it), and the state returns to capped-until-acked. No ack, no rounds — CR9's never-run-past-the-cap and never-silently-merge invariants are unchanged.
4. **Doc drift fix (rides along)**: the review task description template still says the cap is 3; the actual default is 10 (config.round_cap). Align whatever template emits that number — the round-10 reviewer flagged the mismatch in passing.

## Out of scope
- Changing the default cap; artifact-type-aware convergence rules (e.g. approve-on-no-majors for design docs — a separate discussion); devloop changes; auto-merge of any kind.

## Verification expectations
- Unit tests with fakes per repo convention: capped PR + no ack → CAPPED (today's behavior); valid ack → exactly N additional rounds admitted, logged, activity-appended; malformed ack / non-operator author → ignored with a log line, no rounds; consumed grant without approve → one fresh escalation naming the ack, then capped again; ack grammar parsing matrix; the escalation message includes the re-entry instructions.
- run-tests.sh + repo checks green; version file bumped (publish-on-merge); Delivery notes in the PR body.

## Definition of Done

A capped PR can be re-entered by an explicit, allowlisted, bounded operator ack that grants N more review rounds (loudly logged, once per ack, escalating freshly if consumed without approval), the cap-out escalation itself teaches the ack grammar and recommends a single verification round when the head moved past the last verdict, the cap-3-vs-10 template drift is fixed, and CR9's never-past-cap / never-silent-merge invariants hold — test-pinned, version-bumped, merged via a reviewed PR.

## Acceptance criteria

- [ ] A well-formed operator ack on a capped PR admits exactly N more rounds (bounded, per-ack, counted never inferred), loudly logged and activity-appended; no ack → CAPPED unchanged — test-pinned with fakes.
- [ ] Malformed or non-allowlisted acks are ignored with a log line; a consumed grant without approve fires one fresh escalation naming the ack then returns to capped; never-past-cap and never-silent-merge hold on every path — test-pinned.
- [ ] The cap-out escalation message includes the ack grammar and the verification-round recommendation when the head moved past the last verdict's head; the cap-3-vs-10 template drift is fixed — asserted in tests.
- [ ] Repo checks green; version bumped; Delivery notes in the PR body.

Alissa-Task: TASK-1214531903

RHDZMOTA:
Closes #42

**Origin task:** TASK-1214531903 (develop-daemon partition)
**Implementation task:** TASK-647078429 (this actor's partition, downstream of the origin)

## What this does

A capped PR can now be re-entered by an **explicit, allowlisted, bounded operator ack**, and the cap-out page teaches the lever at the moment it is needed.

**The grammar** — one line of a PR comment, backticks optional, prose around it fine:

```
alissa-review: re-enter +N
```

- Honoured only from a login in the new `operators` config (empty = **nothing** is honoured; the lever fails closed). The reviewer identity is *never* an operator however it is configured — the escalation itself quotes the grammar, so a self-ack would lift CR9's cap with no human in the loop.
- `N` from 1 to `MAX_REENTRY_ROUNDS` (5). A bigger re-entry is a second comment, not a bigger number.
- **Counted, never inferred:** one grant per ack comment, keyed by comment id in a new `grants` ledger table, so the ack that sits on the PR forever grants exactly once. Effective cap = `round_cap + Σ grants`.
- Loudly logged (`RE-ENTRY GRANT …`) and appended to the review-loop activity comment (author, comment id, before/after cap). The activity append is retried on later polls until it lands — the grant itself never waits on a comment API call.
- Ignored **with one log line each** (deduped per comment id, not per poll): malformed directives, out-of-range `N`, contradictory directives in one comment, quoted (`>`) lines, non-operator authors.
- **Escalation stays once-only:** a grant consumed without an `approve` fires exactly one fresh cap-out naming the ack, then returns to capped-until-acked. Deduped on `(head_sha, granted total)` via the ping ledger — deliberately not a wall-clock comparison, whose ties fail in the loud direction. A PR with no grant takes the pre-existing head-keyed path unchanged, so no already-escalated PR re-pages on upgrade.
- The GitHub comment scan runs **only when the loop would otherwise be capped** — no extra API call per poll on the common path (the local grant sum is read every pass).

**The verification-round special case (the PR #277 shape).** When the head has moved past the head the last verdict was written against, the escalation says so and recommends exactly one round: the reviewer verifies the fix against its own final findings and flips to approve (or re-requests, consuming the grant).

**Doc drift (item 4).** Both reviewer directives now carry the *effective* cap and tell the reviewer to record **that** number in the review task description, correcting a stale template default — round 1 is the round that creates that description, and it previously carried no cap at all, so the skill's documented `3` propagated while the daemon's default has been `10`.

CR9's invariants are untouched: never a round past the (now effective) cap, never a silent merge; no ack, no rounds.

## Delivery notes

- `loop.py` — `parse_reentry_ack` / `Ack` (pure, table-tested), `_collect_acks`, `_announce_grants`, `_log_ignored_ack`, `_escalation_owed`, `capout_kind`; escalation body rewritten (grammar + ceiling + verification hint + consumed-grant note); effective cap threaded through the directives, the activity lines and the spawn path. `RateLimited` propagates out of the scan rather than being swallowed into a quiet "capped".
- `state.py` — `grants` table (`CREATE TABLE IF NOT EXISTS` *is* the migration; legacy ledgers untouched) + `record_grant` / `granted_rounds` / `newest_grant` / `read_grants`.
- `config.py` / CLI — new `operators` key and repeatable `--operator`; a bare string is rejected (it would iterate into characters and allowlist nobody).
- Container — `ALISSA_REVIEW_OPERATORS` (`|`-separated, pass-through: unset ⇒ key omitted ⇒ library default = fail closed), wired through `revloop-config.sh`, `entrypoint.sh`, Dockerfile ARG/ENV, both entrypoint test scripts and the README env table. The end-to-end entrypoint harness caught a real boot-killer while writing it: `grep` exits 1 on an empty operator list, which under `set -euo pipefail` killed the entrypoint mid-bootstrap — fixed and now covered by the "unset ⇒ omitted" assertion on a real boot.
- `ghclient.py` — `issue_comments` now PAGES (explicitly, not `gh api --paginate`, which this module's gh 2.4.0 target emits as one JSON document per page). Both readers of that list failed silently on a truncated read: the activity-comment finder would fork a second comment, and the ack scan would drop an operator's ack — on exactly the long-lived capped-out PRs where the thread is longest. Bounded at 20 pages, and the truncation is logged rather than assumed away.
- Console — `config_echo` reports the allowlist next to the escalation inbox where the lever gets used.
- Version `0.15.0` → `0.16.0` (publish-on-merge). The Dockerfile's `REVLOOP_VERSION` stays at `0.15.0` on purpose: the release does not exist on PyPI until this merges, and `check-entrypoint.yaml` installs the pinned version. Re-pin in a follow-up, as with 0.5.0 → 0.6.0.
- Docs — daemon README gains an "Operator re-entry after a cap-out" section, two behaviour-table rows and the `operators` setting; `revloop.config.example.json` documents the key.

## Verification

All run locally against the worktree source (`PYTHONPATH=src/main`, package installed to user site):

- `bash tests-unit.sh alissa-tools-github-revloop` — **350 passed** (+52 new, incl. the ack-grammar matrix, cap/no-cap decision paths, refusals, once-only escalation, the escalation body's contents, the state-ledger round trip, and the comment-paging bound)
- `bash tests-coverage.sh alissa-tools-github-revloop` — 89% total; `state.py` 100%, `loop.py` 92%
- `bash check-style.sh alissa-tools-github-revloop` — clean
- `bash check-types.sh alissa-tools-github-revloop` — `Success: no issues found in 28 source files`
- `bash docker/claude/tests-entrypoint-config.sh` — ALL PASS
- `bash docker/claude/tests-entrypoint-ui.sh` — all four boots green (this is the one that caught the `set -e` bug)


alissa-app:
- 2026-07-28 20:07:41 UTC — `review-alissa-github-review-daemon-pr43-r1-bc5922` — round 1 of 10 — spawned
- 2026-07-28 20:23:51 UTC — `review-alissa-github-review-daemon-pr43-r2-75bebd` — round 2 of 10 — spawned
- 2026-07-28 20:35:19 UTC — `review-alissa-github-review-daemon-pr43-r3-ec5865` — round 3 of 10 — spawned

RHDZMOTA:
Note for the reviewer: one more commit landed a few minutes after the review request — `4e1d932` (page the issue-comment read, so an operator ack past comment #100 is still seen; both readers of that list failed silently on a truncated read). Review head `4e1d932`. Delivery notes in the PR body have been updated to match.

RHDZMOTA:
CI note (self-caught, fixed in \`efade8a\`) — the entrypoint-console job failed on the two previous heads, and it was a real signal worth recording rather than a flake:

\`check-entrypoint.yaml\` installs the daemon package **pinned in the Dockerfile** (`REVLOOP_VERSION=0.15.0`, the last published release), and the console sidecar it boots is that older build. `Config.build`'s unknown-key guard is fatal by design, so an older sidecar handed a config carrying this PR's new `operators` key dies at boot — the harness then reported it as "could not find the sidecar process to kill". Verified directly against the 0.15.0 wheel:

```
0.15.0 REJECTS: unknown config key(s): operators. Valid keys: hub_template, poll_interval, …
```

Fix: assert the **set** case (`ALISSA_REVIEW_OPERATORS` → `operators` array) on the console-OFF boot, where nothing real parses the generated config, and keep the **absent-when-unset** assertion on a boot that does run the sidecar — that is the direction that matters there (the default must not disturb it, and it is what the container renders until the release lands). The reasoning is in a comment at the top of case 1, since it applies to every future config-key addition.

The container README's env row now also states the knob needs `REVLOOP_VERSION` ≥ 0.16.0. Latest head `efade8a`; all five checks re-running.

docker/claude/Dockerfile
```diff
@@ -183,6 +183,11 @@ WORKDIR /workspace
 # repo needs no separator). "|" is used because repo slugs contain "/".
 #   e.g. "fahera-mx/studio.alissa.app|fahera-mx/blog.alissa.app"
 ARG ALISSA_REVIEW_REPOS=""
+# GitHub logins allowed to re-open a capped PR with an
+# `alissa-review: re-enter +N` comment, same "|"-separated convention. Empty ON
+# PURPOSE: the lever fails closed, and an operator opts in explicitly.
+#   e.g. "RHDZMOTA"
+ARG ALISSA_REVIEW_OPERATORS=""
```
alissa-app, line 194: [blocker] The image advertises a config key the library it installs will refuse to load.

This PR adds `ALISSA_REVIEW_OPERATORS` (here, plus the `ENV` block, `entrypoint.sh`, `revloop-config.sh` and the README table), so a container built from this commit with the knob set writes `"operators": [...]` into `revloop.config.json`. But line 32 still pins `ARG REVLOOP_VERSION=0.15.0`, and 0.15.0's `Config.build` rejects unknown keys:

```
$ PYTHONPATH=<origin/main src/main> python3 -c 'Config.build(p, {"repos":["a/b"], "operators":["RHDZMOTA"]})'
ValueError: unknown config key(s): operators. Valid keys: hub_template, poll_interval, round_cap, repos, agent_profile, reviewer_login, state_path, on_missing_review_task, on_missing_hub, dry_run
```

`__main__.main` catches that as `config error: …` and returns 2, `entrypoint.sh:464` runs the daemon in the foreground and `wait`s on it, so the container exits during bootstrap. The failure mode for an operator following the new README row is "the daemon stopped booting after the upgrade" — the one thing this feature exists to make less likely.

CI cannot catch it either: `tests-entrypoint-ui.sh:87` stubs `alissa-revloop`, so the boot-4 assertion added in this PR proves the config is *rendered*, not that the pinned daemon can *read* it.

The PR body's reason for holding the pin ("the release does not exist on PyPI until this merges") is already handled by the workflow — `check-entrypoint.yaml:44-47` falls back to installing the repo source tree when the pinned version is not published yet. So bumping this to `0.16.0` in this PR keeps CI green today and makes the container consistent with what it advertises. The 0.5.0 → 0.6.0 precedent cited in the PR body did not add a config key, so it does not cover this case.

If you would rather keep the pin, then the container half (ARG/ENV, `operators_lines`, `revloop-config.sh`, README row) has to land in the follow-up that re-pins, not here — the two must not be split in this direction.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -267,11 +445,26 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
-            if self.state.escalated(pr.full_name, number, pr.head_sha):
+        # The effective cap is the configured one plus every re-entry an
+        # operator has explicitly acked on THIS PR (issue #42). The sum is a
+        # local read, so it costs nothing on the common path; the GitHub scan
+        # that can DISCOVER a new ack runs only when the loop would otherwise
+        # be capped -- exactly the state an ack exists to unstick, and the only
+        # one where an extra comments fetch per poll is worth paying for.
+        granted = self.state.granted_rounds(pr.full_name, number)
+        if completed >= self.config.round_cap + granted:
+            granted = self._collect_acks(pr, granted)
```
alissa-app, line 466: [major] The ack scan runs on every poll of a capped PR even when no ack can ever be honoured.

`operators` defaults to `()` and the container bakes it empty on purpose (fail closed), so on a default deployment `_collect_acks` fetches the PR's entire issue-comment thread every poll and then discards every directive it finds at `_is_operator`. Confirmed against the test fakes with `operators=()`:

```
operators: () -> 5 polls -> 5 issue_comments fetches, 0 grants possible
```

A capped PR keeps its review request pending, so it stays in the `review-requested:@me` result set indefinitely — this is not a one-off cost, it is once per `poll_interval` forever, per capped PR. And with the paging change in this same PR each of those is now up to `COMMENT_PAGE_LIMIT` (20) REST calls rather than one, on exactly the long-lived capped-out threads the paging exists for: ~1200 requests/hour for a single pathological PR against a budget the daemon already treats as scarce (`RateLimited` is propagated all the way out to `run_forever`'s backoff).

The guard is one line — an early return in `_collect_acks` when `self.config.operators` is empty, or folding it into this condition:

```python
if self.config.operators and completed >= self.config.round_cap + granted:
```

(the `_announce_grants` call is local-only, so it can stay unconditional). Worth a test asserting `issue_comments` is never called with an empty allowlist, so the fail-closed default stays free.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -98,19 +114,78 @@
...
     "**Review loop cap-out (CR9)** — {rounds} rounds ran on this PR without "
-    "converging on `approve`. Per the alissa-code-review skill the loop does not "
-    "run past the cap and never silently merges; this needs an operator decision "
-    "(merge with a recorded waiver, direct specific fixes and re-enter with a "
-    "fresh cap, or park it).\n\n"
-    "Last verdict: `{last_state}` at `{sha}`."
+    "converging on `approve` (effective cap {cap}).{grant_note} Per the "
+    "alissa-code-review skill the loop does not run past the cap and never "
+    "silently merges, so it stops here: this needs an operator decision.\n\n"
+    "Last verdict: `{last_state}` at `{sha}`.\n"
```
alissa-app: [minor] The cap-out page now contradicts itself about which commit the last verdict covers.

`{sha}` here is `pr.head_sha[:8]` — the *current* head — while the verification hint added just below reports the head the verdict was actually written against (`last.commit_id`). When the head has moved, the two lines disagree, and this is the exact case the hint exists for. Rendered from the test fakes (`make_pr(sha="f1xedhead000")`, reviews at `0ldhead0000`):

```
Last verdict: `changes_requested` at `f1xedhea`.

**The head has moved since that verdict** (`f1xedhea`, reviewed at `0ldhead0`) — …
```

An operator reading the first line reasonably concludes the last verdict already covers the pushed fixes, which is the opposite of what the page is trying to tell them — and the whole value of this change is that the page is read at decision time.

The `at `{sha}`` phrasing predates this PR, but the contradiction is new, and `_escalate` already has `reviewed` in hand. Either use `reviewed[:8]` for the last-verdict line, or reword it to name the current head explicitly (e.g. `` Last verdict: `{last_state}` (head is now `{sha}`) ``).

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -292,9 +485,161 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
+        """
+        granted = self.state.granted_rounds(pr.full_name, pr.number)
+        for row in self.state.read_grants(pr.full_name, pr.number):
+            kind = grant_activity_kind(row["comment_id"])
+            if self.state.pinged(pr.full_name, pr.number, kind):
+                continue
+            line = (
+                f"- {_now()} — operator `{row['author']}` — re-entry ack "
+                f"(comment {row['comment_id']}) — +{row['rounds']} round(s) — "
+                f"effective cap {self.config.round_cap} → "
```
alissa-app: [minor] With more than one grant, the activity lines report the wrong before/after cap, and land newest-first in an append-only log.

`granted` is read once as the PR's *total* (line 604), so every un-announced row prints `round_cap → round_cap + total` rather than its own transition — and `read_grants` orders newest first, so the appends go out in reverse chronological order into a log whose whole contract is "the daemon appends a line each time". Two acks read in one pass:

```
- 2026-07-28 20:15:13 UTC — operator `operator-1` — re-entry ack (comment 1002) — +2 round(s) — effective cap 3 → 6
- 2026-07-28 20:15:13 UTC — operator `operator-1` — re-entry ack (comment 1001) — +1 round(s) — effective cap 3 → 6
```

The cap is claimed to go 3 → 6 twice, and the +2 is logged before the +1 that preceded it. Same shape whenever a `_append_activity` call fails and is retried on a later poll alongside a newer grant — which is precisely the retry path `grant_activity_kind` was built for, so this is not only the two-acks-at-once corner.

The DoD asks for the grant to be "appended to the activity comment (author, comment id, **before/after cap**)", so the number matters. Iterating `reversed(self.state.read_grants(...))` and carrying a running total (or storing the cumulative total on the row) fixes both halves. The existing single-grant test passes either way, so it needs a two-grant case to hold.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -292,9 +485,161 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
-    def _defer_stale_round(self, pr: PullRequest, round_: int, age: float) -> Decision | None:
+    def _is_operator(self, author: str) -> bool:
+        """Whether an ack from this author may be honoured.
+
+        Allowlist membership, case-insensitively (GitHub logins are), minus the
+        reviewer identity itself: the daemon's own escalation comment carries
+        the grammar, so an identity that could ack its own page would be able
+        to lift CR9's cap without a human ever touching it.
+        """
+        if not author or author.lower() == (self.github.login or "").lower():
```
alissa-app, line 519: [question] Should the PR author be excluded from the allowlist too, the way the reviewer identity is?

The reasoning for excluding `self.github.login` is that the daemon's own escalation quotes the grammar, so a self-ack would lift CR9's cap with nobody in the loop. The implementing agent is one step away from the same shape: it can comment on its own PR, and it has read the escalation. Today that requires an operator to have allowlisted the implementer's login, so it is opt-in — but in this workspace the implementer and the reviewer are both `alissa-app`-family identities, and a shared bot login is exactly the case the reviewer-identity guard already anticipates.

Not filing this as a defect because it is a policy call the origin task leaves open ("reuse the daemon's existing operator/identity notions"), and refusing `pr.author` outright would also block the legitimate case of an operator who opened the PR by hand. But it is worth a sentence somewhere durable — either a rejection of `pr.author` when it is not a human operator, or a comment here recording that allowlisting an agent identity is a deliberate operator choice.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
```diff
@@ -283,6 +298,72 @@ def record_escalation(self, repo: str, number: int, head_sha: str) -> None:
...
+            "WHERE repo=? AND number=?",
+            (repo, number),
+        ).fetchone()
+        return int(row["total"]) if row is not None else 0
+
+    def newest_grant(self, repo: str, number: int) -> sqlite3.Row | None:
+        """The most recently recorded ack for this PR, or None.
+
+        The escalation names it ("the re-entry granted by @x was consumed"),
+        and its timestamp is what `escalated_at` is compared against.
```
alissa-app: [nit] This docstring describes a design the PR deliberately did not take. Nothing compares `granted_at` against `escalated_at` — `capout_kind`'s own docstring says so explicitly ("Deliberately not a timestamp comparison against the escalation row"), and `_escalation_owed` dedupes purely through the ping ledger. Reads like a leftover from an earlier draft; suggest "…and it is the newest ack by `granted_at`, so a second grant supersedes the first in the page's wording."

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
```diff
@@ -283,6 +298,72 @@ def record_escalation(self, repo: str, number: int, head_sha: str) -> None:
...
+        """
+        return self._db.execute(
+            "SELECT * FROM grants WHERE repo=? AND number=? "
+            "ORDER BY granted_at DESC, rowid DESC LIMIT 1",
+            (repo, number),
+        ).fetchone()
+
+    def read_grants(
```
alissa-app: [nit] `read_grants`' unfiltered form and its `limit` parameter have no production caller — `_announce_grants` always passes `repo`/`number` and never a limit. The docstring justifies them as "a console-style read", but `webui/sources.py` gains only the `operators` echo in this PR and never reads the grants table, so the operator inbox still cannot show that a PR was re-entered. Either wire the console read in (which seems like the more useful half of the audit story the DoD asks for) or drop the unused parameters until something needs them; the sibling `read_escalations` shows the shape you would want.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -149,6 +158,18 @@ def build(
             )
 
         repos = tuple(raw.get("repos", ()))
+
+        # A bare string would iterate into single characters and silently
+        # allowlist nothing real, so it is rejected rather than accepted.
+        raw_operators = raw.get("operators", ())
+        if isinstance(raw_operators, str):
```
alissa-app: [nit] Good guard — but `repos` two lines up has exactly the same footgun and no guard: `repos: "fahera-mx/studio.alissa.app"` in the config file iterates into single characters and yields an allowlist of 30-odd one-character repo names, which then silently matches nothing in `watches()`. Since you are already paying for the check here, hoisting it into a small helper and applying it to both would close the pair. Pre-existing, so a follow-up task is a fine answer if you would rather keep this diff scoped to #42.

alissa-app (changes requested):

Fresh-instance review per the alissa-code-review skill (CR1–CR9), judged against the origin task's definition-of-done (TASK-1214531903 / issue #42), review task **TASK-1092201654**.

The library side of this is genuinely good work: the ack grammar is dull and auditable exactly as the task asked, the grant is counted per comment id rather than inferred, the once-only escalation is keyed on `(head, granted total)` instead of a wall-clock comparison, and the paging fix in `issue_comments` is a real bug found while building the feature rather than a drive-by. 350 unit tests pass on the worktree source (`PYTHONPATH=src/main`), the entrypoint config harness is ALL PASS, and the `grants` table migrates onto a legacy `state.db` cleanly. All four DoD items are delivered and test-pinned.

One blocker stops it: **the container ships the previous library**. `docker/claude/Dockerfile` still pins `REVLOOP_VERSION=0.15.0`, but the same image now advertises `ALISSA_REVIEW_OPERATORS` in its ARG/ENV and README table — and 0.15.0's `Config.build` rejects unknown config keys, so setting the new knob makes `alissa-revloop` exit 2 and the container die at boot. Verified below.

### Findings

| severity | count |
| --- | --- |
| blocker | 1 |
| major | 1 |
| minor | 2 |
| nit | 3 |
| question | 1 |

### Rubric

- **Correctness** — the four DoD items are all delivered and pinned by tests (ack grammar matrix, cap/no-cap decision paths, refusals, once-only escalation, escalation-body contents, ledger round trip, comment paging). No logic defects found in the grant accounting: I walked the cap→ack→consume→re-page→re-ack sequence and CR9's never-past-the-effective-cap and never-silently-merge invariants hold on every path I could construct, including a head move mid-sequence. Two message-accuracy defects (`[minor]`, below). The one correctness failure is the container/library version skew (`[blocker]`).
- **Coding practices** — matches the repo's idiom closely (dense why-comments, frozen dataclasses, table-driven tests, ping-ledger dedupe). Commits carry the co-author trailers, the branch is `TASK-647078429-REENTRY-ACK`, the PR names both task refs and `Closes #42`. `pycodestyle`/`mypy` are not installed in this reviewer environment, so I could not independently re-verify those two claims; the 350-test and entrypoint-harness claims I did reproduce. Three `[nit]`s.
- **Security** — the lever fails closed (empty allowlist honours nothing), the reviewer identity is excluded from the allowlist even when configured, quoted (`>`) lines are dropped, the grammar is anchored so a mention in prose is not a directive, `N` is bounded, and the grant is keyed by comment id so a re-read cannot re-grant. Comment bodies are never interpolated into shell, SQL (parameterized) or log format strings. One design `[question]` on who else should be excluded. Noted but not filed as a finding: a collaborator with write access to the repo can edit *another* user's comment, so they could insert an ack into an allowlisted operator's existing comment — but that actor can already merge the PR outright, so the ack adds no privilege.
- **Performance** — one `[major]`: the ack scan is unconditional on capped PRs, so it pays a full issue-comment fetch (now up to 20 paged requests) every poll even when `operators` is empty and no ack could ever be honoured. Everything else is a local SQLite read on the common path, as advertised.
- **Side-effects / blast radius** — `CREATE TABLE IF NOT EXISTS grants` is a safe additive migration and is tested against a pre-existing ledger; `render_revloop_config` gained an *optional* second positional arg so the existing single-arg call sites keep working; `--operator` and the `operators` key are additive; no already-escalated PR re-pages on upgrade (`_escalation_owed` short-circuits when `granted == 0`). I grepped the workspace for other consumers of `revloop.config.json` / `ALISSA_REVIEW_OPERATORS` and found none outside this repo. The blast radius that *is* real is the container one — see the blocker.

Bump `ARG REVLOOP_VERSION` to `0.16.0` (`check-entrypoint.yaml` already falls back to the repo source tree when the pin is not yet on PyPI, so CI stays green while the PR is open), and either guard the ack scan on a non-empty `operators` allowlist or justify the per-poll fetch. The two `[minor]`s are message-accuracy fixes in operator-facing text and are cheap; the nits and the question are yours to triage.

_Reviewer posture: read-only (CR6). Nothing was committed, pushed, merged, or fixed._

docker/claude/Dockerfile
RHDZMOTA: [triage:pursue] Agreed, and the reasoning in the PR body was wrong: I read the `check-entrypoint.yaml` fallback as covering only the *first* release of a renamed dist, but lines 44-47 fall back to `pip install ./alissa-tools-github-revloop` for any unpublished pin, so bumping keeps CI green today. And the split you name is the real defect — the image cannot advertise a knob the library it installs refuses. Bumping `ARG REVLOOP_VERSION` to `0.16.0` in this PR (the `needs ≥ 0.16.0` note in the README env row then becomes redundant against the image and comes back out).

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Correct, and worse than a fixed cost: 20 paged requests per poll per capped PR, on a budget the daemon backs off for. I put the guard inside `_collect_acks` rather than in the `evaluate` condition, so the local grant sum and `_announce_grants` (which can still have a retry pending from when the allowlist was non-empty) keep working, and the early return is next to the docstring that explains it. Adding a test that asserts `issue_comments` is never called with `operators=()`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] The contradiction is real and I introduced it by adding a second, differently-scoped sha to the same page. Taking the second option — naming the current head explicitly on the last-verdict line and the reviewed head in the hint — because dropping to `reviewed[:8]` alone would lose the "what is unreviewed right now" fact an operator needs to decide. The existing assertion moves with it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Both halves are bugs: the running total was read once as the PR total, and the retry path you name (a failed append landing beside a newer grant) is exactly where it bites, not just the two-in-one-pass corner. Fixing by walking `read_grants` oldest-first and carrying a running total so each line reports its own transition, with a two-grant test pinning both the arithmetic and the order.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:answer] Keeping `pr.author` eligible, deliberately. Refusing it outright breaks the ordinary case of an operator who opened the PR by hand — which is the main way a human operator ends up on a PR at all here — and the protection that matters is already structural: the allowlist is opt-in and empty by default, so an agent identity can only ack if an operator put it there. The reviewer identity is different in kind: it is not opt-in, the daemon authors the page that teaches the grammar, and a shared bot login would make it self-granting with nobody in the loop. Recording that distinction where it will be read (a sentence in `_is_operator` and the README section) rather than changing the rule.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
RHDZMOTA: [triage:pursue] Leftover from the first draft, which did compare `granted_at` against `escalated_at`; I moved to the ping-ledger key precisely because the wall-clock comparison ties, and missed the docstring. Rewriting it to your wording.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
RHDZMOTA: [triage:pursue] Dropping the unused `limit` and the unfiltered form rather than wiring the console. Reason for that direction: the payload alone would not show an operator anything (`page.py` renders named fields, so a grants read with no rendering is a half-landed feature), and the audit trail the DoD actually asks for — loud log line plus the activity-comment append — is delivered and pinned. The console inbox showing "this PR was re-entered by @x" is a real follow-up and I would rather it arrive with its rendering; filing it against the origin task rather than smuggling the data half in here.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: [triage:pursue] Fair — the guard is only half a guard if the key next to it has the same footgun. Hoisting it into a small helper and applying it to `repos` and `operators` both; a one-character-repo allowlist that silently matches nothing is the worse failure of the two, since `watches()` then quietly reviews no PR at all.

alissa-app:
**Round 1 addendum — head moved during the review (no change to the verdict).**

My review record above was written against `4e1d932` and submitted pinned to it. `efade8a` ("assert the operators knob on a boot with no sidecar (CI green)") landed at 20:13 UTC, between my diff fetch and my submit, so it is not covered by the review record. I have read it; here is what it does to the findings, so nobody has to guess:

- **`[blocker]` (Dockerfile pin) — stands, and `efade8a` is direct evidence for it.** That commit reaches the same diagnosis I did and answers it with documentation plus a test relocation rather than a pin bump: the README row now says the knob "needs `REVLOOP_VERSION` ≥ 0.16.0 (an older daemon rejects the key it does not know)", and the `ALISSA_REVIEW_OPERATORS`-set assertion moves from the sidecar boot (case 4) to the console-off boot (case 1) with the comment *"A config key newer than that pin is an unknown key to it, and the daemon's unknown-key guard is fatal by design — so asserting the SET case on a boot that starts a sidecar would fail CI for every PR that adds a config key, until the release lands."*

  That is the blocker restated by the author. A README caveat does not make the boot survivable: the image still bakes `ARG`/`ENV ALISSA_REVIEW_OPERATORS` and still installs 0.15.0, so an operator who sets the knob gets `config error: unknown config key(s): operators` and exit 2 during bootstrap. And moving the assertion off the sidecar boot narrows CI's exposure to the skew rather than removing the skew. `check-entrypoint.yaml:44-47` already falls back to the repo source tree when the pinned version is unpublished, so bumping `ARG REVLOOP_VERSION` to `0.16.0` in this PR keeps CI green *and* lets the assertion sit anywhere.

- **`[major]`, both `[minor]`s, the three `[nit]`s and the `[question]` — untouched.** `efade8a` changes only `docker/claude/README.md` and `docker/claude/tests-entrypoint-ui.sh`; no library file moves, so every line I anchored on is unchanged.

Verdict is unchanged: `request_changes`. The CR6 envelope on TASK-1092201654 records `4e1d932` as the reviewed head and carries this delta as a task comment.

RHDZMOTA:
Round-1 fixes pushed as `8a9e7eb` (triage replies are on each thread, posted before this push). Summary of what changed:

- **[blocker]** `ARG REVLOOP_VERSION` `0.15.0` → `0.16.0`. You were right on both counts — the workflow's source-tree fallback covers any unpublished pin, not just a renamed dist's first release, and the image must not advertise a key the library it installs refuses. The `needs ≥ 0.16.0` note I had added to the README env row is gone with it.
- **[major]** `_collect_acks` returns immediately when `operators` is empty, so the fail-closed default costs no comment fetch at all. New test asserts zero `issue_comments` calls across three polls of a capped PR with `operators=()`.
- **[minor]** The last-verdict line now names both shas when the head has moved (`Last verdict: \`x\` on \`0ldhead0\` — the head is now \`f1xedhea\`.`) and keeps the single-sha wording when it has not; both pinned.
- **[minor]** `_announce_grants` walks the grants oldest-first with a running total, so each line reports its own transition (`cap 3 → 4`, then `cap 4 → 6`). Two-grant test pins the arithmetic and the order.
- **[nits]** `newest_grant`'s docstring rewritten to your wording; `read_grants` narrowed to `(repo, number)` with the console follow-up noted rather than half-landed; the string-list guard hoisted into `_string_list` and applied to `repos` too (parameterized test covers both keys).
- **[question]** Answered on the thread and recorded in `_is_operator`: the PR author stays eligible, because the protection is structural (opt-in, empty by default) and refusing them would block the operator who opened the PR by hand — while the reviewer identity is excluded precisely because it is *not* opt-in.

Re-verified locally: 353 unit tests pass, coverage 89% (`state.py` 100%, `loop.py` 92%), `pycodestyle` clean, `mypy` clean, and both entrypoint harnesses green (config renderer ALL PASS, console wiring all four boots).

docker/claude/README.md
```diff
@@ -159,6 +159,7 @@ automatically; locally pass `--build-arg`):
 | ARG / env | default | meaning |
 | --- | --- | --- |
 | `ALISSA_REVIEW_REPOS` | *(required if no manifest mounted)* | allowlist as one `\|`-separated string (see below) |
+| `ALISSA_REVIEW_OPERATORS` | *(empty — no ack honoured)* | logins allowed to re-open a capped PR with `alissa-review: re-enter +N`, one `\|`-separated string; **pass-through** |
```
alissa-app, line 169: [major] The documented build recipe still pins `0.15.0` — the version that refuses the knob this row documents.

`8a9e7eb` correctly bumped `ARG REVLOOP_VERSION` to `0.16.0` and, with it, removed the `needs REVLOOP_VERSION >= 0.16.0` caveat that used to sit in this very row. But the Build section 140 lines up still shows:

```sh
# with configuration baked in (see the Configuration table):
docker build \
  --build-arg REVLOOP_VERSION=0.15.0 \
  --build-arg ALISSA_REVIEW_REPOS="..." \
  ...
```

(`docker/claude/README.md:19`). That is the recipe this table points operators at — the paragraph above it says "Set them at build time … locally pass `--build-arg`" — so the operator who reads this new row and then follows the documented way to bake it in overrides the good default with the one version that cannot read the key. Reproduced against the 0.15.0 library source (`origin/main`, version file `0.15.0`):

```
REJECTED: unknown config key(s): operators. Valid keys: hub_template, poll_interval, round_cap, repos, agent_profile, reviewer_login, state_path, on_missing_review_task, on_missing_hub, dry_run
```

`__main__.main` maps that to `config error: …` / exit 2, and `entrypoint.sh:464-466` backgrounds the daemon and `wait`s on it, so the container exits during bootstrap — the same failure mode as the round-1 blocker, reached through the repo's own docs instead of through the ARG default, and now with no caveat left anywhere to warn about it.

The example line is pre-existing (it came in with #39, where `0.15.0` matched the ARG default and no config key needed a newer daemon). What makes it a finding *now* is this PR: it adds the first key with a version floor, bumps the pin past the example, and removes the note that documented the floor.

Cheapest fix that also stops the drift recurring: drop the `--build-arg REVLOOP_VERSION=...` line from the example entirely. The ARG default is correct by construction and is the value CI reads back out of the Dockerfile (`check-entrypoint.yaml:42`), so an example that repeats it can only ever go stale again at the next release. Bumping it to `0.16.0` also closes the failure, at the cost of one more line to remember each release.

docker/claude/Dockerfile
```diff
@@ -29,7 +29,7 @@ FROM python:3.12-slim-bookworm
 # release this repo publishes on merge; it is the first version whose console
 # script `alissa-revloop-ui` the entrypoint starts (0.14.0 added the script,
 # this release wires it into the container), so the pin must not go below it.
-ARG REVLOOP_VERSION=0.15.0
+ARG REVLOOP_VERSION=0.16.0
```
alissa-app, line 36: [minor] The comment directly above this line still describes the pin it no longer holds.

```
# Version of the daemon to install from PyPI. Bump per release. 0.15.0 is the
# release this repo publishes on merge; it is the first version whose console
# script `alissa-revloop-ui` the entrypoint starts (0.14.0 added the script,
# this release wires it into the container), so the pin must not go below it.
ARG REVLOOP_VERSION=0.16.0
```

Two claims are now false: `0.15.0` is not the release this repo publishes on merge (the version file at this head is `0.16.0`, and that is what `package-publish.yaml` uploads), and "this release wires it into the container" describes #39's work, not this one. The load-bearing half — the 0.15.0 floor, because `alissa-revloop-ui` is what `entrypoint.sh` starts — is still true and worth keeping, but it now reads as a statement about the *current* pin rather than about the floor beneath it.

Suggest something like: "Bump per release; `0.16.0` is the release this repo publishes on merge. The pin must not go below `0.15.0` — the first version shipping the `alissa-revloop-ui` console script the entrypoint starts." A comment that names a version is going to drift every release; naming the *floor* separately from the *current* value is what makes the next bump a one-line edit.

docker/claude/tests-entrypoint-ui.sh
```diff
@@ -175,7 +175,17 @@ info "1. console OFF by default (ALISSA_UI_ENABLED unset)"
 # -----------------------------------------------------------------------------
 LOG1="${TMPROOT}/off.log"
 rm -f "${MARKERS}"/*
-run_entrypoint "${LOG1}"; PID1="${EP_PID}"
+# ALISSA_REVIEW_OPERATORS rides on THIS boot rather than paying for a fifth
```
alissa-app: [minor] This rationale was invalidated by the pin bump in the same commit — and the assertion it justifies is the one that would have caught the round-1 blocker.

The premise is "CI installs the version PINNED IN THE DOCKERFILE (the last published release)". As of `8a9e7eb` the pin is `0.16.0`, which is *this PR's* release, not the last published one: `check-entrypoint.yaml:44-47` falls back to `pip install ./alissa-tools-github-revloop` while `0.16.0` is unpublished, and installs the real `0.16.0` once it lands. Either way the sidecar CI boots now knows the `operators` key, so "asserting the SET case on a boot that starts a sidecar would fail CI" is no longer true here.

That matters beyond the prose, because the SET assertion is now on case 1, where — as the comment itself says — nothing real parses the generated config. `jq` reading back a file the entrypoint just wrote proves the renderer works; it cannot prove the daemon can *load* what was rendered. That gap is exactly how the round-1 blocker reached review: `tests-entrypoint-ui.sh` stubs `alissa-revloop`, so the only component in this harness that would reject an unknown key is the console sidecar, and the SET case no longer runs on a boot that starts one.

With the pin bumped, the assertion can go back to case 4 (or ride along on case 3), where a real sidecar boots against a config carrying `operators` — turning a renderer check into an end-to-end pin/library-skew check, for free.

If you would rather keep the split, the comment still needs rewriting: the general guidance it offers ("every PR that adds a config key") only holds for a PR that adds a key *without* bumping the pin, which is the practice this PR just moved away from. Worth saying that explicitly, since the comment is written to be read by the next person adding a config key.

README.md
```diff
@@ -178,6 +181,47 @@ Leave it on `skip` unless you want unattended clones.
...
+by N. `N` runs from 1 to 5 — a bigger re-entry is a second comment, not a bigger
+number.
+
+- **Counted, never inferred.** One grant per ack comment, keyed by comment id:
+  the ack sits on the PR forever and grants exactly once. A second grant needs a
+  second comment.
+- **Auditable.** The grant is logged loudly (`RE-ENTRY GRANT …`) and appended to
+  the review-loop activity comment with the author, the comment id, and the
+  before/after cap.
+- **Fails closed.** No `operators` allowlist, no honoured acks. The reviewer
```
alissa-app, line 209: [nit] The `[triage:answer]` on the PR-author question said the distinction would be recorded "in `_is_operator` and the README section". The `_is_operator` half landed in `8a9e7eb` and reads well; this bullet still only says the reviewer identity is never an operator, and nothing here tells an operator that the **PR author is deliberately still eligible** — which is the half that is surprising, and the half that matters when someone considers allowlisting an agent identity.

One sentence after "with nobody in the loop" would close it, e.g.: "The PR *author* is not excluded — an operator who opened the PR by hand is the ordinary case — so putting an agent identity on the allowlist is a deliberate choice, not something the daemon does for you."

Nit, not a defect: the answer itself is sound and it is durably recorded in the code.

alissa-app (changes requested):
**Round 2 of 10 — `request_changes`.** Head reviewed: `8a9e7eb4` (base `6348b5a`, = `origin/main`). Supersedes the round-1 verdict written against `4e1d932`.

### Prior-finding reconciliation (CR9)

Every round-1 finding carries exactly one triage reply, so the round does not bounce. Dispositions, each checked against the diff rather than the reply:

| --- | --- | --- |
| `[blocker]` Dockerfile pins 0.15.0 while advertising `ALISSA_REVIEW_OPERATORS` | pursue | **fixed** (`8a9e7eb`) — `ARG REVLOOP_VERSION=0.16.0`, matching the version file; the `needs ≥ 0.16.0` README caveat came back out with it |
| `[major]` ack scan runs every poll with an empty allowlist | pursue | **fixed** (`8a9e7eb`) — `_collect_acks` returns before the fetch when `operators` is empty; `_announce_grants` (local-only) still runs, so the append retry survives |
| `[minor]` cap-out page contradicts itself about which head the verdict covers | pursue | **fixed** (`8a9e7eb`) — `VERDICT_LINE` / `VERDICT_LINE_MOVED`, both shapes pinned |
| `[minor]` multi-grant activity lines report the wrong before/after cap, newest-first | pursue | **fixed** (`8a9e7eb`) — `reversed(read_grants(...))` with a running total; two-grant test pins arithmetic *and* order |
| `[question]` exclude the PR author from the allowlist too? | answer | **accepted** — the protection is structural (opt-in, empty by default) and refusing `pr.author` would block the operator who opened the PR by hand; recorded in `_is_operator`'s docstring |
| `[nit]` `newest_grant` docstring describes a design not taken | pursue | fixed |
| `[nit]` `read_grants`' unfiltered form / `limit` have no caller | pursue | fixed — narrowed to `(repo, number)`, console read deferred |
| `[nit]` `repos` has the same string footgun as `operators` | pursue | fixed — `_string_list` helper applied to both, parameterized test |

Nothing from round 1 is still open.

### New findings on the delta (`4e1d932..8a9e7eb`)

The blocker fix is right, and correct where it counts — but the pin bump left three places in the docs and the harness still describing the world as it was at 0.15.0, and one of them hands an operator the exact boot failure the blocker was filed for.

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 2 |
| nit | 1 |
| question | 0 |

### Dimension report

- **Correctness:** All four round-1 code fixes hold; `353 passed` locally at this head (`PYTHONPATH=src/main python3 -m pytest src/test -q`), matching the PR body, and `bash docker/claude/tests-entrypoint-config.sh` is ALL PASS. Every DoD item from the origin task remains delivered and test-pinned. The one correctness defect left is documentation that contradicts the shipped pin (`[major]` on `docker/claude/README.md`, `[minor]` on `docker/claude/Dockerfile`).
- **Coding practices:** The delta matches repo idiom — the `_string_list` hoist is the right shape, the new docstrings carry the *why*, and the two new tests assert behaviour rather than implementation. Commit `8a9e7eb` carries the co-author trailers and names the task; branch and PR conventions unchanged from round 1. `pycodestyle` / `mypy` are still not installed in this reviewer environment, so those two PR-body claims remain unverified by me; no line in the delta exceeds the repo's 120-column limit.
- **Security:** Nothing found. The empty-allowlist early return is strictly fail-closed (it can only withhold rounds, never invent them), `_is_operator` is unchanged, and `_string_list` on `repos` makes the `on_missing_hub: add` guard *stricter* (a whitespace-only entry now yields an empty allowlist and raises instead of passing the non-empty check). No secrets in the new entrypoint-harness env.
- **Performance:** The round-1 `[major]` is genuinely gone — re-ran the reproduction shape: with `operators=()` a capped PR now issues zero `issue_comments` calls across repeated polls, and the new test pins it. Nothing else in the delta touches a hot path; `_announce_grants` is a local SQLite read.
- **Side-effects / blast radius:** `read_grants`' narrowed signature has no consumer outside this PR (grepped the workspace — the table and all three readers are new here). `_string_list` on `repos` is a deliberate behaviour change on a pre-existing key: a hand-written `"repos": "org/repo"` now fails config validation loudly instead of silently watching nothing — the right direction, and the container renders `repos` as a JSON array so no container config is affected. The pin bump means `docker build` from this branch cannot resolve `alissa-tools-github-revloop==0.16.0` until the merge publishes it; that is the accepted consequence of the fix I asked for (publish-on-merge closes the window), and it is precisely why the README's *hardcoded* `0.15.0` example is now the wrong thing to leave standing.

### Waived majors

None.

Fix the `[major]` (one line in `docker/claude/README.md`). The two `[minor]`s are stale prose introduced by the fix commit and cheap to correct; the `[nit]` is discretionary.

RHDZMOTA:
- 2026-07-28 20:32:25 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr43-r2-a1` — fix edge — spawned

docker/claude/README.md
RHDZMOTA: [triage:pursue] Taking the cheaper-forever option: dropping the `--build-arg REVLOOP_VERSION=...` line from the example rather than bumping it. The ARG default is correct by construction and is the value `check-entrypoint.yaml:42` reads back out of the Dockerfile, so an example repeating it is a second source of truth that can only go stale again — and its staleness is load-bearing now that a config key has a version floor. Good catch that my own removal of the caveat is what un-warned it: I deleted the note because the ARG default made it redundant, without checking what else in that file pins a version.

docker/claude/Dockerfile
RHDZMOTA: [triage:pursue] Fair, and the suggested split is the right shape — the comment should state the floor (0.15.0, because `entrypoint.sh` starts `alissa-revloop-ui`) separately from the current pin, so the next release edits one number instead of re-reasoning the paragraph. Rewriting it that way.

docker/claude/tests-entrypoint-ui.sh
RHDZMOTA: [triage:pursue] Agreed, and the second half of your point is the important one: on case 1 the assertion only proves the renderer wrote a file, and the component that would reject an unknown key is exactly the one that boot does not start. Moving the SET case back to case 4 so a real sidecar boots against a config carrying `operators` — which turns it into the end-to-end pin/library-skew check that would have caught the round-1 blocker, for free — and keeping the absent-when-unset assertion on case 1. The comment is rewritten to say the general rule the pin bump established: a PR adding a config key bumps the pin in the same PR, and this assertion is what proves the two moved together.

README.md
RHDZMOTA: [triage:pursue] Right — I recorded the surprising half only where a reader of the code would find it, and the README section is what an operator reads before deciding to allowlist an identity. Adding the sentence there.

RHDZMOTA:
Round-2 fixes pushed as `19a472f` (triage replies posted first, one per thread). Docs and harness only — no library code changed, so the 353 tests, style and types are unchanged from `8a9e7eb` and were re-run anyway.

- **[major]** The `--build-arg REVLOOP_VERSION=0.15.0` line is gone from the documented build recipe rather than bumped, per your reasoning: the `ARG` default is correct by construction and is what CI reads back out of the Dockerfile, so a second copy in prose can only go stale again — and its staleness is load-bearing now that a key has a version floor. A short paragraph under the recipe says why it is absent and when you would pass it deliberately.
- **[minor]** The Dockerfile comment now states the current pin (`0.16.0`, matching the version file) and the floor (`0.15.0`, the first release shipping `alissa-revloop-ui`) as separate facts, so the next release edits one number.
- **[minor]** The `operators` assertion is back on case 4, where a real sidecar boots against a config carrying the key — an end-to-end pin/library-skew check rather than a renderer check, and the thing that fails if a future config key ships without its pin bump. The comment now states that rule for the next person adding a key, and case 1 keeps the absent-when-unset assertion. All four boots green locally.
- **[nit]** The README's re-entry section now says the PR author is deliberately *not* excluded, and that allowlisting an agent identity is therefore an operator's choice.

Round 3 requested at `19a472f`.

docker/claude/Dockerfile
```diff
@@ -25,11 +25,15 @@
...
-# release this repo publishes on merge; it is the first version whose console
-# script `alissa-revloop-ui` the entrypoint starts (0.14.0 added the script,
-# this release wires it into the container), so the pin must not go below it.
-ARG REVLOOP_VERSION=0.15.0
+# Version of the daemon to install from PyPI. Bump per release: 0.16.0 is the
+# release this repo publishes on merge (it must match the dist's version file,
+# or the image ships a library that cannot read the config this image renders).
+#
+# The FLOOR is separate from the current pin and moves far less often: never
+# below 0.15.0, the first version shipping the `alissa-revloop-ui` console
```
alissa-app, line 33: [minor] The floor this comment establishes is now below the image's real floor — and it is the round-1 blocker's floor.

The split into *floor* vs *current pin* is the right shape and fixes what I filed. But the number is wrong twice, and one of the two is load-bearing:

1. **The image's floor is `0.16.0`, not `0.15.0`.** After this PR the entrypoint can render `operators` into `revloop.config.json`, and `Config.build` rejects that key fatally below 0.16.0 (`config error: unknown config key(s): operators`, exit 2, at boot — I reproduced it against the installed 0.15.0 while running `tests-entrypoint-ui.sh`). "never below 0.15.0" therefore sanctions exactly the skew round 1 blocked on. It reads worse next to `docker/claude/README.md:28-33`, added in this same commit, which tells operators to pass `--build-arg REVLOOP_VERSION=…` "only when you deliberately want a version other than the pinned one": the pair reads as a supported downgrade path to a container that exits 2 as soon as `ALISSA_REVIEW_OPERATORS` is set.

2. **The console-script fact is off by a release.** `alissa-revloop-ui` first shipped in **0.14.0** — `fc41cbe` ("reviewer console sidecar `alissa-revloop-ui` (0.14.0)") is what added the entry point to `setup.py`. 0.15.0 is the release that *wired it into the container*, which is what the comment you replaced said ("0.14.0 added the script, this release wires it into the container"); the rewrite dropped a fact that was already correct.

Suggested shape — it also makes this comment agree with the rule `tests-entrypoint-ui.sh:306-309` states, which is the whole point of naming the floor apart:

```
# The FLOOR is separate from the current pin and moves far less often: never
# below 0.16.0, the first version that accepts the `operators` key this image
# can render. (The `alissa-revloop-ui` console script the entrypoint starts
# arrived earlier, in 0.14.0; the floor is set by the newest config key, so it
# moves with every PR that adds one.)
```

docker/claude/tests-entrypoint-ui.sh
```diff
@@ -295,8 +297,24 @@ info "4. sidecar killed under the container -> fail-VISIBLE, not fail-fatal"
...
 LOG4="${TMPROOT}/died.log"
 rm -f "${MARKERS}"/*
-run_entrypoint "${LOG4}" ALISSA_UI_ENABLED=on ALISSA_UI_PASSCODE="${PASSCODE}"; PID4="${EP_PID}"
+# ALISSA_REVIEW_OPERATORS rides on THIS boot rather than paying for a fifth one,
+# and specifically on a boot that starts a REAL sidecar — that makes it an
+# end-to-end pin/library-skew check rather than a renderer check. The daemon is
+# stubbed here, so the sidecar is the only component that actually LOADS the
+# generated config, and `Config.build` rejects unknown keys fatally: an image
+# whose `REVLOOP_VERSION` predates a config key it renders dies at boot, and
+# this assertion is what fails instead. Hence the rule for the next person
```
alissa-app, line 306: [minor] The skew check fires, but not through this assertion, and the failure names the wrong thing.

I ran the harness on a box with `alissa-tools-github-revloop==0.15.0` installed — which the sidecar resolution above prefers over the source tree — so case 4 boots a real 0.15.0 sidecar against a config carrying `operators`. That is precisely the pin/library skew this comment is about, and the case did fail. But it failed like this:

```
4. sidecar killed under the container -> fail-VISIBLE, not fail-fatal
  ok   generated config carries ALISSA_REVIEW_OPERATORS (split, trimmed)
  FAIL could not find the sidecar process to kill
```

The `jq` assertion **passes** — it is the renderer read-back the last sentence of this comment dismisses, and the sidecar's death does not touch it. What fails is case 4's own supervision check, timing out after 30s of `pgrep`, with a message that mentions neither the pin, the library, nor the config. So "this assertion is what fails instead" is not what happens, and someone reading a red CI run debugs sidecar supervision rather than a version skew. The diagnosis *is* in the log — `run_entrypoint` redirects the entrypoint's stdout/stderr to `${LOG4}` and the backgrounded sidecar inherits it, so `config error: unknown config key(s): operators` is sitting right there — but this `bad` message does not even point at the log the way its siblings do (`"(see ${LOG4})"`).

Cheapest fix that makes the comment true and the failure self-describing: check the log where the death is already certain, in the `else` branch of the `pgrep` loop (race-free, unlike asserting right after the `jq` line):

```sh
else
  if grep -qF "config error" "${LOG4}"; then
    bad "sidecar rejected the generated config — REVLOOP_VERSION predates a key this entrypoint renders (see ${LOG4})"
  else
    bad "could not find the sidecar process to kill (see ${LOG4})"
  fi
fi
```

Keeps case 4's own subject clean and gives the skew its own name.

docker/claude/tests-entrypoint-ui.sh
```diff
@@ -295,8 +297,24 @@ info "4. sidecar killed under the container -> fail-VISIBLE, not fail-fatal"
 # -----------------------------------------------------------------------------
 LOG4="${TMPROOT}/died.log"
 rm -f "${MARKERS}"/*
-run_entrypoint "${LOG4}" ALISSA_UI_ENABLED=on ALISSA_UI_PASSCODE="${PASSCODE}"; PID4="${EP_PID}"
+# ALISSA_REVIEW_OPERATORS rides on THIS boot rather than paying for a fifth one,
```
alissa-app, line 300: [nit] The file header's case map (lines 12-20) no longer matches what these two cases do.

This commit moved a responsibility between cases without updating the map at the top:

- case 1's line still says "(and the pass-through-when-unset config contract still holds **end to end**)" — the assertion it now carries is the right one, but nothing loads the config on that boot, so it is a renderer check, not end to end. That is the exact distinction this commit's new comment draws two hundred lines down.
- case 4's line says only "the monitor logs the death LOUDLY and the daemon + worker keep running" — nothing records that it now also carries the operators pass-through and the pin/library-skew check, which is the second reason it can go red and (per the `[minor]` above) the one whose failure message is least self-explanatory.

Two half-lines keep the map honest for the next reader, who will start at the top of the file.

README.md
```diff
@@ -178,6 +181,50 @@ Leave it on `skip` unless you want unattended clones.
...
+  identity is never an operator, however it is configured — the cap-out comment
+  itself quotes the grammar, and a daemon that could ack its own page would lift
+  CR9's cap with nobody in the loop. The PR *author* is not excluded — an
+  operator who opened the PR by hand is the ordinary case — so putting an agent
+  identity on the allowlist is a deliberate choice, not something the daemon
+  does for you. Malformed directives, out-of-range `N`,
```
alissa-app, line 215: [nit] Paragraph left un-reflowed around the insert — this line breaks at ~38 columns mid-paragraph where the rest of the file fills to ~76.

alissa-app (approved):
## Round 3 review — `19a472f` — approve

Round 2 reviewed `8a9e7eb`. Every round-2 finding carries exactly one `[triage:pursue]` reply posted before the fix push, and every one of them holds:

| --- | --- |
| `[major]` `docker/claude/README.md:19` — the documented build recipe still pinned `0.15.0` | **fixed** — the `--build-arg REVLOOP_VERSION=…` line is gone from the example, with a paragraph saying why it will not come back. Deleting rather than bumping is the better of the two options: it leaves one source of truth (the `ARG` default), which is also the value `check-entrypoint.yaml:41` greps back out of the Dockerfile. |
| `[minor]` `docker/claude/Dockerfile:36` — the comment described a pin it no longer held | **fixed**, and split into floor + current pin exactly as suggested — but the floor it now states is wrong in the direction of the round-1 blocker. New `[minor]` inline. |
| `[minor]` `docker/claude/tests-entrypoint-ui.sh:178` — invalidated rationale, assertion parked on the boot where nothing parses the config | **fixed** — the SET case moved to case 4 (real sidecar), the absent-when-unset case stays on case 1, and the comment now states the general rule. New `[minor]` inline on how that check *reports* when it fires. |
| `[nit]` `README.md:209` — the README half the `[triage:answer]` promised | **fixed** at `README.md:212-215`; it names the surprising half (the PR author stays eligible) where an operator decides whether to allowlist an identity. |

Nothing from round 1 or round 2 is still open. No blocker and no major in the delta.

### Verification

Read-only, in a throwaway worktree at `19a472f` (removed afterwards):

- `PYTHONPATH=src/main python3 -m pytest src/test -q` → **353 passed**
- `bash docker/claude/tests-entrypoint-config.sh` → **ALL PASS**
- `bash docker/claude/tests-entrypoint-ui.sh` → cases 1–3 green; case 4 fails **in this environment** as `could not find the sidecar process to kill`. That is not a defect in the PR — this box has `alissa-tools-github-revloop==0.15.0` installed and the harness prefers the installed console script over the source tree, so the new case-4 assertion caught a real pin/library skew. Confirmed directly: 0.15.0's `alissa-revloop-ui`, pointed at a config carrying `operators`, exits **2** with `config error: unknown config key(s): operators`. CI installs the pin (falling back to the repo source tree while 0.16.0 is unpublished) and will be green. The second `[minor]` below is about how that failure *presents*, not about whether it fires.
- Pin/version cross-check: `ARG REVLOOP_VERSION=0.16.0` matches the dist's `version` file (`0.16.0`); no `0.15.0` pin survives anywhere outside the Dockerfile's floor prose.
- `pycodestyle` / `mypy` are still not installed here, so those two PR-body claims remain unverified by me. No changed line exceeds the repo's limits.

One observation rather than a finding: while `0.16.0` is unpublished, `check-entrypoint.yaml`'s source-tree fallback installs a library that already knows `operators`, so **this** PR's CI run cannot actually prove the pin and the key moved together — the case-4 check starts biting from the next config-key PR, once the previous pin resolves on PyPI. That is inherent to the bootstrap fallback round 1 accepted; worth knowing before leaning on the comment's claim.

### Rubric

- **Correctness:** the three round-2 code/doc fixes hold and the suites above pass; the two `[minor]`s are inaccuracies inside the two comments this round rewrote.
- **Coding practices:** the delta keeps the repo's why-not-what comment idiom; `19a472f` carries the C1 co-author trailers and names the task; branch (C2), PR naming and PR-as-evidence (C3/C4) unchanged. Two `[nit]`s.
- **Security:** nothing found — the delta is documentation plus test wiring; the harness boot carries only the existing stub tokens and a literal allowlist, and the fail-closed default is now the case-1 assertion.
- **Performance:** nothing found — no runtime code changed this round.
- **Side-effects / blast radius:** dropping the example `--build-arg` changes no build behaviour (the `ARG` default already was the effective value). Moving the SET assertion to case 4 makes CI's contract *stricter* — it now boots a real sidecar against a config carrying `operators` — and the fail-closed default is still asserted, on case 1.

**Verdict: approve.** The two `[minor]`s and two `[nit]`s are all documentation/comment corrections; none gates the merge. Fixing them here is cheapest, but carrying them into the release change that publishes 0.16.0 is a legitimate call.

docker/claude/Dockerfile
RHDZMOTA: [triage:later] → **TASK-969871405**. Both corrections accepted as stated: the image's floor is `0.16.0` (the first version that accepts the `operators` key this entrypoint can render), and I dropped a fact that was already right — `alissa-revloop-ui` shipped in 0.14.0, 0.15.0 wired it into the container. Your suggested wording, including the "the floor moves with every PR that adds a config key" clause, goes in verbatim. Deferred rather than pushed because the approve is terminal: convergence merges the APPROVED commit, so anything I push now is orphaned on the branch rather than landing in `main`. It is comment prose with no runtime effect, and the executable guard (case 4) already fires on the skew.

docker/claude/tests-entrypoint-ui.sh
RHDZMOTA: [triage:later] → **TASK-969871405**. You are right that my comment overclaims: the `jq` line passes, and what fails is case 4's supervision check with a message that names neither the pin nor the config. Taking your `else`-branch shape exactly — checking `${LOG4}` for `config error` where the death is already certain avoids the race that asserting right after the render would introduce, and it makes both messages point at the log like their siblings. Also noting your observation that this PR's own CI cannot prove the pin and the key moved together (the source-tree fallback installs a library that already knows the key); the check starts biting from the next config-key PR, and the comment should say so rather than implying it is proven here.

docker/claude/tests-entrypoint-ui.sh
RHDZMOTA: [triage:later] → **TASK-969871405**. Fair — I moved a responsibility between cases and left the map at the top describing the old split, which is the first thing the next reader sees. Case 1's "end to end" is exactly the claim my own new comment refutes two hundred lines down.

README.md
RHDZMOTA: [triage:later] → **TASK-969871405**. Reflow to the file's ~76-column fill.

RHDZMOTA:
**Round 3: approve on `19a472f`** — converged. Recording the close-out here, since the PR body is the cross-actor record.

**Nothing further will be pushed to this branch.** An approve is terminal: convergence merges the *approved* commit, so a post-approve push lands on the branch but not in `main` — it orphans. The four non-gating comments on the approve (2 `[minor]`, 2 `[nit]`, all documentation/comment corrections) are therefore triaged `[triage:later]` on their threads and carried into **TASK-969871405**, filed downstream of the implementation task and scoped to land the way everything else does — a reviewed PR, not a direct push:

1. `docker/claude/Dockerfile` — the floor is `0.16.0` (the first version accepting the `operators` key this image can render), and `alissa-revloop-ui`'s provenance goes back to 0.14.0. My rewrite both understated the floor and dropped a fact that was already correct; the reviewer's wording, "the floor moves with every PR that adds a config key", goes in verbatim.
2. `tests-entrypoint-ui.sh` — the skew check fires, but through case 4's supervision timeout, not the `jq` assertion, and the message names neither the pin nor the config. Taking the suggested `else`-branch shape (grep `${LOG4}` for `config error`, point at the log).
3. The file header's case map still describes the pre-`19a472f` split.
4. `README.md:215` reflow.

Also recording the reviewer's observation, because it qualifies a claim in that comment: while `0.16.0` is unpublished, `check-entrypoint.yaml`'s source-tree fallback installs a library that already knows `operators`, so **this** run cannot prove the pin and the key moved together — the case-4 check starts biting from the next config-key PR. That correction is part of TASK-969871405.

One tooling note for whoever reads the logs: `alissa-pr-review` exited **1** (`request_changes`) on this round while the GitHub review was `APPROVED` on the same head — it read a stale verdict envelope off the review task. `gh pr view --json reviewDecision,latestReviews` is authoritative (`APPROVED`, latest review on `19a472f`), and that is what I acted on.

Task refs unchanged: origin **TASK-1214531903**, implementation **TASK-647078429** (`pending_validation`, PR attached as deliverable evidence), review **TASK-1092201654**, follow-up **TASK-969871405**. Merging is the operator's/convergence's call, not mine.

PR #7, opened 2026-07-20
TASK-1585494684: reviewer directives — register the review + always verdict

RHDZMOTA:
Downstream of **TASK-817578483**. Fixes a reviewer reliability gap seen on re-review. **Cross-repo (R14):** pairs with the skill PR — `fahera-mx/skills.alissa.app` branch `TASK-1585494684-REVIEWER-CLOSE-ROUND`.

## Problem
On round k>1 the reviewer sometimes (a) produces findings but never **registers** them as a GitHub review on the PR — so the daemon's submitted-review count never advances and the round never lands — and (b) ends with **no decisive verdict**. Both stall the loop.

## Fix (the literal prompt)
The skill already specifies the closing steps, so this is a compliance gap — and the directive is what the reviewer reads first. Both `ROUND_1_DIRECTIVE` and `ROUND_K_DIRECTIVE` now carry a shared `_CLOSE_THE_ROUND` contract:
1. **Submit** the review so it lands as one **registered review record** on the PR (verify via the reviews API) — findings left only in-session don't exist.
2. Close with a **decisive verdict** — `approve` or `request_changes`, never neither, never comment-only.
3. Read-only reinforced — a needed fix **is** `request_changes`, never a commit.

## Changes
- `loop.py`: `_CLOSE_THE_ROUND` folded into both directives.
- Package **0.2.0 → 0.4.0** (monotonic; assumes the pr-review 0.3.0 PR #6 lands first — rebase if merge order differs).
- **4 new tests** assert both directives carry the contract and still `.format()` cleanly.

## Validation
pycodestyle clean · mypy clean · **79 tests pass**.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #47, opened 2026-07-29
TASK-1683326771: reap idle reviewer sessions whose PR is terminal (revloop 0.16.2)

### LINKED ISSUE #46: Review-session reaper: kill idle review-pr sessions whose PR is terminal (memory leak fix)
Operational incident 2026-07-28: the shared worker container climbed past 10 GB RAM. Among the live sessions were `review-pr-296`, `review-pr-302`, and `review-pr-302-r2` — all idle, all with their PRs merged hours earlier. Root cause: revloop spawns review sessions (`review-pr-<N>` and round variants `review-pr-<N>-r<K>`) but NOTHING kills them when the PR reaches a terminal state — each idle interactive claude process holds hundreds of MB forever.

**Add a reap edge to the poll loop:**
1. Enumerate live tmux sessions matching ONLY revloop's own naming grammar (`review-pr-<N>` with optional `-r<K>` suffix). Sessions not matching are NEVER touched (other daemons share the container).
2. Resolve the PR's live state via the existing GitHub client.
3. Reap when ALL hold: PR merged or closed AND session idle AND a grace period since last session activity (default ~30 min, configurable). Kill via the same tmux layer used to spawn, per-session only (`kill-session` semantics — NEVER `kill-server`; the container is shared and kill-server nukes every lane's workers).
4. Log each reap with evidence (name, PR state, idle duration). Post-sweep, if live matching sessions exceed a configurable cap, log loudly (page-worthy).

**Locked decisions / guardrails:**
- **Never kill a busy session**, even on a merged PR — scoped post-merge re-reviews of fold commits are an established pattern; busy+terminal is logged, not killed.
- **v1 reaps on terminal PR only.** Do NOT reap superseded rounds (r1 when r2 exists) while the PR is open — the cap-out re-entry mechanism (TASK-1214531903) may re-enter earlier context; superseded-round reaping is a possible v2 with its own analysis.
- Reaping must not confuse round accounting: a reaped session's verdict envelopes already live on the review task/PR; killing the tmux session must not affect how rounds are counted or re-entered.
- Grace period gives a just-merged PR's session time to finish in-session close-out.
- Killed sessions won't do a clean-exit transcript upload — acceptable: verdicts and findings already live in GitHub reviews and task evidence.
- Reap failures (tmux error, GitHub 404) are logged and skipped — never block the main poll walk.

**Ship it:** version-file PATCH bump required in the same PR (publish workflow only fires on version change); tests for the grammar matcher, terminal resolution, and the busy/grace guards.

## Definition of Done

Revloop's poll loop includes a reap edge that kills tmux sessions matching only the review-pr naming grammar whose PR is merged/closed, the session is idle, and the grace period elapsed; per-session kills only (never kill-server); busy sessions never killed; superseded rounds on open PRs never reaped in v1; round accounting and cap re-entry unaffected by reaped sessions; reaps and holdouts logged with evidence; post-sweep cap check logs loudly; tests cover grammar, terminal resolution, and both guards; version file patch-bumped; tests/typecheck green; draft PR per conventions.

## Acceptance criteria

- [ ] Tests pin: terminal+idle+grace reaps; open-PR never reaps (including superseded rounds); busy never reaps; non-matching names never touched; kill is per-session (no kill-server anywhere in the change).
- [ ] Evidence that round counting and cap re-entry behave identically whether or not earlier round sessions were reaped.
- [ ] Deployed container observed reaping a real merged-PR review session (log line with evidence) and review-session count returns to baseline after a merge wave.

Alissa-Task: TASK-793080321

RHDZMOTA:
Closes #46

| | |
| --- | --- |
| Implementation task | `TASK-1683326771` (this actor's partition, soft-dependent on the origin) |

## What changed

Revloop's poll sweep grows a **terminal-PR reap edge** for reviewer sessions it
never spawned, and its notion of "a session I may kill" becomes a grammar rather
than a prefix.

- **Ownership is a parse, not a prefix.** `alissa.parse_session_name` accepts
  exactly two shapes — `review-<repo>-pr<n>-r<k>-<nonce>` (this daemon's
  `session_name`) and `review-pr-<n>[-r<k>]` (the shape the alissa-code-review
  procedures spawn by hand). `list_review_sessions` filters on it, so a name
  belonging to another lane in the shared container is never even enumerated.
  Producer and parser share `session_repo_slug`, so they cannot drift.
- **Ledger-less sessions become reapable, on a terminal PR only.** With no spawn
  row the session name is the only evidence, so the number it carries is
  resolved against the `repos` allowlist — which doubles as the ownership
  boundary. A name carrying a repo picks its allowlist entry; a bare
  `review-pr-<n>` is probed across the allowlist and must hit **exactly one** PR.
  Zero or several hits is a guess, and the sweep spares rather than guesses.
- **Guards.** Busy is never killed, even on a merged PR. Idle-and-quiet for
  `reap_grace_seconds` (new, default 30 min) is required. For a ledger-less
  session the PR must be merged or closed — an open PR spares every round of it,
  superseded rounds included.
- **Observability.** Each reap logs name + PR state + idle duration; holdouts log
  at debug with which guard spared them; a failed fetch or kill is logged and
  skipped, never fatal to the walk. After each pass, more than
  `reap_session_cap` (new, default 6) live reviewer sessions logs `ERROR`.
- Version file `0.16.1` → `0.16.2`; both knobs are container pass-through
  (env var > library default) and documented in both READMEs.

### Judgment calls

- **The issue's `review-pr-<N>` grammar is not what this daemon emits.** The
  daemon has emitted `review-<repo>-pr<n>-r<k>-<nonce>` since the collision-proof
  session rename. I checked the incident container: `review-pr-296`,
  `review-pr-298`, `review-pr-302`, `review-pr-302-r2`, `review-pr-305` were all
  live and idle, and that grammar is written down in the alissa-code-review
  skill's own `spawn-a-reviewer-session.md` / `run-the-review-loop.md`. So the
  issue is describing *hand-spawned* rounds of the same loop. The matcher
  therefore accepts **both** shapes — dropping the daemon's own would have made
  the sweep stop recognising the sessions it spawns.
- **The pre-existing finished-round sweep on open PRs was KEPT.** The issue reads
  as though nothing reaps today; a ledger-based sweep has been merged since
  TASK-1609166327 and reaps a session whose *round* is complete even while the PR
  is open, with the round read off the spawn ledger. AC-1's "open-PR never reaps
  (including superseded rounds)" is implemented for the **new, ledger-less** edge
  — where the name genuinely cannot tell a superseded round from an in-flight
  one — and I did not delete the ledger-based path, which grounds the same
  decision in recorded round accounting rather than in a name. Removing it would
  have deleted ~8 merged regression tests (including "the starved case": an
  approved PR leaves the review-requested search, so nothing else ever frees that
  slot) and re-opened the leak it was written for. **If the intended reading is
  that the ledger path must go too, say so and I will strip it** — it is a
  deliberate, reversible deviation, not an oversight.
- **Grace default 5 min → 30 min, and one knob for two readers.** The old sweep
  used a hardcoded 5-minute quiet period; the issue asks for ~30, configurable.
  The stale-round liveness probe asked the same question ("is this session
  finished?") through the same constant, so it now reads the same config key.
  Effect on the stale path: an idle, dead-looking session is respawned over after
  30 minutes of quiet instead of 5 — strictly the safer direction (it defers
  longer before risking a double-spent round), and the stale timer ahead of it is
  90 minutes regardless.
- **`reap_session_cap` default 6.** The issue does not name a number. Six is an
  alarm threshold, not a capacity limit: a healthy loop runs a couple of
  concurrent rounds, and the incident container held five idle review sessions
  plus other lanes.
- **Busy sessions are logged without resolving their PR.** The issue asks for
  "busy + terminal is logged, not killed". Resolving the PR state of a session we
  will never kill costs one API fetch per busy reviewer per 60-second poll, so the
  log line records "busy — never reaped" without the fetch. The guarantee (never
  killed) is unconditional and tested; only the terminality detail is absent from
  that line.
- **Patch bump, not minor.** The issue mandates a PATCH bump and I followed it,
  though the change adds two config keys — by this repo's recent practice
  (`0.16.0` for the re-entry ack) that would ordinarily be a minor.
- **Container env pass-through is a small, deliberate excursion** beyond the
  issue's declared scope (`ALISSA_REAP_GRACE_SECONDS`, `ALISSA_REAP_SESSION_CAP`
  in `revloop-config.sh` / `Dockerfile` / entrypoint tests). Without it
  "configurable" is unreachable in the deployed container, which renders its
  config file from the environment on every boot.
- **`webui.is_managed` still uses the `review-` prefix.** It labels rows in the
  console's session table and kills nothing, so tightening it would only change
  what an operator sees. Left alone on purpose.

- [ ] **RULING NEEDED — the ledger-backed open-PR reap.** Issue #46's locked
      decisions say "v1 reaps on terminal PR only"; the merged ledger path
      (TASK-1609166327) reaps a *ledger-backed* round whose verdict already
      landed while the PR is still open, and I kept it. Round 1's reviewer read
      the argument, said "on the evidence I'd keep it", and asked for the
      operator's ruling in writing because the origin task sits in a partition
      neither of us can read. **Rule at merge:** keep it (no change), or strip
      it — which is a scope change, and I would take it as its own task rather
      than fold it into this branch. Reasoning is in *Judgment calls* below and
      on the `[question]` thread.
- [ ] **AC-3, the whole of it.** Nothing here was run against the deployed
      container: no real merged-PR reviewer session was observed being reaped, and
      no review-session count was watched returning to baseline after a merge
      wave. There is no docker daemon in this runner and the deployed image pins
      an older `REVLOOP_VERSION`, so this is inherently post-deploy evidence.
- [ ] **The reap log line in production.** The evidence format (name, PR state,
      idle duration) is asserted in a unit test against a fake; it has never been
      emitted by the real logger with real `alissa tmux ls` output.
- [ ] **`alissa tmux ls --json` field contract.** Read live in this container
      (`name`, `status`, `lastActivity`, `live`) and matched against the existing
      parser, but never exercised end-to-end by the daemon in this session. A
      session whose `status` is neither `idle` nor `busy` (e.g. `gone`, which the
      non-`--live` listing does report) is treated as not-idle and spared.
- [ ] **The bare-name allowlist probe against real GitHub.** `_fetch_pr`'s 404
      handling is exercised with a faked `CommandError`; the real `gh api` failure
      shape for a missing PR was not observed here.
- [ ] **The cap `ERROR` line as an actual page.** It is a log line; whether
      anything alerts on `ERROR` in the deployed container is an operator fact I
      cannot check.
- [ ] **Container knobs at build time.** `revloop-config.sh` renders correctly
      (its test suite passes here), but no image was built, so the new `ARG`/`ENV`
      pair has not been through a real `docker build`.
- [ ] **The Dockerfile still pins `REVLOOP_VERSION=0.16.0`.** Deliberately not
      touched here — the re-pin is tracked separately as `TASK-1160071068`, out of
      this branch's scope (that task targets `0.16.1`; it now wants `0.16.2`) — so the new behaviour ships only once that pin moves
      past `0.16.2`.

### Verification

Run from the task worktree against the worktree source
(`PYTHONPATH=…/src/main`), all green:

- `bash tests-unit.sh alissa-tools-github-revloop` — **385 passed** (353 before;
  +32 new, no test deleted, no assertion weakened).
- `bash tests-coverage.sh alissa-tools-github-revloop` — TOTAL 90%;
  `loop.py` 93%, `alissa.py` 81%, `config.py` 97%.
- `bash check-style.sh alissa-tools-github-revloop` — clean (pycodestyle).
- `bash check-types.sh alissa-tools-github-revloop` — `Success: no issues found
  in 28 source files` (the `annotation-unchecked` notes on the test module are
  pre-existing).
- `bash docker/claude/tests-entrypoint-config.sh` — ALL PASS, including the new
  omitted-when-unset and override-wins assertions for both knobs, and the
  existing cross-check that omitted keys resolve to the library default.
- `bash docker/claude/tests-entrypoint-ui.sh` — all checks passed (untouched
  area, run as a regression check).

Not run: `docker build` and any container boot (no docker in this runner); the
GitHub Actions workflows themselves.

New tests worth naming: the grammar's accept/reject table is parametrised over
the **real** session names from the incident container; the round-accounting
invariance test (AC-2) decides the same PR twice, once with its round-1 session
reaped and once without, and asserts an identical decision, round and granted
re-entry total; and the no-`kill-server` test walks every literal `argv` in the
package with `ast` rather than grepping the call site.

### Round 1 (`request_changes`) — what changed

One `[major]`, three `[minor]`, three `[nit]`, one `[question]`. Triage posted on
every thread before this push, per the loop's ordering rule.

| finding | triage | outcome |
| --- | --- | --- |
| `[major]` allowlist probe repeats every poll (~840 REST calls/h per bare name at 7 repos / 30s) | `pursue` | probe answer (incl. a sticky negative) cached per session name, pruned against the live list; 6+1+1 fetches over 3 polls instead of 18 |
| `[minor]` "ownership boundary" claimed more than the code proves | `pursue` | docstring + README say what holds; over-reap **and** the PR-number-collision under-reap named as v1 limits |
| `[minor]` `reap_grace_seconds` silently wedges the stale-round respawn above 90 min | `pursue` | `STALE_ROUND_SECONDS` moved to `config` (re-exported from `loop`), `Config.build` now rejects the coupling |
| `[minor]` holdout lines are debug, invisible in the container | `pursue` | every spare records a reason; the cap `ERROR` interpolates them |
| `[nit]` cap `ERROR` fires 120×/hour | `pursue` | in-process dedupe on the survivor set (not the pings ledger — rationale on the thread) |
| `[nit]` `al and w.github` is a non-guard | `pursue` | collapsed |
| `[nit]` `is_managed` / removed `REAP_QUIET_SECONDS` noted for the record | `answer` | acknowledged, no change |
| `[question]` ruling on the ledger-backed open-PR reap | `answer` | answered on the thread; promoted to the operator's-gate checklist above |

Re-verified after the fixes: **392 passed** (+7), coverage 90% total / `loop.py`
93%, pycodestyle clean, mypy clean, `tests-entrypoint-config.sh` ALL PASS.

### Round 2 (`approve`) — where its findings went

Round 2 verified every round-1 triage and fix, re-ran the suites itself, and
approved on `d73214f`. It attached three non-blocking findings (2 `[minor]`,
1 `[nit]`), all in the round-1 caching code, and noted its actor cannot register
tasks.

**Nothing further was pushed to this branch** — an approve is terminal and
merge-racy. All three are triaged `[pursue]` on their threads and ship as
**PR #48**, registered as **`TASK-2006119747`** (soft-dependent on both task refs
above). The headline one is a regression round 1's fix introduced: the probe
cache removed the per-poll self-healing, so a transient `gh` failure during the
single probe could pin a wrong resolution for the session's lifetime and kill an
unrelated live reviewer. PR #48 makes a definite 404 distinguishable from a
failed call and refuses to cache an error-tainted pass.

`TASK-1160071068` (the `REVLOOP_VERSION` re-pin) now wants `0.16.3`, not
`0.16.2`, since PR #48 bumps again.

### Scope touched

Inside the issue's declared scope (the poll loop, tests, version file):

- `…/revloop/alissa.py` — the grammar (`SessionRef`, `parse_session_name`,
  `session_repo_slug`), `ManagedSession.ref`, tightened `list_review_sessions`.
- `…/revloop/loop.py` — `sweep_sessions` rewritten around the two edges, new
  `_resolve_pr` / `_name_candidates` / `_reap_reason` / `_check_session_cap` /
  `_fetch_pr` (replacing `_sweep_pr`); `_completed_rounds` loses its terminal
  branch (the caller answers that first); `session_name` builds its repo
  component through the shared slug helper; the stale-round liveness probe reads
  the new grace knob.
- `…/revloop/config.py`, `…/revloop/__main__.py` — the two keys, their
  validation, and their CLI flags; `STALE_ROUND_SECONDS` moved here from
  `loop.py` (re-exported there, so no import site changed) so the grace knob
  can be validated against it.
- `…/revloop/version` — `0.16.1` → `0.16.2`.
- `…/test_revloop/test_loop.py` — +39 tests (32 in round 1, 7 in round 2); existing sweep tests re-pointed
  from the removed `REAP_QUIET_SECONDS` constant to `config.reap_grace_seconds`,
  and the `FakeAlissa` session lister now filters on the real grammar.

Excursions, called out:

- `docker/claude/{Dockerfile,revloop-config.sh,tests-entrypoint-config.sh,README.md}`
  — env pass-through for the two new knobs (rationale above).
- `README.md`, `revloop.config.example.json` — the settings table, the example
  config, and a rewritten *Reaping finished reviewer sessions* section.

Untouched on purpose: `webui/` (the console's `is_managed` labelling),
`ghclient.py`, `state.py`, and the `REVLOOP_VERSION` pin in the Dockerfile.


alissa-app:
- 2026-07-29 06:36:09 UTC — `review-alissa-github-review-daemon-pr47-r1-5fb2db` — round 1 of 22 — spawned
- 2026-07-29 06:58:42 UTC — `review-alissa-github-review-daemon-pr47-r2-59203e` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -892,43 +946,186 @@ def sweep_sessions(self) -> int:
...
+            log.debug(
+                "reap sweep: %s has no ledger row and no watched repo matches "
+                "its name — sparing it",
+                ses.name,
             )
-        return reaped
+            return None
+
+        hits = []
+        for repo_slug in candidates:
```
alissa-app, line 1069: `[major]` **This probe repeats every poll, forever, for every session it spares.**

`prs` is a per-*pass* memo, so nothing survives to the next poll. The sessions this loop exists
for — bare `review-pr-<n>` — are spared while their PR is open (`_reap_reason`, correctly), which
means the full allowlist sweep re-runs on each poll for the entire life of the PR. That is the
common case, not the edge: a review loop's PR is open for most of its life, and idle-for-30-min is
the *normal* state of a finished round waiting on the implementer.

Against the deployed `revloop.config.json` (7 repos, `poll_interval: 30`):

- 7 fetches × 120 polls/hour = **840 REST calls/hour per distinct bare-name PR number**
- the 2026-07-28 incident shape (`review-pr-296/298/302/302-r2/305`, 5 distinct numbers) ≈
  **4,200 calls/hour** — on top of the search + per-PR fetches `evaluate()` already spends,
  against a 5,000/hour authenticated limit

and `_fetch_pr` re-raises `RateLimited`, so the failure mode is `run_forever` backing off and the
review loop stalling — a new operational risk of the same class as the one this PR fixes.

Reproduced against this branch (6 repos, one spared bare-name session on an open PR):

```
polls: 3   gh.pr_fetches: 18      # 6 per poll, no decay
```

The per-pass memo and the busy/grace pre-filter are both right; what is missing is memory *across*
passes. A dict on `ReviewWatcher` keyed by session name, holding the resolved `(repo, number)` or a
sticky "unresolvable" marker, makes a spared session cost one probe per lifetime instead of one per
poll — session names are unique per spawn (that is what the nonce is for) and a session's PR never
changes, so the cache needs no invalidation beyond dropping names that left the live list.

The docstring already says a bare name "costs one fetch per watched repo"; it is the *per poll,
indefinitely* part that turns a documented cost into a budget problem.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -892,43 +946,186 @@ def sweep_sessions(self) -> int:
...
+            # Quiet: probing a repo that simply has no PR #n is the expected
+            # outcome for all but one candidate, not a failure worth a warning.
+            found = self._fetch_pr(prs, repo_slug, ref.number, quiet=True)
+            if found is not None:
+                hits.append(found)
+        if len(hits) != 1:
```
alissa-app, line 1074: `[minor]` **The exactly-one rule bounds ambiguity inside the allowlist, but ownership is still inferred from a bare integer** — and the docstring above ("the allowlist ... doubles as the ownership boundary: the sweep only ever reaps a session it can tie to a repo this daemon is responsible for") and the README ("which doubles as the ownership boundary") both claim more than the code establishes.

Two directions, both reachable:

1. **Over-reap.** A `review-pr-<n>` session reviewing a PR in a repo that is *not* in the allowlist
   is killed whenever exactly one watched repo happens to have a terminal PR #n. Confirmed on this
   branch: allowlist `{acme/widgets}`, a merged `acme/widgets#7`, a session `review-pr-7` that
   belongs to something else entirely → `killed == ["review-pr-7"]`. The name carries no repo, so
   nothing in the resolution can distinguish "mine" from "same number, different repo".
2. **Under-reap, which is the one that bites this deployment.** The allowlist now holds 7 repos
   including `forms.alissa.app`, `slides.alissa.app` and `docs.alissa.app` — new repos whose PR
   numbers start at 1 and will overlap `alissa-github-review-daemon` (#47),
   `alissa-github-orchestrate-daemon` (#49) and `alissa-github-develop-daemon` (#71). Once two
   watched repos both have a PR #n, every bare `review-pr-<n>` resolves to 2 hits and is spared
   forever — the reaper silently stops reaping the exact shape issue #46 was filed about, while
   still paying the probe cost from the `[major]` above.

Sparing on ambiguity is the right default and I'm not asking you to guess. What I'd ask for is
(a) softening the two ownership claims to what actually holds ("the number resolves inside the
allowlist"), and (b) treating the collision-blind spot as a known v1 limit in the README's
*Reaping finished reviewer sessions* section, so the next operator who watches two repos with
overlapping PR ranges knows why the count stopped falling. The durable fix is out of scope here:
it needs the repo in the name (a skill-side change to `spawn-a-reviewer-session.md`) or a ledger
row for hand-spawned sessions.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -180,6 +210,16 @@ def build(
                 f"poll_interval must be >= {MIN_POLL_INTERVAL} seconds, got {interval}"
             )
 
+        grace = int(raw.get("reap_grace_seconds", cls.reap_grace_seconds))
+        if grace < 0:
```
alissa-app, line 227: `[minor]` **`reap_grace_seconds` now gates a second, unrelated mechanism, and the validation doesn't know it.**

Folding the stale-round liveness probe onto the same knob is defensible — it is the same question —
but it makes one config value silently control `_defer_stale_round`'s respawn decision, whose own
window (`STALE_ROUND_SECONDS = 90 min`) is a *separate constant*. Set `reap_grace_seconds` above it
and the "idle-finished → dead → respawn" branch becomes unreachable: the session is always
"quieter than the stale window but not quieter than the grace", so every stale round defers.

Reproduced on this branch with `reap_grace_seconds = 4h` and a session that has been idle and quiet
for 2h with its round 4.5h stale:

```
Action.IN_FLIGHT | round 1 is stale (270 min) but session review-widgets-pr7-r1-… is still active
                 | — not respawning over a live reviewer
```

The round never respawns. It degrades gracefully (the `STALLED_DEFER_MULTIPLE` operator ping still
fires, and the default 1800 s is comfortably under 5400 s), which is why this is `[minor]` and not
higher — but the validation here only checks `>= 0`, so the constraint is invisible at config time.
A `reap_grace_seconds >= STALE_ROUND_SECONDS` check (or a documented "must be well under the 90-min
stale window" in the README row) would make the coupling fail loudly instead of quietly.

README.md
```diff
@@ -229,19 +231,58 @@ grant).
...
+| guard | why |
+| --- | --- |
+| the session is **idle** | a busy reviewer is never killed, *even on a merged PR* — scoped post-merge re-reviews of fold commits are an established pattern; busy is logged, not reaped |
+| quiet for `reap_grace_seconds` | a claude session between turns also reports "idle"; the grace period leaves a just-merged PR's reviewer time to finish its close-out (CR6 envelope, task move) |
+| terminal PR, for a ledger-less session | the name's `-r<k>` cannot tell a superseded round from an in-flight one, and an operator re-entry may still want the earlier context — superseded-round reaping on an *open* PR is out of scope |
+
+Kills are always **per session** (`alissa tmux kill <name>`), never a server-wide
+kill: the container is shared, and a test pins that no other kill verb exists
+anywhere in the package. Reaps are logged with evidence (name, PR state, idle
+duration); holdouts and undecidable sessions are logged at debug and retried next
```
alissa-app: `[minor]` **The holdout lines this sentence points operators at are not emitted in the deployed container.**

`__main__.py` sets `level=logging.DEBUG if args.verbose else logging.INFO`, and
`docker/claude/entrypoint.sh:464` launches `alissa-revloop --workspace-root "${WORKSPACE_ROOT}" "$@"`
with no `-v`. So in the container every `log.debug` added by this PR — the busy holdout, the grace
holdout, the "resolves to N watched repos" spare, the quiet `_fetch_pr` miss — is invisible.

That matters more than usual here because the cap-`ERROR` line (which *is* INFO-visible) is the
alarm, and this paragraph plus `_check_session_cap`'s docstring ("the sweep's own holdout lines
(debug) say which guard spared each one") both tell the operator that the explanation for the alarm
is in the log. It isn't. The origin task's DoD asks for "reaps **and holdouts** are logged with
evidence" — reaps are INFO and satisfy it, holdouts as shipped don't.

Cheapest fix that keeps the noise down: when `_check_session_cap` is about to fire, log the spare
reasons for the surviving sessions at the same `ERROR`/`WARNING` level (the alarm is rare, so the
volume is bounded), or add an `ALISSA_LOG_LEVEL` pass-through so `-v` is reachable without a
redeploy edit.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -892,43 +946,186 @@ def sweep_sessions(self) -> int:
...
+        if key not in completed_cache:
+            completed_cache[key] = self._completed_rounds(pr, row["task_ref"])
+        completed = completed_cache[key]
+        if completed is None or row["round"] > completed:
+            log.debug(
+                "reap sweep: round %d of %s is not finished (%s completed) — holding %s",
+                row["round"], pr.slug, completed, ses.name,
+            )
+            return None
+        return f"round {row['round']} of an open PR is done"
```
alissa-app, line 1133: `[question]` **This branch is the deviation you flagged, and it is worth an explicit ruling rather than an implicit one.**

Issue #46's *Locked decisions / guardrails* says: "**v1 reaps on terminal PR only.** Do NOT reap
superseded rounds (r1 when r2 exists) while the PR is open". `TASK-1683326771`'s own
definition-of-done reads the same way — "reaps only sessions ... whose PR is merged/closed". This
line reaps a ledger-backed session on an **open** PR.

Your reasoning in the PR body is good and I am not asking you to strip it on my say-so: the
ledger path decides from recorded round accounting rather than from a name, it predates this task
(TASK-1609166327), removing it would delete ~8 merged regression tests including the starved-slot
case, and CR3 requires a *fresh* instance per round anyway — so reaping a round whose verdict
already landed cannot cost an operator re-entry anything the loop wouldn't discard regardless. On
the evidence I'd keep it.

But it is a documented deviation from a **locked** decision on the origin issue, and the origin
task is in a partition this reviewer can't read, so I can't tell whether the locked wording was
written in ignorance of the ledger path or in spite of it. Please get the operator's ruling on the
thread (`[triage:answer]` is enough if the answer is "keep it") so the next round has it in
writing. If the ruling goes the other way, that is a scope change, not a fix.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -892,43 +946,186 @@ def sweep_sessions(self) -> int:
...
+        minus what it killed, so the number is "sessions this pass could not
+        free". Every idle agent session holds hundreds of MB forever and the
+        worker container is shared, so a count that stays above the cap is the
+        2026-07-28 incident happening again -- and the sweep's own holdout
+        lines (debug) say which guard spared each one.
+        """
+        killed = set(reaped)
+        remaining = sorted(s.name for s in sessions if s.name not in killed)
+        if len(remaining) > self.config.reap_session_cap:
+            log.error(
```
alissa-app: `[nit]` This `ERROR` fires on **every** pass while the count is above the cap — every 30 s in the
deployed config, with the full session list interpolated. Everywhere else the daemon has a
recurring-notification problem it dedupes through `state.pinged(...)` / `record_ping(...)` (the
stalled-round escalation, the deferral activity comment). A `reap_cap_kind(...)` ping keyed on the
surviving-name set would give the same page once per episode instead of 120 times an hour, and
would re-fire when the set changes.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -1432,6 +1459,353 @@ def test_sweep_falls_back_to_review_count_without_a_task_ref(config):
...
+
+def test_cap_check_counts_what_the_sweep_left_behind(config, caplog):
+    """Reaped sessions are not 'live after the sweep' — a pass that cleared
+    itself back under the cap must not page."""
+    cfg = dataclasses.replace(_watched(config, SLUG), reap_session_cap=2)
+    merged = make_pr(state="closed", merged=True)
+    w, _, al = watcher(cfg, merged, [])
+    for number in (1, 2, 3):
+        _live(al, f"review-pr-{number}")
+    _resolve_prs(al and w.github, {(SLUG, n): merged for n in (1, 2, 3)})
```
alissa-app: `[nit]` `al and w.github` evaluates to `w.github` (`al` is always truthy) — it reads as a guard
that isn't one. `_resolve_prs(w.github, ...)` says the same thing.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -44,10 +44,87 @@ def is_open(self) -> bool:
...
 
 
-# The namespace loop.session_name() spawns reviewers into; the reap sweep only
-# ever considers sessions under it.
+# The namespace loop.session_name() spawns reviewers into. Kept as the
+# display-level answer to "is this session in the reviewer namespace?" (the
+# console marks unmanaged sessions with it); the REAP path uses the stricter
+# grammar below, because a prefix is not a strong enough claim of ownership to
+# kill on.
 REVIEW_SESSION_PREFIX = "review-"
```
alissa-app, line 52: `[nit]` Two small consequences of keeping this constant, both fine to leave — noting them so they
are on the record rather than rediscovered:

- `webui/sources.py:108` still answers `is_managed` from the prefix, so the console can now label a
  session "managed" that the sweep will never enumerate, let alone reap. The new comment here says
  it is display-level, which is the right call; the console's own column just now means something
  slightly weaker than it did.
- `REAP_QUIET_SECONDS` was a module-level public name in `loop.py` and is gone in a **patch**
  release, alongside two new config keys. You already argued patch-vs-minor in the body and the
  issue mandates PATCH, so I'm not asking for a change — but the removed export makes the
  minor-bump case slightly stronger than the body puts it.

alissa-app (changes requested):
## Review — round 1 of cap 22 · `request_changes`

Fresh reviewer instance (CR3). Reviewed `70558636e730081207344469df7354831306a6bd` against
`origin/main` @ `09e4d0f`, and against the **origin task's** definition-of-done (issue #46,
hydrated from `TASK-1683326771.upstreamTasks[]` — `TASK-793080321` itself is not readable by this
actor). Review task: **`TASK-426089574`**.

**Verified locally, not just taken from the body** (worktree at the PR head, `PYTHONPATH=src/main`):
`385 passed`, `docker/claude/tests-entrypoint-config.sh` **ALL PASS**, and all six required checks
are green on the PR. The claimed test counts hold.

This is careful, well-argued work — the grammar-over-prefix tightening is the right call, the
producer/parser shared slug closes the drift hole, the AC-2 round-accounting invariance test is
exactly the right test to have written, and the deviations are disclosed rather than buried. One
`[major]` keeps it from approve.

### The blocking one

**`[major]` the bare-name allowlist probe is unbounded in time.** `_resolve_pr` re-probes the whole
`repos` allowlist on *every* poll for every idle ledger-less session, and the memo (`prs`) is
per-pass only. A `review-pr-<n>` session on an **open** PR is spared by design — so it is re-probed
forever. Against the deployed `revloop.config.json` (7 repos, `poll_interval: 30`) that is
**840 GitHub REST calls/hour per distinct bare-name PR number**; the five-session shape from the
2026-07-28 incident is ~4,200/hour against a 5,000/hour limit, and `RateLimited` propagates out of
`sweep_sessions` into `run_forever`'s backoff, stalling the whole loop. I reproduced it: one spared
bare-name session, 6 watched repos, 3 polls → 18 fetches, growing linearly with polls. Details on
the thread.

### Rubric

| dimension | outcome |
| --- | --- |
| Correctness | AC-1/AC-2 hold and are tested; two resolution edges are heuristic — see the `[major]` and `[minor]` on `_resolve_pr`. One `[question]` on a deliberate deviation from a locked decision. |
| Coding practices | Matches repo idiom closely (docstring density, guard-then-log, `dataclasses.replace` in tests). Branch/PR/trailer conventions all hold. 2 nits. |
| Security | Net improvement — the kill surface narrows from a prefix to a parse, and `kill_session` stays an argv list with a per-session verb pinned by an AST walk. Nothing user-controlled reaches a shell. One ownership-inference caveat, `[minor]`. |
| Performance | The `[major]`. Everything else (busy/grace filtered before any fetch, per-pass memo) is measured and right. |
| Side-effects / blast radius | Config/`ARG`/`ENV` additions are purely additive; `list_review_sessions`'s narrowing also feeds `_defer_stale_round` and is safe there. One coupling risk, `[minor]`. |

Bound the bare-name probe across polls (cache the resolution — or the failure — per session name,
so a spared session costs at most one probe per lifetime, not one per poll). The `[minor]`s can be
follow-up tasks if you'd rather; the `[question]` needs an answer, not a fix.

RHDZMOTA:
- 2026-07-29 06:51:06 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr47-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — accepted, and the arithmetic is right: the memo is per-pass, the spared-on-open-PR case is the *common* case, and 7 repos × 120 polls/hour is a budget problem, not a documented cost. Fixing exactly as you describe: a `_probe_cache` on `ReviewWatcher`, keyed by session name, holding the resolved `(repo, number)` **or a sticky `None`** for “this name could not be resolved”. A session name is unique per spawn and its PR never changes, so the only invalidation needed is dropping names that left the live list — done at the top of the sweep against the freshly-listed names.

Cost after the fix, for one spared bare-name session over 6 watched repos across 3 polls: 6 probes on the first poll, then 1 state fetch per poll (the PR state still has to be re-read — that is the fetch that can turn the session reapable), so 8 instead of 18. An *ambiguous* name costs 6 once and then 0 forever, because the sticky `None` short-circuits before any fetch. Both shapes get a test asserting `gh.pr_fetches` across three sweeps.

One deliberate limit I want on the record: the sticky `None` never re-probes within a process, so a name that is unresolvable now stays spared even if the ambiguity later clears. That is the safe direction (spare, never guess), it resets on daemon restart, and re-probing on the chance it improved would reintroduce the cost you just measured.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — you are right that the prose claims more than the code proves, and that is my wording to fix, not yours to work around. The exactly-one rule bounds *ambiguity inside the allowlist*; it does not establish ownership, because a bare integer carries no repo. Both directions you name are real, and I reproduced the under-reap one mentally against the 7-repo allowlist — `forms.`/`slides.`/`docs.alissa.app` starting at #1 will collide with the daemons’ current PR ranges, and from that day every bare `review-pr-<n>` in the overlap resolves to 2 hits and is spared forever.

Doing (a) and (b) exactly as asked: the `_resolve_pr` docstring and the README both drop “doubles as the ownership boundary” for what actually holds — the allowlist *bounds the search*, ownership is inferred, and sparing on ambiguity bounds the blast radius rather than proving anything. The collision blind spot goes into the README’s *Reaping finished reviewer sessions* section as a named v1 limit, with the symptom spelled out (“the count stops falling”) so the next operator does not have to rediscover it, and with the two durable fixes you name (repo in the session name, i.e. a skill-side change to `spawn-a-reviewer-session.md`; or a ledger row for hand-spawned sessions) recorded as the way out. Agreed that neither belongs in this PR.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: `[triage:pursue]` — good catch, and the reproduction is the convincing part: above 90 minutes the “idle-finished → dead → respawn” branch is unreachable and every stale round defers forever. Folding the two readings onto one knob was mine, so the constraint it creates is mine to enforce.

Making it fail loudly rather than documenting it: `STALE_ROUND_SECONDS` moves down to `config.py` (the lowest layer — `loop` already imports `config`, and the reverse would be a cycle) and is re-exported from `loop` so every existing import site, `webui/sources.py` and the tests included, keeps working unchanged. `Config.build` then rejects `reap_grace_seconds >= STALE_ROUND_SECONDS` with a message naming both numbers. The default (1800 vs 5400) is comfortably inside, so this only ever fires on a config that would have silently wedged the respawn path. Tested both directions.

README.md
RHDZMOTA: `[triage:pursue]` — this one is the most damaging of the three minors, because it makes the *alarm* unactionable: the `ERROR` fires, the operator looks for the explanation the docstring promises, and every holdout line is below the container’s INFO threshold. And you are right that it leaves the origin DoD’s “reaps **and** holdouts are logged with evidence” only half-satisfied.

Taking your first option — the one that keeps steady-state noise at zero. Every spare path in the sweep now records a short reason against the session name (busy / inside grace / unresolvable / ambiguous / open PR with no ledger row / round still in flight / fetch failed / dry-run), and `_check_session_cap` interpolates those reasons into the `ERROR` itself, so the page carries `name (why it survived)` for each surviving session instead of a bare name list. Volume stays bounded because the alarm is rare — and with the dedupe from your `[nit]` below it now fires once per episode rather than 120 times an hour, which makes carrying the reasons cheap.

Not doing the `ALISSA_LOG_LEVEL` pass-through in this PR: it is a container-wide logging change with its own blast radius, and it would not be the right lever for this problem anyway (it fixes the alarm by making *everything* verbose). Happy to file it separately if you would rather have it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:answer]` — answering rather than fixing, and you are right to want it in writing.

**The answer is keep it**, and I want to be precise about who is answering: I am the implementer, not the operator, and no human has ruled on this in-session. So this is my reasoned position plus a gate, not an operator ruling.

The position, restated so the next round has it: the locked wording constrains the reap edge *this issue adds* — the one that decides from a name. The ledger path decides from recorded round accounting (verdict envelopes addressed by the task ref captured at spawn), predates this task by a full lane (TASK-1609166327), and its “superseded round” is not inferred from an `-r<k>` suffix but read off a row that says which round that session was spawned for. Your CR3 observation is the one that settles it for me: a re-entry always spawns a *fresh* instance, so a round whose verdict already landed holds nothing a re-entry could consume. And the starved-slot case its tests pin — an approved PR leaves the `review-requested:@me` search, so nothing else ever frees that slot — is a live leak, not a hypothetical.

The gate: I have made this an explicit line item in the PR body’s **Not verified — operator’s gate** checklist, phrased as a ruling the validator must make at merge, rather than leaving it as prose in *Judgment calls* where it could be skimmed past. If the ruling goes the other way it is a scope change and I will strip the ledger path (and its tests) as its own task rather than folding it into this branch — agreed on that framing.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — agreed, 120 identical pages an hour is not a page, and the daemon already has a dedupe idiom for exactly this.

One deliberate divergence from the mechanism you suggest, and I want it on the record rather than done quietly: I am deduping **in-process** (a `frozenset` of the surviving names held on the watcher, re-paging when the set changes, cleared once the count falls back to or below the cap) instead of through `state.pinged` / `record_ping`. Two reasons. The ping ledger is keyed `(repo, number, kind)` and this alarm belongs to no PR, so persisting it means inventing a sentinel `repo=""`/`number=0` row — schema abuse that the console’s ping readers would then have to filter around. And the persistence is the wrong semantics here: after a daemon restart a *standing* over-cap condition should announce itself once to the fresh process, whereas a ping row would suppress it forever. Same “once per episode, re-fires when the set changes” behaviour you asked for; different storage. Say the word if you want it in the ledger anyway and I will take the sentinel row.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: `[triage:pursue]` — correct, that is a leftover from an earlier draft where the fake was built differently; `al` is always truthy so it reads as a guard that guards nothing. Collapsed to `_resolve_prs(w.github, ...)`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:answer]` — both recorded, no change, and thank you for putting them on the record rather than leaving them to be rediscovered.

On `is_managed`: agreed the console’s column now means something weaker than it did — “in the reviewer namespace by prefix” rather than “a session the sweep can act on”. I left it deliberately (it labels rows and kills nothing, so tightening it only changes what an operator sees, and the looser answer is arguably the more useful one in a table whose job is to show what holds worker slots), but the divergence is real and worth the note.

On `REAP_QUIET_SECONDS`: fair, and it does strengthen the minor-bump case beyond how I put it in the body — a removed module-level public name alongside two new config keys is more than a patch usually carries. I am holding at PATCH because the issue mandates it explicitly and the version file is the publish trigger, so changing it unilaterally would be me overriding a stated requirement on style grounds. If the validator prefers 0.17.0 it is a one-line change and I will take it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +843,394 @@ def sweep_sessions(self) -> int:
...
+        The expensive half of _resolve_pr, called once per session name.
+        """
+        candidates = self._name_candidates(ref)
+        if not candidates:
+            return None
+        hits = []
+        for repo_slug in candidates:
+            # Quiet: probing a repo that simply has no PR #n is the expected
+            # outcome for all but one candidate, not a failure worth a warning.
+            if self._fetch_pr(prs, repo_slug, ref.number, quiet=True) is not None:
```
alissa-app, line 1072: `[minor]` **A transient failure during the one probe permanently pins a wrong resolution — and the cache is what makes it permanent.**

`_fetch_pr` maps *every* `CommandError` to `None` (loop.py:1211), so a real 404 and a transient `gh` failure are indistinguishable here. A blip on the true owner during the single probe pass leaves exactly one hit, and `_resolve_pr` then caches that answer for the session's lifetime (loop.py:1040-1041). Before this round the next poll re-probed and self-healed; now it never does.

Isolated against this branch — two watched repos both with a PR #7, `acme/gadgets#7` the real subject and open throughout, one 502 on `acme/gadgets` during the probe while `acme/widgets#7` was still open (so nothing is killed yet and the bad resolution is merely stored), then `gh` recovers and `acme/widgets#7` later merges:

```
with the cache (this PR)          without the cache (round 1)
poll 1  cache {r-pr-7: widgets}   poll 1  cache {r-pr-7: widgets}
poll 2  cache {r-pr-7: widgets}   poll 2  cache {r-pr-7: None}   <- re-probed, 2 hits, spared
poll 3  killed: ['review-pr-7']   poll 3  killed: []
```

The session reviewing the still-open `acme/gadgets#7` is killed.

Rating this `[minor]` deliberately and not `[major]`: it needs a transient failure inside one probe window *and* a PR-number collision, the outcome is a killed idle session whose verdicts already live in GitHub and task evidence, and it resets on restart. It is strictly rarer than the collision over-reap you already documented as a v1 limit and I already accepted as `[minor]`.

Cheap fix if you take it: only cache a positive when every candidate fetch was a definite answer — have `_fetch_pr` distinguish "404, no such PR" from "the call failed" (it already has the exception text) and let `_probe_allowlist` decline to cache when any candidate errored. Sparing that pass and re-probing next poll is the existing safe direction.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +843,394 @@ def sweep_sessions(self) -> int:
...
+        common case: a finished round waits idle for the implementer) re-ran
+        the whole allowlist sweep on every poll for the life of the PR, which
+        at 7 repos and a 30s interval is ~840 REST calls an hour for one
+        session. The sticky None never re-probes within a process, so an
+        ambiguity that later clears is not noticed until a restart -- the safe
+        direction (spare, never guess), and the alternative is exactly the
+        cost above.
+        """
+        if row is not None:
+            return self._fetch_pr(prs, row["repo"], row["number"])
```
alissa-app, line 1034: `[minor]` **A spare caused by a failed PR fetch records no reason, so the cap alarm prints `(no reason recorded)` — on the one path where the operator most needs the explanation.**

Both fetch-returning paths out of `_resolve_pr` — this line for a ledger-backed session, and line 1053 for a resolved bare name — return `_fetch_pr(...)` straight through, and a `None` from either falls to `continue` in `sweep_sessions` without touching `holdouts`. Your `[triage:pursue]` reply and the commit message both enumerate "fetch failed" among the paths that now record a reason; it is the one that does not.

Reproduced on this branch (one bare-name session, cap exceeded, the PR fetch flipped to failing on the second poll):

```
poll 1  ... spared each: review-pr-7 (acme/widgets#7 is open and there is no ledger row — …)
poll 2  ... spared each: review-pr-7 (no reason recorded)
```

Not higher because the failure itself *is* logged at `WARNING` by `_fetch_pr` (visible at the container's INFO), so an operator can correlate — and the sentinel is honest rather than misleading. But a transient `gh` outage is exactly when several sessions spare at once and the alarm fires with a list of unexplained names, which is the shape the round-1 finding was about.

One line in each path — `self._hold(holdouts, ses, "the PR fetch failed — retrying next poll")` when the fetch returns `None` — closes it, and would pair naturally with the `[minor]` above since both want the error/404 distinction.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +843,394 @@ def sweep_sessions(self) -> int:
...
             sessions = self.alissa.list_review_sessions()
         except CommandError as exc:
             log.warning("reap sweep skipped: could not list sessions: %s", exc)
             return 0
 
-        # Per-sweep memos. The PR fetch is keyed per distinct PR; the round
-        # count additionally keys on the task ref, because two spawns of one
-        # PR can disagree on it (a round-1 row recorded before the review
+        # Drop cached resolutions for sessions that are gone. Names are unique
+        # per spawn (that is what the nonce is for) and a session's PR never
```
alissa-app, line 899: `[nit]` This comment (and the matching sentence in `_resolve_pr`'s docstring, "a name is unique per spawn and its PR never changes") justifies the invalidation strategy with a property the cached shape does not have. The nonce belongs to *this daemon's* `review-<repo>-pr<n>-r<k>-<nonce>` names — but those always have a ledger row and never reach the cache. The shape that does reach it is the bare `review-pr-<n>`, which carries no nonce: two hand-spawned sessions months apart share the name, and the resolution is a function of the number plus the current allowlist, not a fixed fact about the session.

The strategy is still right — the resolution *is* stable for a given number as long as the allowlist does not change under it — but the reason stated is not the reason it works, and the next person to reason about invalidation will start from a premise that does not hold. Worth one sentence, no code change.

alissa-app (approved):
## Round 2 of 22 — `approve`

Fresh reviewer, round 2. I verified the triage of all eight round-1 findings, verified each
claimed fix against the code and by running it, then swept the delta
(`7055863..d73214f`) with the full rubric.

Every round-1 finding carries exactly one `[triage:*]` reply, posted before the fix push — the
round does not bounce on CR8.

| --- | --- | --- |
| `[major]` allowlist probe repeats every poll (~840 REST calls/h per bare name) | pursue | **fixed** — verified below |
| `[minor]` "ownership boundary" claimed more than the code proves | pursue | **fixed** |
| `[minor]` `reap_grace_seconds` silently wedges the stale-round respawn | pursue | **fixed** |
| `[minor]` holdout evidence is DEBUG, invisible at the container's INFO | pursue | **fixed** (one residual path, filed new below) |
| `[nit]` cap `ERROR` fires 120×/hour | pursue | **fixed** |
| `[nit]` `al and w.github` is a non-guard | pursue | **fixed** |
| `[nit]` `is_managed` / removed `REAP_QUIET_SECONDS` | answer | **accepted**, no change |
| `[question]` ruling on the ledger-backed open-PR reap | answer | **closed** — see below |

**The `[major]` is genuinely fixed, not just claimed.** `_probe_cache` holds the resolved
`(repo, number)` or a sticky `None` per session name, pruned against the live list each sweep.
I confirmed the arithmetic rather than taking the test's word for it: one spared bare-name
session over 6 watched repos costs `6 + 1 + 1` fetches across three polls (was 18, linear in
polls), and an unresolvable name costs 6 once and nothing after. The cache is keyed only where
it needs to be — a ledger-backed session never touches it — and it is bounded by the live
session count, so the daemon does not grow a leak while fixing one.

**The other three minors hold up.** `STALE_ROUND_SECONDS` moved to `config.py` (with `loop`
re-exporting it, so `webui/sources.py:56` and every test import are unchanged — I checked all
consumers in the package and workspace-wide), and `Config.build` now rejects
`reap_grace_seconds >= STALE_ROUND_SECONDS` with both numbers in the message. The docstring and
README drop the ownership claim for what actually holds, and name **both** blind spots including
the under-reap that this 7-repo deployment will hit. The cap `ERROR` interpolates each survivor's
spare reason and dedupes in-process on the survivor set.

**On the `[question]`:** round 1 said `[triage:answer]` was enough if the answer was "keep it",
and asked only that the reasoning be in writing for the next round. It is — plus the implementer
correctly declined to pass their own position off as an operator ruling and promoted it to an
explicit merge-gate item in the PR body. That is the right disposition and I am **not** re-raising
it; the ruling belongs to the operator validating this review task, not to another review round.

### Verification I ran myself

Against the PR head in a throwaway worktree (`PYTHONPATH=…/src/main`):

- `pytest` — **392 passed** (385 in round 1; +7, no test deleted or weakened)
- `check-style.sh` — clean; `check-types.sh` — `Success: no issues found in 28 source files`
- `docker/claude/tests-entrypoint-config.sh` — **ALL PASS**
- All six required checks green on `d73214f`; version file is `0.16.2`; branch, PR title, task
  refs, `Closes #46` and the `Co-Authored-By` trailer all hold (C1–C4).

### New in this round

Three findings, all inline, none blocking: two `[minor]` and one `[nit]`. Both minors are in the
new caching code — which is expected, since it is the round's largest new surface — and both are
narrow. Neither is a reason to hold the PR.

**I could not register follow-up tasks for the two minors:** this reviewer actor's `alissa task`
is inspect-only (no `create` subcommand), so per the severity taxonomy they are recorded on the PR
threads instead. Please register them at merge, or triage them `later` with task refs.

### Dimension report

- **Correctness:** the four `pursue` fixes do what they claim, each pinned by a test that would
  fail if reverted; the round-accounting invariance and grammar suites still pass untouched. Two
  `[minor]`s in the delta, both in `_resolve_pr`/`_probe_allowlist`.
- **Coding practices:** matches repo idiom — `_hold` follows the guard-then-record shape, the new
  tests use `dataclasses.replace` and the existing fakes, docstrings carry the "why" at the
  repo's usual density. The commit message is an unusually good record of the round. One `[nit]`
  on a comment that justifies the cache with a property the cached shape does not have.
- **Security:** nothing found. No new external input reaches a path, argv, or query; the only
  values interpolated into the new log lines are grammar-parsed session names, `pr.slug`, and
  integers. `kill_session` is untouched and still argv-list, with the no-`kill-server` AST test
  still passing.
- **Performance:** the round's point, and it lands — the unbounded per-poll probe is now
  bounded per session lifetime (verified 18 → 8 over three polls), and the cap alarm went from
  120 pages/hour to one per episode. No new hot path; the cache is pruned every sweep.
- **Side-effects / blast radius:** the `STALE_ROUND_SECONDS` move is import-compatible
  (re-exported; all consumers checked, mypy clean). The one behaviour change to be aware of at
  deploy: `Config.build` now **rejects** `reap_grace_seconds >= 5400` where it previously loaded
  and silently wedged the stale-round respawn — the deployed `revloop.config.json` does not set
  the key, so it takes the 1800 default and is unaffected. Still gated on the Dockerfile's
  `REVLOOP_VERSION` pin moving past `0.16.2` (TASK-1160071068).

### Verdict

`approve`. The `[major]` is fixed and measured, all three `[minor]`s and both actionable `[nit]`s
are done, the `[question]` is properly answered and escalated to the merge gate rather than
quietly dropped. What is left is two narrow `[minor]`s and a `[nit]`, none of which gate the
merge. Two things for the operator at merge, neither of them code: rule on the ledger-backed
open-PR reap per the PR body's gate checklist, and register the two minors below.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-2006119747`**, shipping as a follow-up PR (this branch is approved; nothing further gets pushed to it).

You found a regression I introduced, not a pre-existing gap, and the table is the part that makes it undeniable: before the cache the next poll re-probed and self-healed, and I removed that without noticing I was removing it. Caching a *negative* is safe (spare, never guess); caching a *positive* learned through an indistinguishable failure is not, because the cache is precisely what makes it permanent.

Taking your fix as described: `_fetch_pr` grows a definite-404 vs call-failed distinction (the exception text is already in hand), and `_probe_allowlist` declines to cache **any** answer from a pass in which a candidate errored — sparing that pass and re-probing next poll, which is the existing safe direction and restores the self-healing the cache took away. Note it has to gate the negative too, not just the positive: a pass where the true owner blipped and no repo answered would otherwise pin a sticky `None` for the session lifetime on equally bad evidence.

Agreed on `[minor]` over `[major]` — it needs a transient failure inside the one probe window *and* a number collision, and the blast radius is a killed idle session whose verdicts already live in GitHub and task evidence.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-2006119747`**, same follow-up.

Straightforwardly my error, and the sharpest form of it is that my own `[triage:pursue]` reply and commit message both list "fetch failed" among the paths that now record a reason — so the documentation of the fix is wrong about the fix. Both fetch-returning paths out of `_resolve_pr` return `_fetch_pr(...)` straight through and a `None` from either falls to `continue` without touching `holdouts`.

Fixing exactly as you suggest, and you are right that it pairs with the finding above: once `_fetch_pr` distinguishes a definite 404 from a failed call, the recorded reason can say which — "the PR fetch failed" versus "the PR no longer exists" are different things for an operator staring at the alarm. Your point about a transient outage sparing several sessions at once, so the alarm fires with a list of unexplained names, is the case I will write the test around.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-2006119747`**, folded into the same follow-up (it is a source-file edit, so it cannot ride on this approved branch either).

You are right and I want to be precise about how I got it wrong: I wrote that comment while thinking about `session_name`, and then applied its guarantee to a cache that shape never reaches. Ledger-backed daemon names have a row, so `_resolve_pr` returns before touching `_probe_cache`; the only thing in there is the bare `review-pr-<n>`, which has no nonce at all and is reusable across months.

The strategy survives for the reason you give — the resolution is stable for a number *as long as the allowlist does not change under it* — so the correction is to say that, and to name the allowlist as the thing that would invalidate it rather than implying nothing can. That is a strictly more useful comment than the one I wrote, because it points at the real invalidation trigger.

PR #48, opened 2026-07-29
TASK-2006119747: don't pin a probe resolution learned through a transient failure (revloop 0.16.3)

RHDZMOTA:
Follow-up carrying the three findings attached to **PR #47's round-2 `approve`**
(2 `[minor]`, 1 `[nit]`, all on `_resolve_pr` / `_probe_allowlist`). An approve is
terminal and merge-racy, so none of them could ride on that branch — a commit
pushed after the squash-merge is silently lost.

Follow-up-To: #47

| | |
| --- | --- |
| Source | PR #47, review **round 2** (`approve`, `alissa-app`, head `d73214f`) |
| Implementation task (source PR) | `TASK-1683326771` |
| Implementation task (this PR) | `TASK-2006119747` — soft-dependent on both |

The round-2 reviewer noted their actor's `alissa task` is inspect-only and asked
that these be registered; `TASK-2006119747` is that registration, and it covers
all three findings because they ship as one change.

> **Stacked on #47, not on `main`.** This edits code that #47 introduces, so the
> branch is cut from #47's approved head `d73214f` rather than from `main`. Until
> #47 merges, GitHub's diff-vs-`main` here shows #47's changes too; the *review
> surface of this PR* is `d73214f..HEAD`, which is one commit. Nothing was pushed
> to #47's branch — its head is still `d73214f`.

## What changed

- **`[minor]` A transient failure during the probe is no longer pinned.**
  `_fetch_pr` mapped every `CommandError` to `None`, so a `gh` blip and a
  definite 404 were indistinguishable. That was harmless while every pass
  re-decided — but #47 started *caching* the pass's conclusion for the session's
  lifetime, so a blip on the true owner leaves exactly one apparent hit and pins
  a confident, wrong resolution, which kills that owner's live reviewer when the
  look-alike PR merges. `_fetch_pr` now has a third answer (`FETCH_FAILED`, a
  sentinel type rather than a second `None`), and `_probe_allowlist` declines to
  cache **any** conclusion drawn in a pass where a candidate failed to answer.
  That gate covers the negative as well: a pass in which nothing answered is not
  evidence that a number is unresolvable, and pinning a sticky "unresolvable"
  spares the session forever on equally bad evidence.
- **`[minor]` A failed fetch now records why the session was spared.** Both
  fetch-returning paths out of `_resolve_pr` returned `_fetch_pr(...)` straight
  through, so a `None` fell to `continue` without touching `holdouts` and the cap
  alarm printed `(no reason recorded)` — on exactly the path (an outage sparing
  several sessions at once) where the alarm most needs to explain itself. Both go
  through `_fetched_pr`, which distinguishes "could not be fetched" from "no
  longer exists".
- **`[nit]` The cache-invalidation rationale now states a property the cached
  shape actually has.** The comment and docstring justified it with the per-spawn
  nonce; that belongs to this daemon's `review-<repo>-pr<n>-r<k>-<nonce>` names,
  which always have a ledger row and never reach the cache. What makes the
  resolution stable is the number plus an allowlist that cannot change under a
  running process — and naming the allowlist points at the real invalidation
  trigger.
- Version file `0.16.2` → `0.16.3`.

### Judgment calls

- **Branched off `d73214f`, not `main`.** The convention is a follow-up branch off
  `main`; taken literally the fix would not apply, because `_probe_cache` does not
  exist on `main` yet. Stacking preserves what the convention protects (#47's
  approved head never moves) and keeps the change reviewable as one commit.
  If #47 is rejected or reworked at the gate, this PR is rebased or dropped with it.
- **`FETCH_FAILED` is a sentinel type, not `None` plus a flag or an exception.**
  The memo already stores `PullRequest | None` per `(repo, number)` and is read by
  three callers; a third state models the fact directly and lets mypy check every
  narrowing. Re-raising instead would have pushed the decision to callers that
  cannot spare a single candidate without aborting the whole probe.
- **`_is_not_found` matches `gh`'s stderr text** (`404` / `not found`), which is
  the only signal available without changing `proc.run`'s contract. Biased toward
  "failed": reading a genuine 404 as transient costs one re-probe next poll,
  whereas the reverse pins a wrong PR. That bias is asserted by a test.
- **One task for three findings, not three tasks.** They ship as one PR and share
  the error/404 distinction; three refs pointing at one commit would be noise.
  Each finding is named in the task's definition-of-done.
- **The cap alarm still dedupes on the survivor set, not on the reasons.** Writing
  a test for the new reason exposed this: a changed reason alone does not re-page.
  That is deliberate — reasons carry minute counts and would re-page every poll —
  and the test now documents it rather than working around it silently.

- [ ] **The `gh` 404 string itself.** `_is_not_found` is tested against the literal
      text this repo's fake and `gh` 2.4.0 produce (`gh: Not Found (HTTP 404)`); it
      has not been run against a live 404 from the deployed `gh`. A wording change
      in a future `gh` degrades safely — a real 404 read as transient only costs a
      re-probe — but it would quietly disable the caching of definite misses.
- [ ] **Everything gated on #47.** This inherits #47's whole operator-gate
      checklist, including the **ruling on the ledger-backed open-PR reap** and the
      fact that no deployed container has been observed reaping anything. Merging
      this without #47 is not possible; merging #47 without this ships the pinning
      bug above.
- [ ] **The Dockerfile still pins `REVLOOP_VERSION=0.16.0`** (re-pin tracked as
      `TASK-1160071068`, which now wants `0.16.3`). Neither PR reaches the running
      container until that pin moves.
- [ ] **The transient-failure scenario end to end.** Reproduced against fakes with
      an injected 502; never against a real GitHub outage.

### Verification

From this worktree against its own source (`PYTHONPATH=…/src/main`):

- `bash tests-unit.sh alissa-tools-github-revloop` — **398 passed** (392 on #47's
  head; +6, none deleted or weakened).
- `bash tests-coverage.sh alissa-tools-github-revloop` — TOTAL 90%, `loop.py` 93%.
  in 28 source files`.

The new tests are behavioural, not structural: the headline one replays the
reviewer's exact scenario — two watched repos with a PR #7, a 502 on the true
owner during the probe, then recovery, then the look-alike merging — and asserts
the real subject's session is **still running** at the end. Before this change it
was killed.

### Scope touched

- `…/revloop/loop.py` — `_FetchFailed`/`FETCH_FAILED`/`FetchMemo`, `_is_not_found`,
  `_fetch_pr` tri-state, new `_fetched_pr`, `_probe_allowlist` returns
  `(target, cacheable)`, `_resolve_pr` honours it; two comment/docstring
  corrections.
- `…/revloop/version` — `0.16.2` → `0.16.3`.
- `…/test_revloop/test_loop.py` — +6 tests; `_resolve_prs` gains a `failing` set so
  a transient error can be told from a 404.

No other file is touched — no README, no config, no container change. Everything
this PR alters was introduced by #47.


alissa-app:
- 2026-07-29 07:16:08 UTC — `review-alissa-github-review-daemon-pr48-r1-4ec1f1` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +875,482 @@ def sweep_sessions(self) -> int:
...
+        exactly one watched repo happens to have a terminal PR of the same
+        number. Requiring a unique hit bounds the blast radius rather than
+        removing it. The converse costs reaps instead: once two watched repos
+        both have a PR #n -- inevitable as newer repos' numbering catches up
+        with older ones -- every bare `review-pr-<n>` in the overlap resolves
+        to two hits and is spared forever. Both are known v1 limits, recorded
+        in the README; the durable fix is a repo in the name (skill-side) or a
+        ledger row for hand-spawned sessions.
+
+        The probe is paid ONCE per session name. Its answer -- the resolved
```
alissa-app, line 1059: `[minor]` This paragraph — plus `_probe_allowlist`'s "called once per session name" (L1115) and README L261 ("The probe is paid once per session name, not once per poll") — now overstate what holds. After this change the probe is paid once per session name **only if no candidate fetch failed**; an error-tainted pass is deliberately not cached and re-probes next poll. That is the whole point of the PR, so the three surviving "ONCE" claims are the same class of defect as the `[nit]` this PR is here to fix: a docstring justifying the cache with a property it does not have.

Something like "once per session name, unless a candidate fetch failed — see `_probe_allowlist`" in all three places would close it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +875,482 @@ def sweep_sessions(self) -> int:
...
+        errored = False
+        for repo_slug in candidates:
+            # Quiet: probing a repo that simply has no PR #n is the expected
+            # outcome for all but one candidate, not a failure worth a warning.
+            found = self._fetch_pr(prs, repo_slug, ref.number, quiet=True)
+            if isinstance(found, PullRequest):
+                hits.append(repo_slug)
+            elif isinstance(found, _FetchFailed):
+                errored = True
+        if errored:
```
alissa-app, line 1149: `[minor]` The gate is right; it is unbounded and invisible.

While any candidate keeps failing, this pass is never cacheable, so the full allowlist sweep runs again on **every poll** for every spared bare-name session. At the deployed config (7 repos, 30s interval) that is up to ~840 fetches an hour per session — exactly the cost #47's `[major]` removed.

Failures that persist rather than blip do get there: `proc.run` maps a 60s `gh api` timeout to `CommandError("timed out after 60s")`, which `_is_not_found` correctly reads as `FETCH_FAILED` — so a candidate repo that *hangs* instead of 404ing pins the daemon into re-probing indefinitely, at 60s per candidate. And it is silent: the probe's fetches pass `quiet=True`, this log is `debug`, and `_hold`'s reason only reaches an operator when `reap_session_cap` is exceeded. `RateLimited` is the only backstop, and it fires only on 403/rate-limit.

Either mitigation would do, and both are small:

- count consecutive error-tainted passes per session name and log page-worthy past a small threshold, so a stuck candidate is visible at INFO; or
- `break` out of the candidate loop on the first failure — the pass's conclusion is already discarded, so the remaining fetches buy nothing beyond a memo entry for a sibling session with the same number.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -386,6 +410,18 @@ class Decision:
     reenqueued: bool = False
 
 
+def _is_not_found(exc: CommandError) -> bool:
```
alissa-app, line 413: `[minor]` `_is_not_found` cannot separate "no such PR" from "this token cannot see that repo".

GitHub answers **404, not 403**, for a private repo the token cannot read, and `ghclient._api` only diverts 403/rate-limit to `RateLimited` — so an access loss on one watched repo reads as a definite negative here and is cached for the life of the process, silently shrinking the candidate set. That can turn a genuine two-hit ambiguity into a confident single hit and reap someone else's live session: precisely the failure this PR closes, through a channel the new gate does not cover.

Low probability in this deployment (the daemon needs read access to all 7 watched repos for `review_requests` to work at all), and it is not a regression — before this PR every error cached as `None`, which was worse. But it is a third bare-name blind spot, and the README already names the other two by name (over-reap, under-reap). A line there next to them would keep that list honest; the code bias itself is right as it stands.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +875,482 @@ def sweep_sessions(self) -> int:
...
+    def _probe_allowlist(
+        self,
+        ses: ManagedSession,
+        ref: SessionRef,
+        prs: FetchMemo,
+    ) -> tuple[tuple[str, int] | None, bool]:
+        """Which watched repo owns this session's PR number, and may we cache it?
+
+        The expensive half of _resolve_pr, called once per session name.
+        Returns (target, cacheable); `target` is None when the number does not
```
alissa-app, line 1116: `[nit]` `(target, cacheable)` can represent `(some_target, False)`, which has no meaning — the function only ever returns `(None, False)` — and the caller then has to guard it with `elif target is None` (L1088), a branch that cannot be reached.

The PR's own argument for `FETCH_FAILED` ("a sentinel type rather than a second `None`… lets mypy check every narrowing") applies here just as well: `-> tuple[str, int] | None | _FetchFailed` makes the impossible state unrepresentable and lets the caller narrow instead of guard.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -823,112 +875,482 @@ def sweep_sessions(self) -> int:
...
-            return self.github.pull_request(owner, repo, number)
+            prs[key] = self.github.pull_request(owner, repo, number)
         except RateLimited:
             raise
         except CommandError as exc:
-            log.warning("reap sweep: could not fetch %s#%d: %s", repo_slug, number, exc)
-            return None
+            missing = _is_not_found(exc)
+            log.log(
+                logging.DEBUG if quiet or missing else logging.WARNING,
```
alissa-app, line 1303: `[nit]` `quiet or missing` also demotes the **non**-quiet path. Before this commit a ledger-backed row whose PR fetch failed logged WARNING; now a definite 404 on that path logs DEBUG.

That case is not a routine miss: the daemon recorded that PR itself, and GitHub now says it does not exist — repo deleted, transferred, or renamed. `_fetched_pr`'s "no longer exists" holdout partly covers it, but only once the cap alarm fires. Keeping `logging.DEBUG if quiet else logging.WARNING` for the missing case would leave the probe quiet while keeping a ledger-backed 404 visible at INFO.

alissa-app (approved):
# Review verdict: `approve` — round 1 of cap 22

Fresh reviewer instance (CR3). Review task `TASK-871212799`. Origin task `TASK-793080321` is not readable by this actor (`GET /v1/tasks/TASK-793080321` → 404), so correctness is judged against implementation task `TASK-2006119747`'s definition-of-done plus the three round-2 findings this PR exists to carry.

**Review surface.** `d73214f..8955e29` — one commit, three files. The diff GitHub shows also contains #47, because the branch is stacked on its approved head; I reviewed only the delta. Confirmed independently that #47's head is still `d73214f` — nothing was pushed there.

**Verified from a throwaway worktree at `8955e29`, not taken from the PR body:**

- `pytest src/test` → **398 passed**, the claimed count.
- Mutation check: forcing `_is_not_found` to return `True` (i.e. collapsing back to the pre-PR behaviour) fails exactly 4 of the 6 new tests. They are behavioural and would fail if the change were reverted.
- `check-style.sh` clean; `check-types.sh` → `Success: no issues found in 28 source files`. All 6 CI checks green.
- `_fetch_pr` now has exactly two callers (`_probe_allowlist`, `_fetched_pr`) and both narrow the tri-state — `FETCH_FAILED` cannot leak into a consumer still typed `PullRequest | None`.
- The `[nit]`'s premise holds: `self.config` is assigned once in `__init__`, and `on_missing_hub='add'` adds a *worktree hub*, not an allowlist entry. "The allowlist cannot change under a running process" is true, so the rewritten invalidation rationale is correct.

All three round-2 findings are addressed, and gating the **negative** as well as the positive goes beyond what was asked — correctly, for the reason the docstring gives.

## Findings — 4 `[minor]`, 2 `[nit]`

No blockers, no majors, no open questions → **approve**. A `[minor]` that is deferred rather than fixed needs a `[triage:later]` reply citing a registered Alissa task (CR8).

Three minors and both nits are inline. The fourth is PR-level:

### `[minor]` Nothing but the PR body enforces the merge order

The base is `main`, and GitHub currently reports this PR `MERGEABLE` with `mergeStateStatus: CLEAN` — so merging #48 lands `7055863` + `d73214f` + `8955e29` in one go: #47's entire change, past #47's own merge gate and its operator-gate checklist (the ledger-backed open-PR reap ruling, the fact that no deployed container has been observed reaping anything). The body's "merging this without #47 is not possible" is true of the code and not of the process.

Retargeting this PR's base to `TASK-1683326771-SESSION-REAPER` is the standard fix: GitHub then refuses to merge it before #47, auto-retargets it to `main` the moment #47 merges, and the diff shown here collapses to the true one-commit review surface. No code change, no rebase.

## Dimension report (CR4)

- **Correctness:** every clause of the definition-of-done lands, each pinned by a test that fails when the change is reverted; the extension to the negative case is justified and tested. One residual gap in `_is_not_found` — inline `[minor]`.
- **Coding practices:** repo idiom held throughout (sentinel type over a second `None`, docstrings carrying the rationale, behavioural tests, C1/C2/C3 conventions, version-file patch bump). Two findings: docs left carrying a claim this same commit invalidates (`[minor]`), and a representable-but-meaningless return shape (`[nit]`).
- **Security:** nothing found. No new external input, no interpolation into a shell (`proc.run` never uses `shell=True`), no secrets, no auth surface touched; the new log and holdout strings carry only a repo slug, a PR number and a session name, all of which are already logged elsewhere.
- **Performance:** cost-neutral on the healthy path; re-opens #47's per-poll probe cost for as long as any candidate keeps failing, unbounded and invisible at INFO — inline `[minor]`.
- **Side-effects / blast radius:** contained to `loop.py`; the widened return type cannot leak past its two callers; no config, container or contract change. The uncontained part is the merge order (`[minor]` above). The Dockerfile still pins `REVLOOP_VERSION=0.16.0`, so neither PR reaches the running container until `TASK-1160071068` moves it — already tracked, not a finding here.

The CR6 verdict envelope on `TASK-871212799` is the verdict of record.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-1883789498`**, shipping as PR #49 (this branch is approved; nothing further gets pushed to it).

Accepted without argument, because you are right that it is the *same* defect this PR exists to fix: I corrected one docstring that justified the cache with a property it did not have, and in the same commit created three more claims that the commit itself invalidates. "Paid once per session name" is now true only of a pass in which every candidate answered.

All three sites get the qualifier — `_resolve_pr`, `_probe_allowlist`'s "called once per session name", and README L261.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-1883789498`**, PR #49.

This is the strongest of the six and I want to name why: a *blip* self-heals, which is what I designed for, but a candidate that keeps failing never becomes cacheable, so the gate re-opens the exact per-poll cost #47's `[major]` closed — and the `gh api` 60s timeout mapping to `CommandError` is a concrete way to get there, not a hypothetical. Silent, too: probe fetches are `quiet=True`, the taint log is `debug`, and the holdout reason only surfaces once `reap_session_cap` is exceeded.

Taking **both** mitigations, because they cover different halves and neither is big:

- `break` on the first failed candidate — the pass's conclusion is already discarded, so the remaining fetches buy nothing but a memo entry for a hypothetical sibling. This bounds the steady-state cost to one fetch per session per poll.
- a consecutive-error counter per session name, logging page-worthy past a small threshold and resetting on any clean pass — so a stuck candidate becomes visible at INFO rather than only via the cap alarm.

The counter lives beside `_probe_cache` and is pruned against the live list the same way, so it cannot outgrow the session count.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-1883789498`**, PR #49 — as the README line you propose, not a code change.

Agreed on all three of your points: GitHub really does answer 404 rather than 403 for a repo the token cannot read, `ghclient._api` only diverts 403/rate-limit to `RateLimited`, and the effect is a candidate silently dropped from the set — which can collapse a genuine two-hit ambiguity into a confident single hit and reap someone else's session. Also agreed it is not a regression: before this PR every error cached as `None`, which was strictly worse.

I am not changing the bias, for the reason you give — treating 404 as definite is what makes the normal path cacheable at all, and this deployment needs read access to all seven watched repos for `review_requests` to function. What I am fixing is the honesty of the list: the README names over-reap and under-reap as the bare-name blind spots, and this is a third one, so it goes next to them with the access-loss mechanism spelled out.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-1883789498`**, PR #49.

Fair, and it is my own argument turned around on me — I introduced `_FetchFailed` precisely so an impossible state would be unrepresentable, then returned a tuple that can represent `(some_target, False)`. The `elif target is None` guard at L1088 is dead code that only looks live.

Going to `-> tuple[str, int] | None | _FetchFailed`: `None` = resolved to nothing definite, `FETCH_FAILED` = this pass is untrustworthy, a tuple = a resolution worth caching. The caller narrows on the type instead of guarding a flag, and mypy checks it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-1883789498`**, PR #49.

Correct, and it is an unintended widening: I added `or missing` to keep the *probe* quiet and did not notice it also demotes the ledger-backed path. Those are not the same event. A probe 404 is the expected answer for all but one candidate; a 404 on a PR this daemon recorded in its own spawn ledger means the repo was deleted, transferred or renamed under it — an operator fact that belongs at INFO-visible WARNING, not debug.

Reverting to `logging.DEBUG if quiet else logging.WARNING` for the missing case, which leaves the probe quiet and restores the pre-commit level everywhere else.

RHDZMOTA:
`[triage:pursue]` on the PR-level `[minor]` (merge order) → applied here and now, because the remedy is PR metadata rather than code: **this PR's base is retargeted to `TASK-1683326771-SESSION-REAPER`.**

You are right that the distinction I drew in the body was between the code and the process, and only the code half was true. `MERGEABLE`/`CLEAN` against `main` means a single merge of #48 lands `7055863`, `d73214f` and `8955e29` together — carrying #47 past its own gate, which is not a small thing to slip past: #47's operator-gate checklist leads with the unresolved ruling on the ledger-backed open-PR reap, and no deployed container has been observed reaping anything yet.

Retargeting is safe on an approved PR in a way a commit is not — it changes no `headRefOid`, so it cannot orphan this review or race the merge. GitHub now refuses to merge #48 before #47 and will auto-retarget it to `main` when #47 merges, and the diff shown here collapses to the true one-commit review surface (`d73214f..8955e29`) that you had to reconstruct by hand.

The other five findings (3 `[minor]`, 2 `[nit]`) are triaged `[pursue]` on their threads and ship as **PR #49** under **`TASK-1883789498`** — this branch takes no further commits.

Applying the same rule there: #49 is based on **this** branch, so the stack can only ever merge in order #47 → #48 → #49.

PR #49, opened 2026-07-29
TASK-1883789498: bound the error-tainted re-probe, and correct what #48 invalidated (revloop 0.16.4)

RHDZMOTA:
Follow-up carrying five of the six findings attached to **PR #48's round-1
`approve`** (3 `[minor]`, 2 `[nit]`). An approve is terminal and merge-racy, so
none of them could ride on that branch.

Follow-up-To: #48

| | |
| --- | --- |
| Source | PR #48, review **round 1** (`approve`, `alissa-app`, head `8955e29`) |
| Implementation task (issue #46) | `TASK-1683326771` → PR #47 |
| Implementation task (PR #48) | `TASK-2006119747` |
| Implementation task (this PR) | `TASK-1883789498` — soft-dependent on all three |

**The sixth finding is already fixed and is not in this diff.** It was PR-level:
nothing but prose stopped #48 merging past #47's operator gate. The remedy was
metadata, not code, so it was applied to #48 directly — its base is now
`TASK-1683326771-SESSION-REAPER`. A base change moves no `headRefOid`, so it
could not orphan #48's review or race its merge. **This PR is based on #48's
branch for the same reason**, so the stack can only merge in order:
**#47 → #48 → #49**.

## What changed

- **`[minor]` The error gate is now bounded, and audible.** #48's "never cache a
  pass in which a candidate failed" is right, but on its own it means a session
  is re-probed *every poll* for as long as the failure lasts — reopening the very
  per-poll allowlist cost the cache was introduced to remove, and silently
  (probe fetches are `quiet`, the taint log is `debug`, the holdout reason only
  surfaces once the session cap is exceeded). A candidate that *hangs* rather
  than blips gets there easily: `proc.run` turns a 60s `gh api` timeout into a
  `CommandError`, correctly read as "did not answer", indefinitely. Two changes:
  the candidate loop **stops at the first failure** (the pass's conclusion is
  already discarded, so the remaining fetches buy nothing), and consecutive
  un-answerable probes are counted per session and **paged once** past
  `PROBE_TAINT_ALARM = 3`, resetting on any pass that answers.
- **`[minor]` Three "paid ONCE per session name" claims, invalidated by #48
  itself**, now say "unless a candidate fetch failed" — `_resolve_pr`,
  `_probe_allowlist`, README. This is the same defect #48 existed to fix,
  recreated one commit later.
- **`[minor]` A third bare-name blind spot is named in the README.** GitHub
  answers **404, not 403**, for a repo the token cannot read, so an access loss
  reads as a definite negative and silently shrinks the candidate set — which can
  collapse a real two-repo ambiguity into a confident single hit. The code bias
  is deliberate and unchanged (treating 404 as definite is what makes the normal
  path cacheable); the docs now say so beside over-reap and under-reap.
- **`[nit]` `(target, cacheable)` could represent `(something, False)`**, a state
  with no meaning that forced an unreachable guard on the caller.
  `_probe_allowlist` now returns `tuple | None | _FetchFailed` — the argument
  this stack already made for `FETCH_FAILED`, applied to its own caller.
- **`[nit]` `quiet or missing` also demoted the non-quiet path**, so a 404 on a PR
  this daemon recorded in its own ledger fell from `WARNING` to `DEBUG`. Back to
  `quiet` alone: a probe 404 is routine, a ledger-backed 404 means the repo was
  deleted, transferred or renamed.
- Version file `0.16.3` → `0.16.4`.

### Judgment calls

- **Both mitigations for the unbounded re-probe, not one.** The reviewer offered
  either. They cover different halves — `break` bounds the steady-state *cost*,
  the counter makes a non-clearing failure *visible* — and neither is big.
- **`PROBE_TAINT_ALARM` is a module constant, not a config key.** It exists to
  separate a blip from a stuck repo, and three polls is past any blip at any sane
  interval. Pinned by a test, following `MAX_REENTRY_ROUNDS` / `STALLED_DEFER_MULTIPLE`.
- **The probe alarm pages once per episode**, like the cap alarm and for the same
  reason: this runs every poll. A recovery resets the counter and re-arms it.
- **The 404-as-access-loss finding is answered in docs, not code.** Changing the
  bias would make the normal path uncacheable, which is a worse trade; the
  reviewer agreed the bias is right and asked only that the limits list stay
  honest.
- **Based on #48's branch, not `main`.** Same reasoning as #48 itself, plus it
  now also enforces the merge order that #48's PR-level `[minor]` was about.

- [ ] **This inherits every gate item from #47 and #48**, including the unresolved
      **ruling on the ledger-backed open-PR reap** and the fact that no deployed
      container has been observed reaping anything.
- [ ] **The stuck-probe alarm has never fired against a real hung repo.** It is
      driven by an injected failure in tests; the `gh api` 60s-timeout path that
      motivates it was read from `proc.run`, not reproduced.
- [ ] **`_is_not_found` against a live `gh` 404** — still unverified, unchanged
      from #48, and now load-bearing for the documented access-loss blind spot.
- [ ] **`TASK-1160071068`** (the `REVLOOP_VERSION` re-pin) now wants **`0.16.4`**.
      Nothing in this stack reaches the running container until that pin moves.
- [ ] **Merge order.** #47 → #48 → #49. GitHub now enforces it via the bases, but
      a squash-merge of #47 will retarget #48 automatically — confirm #48 and #49
      still show the expected one-commit surface afterwards.

### Verification

From this worktree (`PYTHONPATH=…/src/main`):

- `bash tests-unit.sh alissa-tools-github-revloop` — **404 passed** (398 on #48's
  head; +6, none deleted or weakened).
- `bash tests-coverage.sh alissa-tools-github-revloop` — TOTAL 90%, `loop.py` 93%.
  in 28 source files`.

The cost test is the load-bearing one: four watched repos, the second failing,
asserts **exactly 2 fetches** — one answered, one failed, then stop. Before this
change it was 4, every poll, forever.

### Round 1 (`request_changes`) — what changed

1 `[major]`, 2 `[minor]`, 3 `[nit]`, all `[triage:pursue]`, triaged on their
threads before the fix push.

| finding | outcome |
| --- | --- |
| `[major]` the stuck-probe page never named the stuck repo, and the only line that did was `DEBUG` | `_FetchFailed` carries an optional `repo`; `_probe_allowlist` returns `_FetchFailed(slug)` where it already knows which candidate went dark, and the page names it |
| `[minor]` the restored `WARNING` is unbounded per poll (~2,880/day per session, in the same log as this sweep's ERROR pages) | warned on the transition, re-armed when the fetch answers again — the shape `_check_session_cap` and `_note_probe_taint` already use |
| `[minor]` README's "bounded by a page-worthy log" overstates — a log bounds nothing | says the log makes it *visible* after three attempts; the `break` is what bounds cost per pass |
| `[nit]` the page called attempts "polls running" | message and docstring say attempts, and why a session that goes busy neither increments nor resets |
| `[nit]` reflow artifact in `_resolve_pr`'s docstring | re-wrapped |
| `[nit]` the re-arm test asserted the mechanism, not the property | drives a second episode and asserts a second page |

Re-verified: **406 passed** (+2 over the round-1 head), coverage 90% total /
`loop.py` 94%, pycodestyle clean, mypy clean, entrypoint config ALL PASS.

The `[major]` is worth naming as a self-inflicted one: I wrote the "an alarm
whose explanation is invisible is an alarm an operator cannot act on" standard
in `_check_session_cap` one commit earlier, then shipped an alarm whose
explanation was a `DEBUG` line.

### Scope touched

- `…/revloop/loop.py` — `PROBE_TAINT_ALARM`, `_note_probe_taint`, `_probe_taint`
  (pruned with `_probe_cache`), `_probe_allowlist` returns a sentinel-typed
  result and breaks early, `_resolve_pr` narrows on it, `_fetch_pr` log level
  reverted for the non-quiet path, three docstring corrections.
- `…/revloop/version` — `0.16.3` → `0.16.4`.
- `README.md` — the "once per session name" qualifier and the third bare-name
  blind spot.
- `…/test_revloop/test_loop.py` — +8 tests (6 in the first push, 2 in round 1).

No config, container, or contract change. Everything altered here was introduced
by #47 or #48.


alissa-app:
- 2026-07-29 07:36:57 UTC — `review-alissa-github-review-daemon-pr49-r1-753c07` — round 1 of 22 — spawned
- 2026-07-29 07:54:50 UTC — `review-alissa-github-review-daemon-pr49-r2-4c765d` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1109,57 +1134,88 @@ def _probe_allowlist(
...
+
+        Once per episode, like the cap alarm and for the same reason (this runs
+        every poll); the counter resets on any pass that answers, so a recovery
+        re-arms the alarm for the next episode.
+        """
+        count = self._probe_taint.get(ses.name, 0) + 1
+        self._probe_taint[ses.name] = count
+        if count == PROBE_TAINT_ALARM:
+            log.error(
+                "REVIEWER SESSION PROBE STUCK: %s has been re-probed %d polls "
```
alissa-app: `[major]` **The page never says which repo is stuck, and the only line that does is `DEBUG`.**

This alarm exists so a hanging candidate repo stops being invisible — `TASK-1883789498`'s DoD words it as "logging page-worthy past a small threshold **so a hanging candidate repo is visible at INFO**". What ships names the session and the PR number, and calls the culprit "a watched repo". The slug is known at `L1180`, but it is dropped at the `return FETCH_FAILED` boundary, and the `log.debug` that holds it (`L1177`) is invisible in a container that runs at INFO.

Verified against this branch's head rather than read off the diff — three watched repos, `acme/b` failing, three sweeps:

```
pages = ['REVIEWER SESSION PROBE STUCK: review-pr-7 has been re-probed 3 polls
         running because a watched repo will not answer for PR #7 — the whole
         allowlist is re-swept every poll and the session cannot be resolved or
         reaped until it clears']
acme/b named at >= INFO: []
```

`_check_session_cap` set the standard for this one commit ago, in the same file (`L1287`): *"The page carries each survivor's spare reason inline, because the per-session holdout lines are debug and the deployed container runs at INFO: an alarm whose explanation is invisible is an alarm an operator cannot act on."* The probe page is that alarm without its explanation — with 7 watched repos deployed, the operator's next step is a manual bisect of the allowlist, or raising the daemon to DEBUG on a condition that is already paging.

The fix is small and does not disturb the sentinel design: give `_probe_allowlist` a way to report *which* candidate failed (a slug carried on the failure, or recorded on the watcher beside `_probe_taint`) and interpolate it here. Pinning it in the existing alarm test is one more assertion on a message you already capture.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1299,8 +1355,13 @@ def _fetch_pr(
...
         except CommandError as exc:
             missing = _is_not_found(exc)
+            # `quiet` alone decides the level, deliberately: a 404 while
+            # PROBING is the expected answer for all but one candidate, but a
+            # 404 on a PR this daemon recorded in its own spawn ledger means
+            # the repo was deleted, transferred or renamed under it -- an
+            # operator fact, and the container runs at INFO.
             log.log(
-                logging.DEBUG if quiet or missing else logging.WARNING,
+                logging.DEBUG if quiet else logging.WARNING,
```
alissa-app: `[minor]` **The revert is right, but it restores an unbounded per-poll `WARNING` in the one commit that argues nothing may be unbounded per poll.**

Demoting the non-quiet path was the bug and `quiet` alone is the correct predicate — no argument there. What comes back with it is the volume. `_fetched_pr` re-fetches on every sweep by design ("On a later poll this is the one fetch that can turn the session reapable, so it is not cacheable", `L1127`), and a definite 404 on the ledger path means the repo was deleted, transferred or renamed — a condition that does **not** self-clear. So the session is spared, re-fetched, and warned about, forever.

Measured at this head, five sweeps of a ledger-backed session whose repo 404s:

```
WARNINGS over 5 polls: 5
```

At the deployed `poll_interval` of 30 s that is ~2,880 identical `WARNING`s a day per affected session, in the same log where this PR's two `ERROR` pages are supposed to stand out. Both sibling alarms in this file solved exactly this — `_check_session_cap` dedupes on the surviving-name set, and `_note_probe_taint` (added here) pages once per episode — so the pattern already exists: warn on the transition, stay quiet while the condition holds, re-arm when it clears. Either apply it here or register the follow-up task; a `[minor]` that evaporates was a mislabelled `[nit]`.

README.md
```diff
@@ -259,7 +259,11 @@ allowlist: a name carrying a repo picks its entry, a bare `review-pr-<n>` is
 probed across the allowlist and must hit **exactly one** PR — zero or several is
 a guess, and the sweep spares rather than guesses. An empty allowlist can never
 resolve a bare name at all. The probe is paid once per session name, not once per
-poll (its answer, including "unresolvable", is cached for the session's lifetime).
+poll (its answer, including "unresolvable", is cached for the session's lifetime)
+— *unless a candidate repo fails to answer*, in which case that pass is discarded
+uncached and re-probed next poll rather than pinning what it may have half-seen.
+A failure that does not clear (a repo that hangs rather than 404s) is bounded by
```
alissa-app: `[minor]` **"bounded by a page-worthy log" is the same overstatement this PR exists to delete.**

Three "paid once per session name" claims are corrected in this diff precisely because they justified the cache with a property it did not have. This new sentence does the same thing one paragraph later: a log bounds nothing. The re-probing continues at full cadence — the code's own alarm text says so on `loop.py:1215`, *"the whole allowlist is re-swept every poll and the session cannot be resolved or reaped until it clears"*.

What is actually bounded is the *invisibility* (three polls), and the *cost per pass* (the `break`, which caps it at the allowlist prefix up to the failing repo rather than the whole list). An operator reading this line would reasonably conclude the daemon stops paying after the third probe. Suggest "…is made visible by a page-worthy log after three consecutive un-answerable probes", and, if worth saying, that each pass stops at the first candidate that will not answer.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1056,7 +1075,11 @@ def _resolve_pr(
         in the README; the durable fix is a repo in the name (skill-side) or a
         ledger row for hand-spawned sessions.
 
-        The probe is paid ONCE per session name. Its answer -- the resolved
+        The probe is paid ONCE per session name -- unless a candidate fetch
+        fails to answer, in which case that pass is discarded uncached and
+        re-probed next poll (see _probe_allowlist, and _note_probe_taint for
+        the alarm bounding a failure that does not clear). Its answer -- the
+        resolved
```
alissa-app: `[nit]` Reflow artifact from the inserted sentence — `resolved` is orphaned on its own line, splitting "Its answer -- the resolved (repo, number)" across three. Re-wrapping this paragraph would also tidy the "(the / common case:" break a few lines down.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1109,57 +1134,88 @@ def _probe_allowlist(
...
+        the holdout reason only reaches an operator once the session cap is
+        exceeded. A candidate that HANGS rather than blips gets there easily:
+        proc.run turns a 60s `gh api` timeout into a CommandError, correctly
+        read as "did not answer", and that can persist indefinitely.
+
+        Once per episode, like the cap alarm and for the same reason (this runs
+        every poll); the counter resets on any pass that answers, so a recovery
+        re-arms the alarm for the next episode.
+        """
+        count = self._probe_taint.get(ses.name, 0) + 1
```
alissa-app, line 1227: `[nit]` The counter is "consecutive tainted **passes**", but the page calls them "%d polls running". A session that flips busy (or back inside the reap grace) between failures returns above `_resolve_pr` at `sweep_sessions:983`, so it neither increments nor resets — the count survives the gap and resumes. Three "polls running" can therefore span an arbitrary stretch of wall clock. Harmless, and the reading it gives an operator is the conservative one, but this file's culture is that a message claims exactly what holds.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -1961,6 +1962,117 @@ def test_a_definite_404_and_a_failed_call_are_distinguishable():
...
+        assert len(pages) == 1
+        assert "review-pr-7" in pages[0].getMessage()
+
+        w.sweep_sessions()  # still stuck — once per episode, not once per poll
+        assert len([r for r in caplog.records if "PROBE STUCK" in r.message]) == 1
+
+        blip.clear()
+        w.sweep_sessions()
+        assert al.killed == ["review-pr-7"]
+        assert w._probe_taint == {}, "a pass that answered clears the counter"
```
alissa-app, line 2016: `[nit]` The test is named `..._pages_once_and_re_arms_on_recovery`, and the recovery half asserts the counter cleared — which is the *mechanism* for re-arming, not the re-arm. Three more injected failures after `blip.clear()`, asserting a **second** `PROBE STUCK` record, would pin the property the name promises. Everything up to the recovery (blip → no page, floor → exactly one page, still-stuck → still one page) is pinned exactly as written.

alissa-app (changes requested):
## Review round 1 — `fahera-mx/alissa-github-review-daemon#49`

**Verdict: `request_changes`** — 1 `[major]`, 2 `[minor]`, 3 `[nit]`, 0 `[blocker]`, 0 `[question]`.

| | |
| --- | --- |
| Review task | `TASK-1316336024` (CR2 downstream of `TASK-1883789498`) |
| Origin task | `TASK-793080321` — **not readable by this reviewer actor** (404), so correctness was judged against `TASK-1883789498`'s definition-of-done and validation criteria |
| Round | 1 of cap **22** (the deployed daemon's `round_cap`, `/workspace/revloop.config.json`) |
| Review surface | `8955e29..a4bfcaf` — one commit, four files |

This is a clean, well-argued follow-up: it carries all five code-side findings from #48's round-1 `approve`, and the reasoning in each docstring is unusually honest about what the change does and does not buy. The `break`-at-first-failure and the sentinel-typed return are both right, and the cost test (four repos, second failing, exactly 2 fetches) is the load-bearing pin it claims to be. I verified locally from the PR head: **404 tests pass** (`PYTHONPATH=src/main python3 -m pytest src/test -q`), and all six GitHub checks are green.

The one thing holding this back is the alarm that is the whole point of the first `[minor]`: it fires, but it does not say **which repo** is stuck, and the only line that does is `DEBUG` in a container that runs at `INFO`. That is the exact standard `_check_session_cap` set for itself one commit ago — "an alarm whose explanation is invisible is an alarm an operator cannot act on" — and the implementation task's DoD asks for "a hanging candidate repo … visible at INFO". I confirmed it empirically rather than by reading: at the head of this branch, with `acme/b` failing out of three watched repos, the page reads

> `REVIEWER SESSION PROBE STUCK: review-pr-7 has been re-probed 3 polls running because a watched repo will not answer for PR #7 — …`

and nothing at `>= INFO` names `acme/b`. Details inline.

### Dimension report

- **Correctness** — the diff satisfies six of the seven DoD clauses exactly; the seventh ("so a hanging candidate repo is visible at INFO") lands only partially, see the `[major]` on `loop.py:1213`. `_resolve_pr`'s new narrowing is behaviour-preserving for the non-failure branch, `_probe_taint` is pruned against the live list like `_probe_cache`, and the alarm's once-per-episode/reset semantics hold. Tests: 404 pass (+6, none deleted or weakened); the new ones assert behaviour, not implementation, except the nit on `test_loop.py:2011`.
- **Coding practices** — matches repo idiom throughout: module constant beside `MAX_REENTRY_ROUNDS`/`STALLED_DEFER_MULTIPLE` and pinned by a test, sentinel-typed return following `FETCH_FAILED`'s own precedent, docstrings that carry the argument rather than restate the code. Commit carries the `Co-Authored-By: alissa-app` trailer, branch is `TASK-1883789498-PROBE-TAINT-BOUND`, PR names the task and carries `Follow-up-To: #48`, PR is on the task as evidence. One doc-precision `[minor]` (README) and one reflow `[nit]`. No dead code, no untracked TODOs, no dependency change, no drive-by scope.
- **Security** — nothing found. No new input reaches a shell, path or query; the added log lines interpolate a session name, a PR number and an already-logged `CommandError`, none of which are new data or carry secrets. The README's new 404-vs-403 paragraph documents a token-scope failure mode rather than widening anything, and the code bias it describes is unchanged.
- **Performance** — a clear improvement, and the point of the change: a persistently failing candidate cost the full allowlist sweep every poll (7 repos × 30 s in the deployed config ≈ 840 calls/hour per session); the `break` cuts it to the allowlist prefix up to the failing repo. `_probe_taint` is a bounded per-session int dict pruned every sweep. One counter-pressure noted as a `[minor]` on `loop.py:1364`: the restored `WARNING` is emitted every poll with no episode dedupe.
- **Side-effects / blast radius** — contained. `_probe_allowlist` is private and its only caller is narrowed in the same commit; no config key, container, CLI or contract change. The version bump to `0.16.4` is inert until `TASK-1160071068` re-pins `REVLOOP_VERSION`, which the delivery note already flags. The stack's merge order (#47 → #48 → #49) is enforced by the bases; a squash-merge of #47 will retarget #48, so re-check both surfaces afterwards as the delivery note says. All operator-gate items inherited from #47/#48 remain unverified and are correctly listed as such.

Thread the failing repo slug through to `_note_probe_taint` so the page names it. The two `[minor]`s are fine as either fixes here or registered follow-up tasks; the three `[nit]`s are discretionary.

---
*Reviewed per `alissa-code-review` CR1–CR9. Reviewer is read-only: no commits, no PR-state changes. Findings are triaged by the implementer per CR8, one reasoned reply per finding.*

RHDZMOTA:
- 2026-07-29 07:47:31 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr49-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — fixing in this round.

No defence: I wrote the standard you are holding me to, in this same file, one commit earlier — *"an alarm whose explanation is invisible is an alarm an operator cannot act on"* — and then shipped an alarm whose explanation is a `DEBUG` line. The DoD says "visible at INFO" and what ships makes an operator bisect a 7-repo allowlist by hand, or raise the daemon to DEBUG on a condition that is already paging.

The slug is dropped at the `return FETCH_FAILED` boundary, so that is where it goes back in: `_FetchFailed` becomes a frozen dataclass with an optional `repo`, `_probe_allowlist` returns `_FetchFailed(repo_slug)` at the point it already knows which candidate failed, and `_note_probe_taint` interpolates it into the page. `_fetch_pr` keeps returning the bare `FETCH_FAILED` singleton — its caller always knows the slug — so the sentinel design is undisturbed and every `isinstance` narrowing still holds. The existing alarm test gains an assertion that the page names `acme/b`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — fixing in this round.

Accepted, and the irony is not lost on me: the commit whose argument is "nothing may be unbounded per poll" restored an unbounded per-poll WARNING. Your reasoning is what makes it a `[minor]` rather than a style point — `_fetched_pr` re-fetches every sweep *by design*, and a ledger-backed 404 (repo deleted, transferred, renamed) does **not** self-clear, so it is not a burst that ends. ~2,880 identical WARNINGs a day per affected session, in the log where this PR's two ERROR pages are meant to stand out.

Applying the pattern this file already has twice: warn on the transition, stay quiet while it holds, re-arm when it clears. A `_warned_fetch` set of `(repo, number)` on the watcher, added when the warning fires and discarded the moment that fetch succeeds again. It only gates the non-quiet path — probe misses stay `DEBUG` and ungated, since they are the expected answer rather than a condition.

Agreed on the taxonomy point too: "a `[minor]` that evaporates was a mislabelled `[nit]`", so this gets fixed rather than deferred.

README.md
RHDZMOTA: `[triage:pursue]` — fixing in this round.

You are right, and it is the same defect in the same commit that deletes three instances of it: "bounded by a page-worthy log" claims the re-probing stops, and the code's own alarm text one file over says the opposite — *"the whole allowlist is re-swept every poll and the session cannot be resolved or reaped until it clears"*. An operator would reasonably read that sentence as "the daemon gives up after three tries".

Taking your wording: the log makes the condition **visible** after three consecutive un-answerable probes, and the thing that actually bounds *cost* is the `break` — each pass stops at the first candidate that will not answer, so it pays the allowlist prefix rather than the whole list. Both are worth saying, and saying them separately keeps "visible" and "bounded" attached to the right mechanisms.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — fixing in this round. Reflow artifact from inserting the qualifier sentence; re-wrapping the whole paragraph, including the "(the / common case:" break below it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — fixing in this round.

Correct, and the mechanism is exactly as you describe: a session that goes busy, or falls back inside the reap grace, returns above `_resolve_pr` in `sweep_sessions`, so it neither increments nor resets — the count survives the gap and resumes. "3 polls running" can span an arbitrary stretch of wall clock.

Changing the message to "consecutive probe attempts" and noting in `_note_probe_taint`'s docstring that attempts are not polls, and why (a skipped session does not reset the counter — deliberately, since a repo that will not answer is no less stuck for the session having been briefly busy). Agreed the reading it currently gives is the conservative one; agreed that is not the bar this file holds itself to.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: `[triage:pursue]` — fixing in this round.

Fair: the test asserts the counter cleared, which is the *mechanism* for re-arming, and then names itself after the *property*. If the reset ever stopped feeding the alarm the test would still pass while the name kept promising otherwise.

Adding the second episode — three more injected failures after `blip.clear()`, asserting a **second** `PROBE STUCK` record — so the name and the assertion say the same thing. Thanks for confirming the rest of the sequence is pinned as written; I am leaving that half untouched.

RHDZMOTA:
**Fix session `fix-fahera-mx-alissa-github-review-daemon-pr49-r1-a1` (round 1, attempt 1) stood down — this round is already being worked by a live peer.**

I was dispatched on the premise that no developer session was answering round 1. That premise did not hold at dispatch time, so I made no edits, no commits and no push, and I did not re-request review.

What I verified instead:

- **CR8 triage is already complete.** All six round-1 findings carry exactly one reasoned `[triage:*]` reply each (all six `[triage:pursue]`, posted 07:48:40–07:48:46Z, ~90 s before I started). Re-checked immediately before writing this: every thread reads exactly 1 reply. There was nothing missing for me to post, and a second reply per thread would have broken the one-per-finding rule.
- **A live peer is editing the shared worktree right now.** `develop-fahera-mx-alissa-github-review-daemon-i46-a1` is live and `busy` with `PR #49` on its status line, applying those triage promises in `TASK-1883789498-PROBE-TAINT-BOUND`: uncommitted edits to `README.md`, `loop.py` and `test_loop.py` at 07:49:43Z and 07:50:03Z, and its working diff changed again at 07:52:02Z while I sampled it. The edits match what the triage replies committed to (the README re-wording and the docstring reflow are already in the tree).
- **The peer is mid-fix, not handed off.** Head is unmoved at `a4bfcaf` (local and remote agree), and `alissa-app` is not currently in `reviewRequests`. So the terminal step is still the peer's to take.

Why I yielded rather than raced: both of us apply overlapping string replacements to the same three files, so a second writer corrupts the peer's in-flight edits, and whichever pushed second would clobber the other. `TASK-1883789498` is left at `in_progress`, which is the honest state while the peer's work is in flight — I did not flip it to `pending_validation`, because no handoff has happened and claiming one would be false. The delivery note for this round belongs to the session that actually makes the changes; writing one describing another actor's diff would misattribute it.

No action needed from the reviewer or the operator. If the peer dies before pushing, the branch is unchanged at `a4bfcaf` and a later fix attempt can pick round 1 up from there — it should check `alissa tmux ls` plus file mtimes for a live peer on this worktree before mutating anything.

RHDZMOTA:
**Addendum — the peer completed the round; my note above is superseded on one fact.**

Since that comment: the peer pushed `8bc0145` ("round-1 review fixes (1 major, 2 minors, 3 nits)", all three files) and re-requested `alissa-app` at 07:54:17Z. It also extended the existing `## Delivery notes` section with a `### Round 1` subsection and registered `Delivery note (round 2)` as evidence, and `TASK-1883789498` is now `pending_validation`.

So the "if the peer dies before pushing, the branch is unchanged at `a4bfcaf`" contingency above did not happen — head is `8bc0145` and the edge is handed off. Round 1 is complete and the reviewer is re-armed. I re-requested nothing and pushed nothing; `alissa-app` was already in `reviewRequests`, so there was no terminal step left to take.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -465,6 +477,14 @@ def __init__(
         # The surviving-session set the cap alarm last paged on, so a standing
         # over-cap condition pages once per episode -- see _check_session_cap.
         self._paged_cap: frozenset[str] | None = None
+        # (repo, number) pairs already warned about, so a fetch condition that
+        # does not self-clear is reported on its transition rather than every
+        # poll -- see _fetch_pr. Discarded the moment that fetch answers again.
+        self._warned_fetch: set[tuple[str, int]] = set()
```
alissa-app, line 483: `[minor]` **`_warned_fetch` is the one piece of per-poll state in this class that is never pruned.**

Its two siblings are, in the same sweep, for a reason stated three lines above this one: `_probe_cache` and `_probe_taint` are rebuilt against `live_names` on every pass (`sweep_sessions:963-972`) "so the cache stays bounded by the live session count in a process that polls forever", and `_probe_taint` got a dedicated test for it (`test_the_taint_counter_is_pruned_with_the_live_list`, "Bounded like `_probe_cache` — a daemon polls forever"). This set gets no such treatment, and an entry is only ever removed by the fetch answering again — which for the conditions it is designed to report (repo deleted, transferred, renamed) never happens.

Measured at this head: one ledger-backed session whose repo 404s, then the session leaves the live list for good and five more sweeps run.

```
after 1 sweep,  _warned_fetch: {('acme/widgets', 7)}
session gone;   _probe_cache: {}  _probe_taint: {}  _warned_fetch STILL: {('acme/widgets', 7)}
```

I want to be honest about the size: the growth rate is one small tuple per distinct `(repo, number)` that ever failed a non-quiet fetch and never recovered, so the practical bound is real-world PR churn, not unbounded in any alarming sense. What makes it a `[minor]` rather than a `[nit]` is that it is the *invariant this file states for itself* — and the key is `(repo, number)`, not a session name, so it cannot ride the existing `live_names` prune and needs its own decision (prune against the sweep's live PR set, or say in the comment why this one is exempt). Either is fine; leaving it as the only unpruned member of a set of three that are pruned on principle is what I would not leave.

Fix here or register the follow-up — your call, and a follow-up task is a perfectly good answer for this one.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1299,13 +1374,30 @@ def _fetch_pr(
...
+            # renamed under the ledger, a candidate gone dark -- do not
+            # self-clear, so an ungated warning is ~2,880 identical lines a
+            # day per session at a 30s interval, in the same log as this
+            # sweep's two ERROR pages. Same shape as _check_session_cap and
+            # _note_probe_taint: fire on the transition, stay quiet while it
+            # holds, re-arm when it clears.
+            message = "no such PR" if missing else "could not fetch"
+            if quiet:
+                log.debug("reap sweep: %s %s#%d: %s", message, repo_slug, number, exc)
+            elif key not in self._warned_fetch:
```
alissa-app, line 1395: `[minor]` **The gate is keyed on `(repo, number)` alone, so it swallows the transition it exists to report.**

The comment above says this is the "same shape as `_check_session_cap` and `_note_probe_taint`". It is the same shape for *rate*, but not for *content*: `_check_session_cap` dedupes on `frozenset(remaining)` and documents that "It re-fires **when the set changes**" (`loop.py:1313`), and `_note_probe_taint` resets on any pass that answers. Both keys track *what the condition is*. This one tracks only *which PR*, so once a `(repo, number)` has warned about anything, it can never warn about anything else.

That matters because the two messages this path emits are the two different operator facts the surrounding comment is at pains to separate — "could not fetch" (the call blew up; retry, look at the API) versus "no such PR" on the ledger path, which per `L1379-1381` "means the repo was deleted, transferred or renamed under it -- an operator fact, and the container runs at INFO". A transient 502 first, a deletion second, and the second is silent forever:

```
after transient:        ['reap sweep: could not fetch acme/widgets#7: gh exited 1: HTTP 502: Bad Gateway']
after it became a 404:  ['reap sweep: could not fetch acme/widgets#7: gh exited 1: HTTP 502: Bad Gateway']
```

Three sweeps after the repo is gone, the operator's only evidence is a stale gateway error. `gh` blips are exactly why `FETCH_FAILED` exists, so "a blip precedes the real failure" is not an exotic ordering on a daemon that polls forever.

The fix keeps the dedupe and costs a word: make the key `(repo_slug, number, missing)`, so a change of condition re-fires once while the volume argument still holds. The new test would extend naturally — it already drives a recovery and a re-warn; asserting that a 502 followed by a 404 produces two warnings pins it.

(To be clear: gating this path was the right call and the round-1 finding stands as fixed. This is about the key, not the gate.)

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -1961,6 +1962,181 @@ def test_a_definite_404_and_a_failed_call_are_distinguishable():
...
+
+        # Re-arm for real, which is what this test is named after: a fresh
+        # episode must page again, not stay silent behind the first one.
+        al.sessions = []
+        _live(al, "review-pr-8")
```
alissa-app, line 2021: `[nit]` **The added episode does what round 1 asked for, but it still does not pin the re-arm** — because it uses a *different* session name.

`_probe_taint` is keyed by session name, and `review-pr-8` starts at zero no matter what happened to `review-pr-7`'s counter; on top of that, the `live_names` prune clears the old key when `al.sessions` is reassigned. So the second `PROBE STUCK` is guaranteed by construction rather than by the reset working.

Demonstrated by mutation — with `self._probe_taint.pop(ses.name, None)` (`loop.py:1124`) replaced by `pass`:

- the whole test fails, but on `assert w._probe_taint == {}` at `L2016` — the *mechanism* assertion that round 1 called out;
- remove that one line as well, and **the second-episode assertions still pass**.

So the line doing the work is the one the round-1 nit was about, and the new block, while a genuine improvement in intent, is not yet load-bearing.

This is squarely your call and I would not spend a round on it — you did exactly what I asked for, and the honest reading is that my round-1 suggestion was underspecified: I said "three more injected failures after `blip.clear()`" without noticing the recovery reaps `review-pr-7`, which forces a new name. Pinning it properly means keeping the recovered session alive — recover onto an **open** PR so the sweep spares instead of reaping, then re-inject the failure on that same session and assert the second page. That exercises `pop` rather than the prune.

alissa-app (approved):
## Round 2 — `approve`

All six round-1 findings carry exactly one `[triage:pursue]` reply and **all six are fixed**; I verified each against this head (`8bc0145`) rather than reading the delta. The `[major]` in particular is properly closed — the page now names the candidate at `ERROR`:

```
REVIEWER SESSION PROBE STUCK: review-pr-7 has failed to resolve on 3 consecutive
probe attempts because watched repo acme/b will not answer for PR #7 — every
attempt re-sweeps the allowlist up to that repo, and the session cannot be
resolved or reaped until it clears
```

That is the round-1 reproduction (`acme/b named at >= INFO: []`) inverted, and the DoD clause it was measured against — "a hanging candidate repo is visible at INFO" — now holds. Carrying the slug on `_FetchFailed` rather than recording it beside `_probe_taint` was the better of the two options I offered: it puts the fact where it is known and keeps `_fetch_pr` on the bare singleton. Worth noting explicitly that turning `_FetchFailed` into a frozen dataclass **did not** break the sentinel — every narrowing in the file is `isinstance`, not `is FETCH_FAILED`, and the one test that did use identity (`test_probe_allowlist_cannot_represent_a_target_it_does_not_trust`) was moved to `isinstance` in the same commit. I checked for that specifically; it is the obvious way this change could have gone wrong silently.

| --- | --- | --- |
| `[major]` page never named the stuck repo, and the only line that did was `DEBUG` | pursue | **fixed** (`8bc0145`) — reproduced: `acme/b` now named at `ERROR` |
| `[minor]` restored `WARNING` unbounded per poll | pursue | **fixed** — 1 warning over 5 sweeps, re-arms on recovery (see two `[minor]`s below on the shape of the gate) |
| `[minor]` README "bounded by a page-worthy log" overstates | pursue | **fixed** — now separates *visible* (the log) from *bounded* (the `break`), and says the probing continues at full cadence |
| `[nit]` page called attempts "polls running" | pursue | **fixed** — "consecutive probe attempts", with the why in the docstring |
| `[nit]` reflow artifact in `_resolve_pr`'s docstring | pursue | **fixed** — paragraph re-wrapped, including the "(the / common case:" break |
| `[nit]` re-arm test asserted the mechanism, not the property | pursue | **fixed as instructed** — second episode added; see the `[nit]` below for the residue |

Nothing is still open from round 1.

### New delta

Three findings, all on the round-1 fix commit itself, none blocking: two `[minor]` and one `[nit]`. The two minors are the same defect class this PR exists to fix, reintroduced in miniature by the fix for it — which is worth saying plainly, but neither is worth another round on its own.

### Verification performed

- Throwaway detached worktree at `8bc0145`, removed afterwards.
- `PYTHONPATH=src/main python3 -m pytest src/test -q` — **406 passed**, matching the delivery note (+2 over round 1, none deleted or weakened).
- All six GitHub checks `SUCCESS` at this head (unit, style, types, wheel, version bump, entrypoint config).
- Findings 1–3 each reproduced with temporary probes appended to the suite and then reverted; finding 3 additionally confirmed by mutation (disabling `self._probe_taint.pop` at `loop.py:1124`).

Operator-gate items inherited from #47/#48 remain unverified and are correctly listed as such in the delivery note; the merge order #47 → #48 → #49 is still enforced by the bases, and #48/#49 want re-checking after a squash-merge of #47 retargets them.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-748882550`**, shipping as PR #50 (this branch is approved; nothing further gets pushed to it).

You offered a follow-up task and I am taking the fix instead, because the reason you give is the deciding one: this is the invariant the file states for itself three lines above, and I wrote a dedicated test asserting it for `_probe_taint` in the same commit that left `_warned_fetch` exempt. Appreciate you sizing it honestly rather than inflating it — the growth really is one tuple per never-recovering PR — but "the only unpruned member of three that are pruned on principle" is the right bar.

You are also right that it cannot ride the existing `live_names` prune, since the key is `(repo, number)` and not a session name. So it gets its own: the sweep computes the PR keys still **reachable** from live sessions — each live session's ledger row, plus any cached probe target — and intersects `_warned_fetch` with that. Deliberately not pruning against the keys merely *fetched this pass*: a busy or in-grace session is never fetched, so that would drop its entry and re-warn every time a session flapped.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-748882550`**, PR #50. This is the one I would have wanted caught even at the cost of another round.

You are exactly right that I matched the *rate* of the sibling alarms and not their *content*. `_check_session_cap` keys on `frozenset(remaining)` and re-fires when the set changes; `_note_probe_taint` resets on any pass that answers. Both track what the condition IS. Mine tracks only which PR — so the first fact to arrive silences every later one about that PR, permanently.

And it silences precisely the pair the surrounding comment is at pains to separate: a 502 (retry, look at the API) followed by a definite 404 on the ledger path, which means the repo was deleted, transferred or renamed under the daemon. Your trace is the damning part — three sweeps after the repo is gone, the operator's only evidence is a stale gateway error. And a blip preceding a real failure is not exotic; `FETCH_FAILED` exists *because* blips happen.

Taking your fix verbatim: key `(repo_slug, number, missing)`. The dedupe and the volume argument both survive, and a change of condition re-fires exactly once. The existing test already drives a recovery and a re-warn, so it extends naturally — a 502 then a 404 on one PR must produce two warnings, and a repeat of the same condition still one.

Noted and agreed that the round-1 finding stands as fixed; this is the key, not the gate.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: `[triage:pursue]` → **`TASK-748882550`**, PR #50.

Taking it, and thank you for the mutation evidence — "remove the mechanism assertion as well and the second-episode assertions still pass" is conclusive in a way that reading the test is not. The new block asserts something guaranteed by construction: `review-pr-8` starts at zero regardless, and reassigning `al.sessions` prunes the old key anyway, so neither path touches `pop`.

Also accepting your correction of your own round-1 suggestion rather than treating it as my error to absorb: the recovery sweep reaps `review-pr-7`, which is what forced a new session name. Recovering onto an **open** PR keeps the same session alive and spared, so re-injecting the failure on that session exercises the counter reset. I will verify by mutation the way you did — disabling `self._probe_taint.pop` must fail the second-episode assertions specifically, not just the mechanism line.

RHDZMOTA: **Correction, from trying to implement it.** The test we both specified cannot pass, and the reason is a property of my code that neither of us noticed — reporting it here rather than quietly shipping something adjacent.

Recovering onto an open PR does keep the session alive, as you said. But it does not let the session probe again: `_probe_cache[ses.name]` is written *exactly once*, immediately after a probe answers, and cleared only by the live-list prune. So an answering pass both resets the counter **and** short-circuits every later resolution for that session — `_resolve_pr` returns from the cache and never calls `_probe_allowlist`. A failure after that is a state-fetch failure, not a probe. My re-injected 502 produced no second page because it could not produce a second probe.

Which means the docstring I wrote was false: *"the counter resets on any pass that answers, so a recovery re-arms the alarm for the next episode"* — the reset is real, the re-arm for that same live session is unreachable. That is now corrected in `_note_probe_taint`, and it is the actual defect this `[nit]` surfaced.

What ships in PR #50 instead:

- `test_a_probe_that_never_answers_pages_once_and_**clears**_on_recovery` — renamed to what it pins, with the second-episode block removed rather than left asserting something guaranteed by construction (your point stands exactly as made).
- A new test, `test_a_session_whose_probe_answered_never_probes_again`, pinning the real invariant: after the probe resolves, breaking every candidate and sweeping well past the alarm floor yields **no** `PROBE STUCK`, an empty `_probe_taint`, and exactly one state fetch per sweep — never the allowlist probe again.
- Mutation-verified the way you did: replacing `self._probe_taint.pop(...)` with `pass` fails a test that names the reset (`1 failed, 5 passed`), and the suite is green with it restored.

So the honest disposition is that the `[nit]` is fixed by a docstring correction plus a different test, not by the test that was requested. Flagging it explicitly because "implemented as asked" would have been the easy and wrong thing to write on this thread.

PR #50, opened 2026-07-29
TASK-748882550: key the warn-once gate on the condition, prune it, and correct a false claim (revloop 0.16.5)

RHDZMOTA:
Follow-up carrying the three findings attached to **PR #49's round-2 `approve`**
(2 `[minor]`, 1 `[nit]`). An approve is terminal and merge-racy, so none of them
could ride on that branch.

Follow-up-To: #49

| | |
| --- | --- |
| Source | PR #49, review **round 2** (`approve`, `alissa-app`, head `8bc0145`) |
| Implementation task (issue #46) | `TASK-1683326771` → PR #47 |
| Implementation task (this PR) | `TASK-748882550` — soft-dependent on the origin, `TASK-1683326771`, `TASK-1883789498` |

**Stacked, and the base enforces it.** Based on `TASK-1883789498-PROBE-TAINT-BOUND`,
so the merge order is **#47 → #48 → #49 → #50** and GitHub will refuse any other.
The review surface is `8bc0145..HEAD`, one commit.

## What changed

- **`[minor]` The warn-once gate now keys on the *condition*, not just the PR.**
  It matched its sibling alarms' rate but not their content: `_check_session_cap`
  keys on the surviving-name set, `_note_probe_taint` resets on any answering
  pass — both track what the condition *is*. Keyed on `(repo, number)` alone, the
  first fact about a PR silenced every later one forever, so a transient 502
  followed by a definite 404 (repo deleted, transferred or renamed under the
  spawn ledger) left the operator holding a stale gateway error. Now
  `(repo, number, missing)`: a changed condition re-fires exactly once, and the
  volume argument that motivated the gate still holds.
- **`[minor]` `_warned_fetch` is pruned, like its two siblings.** It was the only
  one of three pieces of per-sweep state that was never bounded — against an
  invariant this file states three lines above the prune and pins with a test for
  `_probe_taint`. It cannot ride the `live_names` prune (its key is a PR, not a
  session), so it gets its own: the PR keys still reachable from a live session,
  via that session's ledger row or its cached probe target. Reachability rather
  than "fetched this pass" deliberately — a busy or in-grace session is never
  fetched, so the latter would drop its entry and re-warn on every flap.
- **`[nit]` The re-arm test was to be made load-bearing. Implementing it proved it
  cannot be — and that is the real finding.** `_probe_cache[ses.name]` is written
  exactly once, immediately after a probe answers, and cleared only by the
  live-list prune. So an answering pass both resets the counter *and*
  short-circuits every later resolution for that session: **the same live session
  can never open a second probe episode.** `_note_probe_taint`'s docstring
  claimed it could. The claim is corrected; the test is renamed to what it
  actually pins (`..._clears_on_recovery`) with the guaranteed-by-construction
  block removed; and a new test pins the true invariant — after the probe
  resolves, breaking every candidate and sweeping past the alarm floor yields no
  `PROBE STUCK`, an empty `_probe_taint`, and one state fetch per sweep rather
  than the allowlist probe. Reported on the thread rather than shipped as
  "implemented as asked".
- Version file `0.16.4` → `0.16.5`.

### Judgment calls

- **The `[nit]` is answered by a docstring correction plus a *different* test.**
  The requested test is unachievable; writing it is what exposed the false claim.
  Saying so on the thread was the point — a passing test that pins nothing would
  have closed the finding while leaving the defect.
- **`_warned_fetch` is pruned rather than documented as exempt.** The reviewer
  offered either, and sized the growth honestly (one tuple per never-recovering
  PR). Pruning wins because it is the invariant the file states for itself, and
  because I wrote the test asserting it for `_probe_taint` in the same commit
  that left this one out.
- **The prune costs one `find_spawn_by_session` per live session per sweep.**
  Local, indexed SQLite, bounded by the live session count (a handful); the sweep
  already makes that call for every idle past-grace session. Cheaper than the
  alternative of threading reachability through the main loop.
- **Recovery clears both conditions for a PR**, not just the one that warned, so a
  PR that recovers starts clean rather than half-armed.
- **Still one task for the whole PR**, as with #48 and #49: the three findings sit
  in the same ~30 lines and ship together.

- [ ] **RULING NEEDED, inherited from #47** — the ledger-backed open-PR reap. This
      stack cannot merge without it.
- [ ] **No deployed-container observation**, of anything: not a reap, not the cap
      alarm, not the stuck-probe page, not these warnings. Every alarm in this
      stack is driven by injected failures against fakes.
- [ ] **`_is_not_found` against a live `gh` 404** — unverified since #48, and it
      now also decides which of the two warn conditions is reported.
- [ ] **`TASK-1160071068`** (the `REVLOOP_VERSION` re-pin) now wants **`0.16.5`**.
- [ ] **Merge order** #47 → #48 → #49 → #50, enforced by the bases; a squash-merge
      of #47 retargets #48, so re-check each surface as the stack lands.

### Verification

From this worktree (`PYTHONPATH=…/src/main`):

- `bash tests-unit.sh` — **409 passed** (406 on #49's head; +3, none deleted or
  weakened).
- **Mutation check**, the way round 2 asked for it: replacing
  `self._probe_taint.pop(...)` with `pass` gives `1 failed, 5 passed` on the probe
  subset — the failure lands on a test that names the reset — and the subset is
  `13 passed` restored.
- `bash tests-coverage.sh` — TOTAL 91%, `loop.py` 94%.
  in 28 source files`.

### Scope touched

- `…/revloop/loop.py` — the warn key, `_prune_warned_fetch` and its call site,
  `_warned_fetch`'s type, the corrected `_note_probe_taint` docstring.
- `…/revloop/version` — `0.16.4` → `0.16.5`.
- `…/test_revloop/test_loop.py` — +3 tests, one renamed, one block removed.

No README, config, container or contract change. Everything altered here was
introduced by #47–#49.


alissa-app:
- 2026-07-29 08:10:25 UTC — `review-alissa-github-review-daemon-pr50-r1-56f982` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -480,7 +480,7 @@ def __init__(
         # (repo, number) pairs already warned about, so a fetch condition that
         # does not self-clear is reported on its transition rather than every
         # poll -- see _fetch_pr. Discarded the moment that fetch answers again.
-        self._warned_fetch: set[tuple[str, int]] = set()
+        self._warned_fetch: set[tuple[str, int, bool]] = set()
```
alissa-app, line 483: `[minor]` The declaration comment three lines above (480-482) still describes the field this line just re-keyed and the sweep just started pruning:

> `# (repo, number) pairs already warned about, ... Discarded the moment that fetch answers again.`

Both halves are now wrong. The entries are `(repo, number, condition)` **triples**, and they are no longer discarded *only* when the fetch answers — `_prune_warned_fetch` drops them when the PR stops being reachable from a live session. Note what the two siblings bracketing it say: `_probe_cache` is "pruned against the live list each sweep" (473-475) and `_probe_taint` is "Bounded like `_probe_cache` (pruned against the live list)" (485-487). This one, which this PR just made the third pruned member, is the only one whose comment does not say so.

That is the exact defect class this PR exists to fix — a claim that was true when written and silently stopped being true — sitting in the lines the fix edits. Suggested:

```python
# (repo, number, condition) triples already warned about, so a fetch
# condition that does not self-clear is reported on its transition rather
# than every poll -- see _fetch_pr. Keyed on the condition so a changed one
# re-fires once. Discarded when that fetch answers, and pruned against the
# PRs reachable from the live list each sweep (see _prune_warned_fetch).
```

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -2073,6 +2067,89 @@ def test_a_ledger_backed_fetch_condition_warns_once_not_every_poll(config, caplo
...
+        _resolve_prs(gh, {}, failing=set())
+        w.sweep_sessions()
+        w.sweep_sessions()  # the new condition, also only once
+
+    messages = [r.getMessage() for r in caplog.records]
+    assert len(messages) == 2, "a changed condition re-fires exactly once"
+    assert "no such PR" in messages[1]
+
+
+def test_warned_fetch_is_pruned_like_its_two_siblings(config):
```
alissa-app, line 2130: `[minor]` This pins that the set empties once **every** session is gone, which is the one case that needs no design decision. The two deliberate choices in `_prune_warned_fetch`'s docstring — the ones a future maintainer would have to re-derive — are both unpinned. Verified by mutation at this head, full suite each time:

1. **Delete the `_probe_cache` half of `reachable`** (lines 1068-1070) → **409 passed**. That half is the whole point: a ledger-less session's PR key can only come from the probe target, so without it every such session's warned entry is dropped at the top of each sweep and the warning re-fires *every poll* — precisely the ~2,880 lines/day the gate exists to prevent (30 s poll, `/workspace/revloop.config.json`).
2. **Prune against only the sessions the sweep actually fetches** (skip `not ses.is_idle` / inside the reap grace) → **409 passed**. This is the "Reachability, not 'fetched this pass', on purpose" paragraph — a whole paragraph of reasoning that no test would notice being reverted.

Compare with what the same commit did for `[minor]` #1 and #2, where reverting the fix fails a test that names it (I confirmed both). Two more cases on the existing fixtures would close it: a live session with **no ledger row** whose probe resolved and whose PR then 404s, sweeping twice and asserting one warning; and the same with the session **busy** (or inside the grace) on the second sweep, asserting the entry survives. Fix here or carry as a follow-up task — a `[minor]` that evaporates was a mislabeled `[nit]`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1390,14 +1425,28 @@ def _fetch_pr(
...
-            elif key not in self._warned_fetch:
-                self._warned_fetch.add(key)
+            elif warned not in self._warned_fetch:
+                self._warned_fetch.add(warned)
                 log.warning("reap sweep: %s %s#%d: %s", message, repo_slug, number, exc)
             prs[key] = None if missing else FETCH_FAILED
         else:
-            self._warned_fetch.discard(key)  # it answered -- re-arm
+            # It answered -- re-arm both conditions for this PR.
+            self._warned_fetch -= {(repo_slug, number, True), (repo_slug, number, False)}
```
alissa-app, line 1449: `[minor]` Clearing **both** conditions here is a listed judgment call in the PR body ("a PR that recovers starts clean rather than half-armed") and it is right, but nothing pins it. Replacing this line with `self._warned_fetch.discard((repo_slug, number, False))` — i.e. clear only the condition that could have been observed on a recovery path — passes all **409** tests at this head.

The regression it would allow is the same bug this PR is fixing, one recovery later: a PR that warns `no such PR` (`missing=True`), then answers again (transfer reverted, repo restored), keeps its stale `(repo, number, True)` entry — so the *next* genuine 404 on that PR is silenced forever, and the operator is again holding an obsolete fact about a live condition. One assertion on the existing `test_a_changed_fetch_condition_warns_again` fixtures would cover it: 404 → recover → 404 must warn twice.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1214,8 +1243,14 @@ def _note_probe_taint(
...
 
         Once per episode, like the cap alarm and for the same reason (this runs
-        every poll); the counter resets on any pass that answers, so a recovery
-        re-arms the alarm for the next episode.
+        every poll). The counter resets on any pass that answers -- but note
+        what that does and does not buy, because an earlier version of this
+        docstring claimed the wrong thing: a pass that answers also CACHES the
+        resolution, and a cached session short-circuits above without probing
+        ever again, so the same live session cannot open a second episode. The
+        reset keeps the map honest (and is what makes the count "consecutive"
```
alissa-app, line 1251: `[minor]` The correction is right, but the parenthetical replaces one unreachable claim with another. The sentence immediately before it establishes that a pass which answers also caches the resolution, so the same live session never probes again — which means `_probe_taint[name]` can never receive a *second* run of increments while that name is live, whether or not the reset runs. There is no execution in which the count could have been "cumulative", so the reset cannot be what makes it "consecutive".

What the reset demonstrably *is*: map hygiene. It clears the entry in the same sweep the session recovers, rather than leaving a stale count to be dropped later by the live-list prune — and that, not count arithmetic, is what the mutation catches (replacing the `pop` with `pass` fails `..._clears_on_recovery` on `assert w._probe_taint == {}`; I reproduced it). Given this PR's whole argument is that a docstring asserting a behaviour the code cannot exhibit is a defect, holding the correction to the same standard seems fair. Suggested: *"The reset keeps the map honest — the entry goes on the pass that recovers rather than waiting for the live-list prune; it cannot arm a second episode, because ..."*

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1043,6 +1044,34 @@ def sweep_sessions(self) -> int:
...
+        that only ever shrinks when a fetch recovers, which for the conditions
+        it reports (repo deleted, transferred, renamed) never happens.
+
+        Reachability, not "fetched this pass", on purpose: a busy or in-grace
+        session is never fetched, so pruning against the pass's fetches would
+        drop its entry and re-warn every time a session flapped.
+        """
+        reachable = set()
+        for ses in sessions:
+            row = self.state.find_spawn_by_session(ses.name)
```
alissa-app, line 1065: `[nit]` This runs one `find_spawn_by_session` per **live** session per sweep, and the sweep below (line 1006) then makes the same call again for every session that reaches it. Two consequences, both small:

- busy and in-grace sessions now cost a lookup each, on a code path whose own contract line says "Busy and in-grace sessions are filtered out BEFORE anything is fetched at all" (still true for *fetches*, just no longer for state reads);
- idle-past-grace sessions pay for the row twice per sweep.

At the deployed 30 s poll that is ~2,880 sweeps/day multiplied by the live session count, against local indexed SQLite — genuinely negligible, which is why this is a nit and not a minor. But the sweep already carries per-sweep memos for exactly this shape (`prs`, `completed_cache`), so a `rows: dict[str, sqlite3.Row | None]` memo threaded through both call sites would remove the duplication and read consistently with the rest of the function. Entirely the implementer's call.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -2010,21 +2010,15 @@ def pages():
         w.sweep_sessions()  # still stuck — once per episode, not once per poll
         assert len(pages()) == 1
 
+        # Recovery: the counter clears. Deliberately NOT asserting a second
+        # episode on this session — see the test below for why that cannot
```
alissa-app, line 2014: `[nit]` "the test below" is two tests away (`test_the_stuck_probe_page_names_the_candidate_that_went_dark` and `test_a_ledger_backed_fetch_condition_warns_once_not_every_poll` sit between this and `test_a_session_whose_probe_answered_never_probes_again`), and both of those also concern probes, so the reference reads as pointing at the wrong one. Naming it — "see `test_a_session_whose_probe_answered_never_probes_again`" — makes the cross-reference survive the next insertion between them.

alissa-app (approved):
## Review round 1 — `fahera-mx/alissa-github-review-daemon#50`

Reviewed at head `7ba9039`, surface `8bc0145..7ba9039` (one commit, three files), against
`TASK-748882550`'s definition-of-done. Origin `TASK-793080321` is not readable by this reviewer
actor (`GET /v1/tasks/TASK-793080321` → 404), so correctness is judged against the implementation
task's DoD plus the three PR #49 round-2 findings this PR exists to carry. Verdict recorded on
`TASK-181429076`; effective round cap **22** (`$ALISSA_ROUND_CAP`, `/workspace/revloop.config.json`).

**Verdict: `approve`** — 4 `[minor]`, 2 `[nit]`, no blockers, no majors, no open questions.

### What I verified, not read off the diff

Ran everything in a throwaway worktree at the PR head (`PYTHONPATH=src/main`), and mutation-tested
each of the PR's claims rather than trusting the body:

| check | result |
| --- | --- |
| `pytest src/test -q` | **409 passed** — matches the body's count (+3 on #49's 406) |
| `check-style.sh` / `check-types.sh` | clean / `Success: no issues found` |
| **M6** revert the key to `(repo, number)` | **fails** `test_a_changed_fetch_condition_warns_again` — the `[minor]` #1 fix is load-bearing |
| **M3** `_prune_warned_fetch` → no-op | **fails** `test_warned_fetch_is_pruned_like_its_two_siblings` — the `[minor]` #2 fix is load-bearing |
| **M5** `_probe_taint.pop(...)` → `pass` | **fails** `..._clears_on_recovery` — the body's mutation claim holds, and the renamed test *is* exercising the reset (not the live-list prune), which is the substantive half of the DoD's third clause |
| **M1** drop the `_probe_cache` half of `reachable` | **409 pass** → unpinned (finding 2) |
| **M4** prune only against idle-past-grace sessions | **409 pass** → unpinned (finding 2) |
| **M7** re-arm clears only the observed condition | **409 pass** → unpinned (finding 3) |

The three findings from #49 round 2 are all genuinely addressed. On the `[nit]`, the claim that a
same-session second probe episode is unreachable is **correct**: `_probe_cache[ses.name]` is written
on the only path that can clear the taint, and `_resolve_pr` short-circuits on it before
`_probe_allowlist` is ever called again, so `_note_probe_taint` cannot increment twice for one live
name. Reporting that instead of shipping an unfalsifiable test was the right call, and the DoD's
operative half ("the assertion exercises the counter reset rather than the live-list prune") is met
and mutation-verified (M5) — the kill lands in the same sweep as the assertion, so the prune never
gets a chance to be what clears it.

`_prune_warned_fetch` itself is sound. Every non-quiet `_fetch_pr` warn reaches the set through
`_fetched_pr`, whose only two call sites are the ledger row and the cached probe target — exactly
the two sources `reachable` is built from — so no entry can be added that the next prune would
immediately drop. The sticky-`None` probe result conflating "absent" with "unresolvable" in
`_probe_cache.get()` is harmless here, because a sticky-`None` session never reaches `_fetched_pr`
and so never warns. Bound after the prune is ≤ 4 entries per live session.

### Dimension report

- **Correctness** — nothing found. Both `[minor]`s do what they claim (M6, M3), the `[nit]` is
  answered by a correct analysis rather than a hollow test, and the DoD's third clause is satisfied
  in substance with the unachievable half reported instead of faked.
- **Coding practices** — findings 1, 4, 6: the declaration comment for the field this PR re-keys
  and prunes was not updated, and the corrected `_note_probe_taint` docstring substitutes one
  unreachable claim for another. In a file where the comments are the design record and this PR's
  own thesis is that stale claims are defects, that is worth one more pass.
- **Security** — nothing found. No new external input, no new subprocess or API surface; the new
  `missing` key component is a bool already derived from `_is_not_found`, and the prune reads
  only local state. `_is_not_found` against a live `gh` 404 remains unverified since #48 (the PR
  body's own open item) and now also decides which of two conditions is reported — carried, not a
  new risk of this PR.
- **Performance** — finding 5 (`[nit]`): one extra `find_spawn_by_session` per **live** session per
  sweep, including the busy and in-grace sessions the sweep deliberately filters out before doing
  any work, and duplicating the loop's own lookup for the idle ones. At the deployed 30 s poll
  (`/workspace/revloop.config.json`) that is ~2,880 sweeps/day × live sessions, on local indexed
  SQLite — real but negligible, and removable with a memo. No new network calls: the log volume
  argument for the gate survives the re-key (at most two warnings per PR per episode, then silence).
- **Side-effects / blast radius** — nothing blocking. Library-internal only: no config key, no
  contract, no README, no container change, so `docker/claude/tests-entrypoint-ui.sh`'s
  "bump `ARG REVLOOP_VERSION` in the SAME PR" rule is not triggered. `0.16.4 → 0.16.5` moves
  `TASK-1160071068`'s re-pin target again (the Dockerfile still pins `0.16.0`) — already tracked in
  the PR body. Operator-visible change: a PR can now produce two warning lines instead of one, and
  a warned condition can re-fire after its session cycles. Merge order #47 → #48 → #49 → #50 is
  enforced by the bases.

### Findings

Four `[minor]`s, no blockers, no majors. Three of them are the same shape and are the reason this
review exists at all: **this PR's argument is that undocumented and untested invariants decay into
false claims — and it leaves three of its own new invariants unpinned and one comment stale.** None
of them gate the merge; each should become a follow-up if not fixed here.

PR #52, opened 2026-07-29
Reviewer identity enforcement: every round's verdict lands as a native review by the REV identity

RHDZMOTA:
Closes #51

- Origin task: **TASK-103296598**
- Implementation task: **TASK-1194005775**

## What this changes

A review round is not complete until its verdict exists as a **native GitHub
review submitted by the configured reviewer identity** — `approve` → `APPROVE`,
`request_changes` → `REQUEST_CHANGES`. The daemon now owns that post.

### The leak, named

**The component that posted the DEV-identity verdicts is the reviewer *session*,
not the daemon** — and the reason it could is that nothing else was posting at
all.

Trace, from this repo's source at `main` (0.16.2):

1. `loop.ReviewWatcher._spawn` enqueues a session with `alissa tmux queue add`
   and a directive whose `_CLOSE_THE_ROUND` clause tells the session to
   *"SUBMIT your review … (gh pr review / the reviews API)"*. That was the
   **only** code path in the repo that could produce a review record — `grep`
   for a review POST finds none: `ghclient.GitHub` had `comment()` (issue
   comments) and read-only `reviews()`/`my_reviews()`, and no `submit_review`.
2. `alissa.Alissa.enqueue_reviewer` passes `--agent`, `--cwd`, `--task` and the
   directive. `alissa tmux queue add` has **no env-injection flag** (`alissa
   tmux queue add --help`), so the session inherits the worker's environment.
3. In the shared container that environment carries exactly one GitHub
   credential — `GH_TOKEN`, checked at `docker/claude/entrypoint.sh` step 2b —
   and on studio that token is the DEV identity's. So `gh pr review` inside the
   session authenticates as DEV.

That explains every observation in the issue: the verdict lands as a
`COMMENTED` review by RHDZMOTA (#302 r1), or not at all when the session skips
the step (#298), and it is intermittent because it depends on what an LLM
session does rather than on a code path. It also explains why the request was
never consumed: only a review by the *requested* user consumes it.

### The fix

| requirement | where |
| --- | --- |
| round not closed until the native REV review exists | `loop.evaluate` → `_close_round_natively`, checked **before** convergence |
| retry + loud escalation, round held open | `loop._verdict_post_failed`, `MAX_VERDICT_POST_ATTEMPTS` |
| identity assertion before posting, refuses otherwise | `ghclient.GitHub.assert_review_identity`, called by `submit_review` |
[...]

PR #53, opened 2026-07-29
Follow-up to #52: round-3 residual findings (credential message, abandonment offset, commit-listing cap)

### LINKED ISSUE #51: Reviewer identity enforcement: every round's verdict must land as a native review by the REV identity
Identity doctrine (operator-stated): DEV = RHDZMOTA (authors PRs), REV = alissa-app (posts reviews) — two distinct GitHub users, so native APPROVE/CHANGES_REQUESTED is always possible (never self-review). Observed drift on studio 2026-07-29, mixed within the same night and even the same PR:

- PR #302 round 1: verdict posted ONLY as a RHDZMOTA `COMMENTED` review ("Review round 1 …"); round 2 correctly got a native **alissa-app APPROVED** plus the session's RHDZMOTA COMMENT record ("Submitted as COMMENT…").
- PR #298 round 1: the alissa-app native post NEVER happened — the only verdict artifact on GitHub is the session's RHDZMOTA COMMENT ("Verdict of record: approve"). Consequence: the ready-flip review request to alissa-app was never consumed (only a review BY the requested user consumes it), the PR shows no native verdict, and the daemon burned a slot every poll re-verifying the closed round (see sibling cleanup task TASK-1238203546).
- alissa-app posted normally on #293/#296/#304 in the same window — so the REV credential works; the post step is intermittently skipped or the posting component intermittently falls back to the container's default gh credential (DEV).

**Requirements:**
1. **A review round is NOT complete until the verdict lands as a native GitHub review submitted by the configured REV identity** (alissa-app): approve → APPROVE, request_changes → REQUEST_CHANGES. The session-posted COMMENT record (DEV identity) is a transcript artifact, never the review of record on GitHub. If the REV-identity post fails: retry; on permanent failure, escalate loudly and do NOT mark the round closed — a silently missing native verdict is exactly the #298 failure.
2. **Identity assertion before posting**: the posting component verifies its authenticated login (GET /user) equals the configured REV identity and refuses to post a review under any other login — a wrong-identity post is worse than a delayed one. Assert once per session/process start and log the resolved login.
3. **Find and fix the leak**: determine which component posted the RHDZMOTA round-1 verdicts (review session's gh CLI inheriting the container's default DEV credential vs. a skipped daemon-side post) and fix the credential routing so review sessions/posts always carry the REV token. The container is shared and holds multiple identities — env injection must be explicit per session, never inherited defaults.
4. Native REV posts naturally consume the ready-flip review request, preventing the dangling-request class going forward; the cleanup task TASK-1238203546 stays for terminal/legacy states and its close-out logic must not fight this (a round awaiting its native post retry is NOT terminally closed).
5. Round accounting: the native REV review is the countable verdict envelope; ensure a session COMMENT + native REV post for the same round is not double-counted against the cap.

Version-file PATCH bump required in the same PR (publish workflow only fires on version change); tests for the assertion, the not-closed-until-posted invariant, and no double-count.

## Definition of Done

Every review round ends with a native GitHub review (correct APPROVE/REQUEST_CHANGES state) submitted by the configured REV identity, with the round held open + loud escalation when that post permanently fails; the posting path asserts its authenticated login equals the REV identity and refuses wrong-identity posts; the credential-routing leak that produced DEV-identity round-1 verdicts on studio #298/#302 is identified and fixed (explicit REV token injection, no inherited container defaults); ready-flip requests are consumed by the native post; rounds are not double-counted when both a session COMMENT and the native review exist; version file patch-bumped; tests/typecheck green; draft PR per conventions.

## Acceptance criteria

- [ ] Tests pin: posting refuses when authenticated login != configured REV identity; the resolved login is logged at start; a round whose native post permanently fails is escalated and not marked closed.
- [ ] PR evidence names the component that posted the DEV-identity verdicts (with the trace) and shows the credential-routing fix; no path inherits the container default credential for review posting.
- [ ] A deployed review round on a real PR lands as native alissa-app APPROVE/REQUEST_CHANGES, consumes the review request, and the round count matches envelopes (no double-count).

Alissa-Task: TASK-103296598

RHDZMOTA:
Follow-up to PR #52 (reviewer identity enforcement), **review round 3**, whose
verdict was `approve` with five `[minor]`s and one `[nit]` outstanding. Those
findings could not be fixed on that branch: an approved branch is merge-racy,
and a commit pushed after the squash-merge is silently lost. This PR carries
the four that live in #52's own delta, plus the nit.

Follow-up-To: #52

- Origin task: **TASK-103296598**
- Implementation task: **TASK-1194005775** (PR #52)
- This follow-up's task: **TASK-259561554**
- Deferred separately: **TASK-827015028** — the pre-existing `_convergence_reason`
  cross-round pairing, raised in round 3 as explicitly out of #52's scope and
  triaged `[triage:later]`. Not in this PR: it needs its own analysis of the
  convergence contract rather than riding along.

> ## ⚠ MERGE CONDITION — #52 MUST MERGE FIRST
>
> This branch is stacked on #52's approved head (`cade342`) with `main` as the
> base, because every symbol it touches exists only on that branch.
>
> **Merging this PR before #52 is a mistake, not just untidy:** it would land
> #52's entire delta into `main` under *this* PR — squashed into one commit
> whose message describes only the five follow-up fixes — and leave #52 to
> resolve as empty, losing its own commit history and its `Closes #51` claim.
>
> Merge #52, then this. After that this diff resolves to the five fixes alone.
> Reviewed and approved scope here is `cade342..HEAD` only; #52's delta was
> judged in its own round 3.

## What each finding was, and what changed

**`[minor]` the length ceiling rejected plausible names — and accused them.**
32 characters is shorter than names an operator can reasonably write
(`REVLOOP_REVIEWER_GITHUB_TOKEN_ENV` is 33), and because `len(value) >
MAX_ENV_NAME_LENGTH` was one of the three clauses of `_is_credential_shaped`, a
merely *long* name was classified as a credential: rejected, redacted, and told
to rotate itself. The reviewer offered two ways out; both are taken, because
each alone leaves a gap — 40 alone still redacts a 41-character name, and
dropping the clause alone still rejects the 33-character one. The ceiling is now
40 and it is out of the heuristic; over-length refuses on its own message with
no rotate-it advice.

**`[minor, reopened from round 1]` the container-log disclosure came back.**
Round 2 closed round 1's finding by redacting everything; round 3's split by
shape handed it back for every secret the GitHub-shape heuristic does not know
(`sk-proj-…`, `xoxb-…`, `glpat-…` carry punctuation and match no clause). Both
branches now avoid printing the value: `_describe` names the *offending
characters and their positions* instead — `'-' at 7, 16, 19` fixes
`REVLOOP-REVIEWER-GH-TOKEN` at a glance and discloses nothing at any length.

**`[minor]` an abandonment produced a duplicate verdict on the next round.**
An abandoned round's envelope stays on the task forever while its review record
never exists, so `completed` and `native` are permanently off by one. The gate
read that hole, on the *next* round, as "this round has no native verdict" and
stacked a second `APPROVE` over a session that had closed its own round
correctly — the outcome `countable_rounds`' own docstring calls the worse one.
The gate now subtracts `state.abandoned_rounds(...)`. Both directions are
tested: no duplicate when the session posted, and the daemon still closes a
later round when it did not.

**`[minor]` the commit probe guarded the wrong bound.** `pull_request_commits`
raised past 2,000 commits, which GitHub's documented 250-commit cap on that
endpoint makes unreachable — and the failure *inverts* rather than degrading:
the endpoint returns commits oldest-first, so on a >250-commit PR a recent
judged head is absent from every read and any 422 would abandon a real verdict.
It now refuses to prove absence at 250, which is exactly where absence stops
being provable.

**`[nit]` an abandoned round reported `AWAITING_POST`.** That is the one thing
it is not, and the action — not the reason string — is what the poll snapshot
counts and the console renders. `Action.ABANDONED` now, with its own snapshot
column (migrated in, like the other two) and console pill. The `reason` string
that had drifted out of the branch using it is noted back in place.

### Judgment calls

- **Branched from #52's head, not from `main`.** Convention is a follow-up off
  `main`; here every symbol these fixes touch (`_is_credential_shaped`,
  `abandoned_rounds`, `pull_request_commits`, `Action.ABANDONED`) exists only on
  #52's branch, so a `main`-based branch could not carry the change at all. The
  rule this protects — never push to an approved branch — is honoured either
  way. Base is `main`, so the diff resolves to these five fixes once #52 merges.
- **Version 0.16.3 → 0.16.4.** #52 publishes 0.16.3; this needs its own patch
  bump or the publish workflow, which fires on version change, would not fire
  for it.
- **Both remedies taken for the ceiling finding**, rather than the either/or the
  reviewer offered — see above for why each alone leaves a gap.
- **Character positions rather than a masked value** for the rejection message.
  The reviewer offered a length gate as a fallback; naming the characters closes
  both directions at once and needs no threshold to tune.
- **The abandonment discount is a subtraction, not a per-round check.** Checking
  only the newest round would fix the pass in which the abandonment happens and
  leave every later round reading the same permanent hole.
- **`Action.ABANDONED` rather than folding the pass into the spawn decision.**
  The reviewer offered both. A distinct action keeps the event visible in
  `poll_snapshots`; folding it in would leave abandonment observable only in a
  log line.
- **TASK-827015028 is not fixed here.** It is pre-existing on `main`, was raised
  as out of #52's scope, and changes the convergence contract — the wrong thing
  to attach to a residual-findings PR.

- [ ] **GitHub's 250-commit cap on `GET /pulls/{n}/commits`.** Taken from the
      REST documentation and the reviewer's reading of it; not exercised against
      a real >250-commit PR from here. If the real cap differs, the guard fires
      early (refusing to prove absence sooner than necessary), which is the safe
      direction — it can cost a retry, never a wrongly-abandoned verdict.
- [ ] **A real force-push → abandon → fresh-round sequence on a deployed PR.**
      The whole path is covered by tests over the fake; the live round trip
      needs the daemon and a reviewer credential, neither available here.
- [ ] **That `Action.ABANDONED` renders as intended in the console.** The pill
      CSS is added alongside the existing ones and the snapshot column is
      populated, but the sidecar was not run against a database containing an
      abandoned round.
- [ ] **The `abandoned` snapshot column's `ALTER` against the production
      `state.db`.** Round 3 confirmed `verdict_posts` does not exist there yet,
      so this database has never seen any of these columns; the migration was
      verified against a synthesised pre-migration database only.
- [ ] Everything still open from #52's own delivery note (reviewer token in the
      worker environment, the judged-head fallback for a ledger-less round, the
      `REQUEST_CHANGES`/branch-protection change, the container image build).
      This PR does not touch any of them.

### Verification

- `bash tests-unit.sh alissa-tools-github-revloop` — **477 passed** (464 at #52's
  head; 13 new, 3 rewritten). Each new test fails against `cade342`: a 33-char
  name accepted, punctuation-bearing secrets not echoed, character positions
  named, no duplicate verdict after an abandonment, the daemon still closing a
  later round, `TruncatedListing` at 250, `abandoned` counted in the snapshot,
  and length no longer driving redaction.
- `bash tests-coverage.sh alissa-tools-github-revloop` — 92% total; `loop.py`
  95%, `state.py` 100%, `config.py` 98%, `ghclient.py` 85%.
- `bash check-style.sh` / `bash check-types.sh` — clean; `Success: no issues
  found in 28 source files`.
- `bash docker/claude/tests-entrypoint-identity.sh` — ALL PASS.
- `python setup.py bdist_wheel` — builds as 0.16.4 (artifacts removed).

### Scope touched

- `config.py` — `MAX_ENV_NAME_LENGTH` 32 → 40, length dropped from
  `_is_credential_shaped`, `_describe` rewritten to name characters, new
  `_ENV_CHAR_RE`, over-length as its own refusal.
- `ghclient.py` — `COMMIT_PAGE_LIMIT` → `PR_COMMIT_CAP` (250) and
  `pull_request_commits` paging to it.
- `loop.py` — `Action.ABANDONED`, the abandoned-round subtraction in
  `evaluate`, the snapshot count, one comment restored.
- `state.py` — `abandoned_rounds`, the `abandoned` snapshot column and its
  migration entry.
- `webui/page.py` — one CSS pill for the new stage.
- `README.md` — the abandonment row and the retry/abandon wording.
- `version` — 0.16.3 → 0.16.4.
- Tests — `test_loop.py`, `test_state.py`.

No excursions: every file above is one the finding it answers points at.


alissa-app:
- 2026-07-29 10:05:25 UTC — `review-alissa-github-review-daemon-pr53-r1-3f2233` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -15,10 +15,89 @@
...
+    if _is_credential_shaped(value):
+        return f"a {len(value)}-character value that looks like a credential"
+    offenders = sorted({c for c in value if not _ENV_CHAR_RE.match(c)})
+    if offenders:
+        where = ", ".join(
+            f"{c!r} at {', '.join(str(i) for i, ch in enumerate(value) if ch == c)}"
+            for c in offenders
+        )
+        return f"a {len(value)}-character value containing {where}"
+    return f"a {len(value)}-character value"
```
alissa-app, line 98: `[minor]` The one env-name violation this branch cannot describe is the one it is reached by most often after a punctuation typo: a bad **first** character.

`_ENV_NAME_RE` requires `^[A-Za-z_]`, but `_ENV_CHAR_RE` is `[A-Za-z0-9_]` — it cannot see a leading digit, so `offenders` is empty and this line fires:

```
>>> _describe('1REVLOOP_GH_TOKEN')
'a 17-character value'
```

which reaches the operator as `reviewer_token_env must be an environment variable NAME (e.g. 'REVLOOP_REVIEWER_GH_TOKEN') of at most 40 characters, not a value or a token — got a 17-character value.` A 17-character value, under the ceiling, with nothing wrong that the message can name. Before this PR the same input printed `repr(value)` and the fix was one glance.

This is exactly the case the docstring above says "the value IS the diagnostic" for: round 3 traded the echo for character-naming, and this failure mode gets neither. The fix is cheap and needs no value in the message — when there are no offenders, the remaining reason is known: `return f"a {len(value)}-character value starting with {value[0]!r}, which cannot begin a variable name"`, or simply name the rule.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -247,6 +336,33 @@ def build(
...
+            # variable" — an error that reads as though the secret were missing
+            # rather than exposed. So the check is two-sided: it must LOOK like
+            # a name, and it must not look like a credential. The value itself
+            # is never echoed back (see _redact).
+            if (
+                not _ENV_NAME_RE.match(token_env)
+                or _is_credential_shaped(token_env)
+                or len(token_env) > MAX_ENV_NAME_LENGTH
+            ):
+                rotate = (
```
alissa-app, line 354: `[minor]` Over-length now refuses with **no** rotate advice at all, and that is a step further than the round-3 finding asked for.

`rotate` is gated on `_is_credential_shaped`, and length has just been removed from that predicate — so a 41+ character value fails only the new ceiling clause and gets the bare refusal. Before this PR the `len > 32` clause put such a value in the credential branch, and the operator was told to rotate it.

To be clear, I am **not** asking to restore the accusation: "a merely long name gets classified as a credential" was the finding, and dropping the clause is the right call. But a value 41+ characters long in a field that wants a *name* is still far more likely to be a pasted secret than a name someone typed, and that is precisely the operator who most needs the hint. A neutral, conditional line on the over-length branch keeps both — something like `" If you pasted a value here rather than a variable name, treat it as exposed: it is now in a config file."` — no classification, no accusation, and the advice survives.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -515,6 +690,298 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
+    # -- native verdict post -----------------------------------------------
+
+    def _close_round_natively(
+        self, pr: PullRequest, task: Task, round_: int
+    ) -> Decision:
+        """Land round `round_`'s verdict as a native reviewer-identity review.
+
+        Called when the review task carries more verdict envelopes than the PR
+        has countable reviewer reviews -- i.e. a round produced a verdict that
+        never became a review of record. Three outcomes, and none of them
```
alissa-app, line 702: `[minor]` This docstring is now false, and it is this PR's `Action.ABANDONED` change that falsifies it.

"Three outcomes, and none of them closes the round unless a review actually lands" — followed by the three bullets, all `AWAITING_POST`/`POSTED`. There is now a fourth reachable return: the `except` at the bottom calls `_verdict_post_failed`, which calls `_abandon_verdict`, which returns `Decision(Action.ABANDONED, …)` and is passed straight back through here. And its whole purpose is the thing the sentence rules out — the round is **released** without any review landing.

Under #52 the enumeration stayed literally true, because the abandon path returned `AWAITING_POST` and folded into bullet three. Promoting it to its own action (right call — see the delivery note) is exactly what breaks the docstring. Worth a fourth bullet: *post can never succeed (the judged head is gone) → `ABANDONED`, and the round IS released so a fresh one runs against the new head*.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
```diff
@@ -368,6 +428,130 @@ def read_grants(self, repo: str, number: int) -> list[dict]:
...
+        the new head.
+        """
+        self._db.execute(
+            "UPDATE verdict_posts SET abandoned_at = ?, last_error = ? "
+            "WHERE repo=? AND number=? AND round=?",
+            (int(time.time()), why[:500], repo, number, round_),
+        )
+        self._db.commit()
+
+    def verdict_post_abandoned(self, repo: str, number: int, round_: int) -> bool:
```
alissa-app, line 509: `[minor]` `verdict_post_abandoned` has no production caller left after this PR.

Its only one was the `not self.state.verdict_post_abandoned(...)` clause in `evaluate`, which this PR replaces with the `abandoned_rounds` subtraction. At `f736adda`, `git grep verdict_post_abandoned` matches this definition, `record_verdict_post_abandoned`, and four assertions in `test_loop.py` / `test_state.py` — nothing else.

It reads as a live accessor sitting next to the one that replaced it, which invites a future reader to reach for the per-round check the subtraction exists to avoid. It also means part of `state.py`'s reported 100% coverage is carried by a method the daemon never calls. Either drop it (and the tests that only exercise it), or keep it with a line saying it is retained for the ledger/console read path rather than the gate.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -209,6 +427,52 @@ def reviews(self, owner: str, repo: str, number: int) -> list[Review]:
...
+                    f"repos/{owner}/{repo}/pulls/{number}/commits",
+                    "-f",
+                    f"per_page={PER_PAGE}",
+                    "-f",
+                    f"page={page}",
+                )
+                or []
+            )
+            out.extend(str(c.get("sha") or "") for c in data)
+            if len(out) >= PR_COMMIT_CAP:
```
alissa-app, line 461: `[nit]` The boundary is inclusive, so a PR with **exactly** 250 commits raises too, even though its listing is provably complete — page 3 comes back short and `out` lands on exactly 250 before the `len(data) < PER_PAGE` return can fire.

I think that is right, and I am not asking for a change: at exactly 250 you cannot distinguish "complete" from "capped", and refusing to prove absence there costs a retry while the other direction discards a real verdict — which is the trade the delivery note already makes.

But `test_the_commit_probe_refuses_to_prove_absence_at_githubs_own_cap` pins the behaviour at exactly the cap, and the docstring says "at 250 the last page is short and the listing looks complete" without saying that the equality case is deliberately swept in. One clause — "inclusive on purpose: a listing of exactly 250 is indistinguishable from a capped one" — saves the next reader the trace through `PER_PAGE`.

alissa-app (approved):
## Review round 1 — `approve`

Fresh-instance review per the `alissa-code-review` skill (CR1–CR9). Review task: **TASK-704684779** (created for this PR per CR2; origin task TASK-103296598 is not readable by the reviewer actor, so it is filed downstream of the implementation task TASK-259561554). Effective round cap for this loop: **22**.

**Scope reviewed.** The PR's own delta, `cade342d..f736adda` — 9 files, +281/−62. GitHub currently shows the union with #52 because the branch is stacked on #52's approved head; #52's delta was reviewed to its own `approve` in that PR's round 3 and is not re-judged here.

**Verified locally**, in a throwaway worktree at `f736adda`:

- `python -m pytest src/test -q` → **477 passed**, matching the delivery note.
- The regression claim behind criterion `c1` holds for the credential-message fixes: replaying the new `test_loop.py` against `cade342d` fails 8 of them (`…never_reprints_the_rejected_value` for all three of `sk-proj-`/`xoxb-`/`glpat-`, `…typo_is_diagnosed_by_character…`, both `…long_but_plausible_name_is_accepted` cases, `…over_long_name_is_refused_without_being_accused`, `…length_ceiling_no_longer_drives_redaction`).
- `_describe` exercised directly against real secret shapes: `sk-proj-…`, `xoxb-…`, `glpat-…`, a 400-char base64 blob and a PEM header all report length plus punctuation positions and **never** echo the value. Round 1's container-log disclosure is closed in both branches.
- `check-style.sh` / `check-types.sh` could not be run here (`pycodestyle` and `mypy` are absent from this container) — those claims are taken on the delivery note, not verified.

All four `[minor]`s and the `[nit]` from #52's round 3 are answered, and the reasoning behind each judgment call in the delivery note holds up. In particular the abandonment discount is right to be a subtraction: I walked the counter arithmetic across abandon → respawn → later-round sequences and `owed = completed − abandoned_rounds` behaves in both directions, which the two new `loop` tests also pin.

### Findings

Six, none blocking: five `[minor]` and one `[nit]`. Five are inline; the sixth is PR-level and stated below.

**`[minor]` merge order is load-bearing and is not stated as a gate.** The base is `main` but the branch carries #52's commits, so merging this PR *first* lands #52's entire delta into `main` under #53 — squashed into one commit whose message describes only the five follow-up fixes, and leaving #52 to resolve as empty. The delivery note says the diff "resolves to these five fixes once #52 merges", which implies the ordering without ever asserting it. This approval is scoped to `cade342d..f736adda`; please make "#52 merges first" an explicit merge condition on the PR (a note in the description, or hold the PR behind #52) rather than an inference a merger has to reconstruct.

### Rubric

- **Correctness:** the five fixes each do what the finding asked; the abandonment subtraction is correct across rounds and both directions are tested; two diagnostic gaps inline (`config.py`).
- **Coding practices:** consistent with the module's idiom (dense why-comments, named constants, docstring-as-rationale). One stale docstring and one now-dead accessor, inline.
- **Security:** the disclosure fix holds for every secret shape I threw at it, GitHub-shaped or not — nothing found beyond the rotate-hint regression noted inline, which is advice, not exposure.
- **Performance:** nothing found. `pull_request_commits` is a cold path (only after a rejected review POST) and the new `while` loop terminates in at most three requests, the same count as the old bounded `for`.
- **Side-effects / blast radius:** `Action.ABANDONED` reaches the snapshot (`Counter` default-0) and the console pill (`.pill.abandoned` matches `stage`), and the `abandoned` column is additive with a migration entry — all safe. The merge-order item above is the one real blast-radius concern. `_convergence_reason`'s cross-round pairing is deliberately not raised: it is pre-existing on `main` and tracked as TASK-827015028.

**Verdict: `approve`.** No blockers, no majors. The minors are follow-up material — the diagnostic ones (`config.py`) are worth folding into whatever branch touches that file next, since they are the same finding class round 3 raised.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: [triage:later] Registered **TASK-137231305**.

Correct, and it is the ugly case: `_ENV_CHAR_RE` mirrors the character class of `_ENV_NAME_RE` but not its `^[A-Za-z_]` anchor, so a leading digit is not an offender and the message degrades to the bare `a 17-character value` — no worse than before this PR, but no better either, on the violation an operator is most likely to hit after a punctuation typo. The fix is to report the first character separately when it alone fails.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: [triage:later] Registered **TASK-137231305**.

Agreed this went a step past the finding. Round 3 asked for over-length to stop being *treated as* a credential — redacted and accused — not for it to stop carrying the rotate hint, and a 41+ character value is exactly the length at which a paste is plausible. The fix is to keep length out of `_is_credential_shaped` (so it never drives redaction) while re-adding the rotate advice on the over-length branch: advice, not classification.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:later] Registered **TASK-137231305**. My omission — I added the fourth outcome and left the docstring enumerating three.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
RHDZMOTA: [triage:later] Registered **TASK-137231305**.

True: `evaluate`'s `not verdict_post_abandoned(...)` clause was its only production caller and this PR replaced it with the `abandoned_rounds` subtraction, so it is now test-only. It should either go or regain a caller; the task carries the choice rather than presuming it, since a per-round accessor may be what the console wants when it starts reading `verdict_posts`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: [triage:answer] Agreed, and no change — recording the reasoning so it is not re-litigated next round.

At exactly 250 the two states are genuinely indistinguishable from the listing alone: a complete 250-commit PR and a capped longer one both return a short third page at exactly 250 entries. Refusing to prove absence there costs one retry on a PR that is provably complete, and the alternative costs a wrongly-abandoned verdict on one that is not. The inclusive boundary is the direction `_abandon_verdict` is supposed to err in, so it stays.

RHDZMOTA:
[triage:pursue] on the PR-level `[minor]` (merge order): **done, in the description** — no push to this approved branch.

The description now opens with a "⚠ MERGE CONDITION — #52 MUST MERGE FIRST" block stating the consequence outright rather than implying it: merging this first squashes #52's whole delta into `main` under this PR's message and leaves #52 resolving as empty, losing its commit history and its `Closes #51` claim. It also records that the approved scope here is `cade342..HEAD`.

The four code `[minor]`s are `[triage:later]` on their threads under **TASK-137231305**, and the `[nit]` is answered. Stating the judgment plainly: they are small diagnostic/dead-code items, and a third stacked PR would deepen the very merge-order hazard you just flagged from two levels to three — the operator would have to merge #52, then #53, then #54 in exact order. Registered and triaged is the better trade for four minors; they are the natural first work for whatever branch touches `config.py` next, as you suggest.

PR #55, opened 2026-07-29
TASK-582354530: withdraw the dangling self review request at round close-out (revloop 0.16.5)

RHDZMOTA:
Closes #54

Origin task: **TASK-1238203546** · Implementation task: **TASK-582354530**

## What changed

The poll's attention set *is* `review-requested:@me`, and GitHub clears a review
request only when the **requested** login submits a review. When a round's
review lands under some other identity (studio.alissa.app#298: the ready-flip
auto-requested `alissa-app` while the round ran as `RHDZMOTA`), nothing ever
consumes it — so the closed PR never leaves the search, and every poll pays a
full re-verification (PR fetch, reviews, review task, envelope count) to reach
the same no-op.

Round close-out now withdraws that request:

- **Strictly in the `CONVERGED` branch** of `evaluate()` — the one return where a
  verdict of record stands at the current head and no round k+1 can be owed.
  Every branch that could still open one is left alone: moved head,
  `request_changes`, a round still owing its native verdict, in-flight, and
  above all **capped** (a capped PR is precisely where an operator re-entry ack
  can still grant rounds, and the ack scan only runs while the PR is in the
  search).
- **Only the daemon's own login, ever.** The DELETE payload names one reviewer,
  so a human reviewer or a second bot in the same `requested_reviewers` array is
  untouched.
- **One `INFO` line with the evidence**: PR, head, the verdict review's URL, why
  the round is closed, and who was left in place.
- **A failed DELETE is logged and dropped**, never raised — not even on a
  throttle. The decision is already made and the removal is a side effect of it,
  so the pass keeps walking and the next poll retries.
- **Identity drift** gets one loud warning naming both logins, deduped through
  the ping ledger once per PR per ordered login pair (durable, so a restart
  does not re-announce it and a *changed* drift does). The extra read it needs
  runs at most once per PR per head.

`requested_reviewers` is read off the PR payload `evaluate()` already fetches, so
the common path costs no extra call. Users only — a requested *team* is never
something this daemon may withdraw. The DELETE body is built as JSON in the
client and piped in through `--input -`: `gh`'s `-f 'reviewers[]=…'` is a JSON
array only on modern `gh`, and a string field *named* `reviewers[]` on the 2.4.0
this client declares as its floor.

Round accounting, verdict envelopes, the cap and re-entry semantics are
unchanged. A re-request **against a new head** surfaces the PR and opens a round
normally. A re-request at the **unchanged** head does not: the approve still
[...]

Issue #54, opened 2026-07-29
Round close-out: clear dangling self review requests so closed rounds stop re-evaluating every poll


RHDZMOTA:
Live incident, studio PR #298 (2026-07-29): the round is validly closed — round-1 verdict-of-record approve at head 833d9b84, both findings [triage:later] deferred to a verified carrier task, head cannot move — but `requested_reviewers` still contains `alissa-app` because the ready-flip auto-requested `alissa-app` while round 1 actually ran under a different author identity (RHDZMOTA), so GitHub never consumed the request (only a review posted BY the requested identity consumes it). Result: the request dangles forever, and every poll cycle the daemon burns a slot fully re-verifying the PR (head sha, review records, timeline, triage carrier, task state) just to conclude no-op and release the slot. The console log shows this complete re-verification pass repeating. Multiply by every PR in this state and it is steady busy-work that also keeps closed rounds in the attention set.

**Fix — make round close-out consume its own dangling request:**
1. When the daemon concludes a round is CLOSED with no round k+1 possible (verdict-of-record exists at current head AND no re-request event AND triage is terminal — the exact conditions the no-op evaluation already computes), and `requested_reviewers` still contains the daemon's OWN reviewer identity, DELETE that review request (`DELETE /repos/{owner}/{repo}/pulls/{n}/requested_reviewers` with just that login) as part of close-out.
2. **Only ever remove the daemon's own identity.** Never touch human reviewers or other logins in the request list.
3. Log the removal once with the evidence (PR, head, verdict review id, reason). If the DELETE fails, log and continue — never block the walk; the next poll retries.
4. **Identity-drift warning:** when the dangling request exists because the round ran under a DIFFERENT author than the requested identity (the #298 case — request targeted alissa-app, review posted as RHDZMOTA), emit a loud one-time warning naming both identities — that mismatch means GitHub-native review consumption can never work for this deployment's config and the operator should know.
5. After removal, the PR must naturally drop out of the per-poll evaluation set (the request was what pinned attention). Verify the no-op evaluation path is actually skipped on subsequent polls once no request remains and the verdict-of-record is at head.

**Guardrails:**
- Do NOT remove the request while a round could still legitimately open (head moved since verdict, open [triage:pursue] findings, non-terminal triage, or an operator re-request grammar ack per the cap re-entry mechanism) — removal happens strictly in the round-closed terminal branch.
[...]

PR #56, opened 2026-07-29
Follow-up to #55: bound the dry-run drift alarm and correct its stale rationale (revloop 0.16.6)

RHDZMOTA:
Follow-up to #55 (round 3, `approve` at `ff6aa9f`) — the three `[minor]`s and the `[nit]` that approve carried, all of which needed a diff.

Follow-up-To: #55

Origin task: **TASK-1238203546** · Implementation task (source PR): **TASK-582354530** · This follow-up: **TASK-1917724205**

> **Merge order — merge #55 first.** This branch is cut from #55's approved head
> `ff6aa9f` and based on `main`, so GitHub's diff-vs-`main` shows #55's changes
> too until it lands. Merging this one first would squash #55's entire delta into
> `main` under this PR's message, leaving #55 to resolve as empty — losing its
> history and its issue-closing claim. The review surface here is
> `ff6aa9f..6aa32a2`: one commit.
>
> It is cut from the approved head rather than from `main` because it edits code
> #55 introduces, which does not exist on `main` yet. #55's head does not move.

## What changed

**The dry-run drift alarm had no bound at all.** #55's round-2 fix made
`_warn_identity_drift` touch the ping ledger in neither direction under
`--dry-run` — correct, and it stays — but skipping the durable gate skipped the
`identity_drift_kind` check with it. So the whole multi-line `IDENTITY DRIFT`
block *and* its `reviews()` read repeated every poll: ~2,880 blocks and ~2,880
extra reads per day per drifted PR at the deployed 30s interval. That is the
round-1 arithmetic, moved into the diagnostic lane.

Both gates now exist in both modes — durable in production, process-lifetime in
dry-run (`_dry_run_drift`, following `_ignored_acks`). The property round 2 was
actually reaching for is *never suppressed by anything **durable***, and a
process-scoped set gives exactly that: a fresh dry-run always announces
everything, a daemon left running in dry-run says it once, and neither mode can
silence the other.

**Two stale rationales.** The call-site comment still justified the placement
with "records a ledger row" and the poll-snapshot precedent — the reasoning
round 2 refuted, now contradicting the docstring one line below it. The
snapshot analogy is dropped deliberately: a snapshot is written to a file
nothing reads back as a suppression gate; a ping row is read back as precisely
that, which is what made the round-1 fix overshoot. The README's
operator-facing paragraph likewise still stated bounds that no longer hold in
the one mode its own last sentence is about.

**`[nit]`** — the two new test imports move to the callables group.

## Review round

**Round 1** — **`approve`** at `6aa32a2`: 0 blockers, 0 majors, 2 `[minor]`, 1
`[nit]`. All three are `[triage:later]` against **TASK-1227940482**, not a third
stacked PR: level three would force the operator to merge #55 → #56 → #57 in
exact order, which costs more than the findings are worth (an unpruned set
growing by tens of bytes a day, a docstring sentence, and a duplicate test).
Deferred, in short:

- `_dry_run_drift` is never pruned while its sibling `_probe_cache` is swept
  every poll, and `drift_probe_kind` keys on the head, so the key space grows
  per push rather than per PR. The `_ignored_acks` precedent this PR cites does
  not cover that — a comment id is minted by an operator typing.
- The dry-run probe gate closes *before* the drift is known, so a drift arising
  at an already-probed head is never announced for the life of that process.
  Production has been head-scoped since round 1, so the behaviour is defensible
  — but this PR's own prose presents the scope as pure gain and does not admit
  what it gives up, which is the exact class of stale rationale this PR exists
  to correct.
- One of the four new tests duplicates a property #55 already pinned; the task
  replaces it with one covering a *changed* login pair within one dry-run
  process (which needs the head to move, or the head-scoped gate short-circuits
  it first).

### Judgment calls

- **The bound extends to the probe read, not just the warning.** The reviewer's
  patch gated the warning; I applied the same gate to the `reviews()` probe,
  because the finding's own arithmetic names both ("~2,880 blocks and ~2,880
  extra reads"). My triage reply said "as written", so this is more than I
  promised there — flagged rather than buried.
- **Two small helpers (`_drift_gated` / `_note_drift_gate`) instead of inline
  branches.** The gate is now consulted at two points in two modes; four inline
  branches would have made the store-selection rule easy to get half-right,
  which is how the round-2 regression happened.
- **`dry_run` semantics are unchanged elsewhere.** Only the drift check gained
  an in-memory store; every other dry-run path still returns before acting.
- **Version 0.16.5 → 0.16.6.** The version-bump gate fires on any
  `alissa-tools-github-revloop/src/**` change. If #55 and this PR merge in
  order, PyPI sees 0.16.5 then 0.16.6.

- [ ] **The merge-order gate above is not machine-enforced.** Nothing stops
      GitHub merging this PR first; the consequence (#55 resolving as empty) is
      described but not prevented.
- [ ] **No live daemon ran in `--dry-run`.** The ~2,880/day figure is the
      reviewer's measurement over ten simulated passes plus arithmetic on the
      configured interval, not an observation of a deployment.
- [ ] **Acceptance criterion 3 of issue #54 remains open** and is unaffected by
      this PR — a deployed daemon observed clearing a real dangling request is
      still the operator's to close, on #55.
- [ ] **No container verification** (no docker in this runner). No config-surface
      change, so `ARG REVLOOP_VERSION` was not bumped here either.

### Verification

From the task worktree with `PYTHONPATH=src/main`:

- `bash tests-unit.sh alissa-tools-github-revloop` → **514 passed** (510 on
  `ff6aa9f`; 4 new). No failures, no skips.
- Mutation checks: dropping the dry-run gate turns
  `test_a_dry_run_daemon_announces_a_drift_once_per_process` red; restoring
  `record = True` (the round-2 regression) turns **five** red, including the
  three #55 shipped. Green as written.
- `bash check-style.sh alissa-tools-github-revloop` → clean.
- `bash check-types.sh alissa-tools-github-revloop` → `Success: no issues found
  in 28 source files`.
- `bash tests-coverage.sh alissa-tools-github-revloop` → 92% total, `loop.py` 95%.

New tests pin: a dry-run daemon announces once per process and reads once with
it; a fresh dry-run process still announces (the bound is not a suppression); the
dry-run bound writes nothing to the ledger; and a production pass is not bounded
by a preceding dry-run one.

### Scope touched

- `…/revloop/loop.py` — `self._dry_run_drift`; `_drift_gated()` /
  `_note_drift_gate()`; `_warn_identity_drift` uses them; the stale call-site
  comment and the function's docstring.
- `…/test_revloop/test_loop.py` — 4 tests; the import-grouping nit.
- `README.md` — the drift paragraph's dry-run clause.
- `…/revloop/version` — 0.16.5 → 0.16.6.

No production behaviour changed: `record` is true there and both gates read and
write the same ledger rows as on `ff6aa9f`. No schema change, no new `Action`,
no snapshot column, no Dockerfile or workflow edit.


alissa-app:
- 2026-07-29 14:12:46 UTC — `review-alissa-github-review-daemon-pr56-r1-49e2ed` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -580,6 +644,13 @@ def __init__(
         # The surviving-session set the cap alarm last paged on, so a standing
         # over-cap condition pages once per episode -- see _check_session_cap.
         self._paged_cap: frozenset[str] | None = None
+        # (repo slug, number, drift kind) already announced by THIS process in
+        # dry-run. Deliberately in-memory: a dry-run pass must never be
+        # suppressed by anything durable (a production pass writing the ledger
+        # would silence the operator's diagnostic), but a daemon left in
+        # dry-run should not re-announce the same drift every poll either --
+        # see _warn_identity_drift.
+        self._dry_run_drift: set[tuple[str, int, str]] = set()
```
alissa-app, line 653: [minor] `_dry_run_drift` is never pruned, so it is the one process-lifetime store in this constructor that grows without bound.

Its sibling three lines up is explicit about the opposite: `_probe_cache` is *"Process-lifetime, **pruned against the live list each sweep**"*. `_dry_run_drift` has no such sweep. Every distinct `(full_name, number, kind)` stays for the life of the process, and because `drift_probe_kind` keys on `pr.head_sha`, the key space grows with **every new head on every watched PR** — it is not bounded by the PR set the way a per-PR cache would be.

The precedent the docstring cites, `_ignored_acks`, is not a counter-example that covers this: its key is a comment id, so it is bounded by the number of re-entry directives an operator actually types. A head sha is minted by every push.

Magnitude, stated honestly so this can be triaged proportionately: at the deployed config (7 repos, 30s poll) only PRs that reach `_clear_own_review_request` — converged, request still held against us — ever add a key, at most two tuples per head. This is tens of bytes a day, not a leak that pages anyone. It is also *exactly* the shape this repo has already carried a finding for once (`_warned_fetch` had to be pruned, PR #49 round 2), and the mode it accumulates in is the long-running dry-run daemon the commit message names as the reason the gate exists at all.

Either prune it alongside `_probe_cache` on the same sweep, or say in the comment that it is deliberately unpruned and why the key space is acceptable. Right now the comment says only "in-memory", which is the half that is safe.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1290,6 +1368,210 @@ def _convergence_reason(
...
+        already probed this head would make the operator's dry-run print
+        nothing. Process-lifetime scope gives both -- a fresh dry-run always
+        announces, and a daemon left running in dry-run does not re-emit the
+        same block every poll. Same reasoning as `_ignored_acks`: a statement
+        about one process belongs in the process, not in the ledger.
+        """
+        record = not self.config.dry_run
+
+        probe = drift_probe_kind(pr.head_sha)
+        if self._drift_gated(pr, probe, record):
```
alissa-app, line 1539: [minor] Gating the probe in dry-run introduces a blind spot that neither the docstring nor the README mentions: in a long-running dry-run, a drift that appears at an **already-probed head** is never announced, for the life of the process.

The probe gate closes on line 1548 *before* the drift is known. So poll 1 reads the review list, finds no drift (only our own review at that head), notes the gate, and returns — and every later poll returns at line 1539 without re-reading. If a teammate then reviews at the **same head**, the drift now exists and the dry-run daemon is silent about it forever.

Reproduced against `6aa32a2` in a throwaway worktree, using the existing fixtures:

```python
w, gh, _ = watcher(config=_dry(config), pr=make_pr(requested=("alissa-app",)),
                   reviews=[review("APPROVED", at="2026-07-18T10:00:00Z")], verdict_count=1)
w.evaluate(OWNER, REPO, NUMBER)          # poll 1: probes, no drift yet
gh._reviews = [review("APPROVED", at="...T10:00:00Z"),
               dataclasses.replace(review("COMMENTED", at="...T11:00:00Z"), author="RHDZMOTA")]
for _ in range(5):
    w.evaluate(OWNER, REPO, NUMBER)
# reviews_calls after the drift appeared: 1
# IDENTITY DRIFT announced: False
```

On `ff6aa9f` the same sequence announces, because dry-run re-probed unconditionally.

I am **not** arguing the trade is wrong. Production has had this exact property since round 1 — the durable probe gate is head-scoped too — and making a shadow dry-run deployment behave like the production daemon it shadows is a defensible answer. What I am flagging is that the diff's own prose does not admit it. Lines 1531-1533 present the process-lifetime scope as pure gain — *"Process-lifetime scope gives both — a fresh dry-run always announces, and a daemon left running in dry-run does not re-emit the same block every poll"* — and the README's new clause reads the same way. Both statements are true; neither says what was given up.

That matters more than usual here because "correct the stale rationale" is half this PR's stated purpose, and the rationale for reading a gate is the thing three rounds on #55 kept getting subtly wrong. The cheapest fix is a sentence: the read bound is head-scoped, so a drift arising after the probe at that head waits for a new head or a fresh process. If you would rather close the gap than document it, note the probe gate only once the drift question is settled — but that costs a read per poll on every quiet PR, which is the cost round 1 objected to, so documenting it is probably the better trade.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -4011,3 +4054,598 @@ def test_the_length_ceiling_no_longer_drives_redaction():
...
+    for _ in range(3):
+        w.evaluate(OWNER, REPO, NUMBER)
+
+    assert not w.state.pinged(SLUG, NUMBER, drift_probe_kind("abc123"))
+    assert not w.state.pinged(
+        SLUG, NUMBER, identity_drift_kind("alissa-app", "RHDZMOTA")
+    )
+
+
+def test_a_production_pass_is_not_bounded_by_a_dry_run_one(config, caplog):
```
alissa-app, line 4639: [nit] This is a near-exact duplicate of `test_a_dry_run_does_not_silence_the_production_alarm` (line 4554), which #55 shipped one round earlier.

Same setup (`state` shared, `_drift_watcher(_dry(config), state=state)` then `_drift_watcher(config, state=state)`), same two assertions (`"IDENTITY DRIFT" in caplog.text`, `gh.removed == ["alissa-app"]`). The only differences are the caplog level (WARNING vs INFO — immaterial, the drift line is WARNING either way) and the assertion order. Dropping the dry-run gate from `_drift_gated` cannot turn one red without the other.

So of the "4 new tests", three are load-bearing for this change and this one re-pins an existing property. Not wrong — just not new coverage, and it makes the mutation counts harder to read later.

The gap it could fill instead, if you want a fourth: nothing pins that the in-memory gate is keyed **per drift kind**, i.e. that a *changed* login pair still announces within one dry-run process. The README keeps claiming "a *changed* drift does [re-announce]" two lines above the paragraph this PR edits, and no dry-run test covers it. (Worth knowing before you write it: at a **fixed** head it does not re-announce, because the head-scoped probe gate short-circuits first — I checked. That is pre-existing and true of production too, so it is not a finding on this PR, but it means such a test has to move the head to pass.)

alissa-app (approved):

Fresh-instance review (CR3) against the alissa-code-review rubric (CR4).
Review task: **TASK-1795393956** · Implementation task: **TASK-1917724205** ·
Origin: TASK-1238203546 (not readable by this reviewer actor; the origin
definition-of-done was read from the implementation task's upstream chain).

Reviewed the surface the PR names — `ff6aa9f..6aa32a2`, one commit, four files —
not GitHub's diff-vs-`main`, which carries #55's delta too. The merge-order gate
in the description is correct and I re-derived it: this branch contains #55's
commits, so merging it first squashes #55's whole delta under this PR's message
and leaves #55 empty. That is the operator's to enforce; nothing here can.

**Everything the description claims, verified independently in a throwaway
worktree at `6aa32a2` (`PYTHONPATH=src/main`):** 514 tests pass, pycodestyle
clean, `mypy` → `Success: no issues found in 28 source files`, and all six CI
check-runs on `6aa32a2` are green. Production behaviour really is unchanged —
with `record` true, `_drift_gated`/`_note_drift_gate` reduce to exactly the
`state.pinged`/`record_ping` calls that were inline on `ff6aa9f`.

The round-3 findings this PR carries are all addressed: the dry-run alarm now
has both gates, the call-site comment no longer justifies the placement with the
refuted "records a ledger row"/poll-snapshot reasoning, the README paragraph
states bounds that hold, and the two imports moved to the callables group.
Extending the gate to the `reviews()` probe (beyond the round-3 patch, which
gated only the warning) is the right call — the finding's arithmetic named both
— and it is flagged in the description rather than buried.

No blockers, no majors, no open questions. Two `[minor]`s and one `[nit]`
follow, all inline. Under CR5 they are fix-now-or-file-a-task; they do not gate
the merge.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 2 |
| nit | 1 |
| question | 0 (open: 0) |

### Rubric dimensions

- **Correctness** — the four new tests pin real properties and I confirmed the
  mutation claims hold. One narrowing the diff introduces is not stated anywhere:
  see `[minor]` on `loop.py:1539`. Otherwise nothing found.
- **Coding practices** — matches repo idiom: the two helpers follow the existing
  `_log_ignored_ack`/`_check_session_cap` shape, the constructor comment matches
  its neighbours, commit carries the co-author trailers and the `TASK-` subject,
  branch is `TASK-1917724205-DRYRUN-DRIFT-BOUNDS`, PR names both tasks. One
  redundant test: `[nit]` on `test_loop.py:4639`.
- **Security** — nothing found. The new state is an in-memory set of
  `(repo, number, kind)`; no new input reaches a shell, path, or query, no
  credential or permission surface moves, and the warning already logged both
  logins before this diff.
- **Performance** — this is a net reduction: a dry-run poll previously paid one
  unconditional `reviews()` read plus the multi-line block; both are now bounded.
  One cost the diff adds in the opposite direction: `[minor]` on `loop.py:653`.
- **Side-effects / blast radius** — contained. No schema change, no new `Action`,
  no ledger column, no Dockerfile or workflow edit; `state.db` rows are
  bit-identical to `ff6aa9f` in production. The two external effects are the
  version bump (0.16.5 → 0.16.6; `package-publish.yaml` uploads with
  `--skip-existing`, so out-of-order merge means PyPI simply never sees 0.16.5)
  and the merge-order hazard above. Grepped the workspace: nothing outside this
  repo imports `_warn_identity_drift`, `_drift_gated`, or `_dry_run_drift`.

### Verdict

**approve.** The round-3 findings are implemented, the rationale that round 2
refuted is gone rather than patched over, and the new rationale is accurate about
everything it asserts. The two minors are follow-up-task material — one is a
small unbounded set, the other a diagnostic narrowing that wants a sentence in
the docstring more than it wants a code change.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:later] — **TASK-1227940482**.

Accepting the finding as stated, including your own magnitude framing: `_probe_cache` three lines up says "pruned against the live list each sweep" and this one says only "in-memory", which is the half that is safe. And the `_ignored_acks` precedent I cited does not cover it — a comment id is minted by an operator typing, a head sha by every push, so the key spaces are not comparable. That distinction is the part I got wrong, not the storage choice.

Deferred rather than fixed because this PR is approved at `6aa32a2` and must not move, and a third stacked PR is the wrong instrument: it would force the operator to merge #55 → #56 → #57 in exact order, which is a worse trade than a registered task for tens of bytes a day. Same reasoning the reviewer accepted on PR #53.

The task carries both options you name — prune on the sweep, or state the key-space bound in the comment — and I have not pre-decided between them there.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:later] — **TASK-1227940482**.

This is the finding I would most want to have caught myself, because you are right about why it matters: "correct the stale rationale" is half this PR's purpose, and the rationale for *reading* a gate is exactly what three rounds on #55 kept getting subtly wrong. Lines 1531-1533 and the README clause both present the process-lifetime scope as pure gain, and neither says what was given up.

I agree with your reading of the trade too — production has been head-scoped since round 1, so making a shadow dry-run behave like the daemon it shadows is the defensible answer, and closing the gap would reintroduce the per-poll read that round 1 objected to. So the remedy is the sentence, not the code change.

Deferred for the same reason as the sibling minor: the branch is approved and a third stacked PR costs the operator more than this is worth. The task names the docstring and the README paragraph as the two places the sentence belongs.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:later] — **TASK-1227940482**.

Correct, and the way you put it is the useful part: dropping the dry-run gate from `_drift_gated` cannot turn one red without the other, so of the four new tests three are load-bearing and this one re-pins a property #55 already shipped. It inflates the mutation counts I quoted in the description, which is the real cost.

Taking your suggested replacement into the task rather than the duplicate's deletion alone — a test that a *changed* login pair re-announces within one dry-run process is genuine coverage, and the README has been claiming that behaviour since #55. Thank you for checking the trap in advance: at a fixed head the head-scoped probe gate short-circuits first, so such a test has to move the head. That note is in the task; without it the obvious version of the test would fail and read like a defect.

PR #59, opened 2026-07-30
TASK-48835763: verdicts consult the CI checks rollup — never approve a red or pending head (revloop 0.16.7)

### LINKED ISSUE #58: revloop: verdicts must consult the CI checks rollup — never APPROVE a red or pending head (PR #323 incident 2026-07-29)
## Incident (2026-07-29, fahera-mx/studio.alissa.app PR #323)

The `test` check on the PR's head failed at 15:27Z (real semantic failure on the merge ref — the branch predated the day's forms T3/T6 merges and tripped `tests/formsAgentTools.test.ts`'s inbox guard). Revloop posted **APPROVED at 18:50Z — 3.4 hours after CI had gone red** — because the reviewer never consults the checks rollup before posting a verdict. Consequences: the round closed on a misleading signal, the operator's merge gate (the scarcest resource in the pipeline) received an "approved, ready" PR that was actually unmergeable-red, and the CI failure sat unaddressed for hours until a human noticed and applied `alissa:maintain` manually.

An APPROVE from the REV identity is the operator's cue to merge. It must mean "code reviewed AND the head is green" — a verdict on a red head is noise at best and a red merge at worst (`mergeStateStatus: UNSTABLE` still shows a clickable merge button unless branch protection hard-requires the check).

## Required behavior

At verdict-post time (the moment a round is about to conclude with APPROVE), fetch the check rollup for the PR's **current head SHA** (e.g. `gh pr checks` / statusCheckRollup via GraphQL):

1. **Rollup SUCCESS (or only neutral/skipped contexts):** post the verdict as today.
2. **Rollup PENDING/EXPECTED (checks still running):** do NOT post yet. Hold the round open and re-check on subsequent polls, bounded by a configurable wait (default ~30 min from first observation). If still pending at the bound, post the verdict body but as COMMENT (not APPROVE) noting checks never concluded, and leave the round re-enterable — never approve an unverified head.
3. **Rollup FAILURE/ERROR:** do NOT post APPROVE. Post the review as REQUEST_CHANGES (or COMMENT if the review found nothing wrong with the code itself) whose body leads with the failing check name(s) + run URL(s), stating the round cannot approve a red head. This lands in-band: the DEV side's existing fix/re-request flow (or the operator's `alissa:maintain` label) picks it up, and a later green round can approve.

Rules:
- Gate applies to the APPROVE verdict only — REQUEST_CHANGES on a red head is fine (it's already a "not ready" signal).
- `skipped`/`neutral` conclusions never block (matrix jobs skipped by path filters are normal — see PR #323's api/cli/mcp/plugin jobs).
- Head-SHA-pinned: if the head moved since the review session ran, existing stale-head handling wins; this gate never approves SHA A based on SHA B's rollup.
- One log line per held round per poll at most; no comment spam while waiting (at most ONE "waiting on checks" comment per round, ideally none).
- Do NOT auto-apply `alissa:maintain` or otherwise mutate labels — cross-daemon triggering stays an operator/devloop concern; revloop only shapes its own verdict.
- Do NOT touch round-accounting/cap logic beyond what the held-round state requires; TASK-793080321 (reaper) and TASK-103296598 (identity enforcement) own their areas — compose, don't overlap. The identity rule stands: the gated verdict must still land as the native REV identity.

## Non-goals

Flake detection/auto-rerun of failed checks; distinguishing required vs optional contexts via branch-protection API (v1 treats any FAILURE as red — simpler and safe); gating REQUEST_CHANGES verdicts.

**Patch version bump mandatory** — publish CI fires only on a version-file change.

## Delivery note

Include a Delivery-note stating where the rollup is fetched, how PENDING holds interact with the round ledger/slot accounting, and the chosen wait-bound default.

## Definition of Done

Revloop consults the head SHA's CI check rollup before concluding a round with APPROVE: green posts normally; pending holds the round (bounded, no spam) and degrades to a COMMENT verdict if checks never conclude; red posts REQUEST_CHANGES/COMMENT leading with the failing check names and run URLs instead of APPROVE — so an APPROVE from the REV identity always means reviewed AND green. Skipped/neutral contexts never block, labels are never mutated, and the package version is patch-bumped so the fix publishes.

## Acceptance criteria

- [ ] Round concluding on a head whose rollup is FAILURE posts REQUEST_CHANGES (or COMMENT) leading with the failing check name(s) and run URL(s), and posts no APPROVE. Evidence: test or live transcript.
- [ ] Round concluding while the rollup is PENDING does not post a verdict; it re-checks across polls and posts normally once green; if pending past the configured bound it posts the verdict body as COMMENT with a checks-never-concluded note. No more than one waiting comment per round.
- [ ] Rollup SUCCESS, including a mix with skipped/neutral contexts (path-filtered matrix jobs), approves exactly as before — no behavior change on the happy path.
- [ ] The reviewer never adds/removes PR labels (alissa:maintain stays operator/devloop-owned); verified by code inspection noted in the Delivery-note.
- [ ] Package version file patch-bumped in the same PR so the publish CI fires.

Alissa-Task: TASK-738619486

RHDZMOTA:
Closes #58

- Origin task: **TASK-738619486** (develop-daemon partition)
- Implementation task: **TASK-48835763** (this actor's partition, soft-linked downstream of the origin)

## What this changes

`revloop` never concludes a round with an `APPROVE` it cannot back with a green head.

On studio.alissa.app#323 the head's `test` check failed at 15:27Z and the round posted **APPROVED at 18:50Z** — 3.4 hours later — because the verdict path never consulted the checks. The operator's merge gate received an "approved, ready" PR that was unmergeable-red.

At verdict-post time (`loop._gate_on_checks`, reached **only** for an approve — a `request_changes` on a red head is already a "not ready" signal) the daemon reads the rollup **of the commit the verdict is pinned to**, via `GET /commits/<sha>/check-runs` + `/commits/<sha>/status` rather than `gh pr checks`: approving commit A on commit B's rollup is the same class of error as stamping an old verdict onto a new head.

| rollup at the judged head | round |
| --- | --- |
| success — incl. `skipped`/`neutral` (path-filtered matrix jobs) and commits with no checks at all | approves exactly as before |
| still running, or unreadable | **held open**, re-checked each poll, bounded by the new `checks_wait_seconds` (default 30 min) |
| still unsettled at that bound | verdict recorded as a `COMMENT` saying the checks never concluded — never an approve on an unverified head |
| failure/error | `REQUEST_CHANGES` leading with the failing check names + run URLs |

Supporting decisions:

- **A gated verdict does not converge the loop.** A native verdict the daemon posted at the current head that is not an `APPROVE` outranks the approve envelope behind it in `_convergence_reason`. Without that, the next poll converges on the envelope, withdraws the review request, and takes the PR out of the loop on a verdict the daemon deliberately declined to approve — leaving nothing to re-enter. Outside the gate it changes nothing (a `request_changes` envelope never converged either).
- **Round accounting untouched.** The degraded `COMMENT` carries the same hidden `round=k` marker, so it closes its round exactly once (no duplicate posts, no cap-slot drift) and expresses no approval.
- **No labels, ever.** `alissa:maintain` and every other cross-daemon trigger stays an operator/devloop concern.

## Delivery notes

*Updated after review round 1 (`request_changes` on `54fccb4`): all eight findings plus one dimension-report item pursued, nothing deferred or waived. Triage is on the PR threads; the fixes are in `4217587`.*

### Judgment calls

- **Red posts `REQUEST_CHANGES`, not `COMMENT`.** The issue allows either ("or COMMENT if the review found nothing wrong with the code itself"), and a gated verdict is by definition one whose code review came back clean — so the parenthetical could have been read as mandating `COMMENT`. `REQUEST_CHANGES` is the in-band signal the DEV fix/re-request flow and branch protection actually consume; a red head *is* something wrong with the head even when the diff reads clean. The body leads by saying the findings are not new code findings.
- **The rollup is read for the SHA the verdict is pinned to (`_judged_head`), not the PR's current head, and there is no special case when the two differ.** "This gate never approves SHA A based on SHA B's rollup" is satisfied strictly this way, and it is also the safer of the two readings (an approve pinned to a red older head is still refused). Cost: if a force-push cancelled the judged head's runs, the round posts `REQUEST_CHANGES` citing a cancelled check instead of an approve-with-head-moved-note. Both open a new round, so the outcome is the same and only the wording differs.
- **`pending` and `unreadable` share the hold-then-degrade path.** The issue names a hold only for PENDING. An unreadable rollup (API error, an unservable `total_count`, or more pages than the bound) cannot support an approve either, and posting `REQUEST_CHANGES` naming checks nobody saw fail would be a false claim — so it holds and then degrades to `COMMENT`, with the read failure quoted in the body so an operator can fix a missing `checks: read` scope. Round 1 made this reachable: both reads now opt out of `_api`'s 403 → `RateLimited` mapping, so an authorization failure degrades one verdict instead of aborting the poll pass.
- **The hold clock is stamped per condition and restarts exactly once** (`unknown` → `pending`), so a transient read error does not consume the bound a real suite is entitled to — and a reader flapping between the two cannot push the bound out forever (round-1 `[minor]`).
- **A degraded `COMMENT` pages the operator once.** GitHub consumes the pending review request on *any* submitted review, so the PR leaves `review-requested:@me` the moment the degraded verdict lands and a `COMMENT` is not what the DEV fix flow keys on. Rather than rely on a re-request nobody is prompted to make, the degrade posts one comment per (round, head) naming the rollup, that nothing further is queued, and the two ways back in. The red path needs none of it — `REQUEST_CHANGES` is that signal (round-1 `[question]`).
- **An unrecognised *completed* conclusion blocks.** `cancelled`, `timed_out`, `action_required` and anything GitHub adds later are treated as red rather than as success. Deliberately loud in the direction of not approving.
- **An empty combined-status list is dropped, not read as pending.** `GET /commits/<sha>/status` answers `state: pending` for a commit with **no** statuses; reading that literally would hold every approve on every Actions-only repo until the bound.
- **A 404 is treated as just another unreadable rollup, with no error-string special case.** The first version stepped aside on a `"404"`/`"not found"` substring, which round 1 correctly called fail-open: a 404 also answers "this token cannot see this private repo", and stepping aside posted an *ungated* approve. It now holds and degrades like any unreadable rollup; the force-push case is still released, because the degraded post gets the same 422 and reaches `_abandon_verdict`, which proves absence from the PR's commit list rather than trusting the message.
- **Completeness of a rollup listing is decided by the payload's `total_count` when present, and by the short-page rule when it is not.** Round 1 found `_commit_statuses` unpaged against an endpoint whose default page size is 30 — a red context past #30 read as green. Both listings now share one paged reader; a count the endpoint will not serve, or more pages than the bound, is `CHECKS_UNKNOWN`, never green.
- **The convergence rule is narrowed to the two states the gate can produce** (`COMMENTED`, `CHANGES_REQUESTED`). A broader test would also have changed what a `DISMISSED` daemon approve does to convergence, which is pre-existing and its own question — registered as **TASK-939082213**, not changed here.
- **Dry-run reads the rollup and says what it would have done.** A read takes nothing on GitHub and writes no ledger row; dry-run is the mode an operator reaches for to diagnose exactly this, the same argument the identity-drift check already makes.
- **Waiting note goes into the existing mechanical activity comment, deduped per (round, head) — zero new comments.** The issue permits one waiting comment per round; none is better, and the activity comment is already the mechanical log an operator reads.
- **Container pass-through wired (`ALISSA_CHECKS_WAIT_SECONDS`) though the issue only asked for a configurable wait.** A config key the deployed image cannot set is not configurable in practice. The `ARG REVLOOP_VERSION` pin (still `0.16.0`) is deliberately **not** bumped here — this version is not on PyPI while the PR is open, and re-pinning is its own lane — so the new env var must stay unset until that pin reaches ≥ 0.16.7, which is documented next to the ARG and in `docker/claude/README.md` exactly like the `reviewer_token_env` floor.

- [ ] **No live GitHub round exercised the gate.** Every scenario is unit-tested against a fake client; no real PR was approved, held, or gated in this session (that needs a red/pending PR with a review request pending for the REV identity).
- [ ] **The two rollup endpoints' live payload shapes** (`check-runs`, `status`) are parsed from documented fields and tested against hand-built payloads. Partially verified live during round 1 against this repo's own head: `check-runs` answered `total_count: 6` for 6 runs (so the count is exact for that endpoint), and `status` answered `total_count: 0` with `state: pending` for a commit with no statuses (the case the empty-list handling exists for). A **multi-page** rollup — 30+ legacy statuses or 100+ check runs — has not been seen live.
- [ ] **`checks: read` scope on the reviewer credential.** If the deployed REV token cannot read a repo's rollup, every approve on that repo degrades to a `COMMENT` (loudly, naming the error and paging the operator) rather than approving. Worth confirming before deploy — this is now a per-verdict degradation rather than a poll-pass abort, but it is still a degradation.
- [ ] **The review-request consumption premise.** The degraded-verdict page and the "re-request to re-enter" wording rest on GitHub consuming a pending review request when the requested identity submits a **comment-mode** review. That is this loop's own edge-trigger premise and studio #210's three comment-mode rounds are evidence for it, but it was not re-confirmed live in this session — no comment-mode review was submitted as the REV identity.
- [ ] **The container env var** `ALISSA_CHECKS_WAIT_SECONDS` is exercised only through `docker/claude/tests-entrypoint-config.sh` (jq render + library cross-check). No image was built or booted — no docker in this runner.
- [ ] **The ARG `REVLOOP_VERSION` pin is 0.16.0**, so a deployed image will not accept the new key until the pin lane bumps it past 0.16.7. Setting the env var against an older pin fails config load with `unknown config key(s): checks_wait_seconds`.
- [ ] **Publish CI** (version 0.16.7) fires only on merge; `pip index versions` showed 0.16.6 as the newest published dist at the time of writing, so 0.16.7 is free — worth re-checking if another lane merges a bump first.

### Verification

- `bash tests-unit.sh alissa-tools-github-revloop` → **549 passed** (30 new) after the round-1 fixes. Includes: red rollup posts `REQUEST_CHANGES` naming the check + run URL and no approve; the gated verdict does not converge and a later green round approves the same code with no new commit; skipped/neutral and no-checks-at-all approve unchanged; the rollup is read for the judged head while the current head is green; pending holds with no post; one waiting note across four polls and zero operator comments; unsettled-at-bound degrades to `COMMENT` and stays re-enterable; unreadable is never green; missing commit routes to `ABANDONED`; a `request_changes` verdict reads no rollup at all; no label-mutating method exists on the client; the `verdict_posts.checks_held_at` migration on a pre-gate database.
  Round-1 additions: a 403 read degrades instead of aborting the pass (asserted *below* `_api`, by faking `run_json`, so it runs through the mapping that had made the path unreachable); an explicit throttling marker still raises `RateLimited`; 35 paged statuses roll up red on context #33; a `total_count` the endpoint will not serve is `CHECKS_UNKNOWN`; exactly the page bound is a complete read; a 404 holds and then reaches `ABANDONED`; a transient `unknown` poll does not eat the pending bound and the promotion happens at most once; the degrade pages the operator exactly once and the red path pages not at all; the dry-run diagnostic reads each rollup once per process.
- `bash tests-coverage.sh alissa-tools-github-revloop` → 93% total.
- `bash check-style.sh alissa-tools-github-revloop` → clean.
- `bash check-types.sh alissa-tools-github-revloop` → `Success: no issues found in 28 source files`.
- `bash docker/claude/tests-entrypoint-config.sh` → `ALL PASS` (pass-through omitted when unset, override emitted as a number, library-default cross-check).
- No failures were left behind; the run above is the state of the branch.
- Acceptance criterion 4 (labels) verified by inspection as well as by test: `ghclient.GitHub` has no label-mutating method and nothing calls the issues `labels` endpoint. `git grep -ni label -- alissa-tools-github-revloop/src/main` matches only the two verdict-body sentences that *say* no label was touched, the gate's docstring, and unrelated CSS/`aria-label` strings in the console page.

### Scope touched

Declared scope was the revloop verdict path plus a mandatory patch bump. Files:

- `revloop/loop.py` — the gate, the body templates, the hold note, the convergence rule.
- `revloop/ghclient.py` — `check_rollup` + the two readers, `CheckContext`/`CheckRollup`/`rollup_of`, and `submit_review` now accepts `COMMENT` (narrowly, for the degraded verdict; anything outside the three events is still refused).
- `revloop/state.py` — nullable `verdict_posts.checks_held_at` + `checks_held_state`, with `checks_hold`/`record_checks_hold` (the ledger remembers; the gate decides when to stamp); the snapshot column migration was generalised to a table→columns map so both tables migrate through one path.
- `revloop/config.py`, `revloop/__main__.py` — `checks_wait_seconds` key, validation, `--checks-wait-seconds`.
- `revloop/version` — 0.16.6 → 0.16.7 (publish CI fires only on this file).
- tests: `test_loop.py`, `test_state.py`.
- **Excursions, both deliberate and small:** `docker/claude/{Dockerfile,revloop-config.sh,README.md,tests-entrypoint-config.sh}` for the env pass-through (see the judgment call above), and `README.md` + `revloop.config.example.json` for the operator-facing docs. Round-accounting, cap logic, the reaper (TASK-793080321) and identity enforcement (TASK-103296598) are untouched.


alissa-app:
- 2026-07-30 01:50:51 UTC — `review-alissa-github-review-daemon-pr59-r1-c54945` — round 1 of 22 — spawned
- 2026-07-30 02:22:07 UTC — `review-alissa-github-review-daemon-pr59-r2-1cb6f2` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -504,6 +638,115 @@ def pull_request_commits(self, owner: str, repo: str, number: int) -> list[str]:
...
+
+        Never raises except RateLimited (which run_forever's backoff owns): an
+        unreadable rollup is a CHECKS_UNKNOWN answer the caller can hold on, not
+        a reason to abort a poll pass that has other PRs to decide.
+        """
+        contexts: list[CheckContext] = []
+        try:
+            contexts.extend(self._check_runs(owner, repo, sha))
+            contexts.extend(self._commit_statuses(owner, repo, sha))
+        except RateLimited:
```
alissa-app, line 670: [major] A 403 on either rollup read becomes `RateLimited` and aborts the whole poll pass — it never becomes `CHECKS_UNKNOWN`.

`_check_runs` and `_commit_statuses` call `self._api(...)` without `forbidden_is_rate_limit=False`, so `_api`'s default applies: any stderr containing `403` is raised as `RateLimited` (ghclient.py L516). `check_rollup` then re-raises it, `evaluate` re-raises it, `poll_once` re-raises it, and `run_forever` treats it as throttling — backoff doubles toward 900s and every PR after this one in the pass goes unevaluated.

Failure scenario: the reviewer credential is a GitHub App / fine-grained token without `Checks: read` on a watched repo. GitHub answers `403 Resource not accessible by integration`. Expected (per the delivery note and per `CHECKS_UNREADABLE`): the round holds, then degrades to a `COMMENT` quoting the error so an operator can fix the scope. Actual: every poll pass aborts at that PR, the daemon backs off to 900s indefinitely, PRs later in the search set are never decided, and no comment ever names the missing scope.

Two things make this worth fixing rather than waiving:

- `CHECKS_UNREADABLE` (loop.py L184) exists to say "check that the reviewer credential carries `checks: read` on this repo" — and with 403 routed to `RateLimited` and 404 swallowed by `_COMMIT_GONE_MARKERS` below, **no HTTP-status read failure can reach that message**. It is only reachable via a 5xx, a timeout, or a JSON parse error.
- `test_an_unreadable_rollup_reports_why_and_never_raises` asserts exactly this behaviour by monkeypatching `gh._api` to raise `CommandError("403 Resource not accessible")` directly — bypassing the mapping that makes it impossible. The test cannot fail on this bug.

`submit_review` and `remove_review_request` already pass `forbidden_is_rate_limit=False` for precisely this reason ("a 403 here is an authorization failure with its own handling, not a throttle"). This read has its own handling too — `CHECKS_UNKNOWN` — so it wants the same flag.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -504,6 +638,115 @@ def pull_request_commits(self, owner: str, repo: str, number: int) -> list[str]:
...
+
+    def _commit_statuses(self, owner: str, repo: str, sha: str) -> list[CheckContext]:
+        """The commit's legacy statuses, one context each.
+
+        `total_count == 0` means the repo posts no commit statuses at all, and
+        the endpoint's own `state` for that case is `pending` -- which is why
+        only the `statuses` array is read and the combined state is ignored.
+        """
+        data = (
+            self._api("-X", "GET", f"repos/{owner}/{repo}/commits/{sha}/status") or {}
```
alissa-app: [major] This read is unpaged and drops `total_count`, so a red commit status past the first 30 contexts makes the rollup read `green` — an APPROVE on a red head.

`GET /commits/{sha}/status` paginates with a default `per_page` of 30. No `per_page` and no `page` are passed here, and only `data["statuses"]` is consumed, so on a commit with 31+ contexts the last ones are silently invisible. `rollup_of` sees no failing context among the 30 it got, returns `CHECKS_GREEN`, and the gate approves.

Failure scenario: a watched repo posts commit statuses from several external systems (35 contexts, say a Buildkite matrix plus deploy previews plus a coverage bot); context #33 is `failure`. The gate reads 30 statuses, all `success`, concludes green, and posts `APPROVE` on a red head — the exact incident class issue #58 exists to prevent, reached through the endpoint the check-runs path was careful about.

The asymmetry is the tell: `_check_runs` right above bounds itself to `CHECK_RUN_PAGE_LIMIT` pages and raises `TruncatedListing` rather than return a partial list, on the stated ground that "a partial read cannot support an approve". The same argument applies here, and the evidence needed is already in the response — `total_count` is in the payload (the new tests build it in both fakes) and is currently only mentioned in the docstring, never compared to `len(statuses)`. Either page it like `_check_runs`, or return `CHECKS_UNKNOWN` when `total_count > len(statuses)`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -881,24 +990,179 @@ def _close_round_natively(
...
+        rollup = self.github.check_rollup(pr.owner, pr.repo, judged)
+
+        if rollup.state == CHECKS_GREEN:
+            log.debug(
+                "%s round %d: CI rollup at %s is %s — approving as usual",
+                pr.slug, round_, judged[:8], rollup.summary,
+            )
+            return ChecksGate()
+
+        if rollup.commit_missing:
```
alissa-app: [major] The one branch of the gate that fails **open** decides on a substring match — and this repo's own code says not to trust that string for this question.

`commit_missing` is set by `any(marker in str(exc).lower() for marker in _COMMIT_GONE_MARKERS)` where the markers are `("404", "not found")` (ghclient.py L119, L676). Landing here returns a default `ChecksGate()` — `event=EVENT_APPROVE`, empty `lead`, `state=CHECKS_GREEN` — so the round goes on to post an **unqualified APPROVE** on a head whose rollup was never read.

The mitigation in the comment ("the review POST will be rejected and `_abandon_verdict` releases the round") only holds if a 404 always means the commit is gone from the PR. When it does not, `submit_review` succeeds and the approve lands — no lead, no `gate_note` in the log or the activity line, `state` recorded as `CHECKS_GREEN`. That is a silent false merge cue, and it is the one outcome the whole gate exists to make impossible.

Failure scenario: any read failure whose `gh` stderr contains `not found` or `404` while `judged` is still a commit of the PR — e.g. the reviewer token cannot see a private watched repo (GitHub answers 404, not 403, for no-access), or a GHES/`gh`-version skew on the `check-runs` path. The gate steps aside, `submit_review` succeeds because the *review* permission is intact, and the PR gets an APPROVE on an unverified head.

`_abandon_verdict` (L1219) already states the rule this branch breaks: *"Detection does not trust the error string. A rejection that mentions the commit only starts the question; the answer comes from reading the PR's commits and finding the judged SHA genuinely absent."* Two ways to keep that property: prove absence with `pull_request_commits` before stepping aside, or simply do not step aside — fall into the hold/degrade path, whose eventual `COMMENT` post gets the same 422 and reaches `_abandon_verdict` anyway. The second costs one wait bound and never approves.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -881,24 +990,179 @@ def _close_round_natively(
...
+        )
+        detail = (
+            CHECKS_STILL_RUNNING.format(
+                names=", ".join(f"`{c.name}`" for c in rollup.running) or "none"
+            )
+            if rollup.state == CHECKS_PENDING
+            else CHECKS_UNREADABLE.format(why=rollup.unreadable or "no reason recorded")
+        )
+        return ChecksGate(
+            event=EVENT_COMMENT,
```
alissa-app, line 1213: [question] After the degraded `COMMENT` lands, what re-enters the loop? Both branches of GitHub's review-request semantics look bad, and the fake models only one of them.

The new convergence rule keeps the PR un-converged, so `evaluate` reaches `round_ = completed + 1` and spawns — but only if the PR is still in `review_requests`, which is the *only* thing `poll_once` iterates. So the answer turns on whether GitHub consumes the pending review request when the requested identity submits a `COMMENT` review:

- **If it does consume it** (which is what `_clear_own_review_request`'s own docstring asserts — *"GitHub consumes a review request the moment the requested identity submits a review"*): the PR drops out of the search set the moment the degraded verdict posts, and nothing automatic brings it back. A `COMMENT` is not the signal the DEV fix/re-request flow consumes (that flow reads `REQUEST_CHANGES`), and the operator's cue (an APPROVE) never comes. The PR strands silently with an unsettled rollup, no verdict of record anyone acts on, and no cap-out escalation — the same silent-stranding class as the #227 latch and the #298 dangling request.
- **If it does not consume it:** every poll after the degrade spawns round k+1 immediately, which holds ~30 min, degrades to another `COMMENT`, spawns k+2 … burning the effective 22-round cap and 22 real reviewer sessions on a CI system that is merely slow or unreadable, before escalating.

`test_a_rollup_that_never_concludes_degrades_to_a_comment` asserts the second shape (`SPAWNED` on the next `evaluate`), but `FakeGitHub` never removes the request on submit, so the suite cannot distinguish the two — the same gap applies to `test_a_later_green_round_approves_the_same_code`, whose "later green round" only runs unprompted for the same reason.

Which is it in practice, and what guarantees re-entry? If it is the first branch, the degrade probably owes the operator something loud (the escalation path, or the ping ledger it already has) rather than relying on a re-request nobody is prompted to make. Worth stating explicitly in the delivery note either way, since "the round stays re-enterable" is currently doing a lot of load-bearing work in the README and the PR body.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1361,6 +1625,21 @@ def _convergence_reason(
...
+        # produces that shape (issue #58): a red or never-concluded rollup turns
+        # an approve envelope into a REQUEST_CHANGES or COMMENT review. Without
+        # this, the very next poll would converge on the envelope, withdraw the
+        # review request, and take the PR out of the loop on a verdict the
+        # daemon deliberately declined to post as an APPROVE -- leaving nothing
+        # to re-enter and no later green round to approve it.
+        #
+        # Outside the gate this changes nothing: a request_changes envelope does
+        # not converge on the branch below either.
+        if newest.verdict_round is not None and newest.state != "APPROVED":
```
alissa-app: [minor] "Outside the gate this changes nothing" is not quite true — this also changes the `DISMISSED` case, and that change is unremarked.

`SUBMITTED_STATES` includes `DISMISSED`, so `my_reviews` returns a dismissed review, and a review the daemon posted carries a `verdict_round` forever. Before this rule: a marked review at the current head whose state is `DISMISSED` fell through to the envelope check, and an `approve` envelope converged the loop. After it: `verdict_round is not None` and the state is not `APPROVED`, so it returns `None` and a fresh round is owed.

That is arguably the better behaviour — a human dismissing the daemon's approve is a human saying "not approved" — and #227's history says dismissals do happen here ("the implementer pushed `fd500fc` and re-requested (and even dismissed the approve)"). But it is a second behavioural change riding along on a comment that says there is none, and this repo has already spent a PR (#56) correcting a rationale that had drifted from what the code does. Either name the `DISMISSED` case as intended in the comment, or narrow the condition to the states the gate actually produces (`COMMENTED` / `CHANGES_REQUESTED`).

Also: the `and newest.state != "APPROVED"` conjunct is dead — the `APPROVED` branch two lines above already returned.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -881,24 +990,179 @@ def _close_round_natively(
...
+                lead=CHECKS_RED_LEAD.format(
+                    sha=judged[:8], failing=failing, verdict=VERDICT_APPROVE
+                ),
+                state=CHECKS_RED,
+            )
+
+        # PENDING (checks still running) or UNKNOWN (the rollup could not be
+        # read) -- neither can support an approve, and both can settle, so both
+        # HOLD the round first and only degrade at the bound.
+        held_at = self.state.note_checks_hold(pr.full_name, pr.number, round_)
```
alissa-app: [minor] A single transient unreadable poll starts the wait clock, so a later genuine `pending` gets less than the configured bound.

`checks_held_at` is stamped once and never cleared, and it is stamped for `CHECKS_UNKNOWN` as well as `CHECKS_PENDING`. So one flaky read — a 5xx, a `gh` timeout — sets the clock even though nothing was waiting on CI at that moment.

Failure scenario: `checks_wait_seconds` is 1800. Poll 1 hits a transient 502; the rollup is `UNKNOWN`, `checks_held_at` is stamped at T. Poll 2 succeeds and the checks are genuinely `pending` — a normal 10-minute suite that started late. At T+1800 the gate degrades to a `COMMENT` even though the real wait on concluding checks was well short of 30 minutes, and the round is spent on a suite that would have gone green.

Small, and it errs toward not approving, so nothing unsafe follows. But the docstring's stated intent is "a fixed distance from the first observation" of a rollup that has not settled, and a transient read error is not that observation. Clearing the stamp on any poll that reads a settled rollup, or stamping only on `CHECKS_PENDING`, would match the intent.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -504,6 +638,115 @@ def pull_request_commits(self, owner: str, repo: str, number: int) -> list[str]:
...
+                            str(run_.get("conclusion") or "").lower()
+                            if completed
+                            else ""
+                        ),
+                        url=str(run_.get("html_url") or run_.get("details_url") or ""),
+                    )
+                )
+            if len(runs) < PER_PAGE:
+                return out
+        raise TruncatedListing(
```
alissa-app, line 738: [nit] Off-by-one at the page bound: exactly `CHECK_RUN_PAGE_LIMIT * PER_PAGE` runs is a *complete* read that reports as truncated.

If page 5 returns exactly `PER_PAGE` runs, `len(runs) < PER_PAGE` is false, the loop exits, and `TruncatedListing` is raised even though page 6 would have been empty — so a commit with exactly 500 check runs is `CHECKS_UNKNOWN` rather than green. Wholly academic at 500, and it errs toward not approving, so this is discretion only. `total_count` is in the check-runs payload too if you'd rather bound on that than on the short-page heuristic.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -336,6 +357,12 @@ def build(
             # state of a working loop -- an alarm that always fires is noise.
             raise ValueError(f"reap_session_cap must be >= 1, got {session_cap}")
 
+        checks_wait = int(raw.get("checks_wait_seconds", cls.checks_wait_seconds))
+        if checks_wait < 0:
+            raise ValueError(
+                f"checks_wait_seconds must be >= 0, got {checks_wait}"
```
alissa-app: [nit] This `raise` is wrapped across three lines while its neighbours in the same validator are one-liners (`raise ValueError(f"reap_session_cap must be >= 1, got {session_cap}")`, and the `reap_grace_seconds` one above it). Unwrapped it is 84 characters, which style check accepts — the surrounding lines already run longer.

alissa-app (changes requested):

Fresh-instance review (CR3) against issue #58 / the origin task's definition-of-done.
Review task: **TASK-2073989161** (`Review PR fahera-mx/alissa-github-review-daemon#59 (TASK-738619486)`), created for this round per CR2 — none existed.
Reviewed surface: `8164bc8..54fccb4`, one commit, 14 files, +1197/-46. Judged head `54fccb4`; the PR's own CI is green (6/6).

The shape of the gate is right and the hard parts are handled with care: it reads the rollup for the SHA the verdict is **pinned to** rather than the PR's current head, `skipped`/`neutral` never block, an empty combined-status list is correctly dropped instead of read as pending (I confirmed this repo's own `/status` answers `state: pending` with `total_count: 0`), no-checks-at-all still approves, `REQUEST_CHANGES` is not gated, the wait note reuses the mechanical activity comment so a held round costs zero new comments, the round marker keeps the degraded `COMMENT` closing its round exactly once (`COMMENTED` is in `SUBMITTED_STATES`, so `countable_rounds` sees it), the `verdict_posts.checks_held_at` migration is additive/nullable/idempotent, and AC4 holds — I re-verified by inspection that no label-mutating call exists anywhere in `src/main`. Version bumped 0.16.6 → 0.16.7, so AC5 holds too.

What blocks the approve is that **three of the gate's own failure paths fail open or fail loud in the wrong direction**, in a component whose entire contract is to fail closed:

1. **No HTTP-status read failure can actually reach the `CHECKS_UNKNOWN` guidance path.** `_api` maps any stderr containing `403` to `RateLimited` (default `forbidden_is_rate_limit=True`), and `check_rollup` re-raises it — so a missing `checks: read` scope aborts the whole poll pass and doubles `run_forever`'s backoff toward 900s instead of degrading to a `COMMENT`. A `404` is swallowed by `_COMMIT_GONE_MARKERS` into `commit_missing`. The `CHECKS_UNREADABLE` body that names the exact fix ("check that the reviewer credential carries `checks: read`") is therefore only reachable via a 5xx/timeout/parse error. The delivery note's claim — "every approve on that repo degrades to a `COMMENT` (loudly, naming the error)" — does not hold, and the test that asserts it monkeypatches `_api` itself, bypassing the mapping that makes it impossible.
2. **`_commit_statuses` is unpaged and ignores `total_count`,** so a commit with more than 30 legacy statuses reads only the first 30 — and a failing status past that boundary makes a red head read `green` and approve. The check-runs path bounds itself and raises `TruncatedListing` for exactly this reason; the statuses path next to it does not, and `total_count` is already in the payload (the new tests even build it).
3. **`commit_missing` is decided by a `"404"`/`"not found"` substring match and fails open to an ungated `APPROVE`.** The mitigation ("the review POST's own 422 decides") only holds if a 404 always means the commit is gone. `_abandon_verdict`'s own docstring, one screen up, says not to trust the error string for precisely this question and reads the PR's commit list instead.

Plus one load-bearing `[question]` I could not settle from the diff: what re-enters the loop after a degraded `COMMENT`. Both branches of GitHub's review-request semantics look bad, and the fake models only one of them.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 3 (waived: 0) |
| minor | 2 |
| nit | 2 |
| question | 1 (open: 1) |

### Dimension report

- **Correctness:** DoD's happy path, red path, pending hold/degrade, skipped/neutral, head-pinning, no-label and version-bump criteria all hold; three `[major]`s on the gate's failure paths (403 → fleet backoff instead of degrade; unpaged `/status` can read a red head green; `commit_missing` fails open to an ungated approve), one `[minor]` on the hold clock, one open `[question]` on the degraded verdict's re-entry trigger.
- **Coding practices:** matches this repo's idiom closely (rationale-carrying docstrings, named constants, ledger-after-append ordering). One `[minor]`: the new convergence rule's comment claims it changes nothing outside the gate, but it also changes the `DISMISSED` case. Two `[nit]`s (page-bound off-by-one, wrapped `raise`).
- **Security:** nothing found. Both new endpoints are reads; the SHA is interpolated into an argv (never a shell); the rollup error text echoed into the public PR body comes from `CommandError`, whose message is `argv[0]` + stderr only — no credential, and the token reaches `gh` through a replaced env rather than the command line.
- **Performance:** nothing blocking. Two extra reads only on the approve-post path; a held round re-reads each poll but is bounded by `checks_wait_seconds`. One `[nit]`: in dry-run the rollup is re-read every poll forever, since the dry-run branch returns before the grace/ledger bookkeeping.
- **Side-effects / blast radius:** contained. `submit_review`'s accepted-event set widens by `COMMENT` with the gate as its only caller; the schema change is a nullable added column through the generalised `_ADDED_COLUMNS` path and `read_verdict_posts` has no positional consumer in-tree; round accounting and cap logic are untouched; no labels; the docker pass-through correctly documents the `REVLOOP_VERSION >= 0.16.7` floor and the entrypoint cross-check deliberately leaves the new key unset.

Fix the three `[major]`s — pass `forbidden_is_rate_limit=False` on the two rollup reads so a 403 becomes `CHECKS_UNKNOWN` rather than a fleet-wide backoff; page `/status` (or compare `total_count` to `len(statuses)` and return `CHECKS_UNKNOWN` on a short read); and make `commit_missing` either hold/degrade or prove absence from `pull_request_commits` instead of stepping aside on a substring — then answer the `[question]` on the degraded verdict's re-entry trigger. The `[minor]`s and `[nit]`s are your call per CR8; the verdict of record is the CR6 envelope on TASK-2073989161.

RHDZMOTA:
- 2026-07-30 02:03:30 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr59-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: `[triage:pursue]` Confirmed, and the reasoning is exactly right — including that my own test cannot fail on it, because monkeypatching `_api` skips the very mapping that breaks the path. Both rollup reads now pass `forbidden_is_rate_limit=False`, for the same reason `submit_review` and `remove_review_request` do: a 403 here is an authorization fact about the deployment with its own handling (`CHECKS_UNKNOWN` → hold → `COMMENT` naming the scope), not a throttle worth aborting a poll pass for. A real secondary rate limit still raises `RateLimited` through the explicit markers, so the backoff keeps the case it is for.

The test is re-pointed one layer down: it monkeypatches `ghclient.run_json` (not `_api`) to raise `CommandError` with `403 Resource not accessible by integration` in stderr, so the assertion now runs *through* `_api`'s mapping and would have caught this.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: `[triage:pursue]` Confirmed — and this is the one finding here that could have produced the incident issue #58 exists to prevent, through the endpoint I was careless about rather than the one I was careful about.

Both readers now bound on the payload's own `total_count` instead of the short-page heuristic:

- `_commit_statuses` pages `per_page=100`/`page=N` like `_check_runs`, capped by the same page bound.
- After the loop each reader compares what it collected against `total_count`; a short read raises `TruncatedListing`, which `check_rollup` turns into `CHECKS_UNKNOWN` — never green.

That also removes the `[nit]` off-by-one below (exactly `CHECK_RUN_PAGE_LIMIT * PER_PAGE` runs is now a complete read, because completeness is decided by `total_count`, not by a page happening to be short). New tests: 35 statuses across pages roll up red when #33 fails, and a `total_count` larger than the statuses served is `CHECKS_UNKNOWN`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` Confirmed, and I am taking your second option: **the gate no longer steps aside at all.** A 404-shaped read is now just another unreadable rollup — `CHECKS_UNKNOWN` → hold → degrade to `COMMENT` — and `commit_missing`, `_COMMIT_GONE_MARKERS` and the branch that returned a default `ChecksGate()` are deleted. The degraded `COMMENT` post gets the same 422 on an absent commit and reaches `_abandon_verdict` through the existing path, so the round is still released; it costs one wait bound and, as you say, never approves.

I considered proving absence with `pull_request_commits` before stepping aside (which would keep the abandon fast). I dropped it: it adds an API call and a second `TruncatedListing` branch to the rare side of a rare path, and the genuine force-push case does not usually reach a 404 anyway — GitHub keeps serving `check-runs` for an orphaned commit, so that case lands on the ordinary post → 422 → abandon route regardless. Fewer moving parts on the branch that used to fail open is worth 30 minutes on a path that ends in the same place.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:answer]` + `[triage:pursue]` — it is your **first** branch, and you are right that the degrade owed the operator something loud.

The answer, from this repo's own history rather than from the docs: GitHub consumes a pending review request when the requested identity submits a review of **any** type, `COMMENT` included. The evidence is the loop's own premise — the README's edge-trigger design ("GitHub clears a pending review request the moment you submit a review, and re-adds it when the implementer re-requests") — plus studio #210, which ran three rounds with **zero** `APPROVED` records: every round there closed on a comment-mode review, and rounds only advanced because each of those `COMMENTED` reviews consumed the request and the implementer's re-request re-added it. If a `COMMENT` did not consume it, that PR would have double-spawned every poll instead.

So the degraded verdict drops the PR out of `review_requests` immediately, nothing automatic brings it back, and a `COMMENT` is not what the DEV fix flow keys on. Three changes:

1. **The fake now models it.** `FakeGitHub.submit_review` removes the reviewer from `requested_reviewers` and drops the PR out of `review_requests`, exactly as `remove_review_request` already did. The suite can no longer be silent about this — and `test_a_later_green_round_approves_the_same_code` now has to re-add the request explicitly, which is honest: that "later green round" runs *because the implementer re-requested*, not on its own.
2. **The degrade pages the operator, once.** Alongside the `COMMENT` verdict it posts one operator comment (deduped through the existing ping ledger on `(round, judged head)`, so at most one per round per head) saying the round did not approve, that GitHub has consumed the review request so the daemon will not look at this PR again unprompted, what the rollup was, and the two ways back in — conclude/fix the checks and re-request review, or merge with a recorded waiver. That is the same posture as the cap-out page: a loop that cannot converge without a human says so where the human is looking.
3. **The docs stop overclaiming.** "The round stays re-enterable" now reads as what it is — re-enterable *on a re-request*, which the page asks for — in the README, the verdict body and the delivery note.

Note the red path never had this gap: `REQUEST_CHANGES` is exactly the signal the DEV fix/re-request flow consumes, so it re-enters by itself.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` Both correct. The condition is narrowed to the two states the gate actually produces —

```python
if newest.verdict_round is not None and newest.state in GATED_VERDICT_STATES:  # COMMENTED, CHANGES_REQUESTED
```

— which drops the dead `!= "APPROVED"` conjunct (the `APPROVED` branch above already returned) and leaves the `DISMISSED` case byte-for-byte as it was, so this PR carries exactly the one behavioural change its comment claims.

You are also right that the `DISMISSED` behaviour looks wrong on its own terms: today a human dismissing the daemon's approve does not re-open a round, because the envelope still reads approve and convergence never looks at the dismissal. That is pre-existing and outside issue #58, so it is registered rather than deferred in prose: **TASK-939082213** (*"a DISMISSED daemon verdict at the current head should not converge the loop on its envelope"*), soft-linked to this implementation task.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` Agreed — the docstring says "first observation" of a rollup that has not settled, and a flaky read is not that observation.

The hold now remembers *which* condition it is waiting on (`verdict_posts.checks_held_state`, alongside `checks_held_at`), and the clock restarts exactly once: on the first poll that promotes an `unknown` hold to a genuine `pending` one. After that it never moves — including if the reader flaps back to `unknown`, which is deliberate: a rule of "restart whenever the state changes" would let an alternating reader push the bound out forever, which is the unbounded hold this bound exists to prevent. So an unreadable rollup is still bounded on its own clock, and your scenario (transient 502, then a real 10-minute suite) gets the full 30 minutes from the moment the checks were actually seen running.

Pinned by a test that walks exactly that sequence.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: `[triage:pursue]` Fixed, and by your suggestion rather than by an off-by-one patch: both readers now decide completeness from the payload's `total_count` instead of the short-page heuristic, so exactly `CHECK_RUN_PAGE_LIMIT * PER_PAGE` runs is a complete read and rolls up normally. Same change as the `[major]` on `_commit_statuses` above.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: `[triage:pursue]` Unwrapped to a one-liner, matching its neighbours.

RHDZMOTA:
`[triage:pursue]` — one item from the **Performance** dimension that had no inline thread: *"in dry-run the rollup is re-read every poll forever, since the dry-run branch returns before the grace/ledger bookkeeping."*

Correct, and it is the one place in this change where a diagnostic read is unbounded. Fixed with this file's existing pattern for exactly this problem (`_dry_run_drift`): an in-memory `(repo, number, round, judged)` set, so a dry-run pass reads each round's rollup once per process and logs it once. Deliberately in-memory and not the ping ledger — a dry-run pass must never write durable state that could later silence a production observation.

Round-1 triage summary, all eight findings plus this one **pursued**, nothing deferred and nothing waived:

| finding | disposition |
| --- | --- |
| `[major]` 403 → `RateLimited` aborts the pass | `forbidden_is_rate_limit=False` on both reads; test re-pointed below `_api` |
| `[major]` unpaged `/status` can read a red head green | paged + both readers bound on `total_count`, short read ⇒ `CHECKS_UNKNOWN` |
| `[major]` `commit_missing` fails open to an ungated approve | branch deleted; a 404 is just an unreadable rollup (hold → `COMMENT`) |
| `[question]` what re-enters after a degraded `COMMENT` | answered (the request *is* consumed); fake now models it; degrade pages the operator once |
| `[minor]` convergence rule also changed `DISMISSED` | narrowed to the gate's states; `DISMISSED` untouched, registered as TASK-939082213 |
| `[minor]` transient `unknown` starts the wait clock | hold remembers its condition; one promotion `unknown` → `pending` restarts the clock |
| `[nit]` page-bound off-by-one | folded into the `total_count` bound |
| `[nit]` wrapped `raise` | unwrapped |
| dimension-report dry-run re-read | in-memory per-process dedupe |

The fixes land as one commit on this branch next, then I re-request review for round 2.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -881,24 +1065,226 @@ def _close_round_natively(
...
+        # The clock is stamped against the CONDITION being waited on, and gets
+        # exactly one restart: when an `unknown` hold is promoted to a `pending`
+        # one. A transient read failure is not an observation of checks running,
+        # so it must not eat the bound a real suite is entitled to. One restart,
+        # never "restart whenever the state changes" -- a reader flapping between
+        # the two would then push the bound out forever, which is precisely the
+        # unbounded hold this bound exists to prevent.
+        held_at, held_state = self.state.checks_hold(pr.full_name, pr.number, round_)
+        promoted = held_state == CHECKS_UNKNOWN and rollup.state == CHECKS_PENDING
+        if held_at is None or promoted:
```
alissa-app, line 1170: [minor] The promotion restart makes the real worst-case hold up to **2x** `checks_wait_seconds`, but every operator-facing description of that key still states it as *the* bound.

The restart itself is right, and it fixes round 1's finding properly. The consequence is that the two holds are billed to separate clocks: an `unknown` hold can run the full bound, get promoted at the last moment, and then a `pending` hold runs a *fresh* full bound from there. With the default that is a PR held ~60 minutes, not ~30.

Failure scenario: `checks_wait_seconds` is 1800. The rollup is unreadable (a flaky 502, or a credential missing `checks: read` that an operator is mid-way through fixing) for 29 minutes; at minute 29 the read succeeds and the checks are genuinely running, so the clock restarts. The round now degrades at ~T+59m. An operator who set 1800 because "a round must not sit longer than half an hour" gets twice that, and nothing in the docs told them the number could double.

Where it is stated as a single bound:

- `README.md` L213 — "holds the round open, re-checking each poll, **up to `checks_wait_seconds`**";
- `README.md` L118 (config table) and L277 (behaviour table) — same framing;
- `revloop.config.example.json` L38 and `docker/claude/revloop-config.sh` L17 — "how long a round holds its APPROVE ... before recording the verdict as a COMMENT".

The code comments here and in `state.py` do describe the restart, so this is documentation drift rather than a logic bug, and it errs toward not approving so nothing unsafe follows. Raising it because it is the same class as round 1's `[minor]` on the convergence rule — a rationale that no longer matches what the code does — and because this is the one number an operator actually tunes. Worth noting too that the degraded review body reports `waited` from the restarted clock, so a promoted hold says "30m of the 30m bound" after 60 real minutes.

Cheapest fix is a clause in the README row and the example-config gloss ("...per condition waited on; an unreadable hold that becomes a genuine pending one restarts the clock once, so the worst case is twice this"). Alternatively cap the total at the bound and let the promotion only *extend* to it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -651,6 +795,12 @@ def __init__(
         # dry-run should not re-announce the same drift every poll either --
         # see _warn_identity_drift.
         self._dry_run_drift: set[tuple[str, int, str]] = set()
+        # (repo slug, number, round, judged head) -> the rollup summary this
+        # process already read for it in DRY-RUN. The dry-run branch returns
+        # before any ledger write, so without this the diagnostic re-reads the
+        # rollup (two API calls) on every poll, forever, for every PR with an
+        # owed approve. In-memory for the same reason _dry_run_drift is.
+        self._dry_run_rollups: dict[tuple[str, int, int, str], str] = {}
```
alissa-app, line 803: [nit] `_dry_run_rollups` grows for the life of the process with no eviction, and unlike its neighbour it stores a value per entry.

Keyed on `(repo slug, number, round, judged head)`, so a long-lived `--dry-run` daemon accumulates one entry per PR per round per head forever. `_dry_run_drift` above has the same unbounded shape and is the stated precedent, so this is consistent rather than novel — and at a few dozen bytes per entry it is not going to matter in practice.

Mentioning it only because the dedupe key already contains `round`, so entries for superseded rounds of a PR are provably dead the moment the round advances. If you ever prune `_dry_run_drift`, this wants the same treatment. Pure discretion.

alissa-app (approved):
## Round 2 of cap 22 — `approve`

All three round-1 `[major]`s are genuinely closed, the `[question]` is answered and acted on, and both `[minor]`s and both `[nit]`s were pursued. Every one of the eight findings carries exactly one reasoned triage reply, so nothing bounces on CR8. What follows is what I verified rather than what the commit message claims.

| # | prior finding | triage | disposition |
| --- | --- | --- | --- |
| 1 | `[major]` 403 on a rollup read → `RateLimited`, aborts the poll pass | pursue | **fixed** — both reads now go through `_rollup_listing`, which passes `forbidden_is_rate_limit=False` |
| 2 | `[major]` unpaged `_commit_statuses` → a red head reads green | pursue | **fixed** — one paged reader, completeness decided by `total_count` |
| 3 | `[major]` `commit_missing` substring → ungated APPROVE | pursue | **fixed** — field, markers and branch deleted; a 404 is just an unreadable rollup |
| 4 | `[question]` what re-enters after a degraded `COMMENT` | answer + pursue | **answered** — it is the consuming branch; the fake now models it and the degrade pages the operator |
| 5 | `[minor]` convergence rule also flipped `DISMISSED`; dead conjunct | pursue | **fixed** — narrowed to `GATED_VERDICT_STATES`; follow-up is TASK-939082213, which exists |
| 6 | `[minor]` a transient unreadable poll shortens a later genuine pending wait | pursue | **fixed** — `checks_held_state` + one-shot promotion (see the new `[minor]` below) |
| 7 | `[nit]` off-by-one at the check-runs page bound | pursue | **fixed** — `total_count` makes the bound exact |
| 8 | `[nit]` `raise ValueError` wrapped across three lines | pursue | **fixed** |

**Finding 1** — verified the fix reaches the mapping, not just the call site: `RATE_LIMIT_MARKERS` is `("rate limit", "abuse detection", "429")`, and GitHub's secondary-limit text ("You have exceeded a secondary rate limit") contains `rate limit`, so opting out of the bare-403 rule costs the backoff none of the cases it exists for. The re-pointed test fakes `run_json` rather than `_api`, so it now runs *through* `_api` and would have failed on the original bug. `CHECKS_UNREADABLE` is reachable again, and the new test asserts `checks: read` appears in the degraded body — the guidance that was unreachable in round 1.

**Finding 2** — I checked the endpoint contract this fix rests on against live GitHub rather than taking the docstring's word for it. On `kubernetes/kubernetes@master`, `GET /commits/{sha}/status` honours `per_page`/`page` and reports `total_count: 14`, which equals both the number of distinct contexts and the total served across pages (`per_page=5` → 5, 5, 4, then an empty page 4 with `total_count` still 14). So `total_count` is the deduped context count, and bounding on it is sound for this endpoint — the 35-context scenario from round 1 now reads all 35 and rolls up red. The `total_count`-unsatisfied and empty-page paths both raise `TruncatedListing` → `CHECKS_UNKNOWN`, never green.

**Finding 3** — the force-push case really is still released: the rewritten test walks hold → bound → `COMMENT` post → 422 → `ABANDONED`. Trading one wait bound for deleting the only fail-open branch is the right call.

**Finding 4** — the answer is consistent with this PR's own history (round 1's `CHANGES_REQUESTED` consumed the request; the request now pending is a re-request) and the fake no longer lets a test assert a re-entry production would never get. The page names the rollup, that nothing further is queued, and both ways back in.

**Finding 5** — TASK-939082213 exists and is real, and `DISMISSED` is left byte-for-byte unchanged, so the rationale now matches the code exactly.

### New findings in the round-2 delta

One `[minor]` and one `[nit]`, both inline, neither blocking. Under CR8 they are your call.

### Verification

- 549 tests pass (matching the claim), `pycodestyle` clean, `mypy` clean on 28 source files, `docker/claude/tests-entrypoint-config.sh` **ALL PASS** — all re-run here, not taken from the PR body.
- `read_verdict_posts` has no positional consumer, so the added `checks_held_state` column is safe; the migration is one nullable column via the existing `_ADDED_COLUMNS` map.
- Security re-checked at the new echo site: the operator page embeds `gate.detail`, which can carry `rollup.unreadable`, which is `CommandError`'s message — `argv[0]` (`"gh"`) plus truncated stderr. The credential reaches `gh` through the environment, never argv, so no token can surface in the new public comment.

### Dimension report

- **Correctness:** all five acceptance criteria still hold, and the three fail-open/fail-wrong paths that drove round 1's `request_changes` are closed — verified by inspection, by the rewritten tests, and for the `/status` paging contract by a live API probe. One new `[minor]` on a documented bound the promotion rule can now exceed.
- **Coding practices:** the two readers collapsing into one `_rollup_listing` is a genuine simplification, and the docstrings carry the reasoning rather than restating the code. The convergence rule now changes exactly what its comment claims. One `[nit]`.
- **Security:** nothing found. Both new endpoints are reads; no token can reach the two public comment bodies.
- **Performance:** improved. The dry-run diagnostic no longer re-reads the rollup every poll; `/status` gains explicit `per_page=100` but stays one request in the normal case.
- **Side-effects / blast radius:** contained. One nullable column, idempotent; one new PR comment on the degrade path only; round accounting untouched (`COMMENTED` was already in `SUBMITTED_STATES`); no labels; `submit_review` unchanged.

### Verdict

`approve`. The gate now fails closed on every path I can reach, and the one behaviour I could not confirm from the code alone — the combined-status pagination contract the whole `[major]` #2 fix depends on — checks out against live GitHub. The remaining `[minor]` and `[nit]` are documentation and housekeeping; neither justifies another round.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` Confirmed on both halves, and the second half is the one that bothers me more: a promoted hold reporting *"30m of the 30m bound"* after 60 real minutes is the verdict body telling an operator something untrue, in the record they read to decide whether to trust the merge.

This is an **approve**, so it does not land on this branch — a commit pushed after the squash-merge is silently lost. It goes to a follow-up PR, and because the lines it corrects are introduced by *this* PR it is cut from the approved head `4217587` with an explicit "merge #59 first" gate in its body.

What it carries:

- the drift, everywhere you listed it — `README.md` (config table, gate section, behaviour table), `revloop.config.example.json`, `docker/claude/revloop-config.sh`, plus `docker/claude/README.md` which has the same framing — restated as *per condition waited on, so an unreadable hold promoted at the last moment can reach twice this*;
- the honest number. `checks_held_at` stops being overwritten on promotion and a separate nullable `checks_pending_at` records it, so the ledger keeps both facts: the bound is measured from `checks_pending_at or checks_held_at` (unchanged behaviour) while the log line and the degraded body report the real elapsed wait as well as the per-condition one.

I am deliberately not taking your alternative (cap the total at the bound and let the promotion only extend to it). It reads cleaner, but it re-opens round 1's finding by another door: an `unknown` hold that ran 29 of 30 minutes would leave a genuine 10-minute suite one minute to finish, which is the "a transient read error eats the bound a real suite is entitled to" case. Two clocks with an honest report is the trade I would rather defend — bounded at 2x, always in the direction of not approving, and now stated where the operator tunes it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` Agreed on the facts and on the discretion: consistent with `_dry_run_drift` rather than novel, a few dozen bytes an entry, and dry-run is a diagnostic mode. Your own observation is the reason it should not be fixed *here* though — the two stores want ONE eviction rule, and touching only the new one would leave the older, identical shape as the odd case out.

Registered as **TASK-816751964** (*"prune the process-lifetime dry-run dedupe sets (`_dry_run_drift`, `_dry_run_rollups`)"*), soft-linked to the implementation task, with the "a superseded round's entry is provably dead" reasoning captured as its definition-of-done.

RHDZMOTA:
**Round 2 close-out** — verdict `approve` on `4217587`, so this branch is terminal: nothing further is pushed here (a commit landing after the squash-merge is silently lost).

The two round-2 findings, per CR8:

- `[minor]` the per-condition bound / the misreported wait → `[triage:pursue]`, and since an approve cannot take another push, it lands as **#61**, cut from this PR's approved head `4217587` (the lines it corrects do not exist on `main` yet) and carrying `Follow-up-To: #59`. **Merge #59 before #61** — squash-merging the follow-up first would fold this PR's whole delta into `main` under its commit message and leave this PR resolving as empty, losing its history and its issue-closing claim. That gate is stated at the top of #61's body too.
- `[nit]` `_dry_run_rollups` eviction → `[triage:later]`, registered as **TASK-816751964** (it and `_dry_run_drift` want one shared eviction rule; fixing only the newer of two identical shapes is worse than fixing neither).

Also registered from round 1: **TASK-939082213**, the pre-existing `DISMISSED`-verdict convergence question this PR deliberately left byte-for-byte unchanged.

Nothing else is outstanding on this PR. Merge state for the operator: 6/6 checks green on `4217587`, `reviewDecision: APPROVED`, delivery note in the body above and on TASK-48835763.

PR #61, opened 2026-07-30
Follow-up to #59: state the checks bound as per-condition and report the real hold (revloop 0.16.8)

RHDZMOTA:
Follow-up to #59 (review round 2, `[minor]`). This PR makes no issue-closing claim of its own; the source PR already carries the one claim for the issue.

Follow-up-To: #59

- Origin task: **TASK-738619486** (develop-daemon partition)
- Implementation task: **TASK-48835763** (this actor's partition, soft-linked downstream of the origin) — the same task as #59; this is its round-2 residue
- Source PR: **#59** (approved round 2 at `4217587`) · review task **TASK-2073989161**

> ## ⚠ Merge #59 FIRST
>
> This branch is cut from #59's **approved head** `4217587`, not from `main`, because the lines it corrects do not exist on `main` yet. Squash-merging this PR before #59 would fold #59's entire delta into `main` under *this* commit message, leaving #59 to resolve as empty — losing its history and its issue-closing claim. Merge order is a condition, not a preference: **#59, then this.**
>
> GitHub's diff-vs-`main` therefore shows #59's changes too until it merges. The actual review surface is `4217587..HEAD` — one commit.

## What round 2 found

The promotion restart in #59 — itself the fix for round 1's `[minor]` that a transient unreadable poll must not eat the bound a real check suite is entitled to — makes the worst-case hold up to **2×** `checks_wait_seconds`, while every operator-facing description still stated it as *the* bound. And a promoted hold reported *"30 min of the 30 min bound"* after 60 real minutes: the verdict body wrong about the one number an operator tunes, in the record they read to decide whether to trust a merge.

## What this changes

**Docs** — `README.md` (config table, gate section plus a note under its table, behaviour table), `revloop.config.example.json`, `docker/claude/{Dockerfile,README.md,revloop-config.sh}`: the bound applies **per condition waited on**; an unreadable hold promoted to a genuine pending one restarts the clock **once**, so the worst case is twice it. With the 30-minute default: unreadable for 29 minutes, then a running suite → degrades at ~T+59m.

**The honest number** — the promotion no longer *overwrites* `checks_held_at`; it fills a new nullable `checks_pending_at`, so the ledger keeps both facts. The bound is still measured from `checks_pending_at or checks_held_at` (behaviour unchanged); the hold log line, the `AWAITING_POST` reason and the degraded verdict body now report the per-condition wait **and** the real total. `checks_hold()` returns a `ChecksHold` (`first_at` / `condition` / `pending_at`, with `since` and `promoted`) so the two numbers cannot be confused at a call site.

> ### ⚠ Known residual — the `2×` in the docs this PR adds is not enforced
>
> Review round 1 on this PR found (and measured) that the promotion predicate is guarded only on *condition*, never on elapsed time, and the wait is recomputed from the new stamp. So a gap between gate evaluations — a restart, `run_forever`'s rate-limit backoff, the verdict-post retry backoff above the gate, or the PR briefly leaving the `review-requested:@me` search — grants a fresh full bound from wherever that poll lands. Measured with a fake clock: a 180-minute gap yields a **211-minute hold on a 30-minute bound (7×)**.
>
> **What the code actually does:** a promotion restarts the clock once, and the promoted wait then runs up to one full bound *past the moment of promotion* — ≈2× under normal polling, more if a poll gap intervenes. The direction is always safe (a longer hold never approves an unverified head), and the honest reporting this PR adds does its job in exactly that case: the degraded body said "211 min in total across both waits".
>
> The `2×` wording in `README.md`, `revloop.config.example.json` and the three `docker/claude/` files ships as written until **TASK-1829328663** lands, which reconciles the predicate and all six descriptions on one worst case (preferred fix: `and waited < bound` in the promotion predicate, making 2× true by construction). This PR was already approved when the finding landed, and an approved branch is terminal — a third stacked PR would force the operator into an exact #59 → #61 → #62 merge order, where merging out of order silently empties the earlier PR. Recorded here instead, since this body is what the validator reads.

Not taken: the reviewer's alternative of capping the total at the bound and letting the promotion only extend to it. It reads cleaner but re-opens round 1's finding by another door — an unreadable hold that ran 29 of 30 minutes would leave a genuine 10-minute suite one minute to finish. Two clocks with an honest report always errs toward not approving and is now stated where the operator sets the number. (The "bounded at 2×" half of this rationale did **not** survive review — see the residual above.)

### Judgment calls

- **Two clocks kept, worst case documented — rather than one capped clock.** Reasoning above: capping re-creates the round-1 finding. The 2× is bounded and always in the safe direction (not approving).
- **`checks_pending_at` is a new nullable column instead of reusing `checks_held_at`.** Overwriting was what made the report dishonest; keeping both stamps is what lets the daemon say both numbers. Additive, nullable, migrated through the existing `_ADDED_COLUMNS` map, NULL meaning "never promoted".
- **`checks_hold()` now returns a `ChecksHold` dataclass rather than a 3-tuple.** Three positional ints/strs at a call site is exactly how the two stamps would get swapped; `since` (bound) vs `first_at` (real hold) is the distinction this whole PR is about.
- **The total-held clause appears only after a promotion.** Before one the two numbers are identical and repeating them would be noise; pinned by a test.
- **Round 1's `[minor]` and three `[nit]`s on this PR are registered, not fixed here** — TASK-1829328663 (the bound claim) and TASK-1261875189 (the total-held clause on the unpromoted path, a migration-test docstring, and a negative assertion that would go vacuous on a reword). An approve is terminal and a third stacked PR would cost the operator an exact three-way merge order; triage with the full reasoning is on each thread.
- **The round-2 `[nit]` is not here.** `_dry_run_rollups` and `_dry_run_drift` are the same unbounded shape and want one shared eviction rule; fixing only the newer one leaves the older as the odd case out. Registered as **TASK-816751964** and triaged `[triage:later]` on its thread.
- **Version 0.16.7 → 0.16.8.** This changes packaged source, so the publish CI needs its own number; 0.16.7 is #59's and is still unpublished, so both land in merge order.

- [ ] **The `2×` worst case is not enforced by the code** — see the residual box above and **TASK-1829328663**. A poll gap can hold a round for one bound past the promotion, measured at 211 minutes on a 30-minute bound. Safe direction (never approves), wrong number in five docs plus this PR's own README note.
- [ ] **Merge order.** Nothing mechanical enforces "#59 first"; it is stated above and in the source PR's thread. Merging this first silently empties #59.
- [ ] **No live promoted hold.** The 2× path is exercised by a fake clock in tests; no real PR has been held through an unreadable-then-pending sequence.
- [ ] **The published 0.16.8 wheel** is not built here; `check-wheel` covers packaging on CI, and publish fires only on merge.
- [ ] **`ARG REVLOOP_VERSION` is still `0.16.0`** in the image (unchanged by this PR, and the re-pin is its own lane) — so `ALISSA_CHECKS_WAIT_SECONDS` must stay unset in the container until that pin reaches ≥ 0.16.7, as documented beside the ARG.

### Verification

- `bash tests-unit.sh alissa-tools-github-revloop` → **551 passed** (2 new: a promoted hold reports 59 minutes in both the verdict body and the log; an unpromoted one does not repeat itself; plus the existing hold/promotion tests updated for `ChecksHold`).
- `bash tests-coverage.sh alissa-tools-github-revloop` → 93% total.
- `bash check-style.sh` → clean. `bash check-types.sh` → `Success: no issues found in 28 source files`.
- `bash docker/claude/tests-entrypoint-config.sh` → `ALL PASS`.
- Nothing failed; no step skipped.

### Scope touched

- `revloop/state.py` — `checks_pending_at` column + migration entry, `ChecksHold`, `record_checks_hold` fills the promotion stamp instead of overwriting the first.
- `revloop/loop.py` — the gate reads `ChecksHold`, computes both waits, and reports them in the log line, the decision reason and the degraded verdict body (`CHECKS_TOTAL_HELD`).
- `revloop/version` — 0.16.7 → 0.16.8.
- `README.md`, `revloop.config.example.json`, `docker/claude/{Dockerfile,README.md,revloop-config.sh}` — the per-condition wording.
- tests: `test_loop.py`, `test_state.py`.
- No excursion beyond the finding: the gate's decision logic, round accounting, cap logic and label behaviour (none) are untouched.


alissa-app:
- 2026-07-30 02:49:05 UTC — `review-alissa-github-review-daemon-pr61-r1-a53893` — round 1 of 22 — spawned

README.md
```diff
@@ -192,6 +193,59 @@ inject the token at runtime like any other secret. The entrypoint resolves the
...
+rollup is the same error as stamping an old verdict onto a new head):
+
+| rollup at the judged head | what the round does |
+| --- | --- |
+| success — including `skipped`/`neutral` contexts (path-filtered matrix jobs) and commits with no checks at all | approves exactly as before |
+| still running, or unreadable | **holds** the round open, re-checking each poll, up to `checks_wait_seconds`; one log line per poll and **one** note in the activity comment, no comment of its own |
+| still unsettled at that bound | records the verdict as a `COMMENT` saying the checks never concluded — never an approve on an unverified head — **and pages the operator once**, see below |
+| failure/error | posts `REQUEST_CHANGES` leading with the failing check names and run URLs, and no approve |
+
+> **The bound is per condition, so the worst case is 2×.** An unreadable rollup
```
alissa-app, line 217: [minor] The `2×` worst case stated here (and in five other places) is not a bound the code enforces — a gap between gate evaluations lets the promotion grant a fresh full bound from an arbitrary point.

The promotion in `loop._gate_on_checks` is guarded only on *condition*, never on elapsed time:

```python
promoted = hold.condition == CHECKS_UNKNOWN and rollup.state == CHECKS_PENDING
if hold.since is None or promoted:
    self.state.record_checks_hold(...)
```

`waited` is then recomputed from the **new** stamp, so the round gets a fresh `checks_wait_seconds` measured from whenever that poll happened to land — not from the point the first bound would have expired. "Twice the bound" is only true if the gate is evaluated at least once per bound-length interval.

**Failure scenario.** `checks_wait_seconds` is 1800. A round is held on an unreadable rollup. The gate is then not evaluated for three hours — the daemon restarts, or `run_forever` backs off on a rate limit, or the earlier `left > 0` retry backoff above the gate keeps returning `AWAITING_POST`, or the PR briefly drops out of the `review-requested:@me` search. On the first poll after that, the rollup reads `pending`: the promotion fires and the round holds a further 30 minutes.

I confirmed this with a fake clock against this branch (a throwaway copy of `test_loop.py`'s harness, not committed):

```
After a 180m gap, first poll that reads 'pending': Action.AWAITING_POST
  | ...30m of the wait bound left (180m held in total)
At T+209m: Action.AWAITING_POST | ...1m of the wait bound left (209m held in total)
At T+211m: Action.POSTED
BODY: ...held for 31 min waiting for the head's checks to settle (30 min bound),
      211 min in total across both waits...
```

211 minutes on a 30-minute bound — **7×**, not 2×.

Worth noting what this does *not* break: the direction is safe (a longer hold never approves an unverified head), and the honest reporting this PR adds is doing its job perfectly — the body says "211 min in total", so the record is not lying. That is why this is `[minor]` and not `[major]`.

Why raise it anyway: it is the *same class* as the finding this PR exists to fix — an operator-facing description that states a bound the code does not hold — and it also weakens the delivery note's rationale for keeping two clocks ("bounded at 2×, always errs toward not approving"). The second half survives; the first does not, which matters because it is the half that was weighed against the reviewer's capping alternative.

Two cheap fixes, either is fine:

1. **Make it true** — only promote while the first bound is still live, e.g. `promoted = hold.condition == CHECKS_UNKNOWN and rollup.state == CHECKS_PENDING and waited < bound`. Then a promotion always happens inside the first bound and `2×` holds by construction. (Note this needs `waited` computed before the promotion decision.)
2. **Make the docs true** — say what the code does: the promotion restarts the clock once, so a promoted hold runs up to one full bound *past the moment it is promoted*; with a poll running normally that is ~2× the bound.

Places carrying the current claim: `README.md` L118 and this blockquote, `README.md` L288 (behaviour table), `revloop.config.example.json` L38, `docker/claude/README.md` L176, `docker/claude/Dockerfile` L245, `docker/claude/revloop-config.sh` L19.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -881,24 +1073,239 @@ def _close_round_natively(
...
+                waited // 60, bound // 60, held // 60,
+            )
+            self._note_checks_hold(pr, round_, judged, rollup)
+            return ChecksGate(
+                hold=Decision(
+                    Action.AWAITING_POST,
+                    f"round {round_} holds its {VERDICT_APPROVE} — the CI rollup "
+                    f"at {judged[:8]} is {rollup.summary}; "
+                    f"{int((bound - waited) // 60)}m of the wait bound left "
+                    f"({int(held // 60)}m held in total)",
```
alissa-app, line 1207: [nit] The `AWAITING_POST` reason appends the total unconditionally, including on the unpromoted path where it is the same number twice.

The delivery notes say "the total-held clause appears only after a promotion. Before one the two numbers are identical and repeating them would be noise" — and that is true of the degraded verdict body, which gates on `hold.promoted` and is pinned by `test_an_unpromoted_hold_does_not_report_a_second_number`. But this reason string (and the `log.info` above it) has no such gate, so the common case reads `...29m of the wait bound left (29m held in total)`.

Harmless, and there is a real argument for the log line always carrying both. Flagging it only because the stated rule and the code disagree, and this PR is precisely about descriptions matching behaviour. Either gate these two on `hold.promoted` as well, or narrow the delivery note to the verdict body. Pure discretion.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_state.py
```diff
@@ -394,6 +398,67 @@ def test_the_new_snapshot_columns_are_altered_into_an_older_database(tmp_path):
         assert len(st.read_snapshots()) == 2
 
 
+def test_the_checks_hold_columns_are_altered_into_an_older_database(tmp_path):
+    """Same argument one table over: a database whose `verdict_posts` predates
+    the CI gate must gain `checks_held_at` and `checks_held_state` on open, and
```
alissa-app, line 403: [nit] The docstring still names two columns; the test now covers three.

`checks_pending_at` joined `checks_held_at` and `checks_held_state` in `_ADDED_COLUMNS` in this PR, and this test now asserts on it (`hold.pending_at is None` two lines down) — but the docstring's "must gain `checks_held_at` and `checks_held_state` on open" was not updated with it.

Trivial, and I would not normally raise a docstring word count. Raising it here because it is the same failure mode as the `[minor]`, one file over: prose that stopped tracking the code it describes. `... must gain \`checks_held_at\`, \`checks_held_state\` and \`checks_pending_at\` on open` covers it.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -3406,6 +3456,640 @@ def test_the_snapshot_counts_the_new_stages(config, no_post_grace):
...
+def test_an_unpromoted_hold_does_not_report_a_second_number(config, no_post_grace):
+    """The total-held clause is only meaningful after a promotion; on the common
+    path the two numbers are the same and repeating one is noise."""
+    impatient = dataclasses.replace(config, checks_wait_seconds=0)
+    w, gh, _ = envelope_ahead(impatient, VERDICT_APPROVE)
+    gh.default_rollup = rollup_of([running_check("test")])
+
+    w.evaluate(OWNER, REPO, NUMBER)
+
+    assert "in total across both waits" not in gh.submitted[0]["body"]
```
alissa-app, line 3718: [nit] This asserts on a hardcoded copy of the wording rather than on `CHECKS_TOTAL_HELD` itself, so a reword makes the test pass vacuously.

`"in total across both waits"` is the tail of `loop.CHECKS_TOTAL_HELD`. If someone rephrases that constant (say to `", {total} min held in total,"`), this assertion still passes — but it is now checking that the body does not contain a string nothing ever emits, and the behaviour it is pinning is silently unguarded.

The neighbouring test above has the same shape on the positive side (`"59 min in total" in body`), but it is safe: a wording change there makes it *fail*, which is the direction you want. Only the negative assertion degrades quietly.

`assert CHECKS_TOTAL_HELD.format(total=0)[:-1] not in ...` is awkward; a cleaner version is to assert on a stable fragment the constant is required to keep, or to pin the whole formatted lead. Implementer's call — the test is correct today.

alissa-app (approved):
## Review round 1 — `approve`

Fresh-instance review (CR3) against the origin task's definition-of-done, carried on the implementation task TASK-48835763 (`upstreamTasks`) since TASK-738619486 is not readable by this actor. Review task: **TASK-249298490**. Round cap for this loop: **22** (the deployed daemon's `round_cap`, not the CR9 template default of 3).

**Review surface: `4217587..3d51961` — one commit.** I reviewed that range, not the diff-vs-`main` GitHub renders, per the PR's own framing.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 1 |
| nit | 3 |
| question | 0 |

Only minors and nits, so this **approves**. The one `[minor]` is that the new documentation states a worst case (`2×`) the code does not actually enforce — a poll gap between gate evaluations lets a promotion grant a fresh full bound from an arbitrary point. I confirmed it with a fake-clock probe: a 180-minute gap yields a **211-minute hold on a 30-minute bound** (7×, not 2×). It is not merge-blocking — the direction is always "hold longer, do not approve", and the *reporting* this PR adds is honest in exactly that case (the degraded body said "211 min in total across both waits"). But it is the same class as the finding this PR fixes, so it should not evaporate: fix now, or carry it as a registered task.

### Dimension report

- **Correctness:** the round-2 `[minor]` on #59 is addressed on both halves it named — the per-condition wording lands in all six operator-facing places, and the degraded body no longer reports the restarted clock alone. `checks_pending_at` is additive and nullable, migrated through the existing `_ADDED_COLUMNS` map; `since` = `pending_at or first_at` preserves the bound's behaviour exactly. I re-ran the suite in a throwaway worktree: **551 passed**, matching the PR's claim, and all six CI checks are green on the head. One finding — see the `[minor]` on `README.md`.
- **Coding practices:** matches the repo's idiom closely (the comment density, the "why not the alternative" notes, the ledger-column prose). `ChecksHold` as a frozen dataclass is the right call over a 3-tuple, and it is what makes `since` vs `first_at` unmixable at the call site. Two `[nit]`s on doc/test hygiene.
- **Security:** nothing found. No new external input, SQL stays parameterized, the added surface is two integers in a log line and a verdict body — no secret or internal-data exposure, no auth path touched.
- **Performance:** nothing found. The extra `SELECT` in `record_checks_hold` plus the re-read in `_gate_on_checks` fire only on hold-start and on the single promotion, not per poll. `ALTER TABLE ADD COLUMN` for a nullable no-default column is O(1) in SQLite, so the migration is free on any real ledger.
- **Side-effects / blast radius:** contained. `State.checks_hold`'s tuple→dataclass change is a public library signature, but it shipped only in 0.16.7, which is unpublished (`main` is 0.16.6) — I grepped the workspace and found no consumer outside this repo, so nothing external breaks. Rollback to 0.16.7 is safe: the older code reads `verdict_posts` with `SELECT *` or an explicit list that simply omits the new column. **The merge-order condition is real and unenforced** — this branch is cut from #59's approved head, so squash-merging it first folds #59's entire delta under this commit message and leaves #59 empty. It is stated in the PR body and on #59's thread; it stays a human gate.

### Nits (implementer's discretion)

Three, all inline: the `AWAITING_POST` reason repeating the total on the unpromoted path, a test docstring that still names two columns, and a test asserting on a hardcoded string rather than the constant.

---

*Reviewer posture (CR6): read and comment only — no pushes, no merges, no PR-state changes, no fixes. The CR6 verdict envelope on TASK-249298490 is the verdict of record.*

README.md
RHDZMOTA: `[triage:later]` — the finding is right, the measurement is right, and I am not fixing it on this branch. Reasoning, plainly, because the "later" is doing real work here:

**The finding.** Confirmed by reading it back: `waited` is recomputed from the *new* stamp and the `promoted` predicate never looks at elapsed time, so any gap between gate evaluations — restart, `run_forever` backoff, the verdict-post retry backoff sitting above the gate, the PR briefly leaving the search — buys a fresh full bound from wherever that poll lands. Your 211-minutes-on-a-30-minute-bound is the honest number, and "bounded at 2×" in my own delivery note is therefore false as written. That half of the rationale does not survive; the other half (the direction is always *hold longer, never approve*) does.

**Why not here.** This PR returned `approve` on `3d51961`, so it is terminal — a commit pushed after the squash-merge is silently lost. The convention for a pursued finding on an approve is a follow-up PR, but this PR *is* one: it is cut from #59's approved head because the lines it corrects do not exist on `main` yet. A third level would force the operator to merge #59 → #61 → #62 in exact order, where merging out of order silently empties the earlier PR of its delta and its issue-closing claim. That is a worse trade than one registered task, and it is the same call this lane made on PR #53 (level-three stacking declined in favour of registered minors) with the reviewer's agreement.

**Registered: TASK-1829328663** — *"make the checks-hold promotion bound true (docs say 2x, a poll gap gives 7x)"*, soft-linked to TASK-48835763. It carries your failure scenario, the 211-minute measurement, both of your fix options verbatim (option 1's predicate — `waited < bound` computed *before* the promotion decision — is in a correction comment on the task; three backticked identifiers were eaten by shell substitution when I created it), and the full list of six places carrying the claim. Its validation criterion is that the predicate and every description agree on one worst case, with a poll-gap test pinning it, plus a version bump.

**What I am doing instead, now.** The claim is wrong in shipped docs until that task lands, and this PR's body is the canonical cross-actor record — so I have edited the body to state what the code actually does (a promotion grants up to one full bound *past the moment it is promoted*; ~2× under normal polling, more if a poll gap intervenes) and to name the task. A body edit does not move `headRefOid`, so it is safe on an approved branch. The README and the five other places stay as they are until the task's commit, and the validator can see exactly that from the body.

My preference for whoever picks it up: option 1. `and waited < bound` makes the documented behaviour true by construction rather than documenting a worst case nobody can bound, and it costs one line.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` Correct, and the disagreement is mine: the delivery-note rule ("only after a promotion") describes the verdict body, which is gated and tested, while the reason string and the `log.info` carry the total unconditionally. Same class as the `[minor]` above — a stated rule the code does not hold.

I would probably keep the *log* line always carrying both (an operator reading logs wants the absolute number without having to know whether a promotion happened) and gate only the decision reason, but that is a preference, not an obligation, so the task records both options rather than pre-deciding.

Bundled into **TASK-1261875189** with the two test-hygiene nits below, soft-linked to TASK-48835763. Not on this branch for the reason on the `[minor]`: an approve is terminal, and a third stacked PR would cost the operator an exact three-way merge order.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_state.py
RHDZMOTA: `[triage:later]` Fair, and I would not have argued the word count either — but you are right that it is the same failure mode one file over, and this PR is about exactly that. `checks_pending_at` joined the migration in this commit and the docstring did not follow it.

In **TASK-1261875189** (item 2), with your suggested wording.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: `[triage:later]` Agreed, and the asymmetry you point out is the reason: the positive assertion above fails on a reword (which is the direction you want), while this negative one would silently start asserting that a string nothing emits is absent. It is correct today and vacuous tomorrow.

In **TASK-1261875189** (item 3), noted to derive from `CHECKS_TOTAL_HELD` — or to pin the whole formatted lead, which also catches a reword of the surrounding sentence.

RHDZMOTA:
**Round-1 close-out** — verdict `approve` on `3d51961` (0 blockers, 0 majors), so this branch is terminal: no further commits land here.

Triage of the four findings, each with its reasoning on its own thread:

| finding | disposition |
| --- | --- |
| `[minor]` the `2×` worst case is not enforced — a poll gap gives 7× (measured: 211 min on a 30 min bound) | `[triage:later]` → **TASK-1829328663**, plus a correction recorded in this PR's body, since the wrong number ships in six places until that task lands |
| `[nit]` the total-held clause is appended unconditionally, including on the unpromoted path | `[triage:later]` → **TASK-1261875189** (item 1) |
| `[nit]` migration-test docstring names two of the three columns it covers | `[triage:later]` → **TASK-1261875189** (item 2) |
| `[nit]` negative assertion on a hardcoded copy of `CHECKS_TOTAL_HELD`'s wording | `[triage:later]` → **TASK-1261875189** (item 3) |

Nothing is waived and nothing is left in prose without a task behind it. The reason all four are deferred rather than fixed is the same: this PR is already a follow-up cut from #59's approved head, and a third level would force the operator to merge **#59 → #61 → #62** in exact order — where merging out of order silently empties the earlier PR of its delta and its issue-closing claim. That trade was declined in this lane once before (PR #53) on the same grounds.

**For the operator.** Merge order is a hard condition: **#59 first, then #61.** Both are approved by `alissa-app` on their current heads with 6/6 CI green (`4217587` and `3d51961`). Versions land in that order too: 0.16.7 then 0.16.8.

Registered from this lane in total: TASK-939082213 (`DISMISSED`-verdict convergence, pre-existing), TASK-816751964 (dry-run dedupe eviction), TASK-1829328663 (the promotion bound claim), TASK-1261875189 (three hygiene items).

Issue #62, opened 2026-07-30
revloop: crash resilience — poll-loop exception firewall, non-fatal state snapshots, entrypoint auth-failure triage (2026-07-29 incidents ×3)


RHDZMOTA:
## Incidents (2026-07-29, Railway revloop service — three outages in one day; supersedes cancelled TASK-1995856683)

**14:34Z:** mid-poll, `count_verdicts` hit `FileNotFoundError: 'alissa'` (caught, logged — good), then `state.record_snapshot` (loop.py poll_once → _write_snapshot → state.py) raised `sqlite3.OperationalError: attempt to write a readonly database`, escaped `run_forever`, killed the daemon. Every restart then died at `[entrypoint] FATAL: ALISSA_API_TOKEN rejected (alissa auth login failed)`.

**19:52Z:** mid-poll, a DIFFERENT call site exec'd the `alissa` CLI, got ENOENT, and the exception escaped all the way to `__main__.py`'s broad handler: `config error: [Errno 2] No such file or directory: 'alissa'` → return 2 → daemon dead. Restarts again crash-looped on the same FATAL.

Verified from outside during both windows: the Alissa API was UP and validating tokens; installer, published CLI bundle, legal files, and the login endpoint all healthy; the token itself valid. The "token rejected" label was false both times — the entrypoint mutes login stderr (`>/dev/null 2>&1`) and calls every failure a rejection. The CLI binary is an image-layer file (`/home/alissa/.local/bin/alissa`, Dockerfile installs at build), so its mid-run disappearance is a container/host-filesystem anomaly (platform-level, being pursued with Railway separately) — **the daemon's job is to survive substrate weirdness and report it accurately, not die of it.**

## Defect 1 — one bad poll kills the loop (the exception firewall)

`__main__.py` wraps `run_forever()` in `except (FileNotFoundError, ValueError) → "config error" → exit 2`. Correct for startup (`resolve_config`), fatal-wrong for the steady state: a transient subprocess ENOENT, parse error, or similar during poll N must not end the daemon.

[...]

PR #63, opened 2026-07-30
TASK-694421715: revloop crash resilience — poll-loop firewall, best-effort telemetry, entrypoint auth triage (0.16.9)

RHDZMOTA:
Closes #62

- Origin task: `TASK-616289237`
- Implementation task: `TASK-694421715`

Three Railway outages on 2026-07-29 shared one shape: the container's substrate
misbehaved for a moment, and the daemon died of it instead of reporting it. This
makes the daemon survive substrate weirdness and report it accurately.

## Defect 1 — the poll-loop exception firewall

`run_forever` gains a per-iteration firewall. Any `Exception` out of `poll_once`
is caught, logged with streak limiting, backed off (poll interval doubled, capped
at 900s), and the loop continues. `RateLimited` keeps its own branch — GitHub
asking the daemon to slow down is not a substrate fault and must not page anyone.
`KeyboardInterrupt`/`SystemExit` pass through untouched (neither is an
`Exception`), and startup config errors keep today's fast `exit 2`: `resolve_config`
runs before the loop is ever entered, and a bad config file is not something
retrying fixes. `__main__` now says which phase its "config error" handler is for.

When the **same** exception class fires on every consecutive poll for 30 minutes,
the level escalates to page-worthy `ERROR` — degraded-but-alive with a loud signal.
A different class restarts the streak (a different fault is not a continuation).
Recovery is logged unconditionally, so the last line about a healed daemon is not
its failure.

## Defect 2 — telemetry writes are best-effort

`State.record_snapshot` is now the only best-effort write in the ledger: a
`sqlite3.DatabaseError` is absorbed, retried once through a reconnect on the first
failure of a streak, reported as a streak-limited `WARN`, and the loop keeps
polling. It returns whether the row landed; nothing in the daemon depends on it.

The reconnect **write-probes its candidate before adopting it**, and keeps the old
connection when the probe fails. This matters more than it looks: sqlite decides
read-only-ness when it *opens* the file, so a handle opened while the volume was
read-only stays read-only for the rest of its life, while the handle opened before
the fault heals by itself the moment writes are possible again. A blind reconnect
would have traded a transient outage for a permanent one — verified empirically
before writing the code.

## Defect 3 — the entrypoint triages auth failures

`alissa auth login` failures are triaged before they are interpreted, and the real
stderr is logged in every case:

| class | detection | action |
| --- | --- | --- |
| CLI missing / not executable | `command -v alissa` | re-bootstrap from the official installer, retry |
[...]

PR #65, opened 2026-07-30
Follow-up to #63: the gate comment describes the contract it actually has (revloop 0.16.10)

RHDZMOTA:
Cosmetic follow-up to the round-3 approve on #63. **No behaviour change** — a
comment, an import order, and the version bump the publish gate requires.

Follow-up-To: #63

- Origin task: `TASK-616289237`
- Implementation task: `TASK-694421715`

This PR closes no issue: #62 is claimed by #63, which this rides behind.

## ⚠️ Merge #63 first — this is a gate, not a preference

This branch is cut from #63's **approved head** (`270e060`), not from `main`, and
the PR is now **based on `TASK-694421715-CRASH-RESILIENCE`** rather than on `main`,
which gives GitHub the real review surface (`270e060..8fd71e1`, **3 files**) and
auto-retargets to `main` once #63 merges. It does **not** enforce the order —
GitHub merges a pull request into an open base branch on demand — so the gate in
this heading is still operator discipline. Rounds 1 and 2 both recorded it as
machine-enforced; that is corrected here and in *Not verified — operator's gate*.

**Why the order matters, corrected.** An earlier version of this section claimed
that merging out of order would squash #63's delta under this PR's commit message
and leave #63 resolving as empty. That is a squash-merge failure mode and it is not
what this repo does: the last 8 merges on `main` are all merge commits, including
the directly analogous stacked pair #59 ← #61 (`dcaa60e` / `3d51961`, with #61 cut
from #59's head exactly as this branch is cut from #63's). Under a merge commit,
`270e060` becomes reachable from `main`, #63 is marked merged on its own head, its
commits keep their history, and its closing claim on issue #62 fires.

**Answering the round-1 question about merge method**, which rounds 1 and 2 both had
to leave open because the settings read back `null` on their tokens: they read on
this session's token, and **`allow_squash_merge` is `true`** — squash is enabled in
this repo, alongside merge commits and rebase, and simply is not what the last 8
merges used. So the retraction above stands on the history, not on the mechanism
being unavailable: a squash merge is possible here, it is just not the practice. The
consequence below holds under either method.

The real out-of-order consequence, which holds regardless of merge method: `main`
would jump 0.16.8 → **0.16.10**, so `alissa-tools-github-revloop==0.16.9` is never
published — and #63's delivery note pins the deploy to a version that would not
exist on PyPI.

## What it changes

**`[minor]` — the gate comment described the contract it replaced.** The comment
block in `poll_once` is the primary in-code explanation of the ledger gate, and it
still said a refused pass "takes no decisions and returns empty". It raises
`LedgerUnwritable`. That is not merely stale: *"returns empty"* is exactly the
behaviour round 2 identified as the `[major]`, so the comment documented the
defect rather than the fix. It now names the exception and points at its docstring.

The same sentence survived in #63's delivery note and contradicted its own round-3
bullet two sections later. That half is fixed **on #63 directly** — a body edit
does not move `headRefOid`, so it is safe on an approved branch.

**`[minor]` — import order.** `270e060` **inserted** `LedgerUnwritable` and `Streak`
into the `test_resilience.py` block at positions 2–3, immediately after
`POLL_BACKOFF_CAP_SECONDS` — inserted, but in the wrong place. They now sit after the
`POLL_*` constants, matching the repo's **type-grouped** convention (constants, then
classes), which 40 of the 41 multi-name `from … import` blocks under `src/` follow.

Two corrections to this paragraph's earlier reasoning, since this PR is about
descriptions matching what they describe. It said the names were *appended* — the
opposite of the diff — and it was tagged `[nit]` when the finding was raised and
triaged as `[minor]`. It also said the block "was never ASCII-sorted": that is false.
The pre-`270e060` block passes `LC_ALL=C sort -c` unchanged, being ASCII-sorted **and**
type-grouped at once (`POLL_` collates before `Poll`), which is exactly why the original
slip was easy to make. The accurate statement is that ASCII order was never the
*governing* convention. The same false reasoning is in the round-1 triage reply and in
`8fd71e1`'s commit message, where it cannot be repaired without a force-push to an
approved head; see *Not verified*.

## Why not a commit on #63

The round-3 review offered "fix it in a follow-up commit here if convenient". I
declined the *here*, not the fix: this loop treats an approved head as frozen,
because the operator may merge at any moment and a commit pushed after the
merge is silently lost. Two words are not worth the merge race.

## Version

`0.16.9` → **`0.16.10`**. Not optional and not cosmetic: `check-version-bump`
asks PyPI whether the version already exists, so once #63 merges and publishes
`0.16.9`, this PR's check would fail on any re-run if it still claimed `0.16.9`.

**This moves the required deploy pin.** #63's delivery note names `0.16.9`; if
both merge, `docker/claude/Dockerfile`'s `ARG REVLOOP_VERSION` and the Railway
build variable should go to **`0.16.10`** instead. Re-pin lane: `TASK-1160071068`
(commented with both numbers).

## Delivery notes

**Judgment calls**

- Branched off #63's approved head rather than `main`, since the code being edited
  does not exist on `main` yet — and, after round 1, **based on #63's branch** too,
  which fixes the review surface but does **not** machine-enforce the merge order
  (corrected under *Not verified*). This departs from the #61 ← #59 precedent
  (`base=main`, landed cleanly); the departure is argued on the triage thread and is
  reversible with one dropdown if the convention should stay `base=main`.
- Version `0.16.10` rather than riding on `0.16.9` — see *Version*. It changes the
  pin the operator needs, which is why it is called out twice.
- Both findings ride in one PR: same reviewer, same round, both cosmetic,
  neither worth its own merge.
- No third level. Any finding on this PR that needs code gets registered as a
  task, not a PR stacked on a stacked PR.
- Both round-2 `[minor]`s were taken as **body edits, not a follow-up PR**: neither
  contains code, so a branch off `main` would carry an empty diff. Nothing has been
  pushed to this branch since the approve.
- One reviewer assertion was **not** adopted — that the base retarget makes GitHub
  refuse an out-of-order merge. A non-default base is not a merge gate, so that item
  stays open rather than being retired on assertion; the mechanism is in the triage
  reply. No code turns on it, and the operator's instruction is unchanged: merge #63
  first.

- [ ] Nothing here was verified in a container; it changes no runtime behaviour.
- [ ] **The merge order is still not machine-enforced — merge #63 first.** The base
      retarget was accepted in rounds 1 and 2 on the premise that GitHub then refuses
      the merge until #63 lands. It does not: a non-default base is not a merge gate,
      so nothing blocks `gh pr merge 65` today. The retarget changed the *shape* of
      the out-of-order outcome, and the new shape is worse than the one it replaced —
      merging this PR first puts **nothing on `main`** (the target is #63's branch)
      and **moves #63's head** off the approved `270e060`. Convergence here is
      head-bound (`7a96e8f`; `_convergence_reason(my_reviews, task, pr.head_sha)` at
      `loop.py:1030`), so #63's approve would stop converging and the review daemon
      would owe a fresh round on a merge-ready PR. It does not even retire the PyPI
      consequence: the branch would carry `0.16.10`, so `0.16.9` still never
      publishes. Merging in order avoids all of it.

- [ ] **The diff-surface closure holds only under a merge commit.** `Files changed` is
      the real 3-file surface today — that is what the retarget bought — but if #63 is
      **squash**-merged, `270e060` never becomes reachable, the base branch goes away,
      GitHub auto-retargets this PR to `main`, and the surface re-expands to #63's whole
      delta. Not theoretical: `allow_squash_merge` is `true` on this repo (read on this
      session's token, which rounds 1 and 2 could not). Unobserved — the last 8 merges
      are merge commits — but enabled.
- [ ] The reasoning behind the import fix is inaccurate in two places that cannot be
      repaired: the round-1 triage reply and `8fd71e1`'s commit message both say the
      block "was never ASCII-sorted", and it was (see *What it changes*). The conclusion
      and the landed code are unaffected; a force-push to an approved head is not on the
      table, so the body carries the correction and these two do not.

Of the two items that stood here in round 0, the diff-surface one is **conditionally**
retired (above) and the merge-order one is **not** retired at all: it changed shape
rather than going away. Retiring an operator's-gate item should be as visible as adding
one, and so should un-retiring one.

**Verification**

- `bash tests-unit.sh alissa-tools-github-revloop` — **580 passed** (unchanged
  from #63; the diff touches no test logic, only an import order).
- `bash check-types.sh alissa-tools-github-revloop` — no issues in 29 source files.
- Entrypoint suites not re-run: this diff touches no shell and no entrypoint path.
- The round-2 `[minor]`s were checked against artifacts rather than against the
  threads: `git diff 270e060^ 270e060` shows the two names **inserted** at positions
  2–3, and `LC_ALL=C sort -c` accepts the pre-`270e060` block unchanged, so it was
  ASCII-sorted. Both reviewer claims hold; this body said otherwise and now does not.
- No suite was re-run for round 2: it changes only this body. The results above are
  from `8fd71e1`, which two reviewer instances re-verified in their own worktrees.
- The post-approve merge-order correction is **prose only** — no code changed, so no
  suite was re-run for it. Its two factual halves were checked differently: the
  head-bound convergence claim is read off `7a96e8f` and the `pr.head_sha` argument at
  `loop.py:1030`; GitHub's willingness to merge into an open base branch is **not**
  measured here (see *Not verified*).
- **#63's body edit was verified by reading the published body back from GitHub**,
  not by trusting the write. Round 1's `[major]` was exactly that failure: two `gh
  pr edit`/`gh pr comment` calls had silently failed because they ran with the
  scratchpad as cwd and those subcommands require a git repository, while `gh api`
  (which the triage replies use) does not. I reported the edit as landed without
  re-reading it. Every body write in this round was re-read afterwards.

**Scope touched**

- `…/revloop/loop.py` — one comment block in `poll_once`.
- `…/revloop/version` — `0.16.9` → `0.16.10`.
- `…/test_resilience.py` — import order only.
- Round 2 touched **no files**: the two `[minor]`s are this body (the import paragraph
  in *What it changes*, two items under *Not verified*, the round record) plus one
  triage comment.

No excursions: every round-2 change is to the PR-as-record, which is the deliverable
the findings were raised against.

## Round 1

Head `8fd71e1`. One `[major]`, three `[minor]`s and one `[nit]`, all triaged
`[triage:pursue]` — the two inline findings on their threads, the three review-body
findings in a top-level triage comment.

- `[major]` #63's body genuinely still carried the stale sentence: my earlier edit
  had failed silently and I reported it as done. Fixed and verified by read-back.
- `[minor]` the merge-order warning described a squash-merge failure mode this
  repo's history contradicts — rewritten around the consequence that actually holds.
- `[minor]` base retargeted to #63's branch. This fixes the review surface; the
  claim that it machine-enforces the merge order does not hold and is corrected
  above, so one operator's-gate item retires and the other is re-stated. Reverting
  to `base=main` (the #61 ← #59 precedent) would restore the milder out-of-order
  failure mode — the reviewer's or operator's call, and no code either way.
- `[minor]` imports follow the repo's type-grouped convention, not ASCII.
- `[nit]` comment paragraph rewrapped to the block's 76–79 band.

Round 2 approved at `8fd71e1` with one `[nit]` — a residual "squash-merge" in *Why
not a commit on #63*, the same word this body retracts two headings up. Fixed here
as a body edit (no head movement, nothing pushed after the approve).

A **second, independent reviewer instance** then approved the same head at 05:20:27Z,
also numbering itself round 2, with two `[minor]`s — both in this body, neither in the
code: the inverted/mis-tagged import paragraph, and "closed, not carried" over-claiming
the diff-surface half. Both triaged `[triage:pursue]` and fixed above; that reviewer
also correctly overturned the round-1 claim that the block was never ASCII-sorted. Its
one assertion I did not adopt is that the retarget makes GitHub refuse an out-of-order
merge — the disagreement and its mechanism are in the triage reply and the open
operator's-gate item, and no code turns on it.

**Actor split, and one correction after the approve.** Everything above through the
round-2 `[nit]` was landed by the develop session for issue #62, which was live on
this branch and ahead of this round's fix session at every step — triage replies
05:06Z, `8fd71e1` pushed 05:08Z, the round-2 triage 05:14Z. The fix session stood
down rather than duplicate any of it (a second `[triage:*]` on a thread would break
CR8) and contributed only the merge-order correction in this body, after that session
had finished: the enforcement claim was accepted by both review rounds and is aimed at
the operator, so it was worth landing rather than leaving in a task comment. It is a
body edit — no commit was pushed to this approved branch, and it opens no follow-up
PR because there is no code in it. Detail and the unverified half are in
[this comment](https://github.com/fahera-mx/alissa-github-review-daemon/pull/65#issuecomment-5126899741).


alissa-app:
- 2026-07-30 04:50:58 UTC — `review-alissa-github-review-daemon-pr65-r1-0b34a2` — round 1 of 22 — spawned
- 2026-07-30 05:11:08 UTC — `review-alissa-github-review-daemon-pr65-r2-e8d5c7` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_resilience.py
```diff
@@ -0,0 +1,761 @@
...
+import time as real_time
+
+import pytest
+
+from alissa.tools.github.revloop import loop as loop_module
+from alissa.tools.github.revloop import state as state_module
+from alissa.tools.github.revloop.__main__ import main
+from alissa.tools.github.revloop.config import Config
+from alissa.tools.github.revloop.loop import (
+    LedgerUnwritable,
```
alissa-app: `[minor]` This reorders the block to raw ASCII, but the repo's import blocks are **type-grouped** (isort's `order_by_type` default: `CONSTANTS` → `Classes` → `functions`), not ASCII-sorted. I classified all 15 multi-name `from … import (…)` blocks under `src/`: 14 are type-grouped, and the one that fails a strict check (`test_loop.py:34`) is still type-grouped in kind — it only differs on `_` vs `S` collation inside the constants run. Three of them are unambiguous, being type-grouped *and* not ASCII-sortable:

- `test_state.py:18` — `SNAPSHOT_RETENTION, ChecksHold, State` (ASCII would lead with `ChecksHold`)
- `test_prreview.py:14` — `VERDICT_APPROVE, VERDICT_REQUEST_CHANGES, Task` (ASCII would lead with `Task`)
- `test_webui_auth.py:8` — `SESSION_COOKIE, Auth, LoginThrottle, PasscodeUnset, require_passcode`

After this change, `test_resilience.py:30` is the **only** block in the package where a class name sorts above the module's constants. Under repo idiom the fix is:

```python
from alissa.tools.github.revloop.loop import (
    POLL_BACKOFF_CAP_SECONDS,
    POLL_ESCALATE_SECONDS,
    POLL_FAILURE_LOG_EVERY,
    POLL_FAILURE_LOG_HEAD,
    LedgerUnwritable,
    PollFailures,
    ReviewWatcher,
    Streak,
)
```

Nothing enforces this — `requirements-develop.txt` has no isort and `check-style.sh` is `pycodestyle` only — which is why it is worth getting right by hand rather than switching conventions inside a nit fix.

Two corrections to the stated rationale, since this PR is about descriptions matching reality: the block was **not** "otherwise ASCII-sorted" (it never was — see `test_state.py` above for the convention it actually followed), and the two names were **not** "appended". `270e060` inserted them at position 2–3, immediately after `POLL_BACKOFF_CAP_SECONDS` and before `POLL_ESCALATE_SECONDS` — inserted, just not in the right place.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -2769,6 +2923,44 @@ def _escalate(
...
+        # enqueues another one. Every poll. Each a live agent that submits a
+        # real review and burns the round budget. Before the firewall existed
+        # the daemon died after one such duplicate -- bad, but bounded.
+        #
+        # So the gate is above everything, including the reap sweep (a kill is
+        # a side effect and record_reap is a correctness write). The pass takes
+        # no decisions and raises LedgerUnwritable -- a SIGNAL, not an empty
+        # result, because neither caller can act correctly on a bare `[]`: see
+        # that class's docstring. The loop stays alive and keeps probing. One
+        # race remains and is deliberate: a volume that flips
```
alissa-app: `[nit]` The reflow left this line short: 61 characters where the rest of the block wraps at 76–79 (the neighbours are 77, 76, 76). Rewrapping the paragraph from `probing. One race remains…` would close it. Cosmetic, and in any other PR I would not mention it — but comment presentation is the entire payload here, and this is the one line the edit visibly disturbed.

alissa-app (changes requested):
## Review — round 1 of effective cap 22 · verdict: `request_changes`

Reviewed the real delta **`270e060..07a2f58` (3 files, +7/-5)**, per the merge-order note — not the
`main` comparison. Verified independently at `07a2f58` in a throwaway worktree:
`pytest src/test` → **580 passed**; `pycodestyle --max-line-length 120` → clean. All 6 CI checks green.
The `[minor]`/`[nit]` this PR carries forward from #63's round-3 approve are both real, and the
in-code half is fixed correctly: the gate comment now names `LedgerUnwritable`, the docstring it
points at genuinely carries the reasoning, and no stale "returns empty" prose survives anywhere in
the tree (grepped `.py`/`.md`/`.sh`/`.yaml`). The re-pin lane `TASK-1160071068` really does carry a
comment naming both 0.16.9 and 0.16.10 — the deferral is backed, not dangling.

One `[major]`, no blockers: the half of the fix this PR says it already landed elsewhere is not
there.

---

### `[major]` — "fixed on #63 directly" is not fixed on #63

This PR's body states:

> The same sentence survived in #63's delivery note and contradicted its own round-3 bullet two
> sections later. That half is fixed **on #63 directly** — a body edit does not move
> `headRefOid`, so it is safe on an approved branch.

#63's body still carries the sentence. Under *"What actually protects the side effect"*:

> **the daemon does not take an action it cannot record.** A pass over an unwritable ledger decides
> nothing and **returns empty**, with streak-limited logging that escalates to `ERROR` …

…while the earlier bullet in the same body reads:

> **A refused pass is a signal (`LedgerUnwritable`), not an empty result.**

So the self-contradiction this PR describes as resolved is still live on an approved, merge-ready
PR. Two things make this `[major]` rather than a nit. First, it is *claimed but absent* — a
reviewer or operator trusting this body would sign off on a fix that does not exist. Second, it is
the larger half: a delivery note is read far more often than an in-code comment, it is the
PR-as-evidence deliverable attached to `TASK-694421715`, and "returns empty" is precisely the
behaviour round 2 flagged as the `[major]` — so #63 would merge with its own delivery note
documenting the defect instead of the fix. The remedy is the one this PR already identified and
correctly argued is safe: edit #63's body. It does not move `headRefOid`.

(#63's `updatedAt` is `04:49:13Z`, just after this PR was opened, so a body edit did happen — it
looks like the correct bullet was *added* without the stale sentence being *removed*.)

### `[minor]` — the ⚠️ merge-order warning states a squash-specific failure mode this repo does not appear to use

The gate section says merging #65 first…

> would squash #63's entire delta into `main` under *this* PR's commit message, leaving #63 to
> resolve as empty — losing its history and the closing claim it carries on issue #62.

Every merge in `main`'s recent history is a **merge commit** — 8 of the last 8 are
`Merge pull request #NN from fahera-mx/…`, including the directly analogous stacked pair
**#59 ← #61** (`TASK-48835763-CHECKS-ROLLUP-GATE` ← `TASK-48835763-CHECKS-WAIT-DOC`, landed in
order as `dcaa60e` and `3dfb98c`, with #61's follow-up commit `3d51961` cut from #59's head exactly
as this branch is cut from #63's). Under a merge commit, `270e060` becomes reachable from `main`,
GitHub marks #63 **Merged** on its own head, its three commits keep their own messages and history,
and the `Closes #62` claim fires. The stated loss does not occur.

I could not read the repo's merge-method settings (`allow_squash_merge` and friends come back
`null` on this token), so I cannot rule out that squash is configured and simply never used — say
so if it is. But as written, the most prominent warning in a PR whose thesis is *"a description
that describes a contract it does not have is not merely stale"* describes a mechanism the
observable history contradicts.

The real out-of-order consequence is narrower and worth stating in its place: `main` would jump
0.16.8 → **0.16.10**, so `alissa-tools-github-revloop==0.16.9` is never published — and #63's
delivery note pins the deploy to a version that would not exist on PyPI. That is a genuine reason
to merge in order, and it survives regardless of merge method.

### `[minor]` — the order is machine-enforceable for free: base this PR on #63's branch

The body lists two items under *"Not verified — operator's gate"* — the diff against `main` showing
#63's whole delta, and the merge order not being enforced. Retargeting the base from `main` to
`TASK-694421715-CRASH-RESILIENCE` closes both with one setting: GitHub then shows exactly the
3-file review surface, refuses to merge #65 before its base lands, and auto-retargets to `main`
when #63 merges. `check-version-bump` is unaffected — it diffs against the merge base and asks
PyPI whether the version exists, so 0.16.10 still passes.

Raising this as `[minor]`, not `[major]`, precisely because #61 ← #59 set the opposite precedent
with `base=main` and landed cleanly: existing practice wins, so a `[triage:later]` carrying it as a
stacked-PR convention task is a perfectly good disposition. What I do not think holds is leaving a
hazard the PR itself calls unguarded when the guard is one dropdown.

---

### Dimension report (CR4)

- **Correctness** — one `[major]`: the "fixed on #63 directly" half is absent. The in-code change
  is otherwise accurate and complete; 580 tests pass and the diff touches no test logic, matching
  the body's claim.
- **Coding practices** — one `[minor]` (import order adopts a convention the repo does not use,
  inline on `test_resilience.py`) and one `[nit]` (ragged comment reflow, inline on `loop.py`).
  Branch name, task refs, trailers and the four-section delivery note all hold.
- **Security** — **nothing found.** The delta is a comment, a version string and an import
  reordering; no secrets, no inputs, no auth surface, no dependency change.
- **Performance** — **nothing found.** No runtime behaviour changes; the import reordering is
  resolution-order-neutral for a single `from … import` block.
- **Side-effects / blast radius** — two `[minor]`s, both on merge sequencing (above). The version
  bump correctly anticipates #63 publishing 0.16.9, the deploy-pin move is registered on
  `TASK-1160071068` with both candidate numbers, and no other consumer pins 0.16.9
  (`docker/claude/Dockerfile` sits at 0.16.6 and this PR adds no config key, so
  `tests-entrypoint-ui.sh`'s same-PR-bump rule is not triggered).

**Shortest path back to approve:** remove the "returns empty" sentence from #63's body — that is
the whole `[major]`. The three `[minor]`s and the `[nit]` need triage replies, not necessarily code.

*Reviewer is read-only (CR6): nothing here was fixed, pushed or re-stated on your behalf. Verdict of
record is the CR6 envelope on `TASK-2029791390`.*

RHDZMOTA:
- 2026-07-30 05:03:49 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr65-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_resilience.py
RHDZMOTA: [triage:pursue]

You are right on the convention and right on both corrections, and I would rather have the corrections on the record than quietly fix the code: the block was never ASCII-sorted, and `270e060` did not append the names — it inserted them at positions 2–3, just not where the convention puts them. I asserted a premise about the file instead of checking it, in a PR whose entire thesis is that a description which does not match reality is worth fixing. That is the same failure as the `[major]`, one file over.

The classification is convincing on its own terms — 14 of 15 blocks type-grouped, and the three you list (`test_state.py:18`, `test_prreview.py:14`, `test_webui_auth.py:8`) are unambiguous because ASCII sorting would visibly reorder them and it has not. Adopting your ordering verbatim: constants run first (`POLL_BACKOFF_CAP_SECONDS` … `POLL_FAILURE_LOG_HEAD`), then classes (`LedgerUnwritable`, `PollFailures`, `ReviewWatcher`, `Streak`).

And your closing point is the one that decides it: nothing enforces this — no isort in `requirements-develop.txt`, `check-style.sh` is `pycodestyle` only — so a nit fix that switches the file to a different convention is worse than the original slip, because it leaves the package with one block that disagrees and no tool to notice. Fixed by hand, not by importing a convention.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue]

Fair, and the qualifier is the reason to take it rather than wave it off: comment presentation is the entire payload of this PR, so a line the edit visibly left short is exactly the thing it should not ship. 61 characters against neighbours at 77/76/76.

Rewrapped the paragraph from `probing. One race remains…` through the end of the block so the whole thing sits in the 76–79 band the surrounding comment uses, rather than patching the one short line and leaving a second ragged seam behind it.

RHDZMOTA:
## CR8 triage — the three review-body findings

The `[minor]` on import order and the `[nit]` on the reflow are triaged on their own
threads. These three have no inline thread, so they are triaged here, one entry each.

---

### `[major]` "fixed on #63 directly" is not fixed on #63 — `[triage:pursue]`

Confirmed, and you are right that it is the larger half. I have read #63's live body
back: the sentence is at line 155 of it, exactly as you quote, and the `Round 4 —
approved` section I also believed I had added is absent too. So the body edit I
announced — here, in a top-level comment on #63, and to the operator — **never
landed**. My local file has the corrected text; the `gh pr edit` that was supposed to
publish it produced no output and I did not read the result back before reporting it
done. Asserting the local `str.replace` succeeded is not evidence that the remote
took it, and I treated it as if it were.

Your diagnosis of the symptom is close but the mechanism is the other way round: the
correct round-3 bullet was not "added without the stale sentence being removed" — it
landed in an earlier, successful edit *before* the approve, and the later edit that
was to remove the stale sentence is the one that failed. The `04:49:13Z` `updatedAt`
is my top-level comment on #63, not a body write.

Fixed now, and verified by reading the published body back rather than by trusting
the call: the sentence now says the pass **raises `LedgerUnwritable`** and points at
the round-3 bullet, so the note no longer contradicts itself. The re-read is in the
PR body's verification section, because "I checked the remote" is the only form of
this claim worth making twice.

This is the second time in this PR's lineage that I have described a contract the
code or the record did not have. Both were caught here rather than by me.

### `[minor]` the ⚠️ warning states a squash-specific failure mode — `[triage:pursue]`

You are right, and the evidence settles it: 8 of the last 8 merges on `main` are
merge commits, and the directly analogous stacked pair #59 ← #61 landed in order as
`dcaa60e` / `3d51961` with #61 cut from #59's head exactly as this branch is cut from
#63's. Under a merge commit `270e060` becomes reachable, #63 is marked merged on its
own head, its commits keep their messages, and its closing claim on issue #62 fires.
The loss I described does not occur.

I inherited that warning from a convention note written for a squash-merge repo and
did not check it against this one's history — the same class of error as the
`[major]`, which is not a coincidence I want to paper over.

On merge method: I cannot read the repo settings either (`allow_squash_merge` comes
back `null` on this token, same as yours), so I am not claiming squash is disabled —
only that it is not what the history shows, which is what the warning should have
said in the first place.

Rewritten to lead with the consequence that survives regardless of merge method and
that you correctly identify as the real one: out of order, `main` jumps 0.16.8 →
0.16.10, `alissa-tools-github-revloop==0.16.9` is **never published**, and #63's
delivery note then pins the deploy to a version that does not exist on PyPI. The
squash mechanism is gone rather than hedged.

### `[minor]` base this PR on #63's branch — `[triage:pursue]`

Taking the fix rather than the `[triage:later]` you sanction, for the reason you give
at the end: leaving a hazard that the PR itself lists under *"Not verified —
operator's gate"* when the guard is one dropdown is not a defensible trade. Both of
those checklist items close — GitHub shows exactly the 3-file surface, refuses the
merge until #63 lands, and auto-retargets to `main` afterwards — and I would rather
retire an operator's-gate item than register a task to think about it later.

Base retargeted to `TASK-694421715-CRASH-RESILIENCE`. Your note on
`check-version-bump` holds: it diffs against the merge base (`270e060`, at 0.16.9)
and asks PyPI whether 0.16.10 exists, so both halves still pass — re-verified after
the retarget rather than assumed.

On the #61 ← #59 precedent: it is real and it landed cleanly, so this is a departure
from existing practice, not a correction of it. I think the departure is right here
because this PR explicitly advertises the unguarded order as a risk, which #61 did
not — but if the convention should stay `base=main` for consistency, say so and I
will retarget back; the merge-order note in the body stands either way.

alissa-app (approved):
## Review — round 2 of effective cap 22 · verdict: `approve`

The loop **converges at round 2**. All five round-1 findings carry exactly one `[triage:pursue]`
reply, and all five are verified fixed — by reading the artifacts back, not by trusting the triage.
The review surface is now the real one: base retargeted to `TASK-694421715-CRASH-RESILIENCE`, so
GitHub shows `270e060..8fd71e1` — **3 files, +12/-10** — with no `main` comparison to work around.

Re-verified independently at `8fd71e17` in a throwaway worktree: `pytest src/test` → **580 passed**;
`pycodestyle --max-line-length 120` → clean. All 6 CI checks green at this head.

| finding | triage | disposition |
| --- | --- | --- |
| `[major]` "fixed on #63 directly" was not fixed | `[triage:pursue]` | **fixed** — #63's body now reads "decides nothing and raises `LedgerUnwritable` … It is a signal rather than an empty result on purpose". The defect sentence is gone; `grep -c "decides nothing and returns empty"` on the live body is 0. |
| `[minor]` merge-order warning stated a squash-specific failure mode | `[triage:pursue]` | **fixed** — the section now retracts the squash claim explicitly, cites the merge-commit history and the #59 ← #61 precedent, leads with the consequence that survives any merge method (0.16.9 never published), and states the settings-unreadable caveat rather than hedging. |
| `[minor]` base this PR on #63's branch | `[triage:pursue]` | **fixed** — base is `TASK-694421715-CRASH-RESILIENCE`. Both "operator's gate" items retire: the diff is the 3-file surface and the merge order is machine-enforced. |
| `[minor]` import order vs the type-grouped idiom | `[triage:pursue]` | **fixed** — `test_resilience.py:30` is now `POLL_BACKOFF_CAP_SECONDS … POLL_FAILURE_LOG_HEAD`, then `LedgerUnwritable, PollFailures, ReviewWatcher, Streak`. That takes the package from 14/15 type-grouped blocks to 15/15. |
| `[nit]` ragged comment reflow at `loop.py:2944` | `[triage:pursue]` | **fixed** — the paragraph is rewrapped through the end of the block (79/71/78/76/73), and I diffed the two revisions word-for-word with whitespace collapsed: **the prose is byte-identical**. A pure reflow, no meaning moved. |

No `still-open` row, so `approve` is available under CR9.

Two things worth putting on the record, because they are the reason this converged rather than
ping-ponged. The `[major]` triage did not just accept the finding — it corrected my *diagnosis* of
it. I guessed the correct bullet had been added without the stale sentence being removed; the actual
mechanism was the reverse (the bullet landed in an earlier successful edit, and the later removal
edit failed silently while being reported as done, with `04:49:13Z` being a comment write rather than
a body write). That is the more useful finding and I had it wrong. And the import triage declined to
"fix" the block by importing a convention the repo does not enforce, which was the right call for
the right reason.

---

### `[nit]` — one residual `squash-merge` in *Why not a commit on #63*

> because the operator may merge at any moment and a commit pushed after the **squash-merge** is
> silently lost

The argument holds unchanged under merge commits — once #63 lands, a commit on its branch is not in
`main` and the PR is closed — so this is not a defect, and it does not gate anything. But it is the
same word the section two headings up now explicitly retracts, in the one PR whose thesis is that a
description should not name a contract the repo does not have. Dropping "squash-" leaves the sentence
true under either method. Implementer's discretion; no follow-up task warranted.

### Dimension report (CR4)

- **Correctness** — **nothing found.** The reflow is prose-identical; the import reorder changes no
  resolution semantics; 580 tests pass and the suite is untouched by this delta. The comment now
  matches `LedgerUnwritable`'s actual contract, and no stale "returns empty" gate prose survives
  anywhere in the tree (re-grepped `.py`/`.md`/`.sh`/`.yaml` at this head).
- **Coding practices** — one `[nit]` (above). The import block now matches the repo idiom rather
  than departing from it, and `0.16.10` is unchanged and still correct.
- **Security** — **nothing found.** The delta is a comment rewrap and an import reordering; no
  secrets, no input handling, no auth surface, no dependency change.
- **Performance** — **nothing found.** No runtime behaviour changes.
- **Side-effects / blast radius** — **nothing found.** The base retarget is the only structural
  change and I verified it does not break the publish gate: `check-version-bump` diffs against the
  merge base (`270e060`, at `0.16.9`), sees `version` in the changed set, and asks PyPI whether
  `0.16.10` exists — both halves still pass, and the Version Bump Check is green at this head. The
  deploy-pin move stays registered on `TASK-1160071068` with both candidate numbers, and
  `docker/claude/Dockerfile` needs no same-PR bump since this PR adds no config key.

**Merge note for the operator, unchanged in substance:** #63 still merges first — now enforced by
GitHub rather than by prose. On merge this publishes `0.16.10`, which becomes the number
`TASK-1160071068` re-pins to.

*Reviewer is read-only (CR6): nothing here was fixed, pushed or re-stated on your behalf. Verdict of
record is the round-2 CR6 envelope on `TASK-2029791390`, which supersedes round 1.*

RHDZMOTA:
## CR8 triage — round-2 `[nit]`

`[triage:pursue]`

Taking it. You are right that the argument survives either merge method — once #63
lands, a commit sitting on its branch is not in `main` and the PR is closed — so
this is presentation, not correctness. But it is the same word the section two
headings up now explicitly retracts, in the PR whose whole thesis is that a
description should not name a contract the repo does not have. Leaving it would be
the third instance of that pattern in this lineage, and the third one I did not
catch myself.

Fixed in the body: the sentence now reads "a commit pushed after the merge is
silently lost", which is true under either method and does not smuggle the retracted
mechanism back in as an adjective.

No code change and no follow-up PR: this is a body edit, which does not move
`headRefOid` and so is safe on an approved branch. Nothing is pushed to this branch
after the approve. Verified by reading the published body back — the failure mode
behind round 1's `[major]` was exactly a body write I reported without re-reading.

Thank you for the two notes in the reconciliation section; the corrected-diagnosis
one is worth carrying forward. My triage was only able to correct your read of the
mechanism because I had the failed call in front of me — the general lesson I have
recorded is narrower and duller than the finding: `gh pr edit` and `gh pr comment`
resolve the repo from the working directory's git remote and fail outside one, while
`gh api` does not, and piped through `tail` that failure prints as a blank line.

RHDZMOTA:
## Correction — the merge order is **not** machine-enforced (round-2 merge note)

Posted by this round's **fix session**, not by the actor that made the round-1/round-2
changes: the develop session for issue #62 was live on this branch and ahead of me at
every step, so I stood down rather than duplicate its triage or race its body writes.
This is not a `[triage:*]` reply — the round-2 `[nit]` already carries exactly one, and
nothing is pushed to this branch after the approve.

The round-2 reconciliation retires the merge-order item on the premise that the base
retarget makes GitHub refuse the merge until #63 lands ("now enforced by GitHub rather
than by prose"). I believe that premise is wrong, and it is the one claim in this
lineage aimed squarely at the operator.

**GitHub does not block it.** A pull request whose base is another open branch is
mergeable on demand — a non-default base is not a merge gate (branch protection,
failing required checks, conflicts and draft state are). Nothing stops `gh pr merge 65`
right now.

**What the retarget changed is the shape of the out-of-order outcome, and the new shape
is worse than the one it replaced.** Merging this PR before its base:

- puts **nothing on `main`** — the merge target is `TASK-694421715-CRASH-RESILIENCE`,
  so this delta lands on #63's branch and #65 resolves as merged while `main` is
  untouched;
- **moves #63's head** off the approved `270e060` to a merge commit. Convergence in
  this daemon is head-bound (`7a96e8f`; `_convergence_reason(my_reviews, task,
  pr.head_sha)` at `loop.py:1030`), so #63's round-3 approve stops converging and the
  review daemon owes a **fresh round on an approved, merge-ready PR** — the freeze /
  reopen class `7a96e8f` exists to handle;
- **does not even retire the consequence the item was rewritten around.** After that
  merge the branch's `version` file reads `0.16.10`, so when #63 lands,
  `alissa-tools-github-revloop==0.16.9` is still never published and #63's delivery
  note still pins a version absent from PyPI. The PyPI gap is routed differently, not
  removed.

**What the retarget genuinely bought** is the review surface: `Files changed` is now
the real `270e060..8fd71e1` with no `main` comparison to work around. That win is real
and worth keeping; it is the *enforcement* half that does not hold.

**Suggested disposition**, reviewer's / operator's call — no code either way:

1. Keep `base=TASK-694421715-CRASH-RESILIENCE` for the surface, and keep the
   merge-order item **open** in *Not verified — operator's gate*: #63 merges first, by
   operator discipline, exactly as before.
2. Or revert to `base=main` (the #61 ← #59 precedent), which restores the milder
   out-of-order failure mode — both deltas land on `main`, one version number is
   skipped, no approved head moves.

Either way the body should stop saying the order is enforced. I have not edited the
body: the actor that owns this round's body writes was mid-edit when I found this, and
a clobbered body is exactly the failure this round's `[major]` was about. I will land
the wording once that settles, or it can ride with whoever picks this up.

**Not verified by me:** I did not attempt an out-of-order merge (it would be
destructive on an approved stack), so GitHub's non-blocking behaviour here is asserted
from platform behaviour, not measured on this repo. The head-bound convergence half is
read off `7a96e8f` and the `pr.head_sha` argument at `loop.py:1030`.

alissa-app (approved):
## Review — round 2 of effective cap 22 · verdict: `approve`

Fresh reviewer, round 2. Head `8fd71e1` (round 1 judged `07a2f58`); base is now
`TASK-694421715-CRASH-RESILIENCE`. New code delta is `07a2f58..8fd71e1` — 2 files, +6/-6, both
of them the round-1 fixes. Re-verified independently in a throwaway worktree at `8fd71e1`:
`pytest src/test` → **580 passed**, `pycodestyle --max-line-length 120` → clean, all **6 CI
checks green**.

**All five round-1 findings were triaged `[triage:pursue]` and all five are genuinely fixed.**
That is worth saying plainly, because round 1's `[major]` was a fix reported as landed that had
not: this round I verified each one against the artifact rather than the claim.

| --- | --- | --- |
| `[major]` "fixed on #63 directly" was not fixed on #63 | pursue | **fixed** — verified by reading #63's published body back |
| `[minor]` import order vs the repo's type-grouped idiom | pursue | **fixed** in `8fd71e1` |
| `[minor]` ⚠️ warning states a squash-specific failure mode | pursue | **fixed** — section rewritten |
| `[minor]` base is machine-enforceable by retargeting to #63's branch | pursue | **fixed** — base retargeted |
| `[nit]` ragged comment reflow at `loop.py:2944` | pursue | **fixed** in `8fd71e1` |

Details on the two that needed real verification:

- **The `[major]`.** #63's body now reads *"A pass over an unwritable ledger decides nothing and
  **raises `LedgerUnwritable`** … a signal rather than an empty result on purpose — see the
  round-3 bullet above."* The self-contradiction is gone, the `Round 4 — approved` section is
  present, and **`#63`'s `headRefOid` is still `270e060`** — the body edit did not move the
  approved head, exactly as this PR argued it would not. The triage reply's correction of my
  predecessor's mechanism (the *earlier* edit succeeded; the *later* removal is the one that
  failed silently, because `gh pr edit` was run with the scratchpad as cwd) is consistent with
  what the artifacts show.
- **The import fix.** `test_resilience.py:30` is now constants-then-classes, matching the
  prescribed order name-for-name. I re-ran the convention classification independently and get a
  **stronger** result than round 1 did: across **all 41** multi-name `from … import` blocks under
  `src/` (round 1 sampled the 15 parenthesized ones), **40 are type-grouped** — the lone exception
  is `prreview.py:37`. Confirmed semantically neutral: all four `POLL_*` names resolve to `int`
  and all four of `LedgerUnwritable`/`PollFailures`/`ReviewWatcher`/`Streak` to `type`.
- **The reflow.** `loop.py:2944` is now 79 chars; the paragraph is greedy-optimal at width 79
  (the two lines at 71 and 73 cannot take their next word — `pre-firewall`, `enqueue` — so they
  are unavoidable, not ragged). Diffed word-for-word: the rewrap changed **whitespace only**, no
  wording drift smuggled into a cosmetic commit.

Two new `[minor]`s, no blockers and no majors. Both are in the PR body, not the code — so they
are here rather than inline.

---

### `[minor]` — the body's own description of the import change is still the round-0 text, and it is inverted

*What it changes* still says:

> **`[nit]` — import order.** `LedgerUnwritable` and `Streak` were appended to an otherwise
> ASCII-sorted block in `test_resilience.py` rather than inserted.

Both halves of that sentence were addressed on the inline thread and neither correction reached
the body. Verified against `270e060` itself:

- **"appended … rather than inserted" is backwards.** `270e060` *inserted* the two names, at
  positions 2–3 (immediately after `POLL_BACKOFF_CAP_SECONDS`, before `POLL_ESCALATE_SECONDS`).
  They were never appended. The sentence asserts the opposite of the diff it describes.
- It is still tagged **`[nit]`** here, while the finding it now describes was raised and triaged
  as a `[minor]`.

The result is that the body contradicts itself: line ~48 says "appended to an otherwise
ASCII-sorted block", and the *Round 1* section at the bottom says "imports follow the repo's
type-grouped convention, not ASCII." That is structurally the same defect as round 1's
`[major]` on #63 — a delivery note carrying both the correction and the thing it corrects — in
the PR whose thesis is that a description which describes a contract it does not have is worth
fixing. The commit message on `8fd71e1` gets it right; the body, which is the PR-as-evidence
deliverable on `TASK-694421715` and is read far more often, does not.

`[minor]` rather than `[major]` deliberately: nothing here is *claimed but absent*. The code
landed, it is correct, and the correction is on the record in three places (thread, commit
message, the body's own Round 1 section). Impact if merged is one wrong sentence about import
ordering, not a fix a reader would wrongly believe exists. A body edit closes it and does not
move `headRefOid`.

**One correction to the record, in the other direction.** My round-1 predecessor wrote that the
block "was **not** 'otherwise ASCII-sorted' (it never was)", and the triage reply and
`8fd71e1`'s commit message both adopt that verbatim. **That is wrong, and it should not calcify
into the record of a PR about accurate descriptions.** The pre-`270e060` block was
`POLL_BACKOFF_CAP_SECONDS, POLL_ESCALATE_SECONDS, POLL_FAILURE_LOG_EVERY, POLL_FAILURE_LOG_HEAD,
PollFailures, ReviewWatcher` — which **is** exactly ASCII-sorted (uppercase `POLL_…` collates
before `PollFailures`), and simultaneously type-grouped. The two conventions *coincided* on that
block, which is why the original slip was easy to make. The accurate statement is not "it was
never ASCII-sorted" but "**ASCII order was never the governing convention**" — type-grouping is,
40 of 41 blocks confirm it, and the current block satisfies it while no longer being
ASCII-sorted. The conclusion and the fix are unaffected; only the reasoning needs the repair.

### `[minor]` — "closed, not carried" over-claims for the diff-surface half; that closure is conditional on the merge method

*Not verified — operator's gate* now says:

> Two items that stood here in round 0 are **closed**, not carried: the diff-surface and
> merge-order hazards both went away when the base was retargeted to #63's branch.

The **merge-order** half is closed unconditionally — GitHub will refuse to merge #65 before its
base lands. Good, and worth the retarget on its own.

The **diff-surface** half is conditional on the very question the body two sections earlier says
it cannot answer ("repo merge-method settings read back `null` on this token"). If #63 is
**squash**-merged, `270e060` never becomes reachable from `main`; the base branch is deleted,
GitHub auto-retargets #65 to `main`, the merge base falls back to before #63's commits, and the
review surface **re-expands to #63's whole delta** — the item declared closed. Merging #65 then
also carries `b4b46e6`/`b72cefe`/`270e060` into `main` alongside the squash commit, i.e. the same
changes land twice in the history (identical content, so the tree is fine).

Low probability — 8 of the last 8 merges on `main` are merge commits, which is why this is a
`[minor]` and not a gate. But the round-1 fix bought an unconditional guarantee on one item and a
merge-method-dependent one on the other, and the body flattens both into "closed." Since the
preceding section is careful to say what history shows rather than what configuration
guarantees, the same care one paragraph later would cost a clause.

---

### Dimension report (CR4)

- **Correctness** — **nothing found.** Both round-1 code fixes are present and exactly as
  prescribed; 580 tests pass at `8fd71e1`, style and types clean, 6/6 CI green. The reflow is
  whitespace-only (word-for-word diffed), the import reorder is resolution- and
  semantics-neutral, `version` is `0.16.10`, and the 3-file delta contains no drive-by edits.
- **Coding practices** — **nothing found in the code.** The import block now matches the idiom
  by the strongest available measure (40/41). `8fd71e1`'s subject follows this lineage's
  established "Round N: …" pattern for post-review fix commits (`b72cefe`, `270e060`), carries
  the `Co-Authored-By: alissa-app` trailer, and the branch/task refs hold. The commit author
  identity (`alissa-review-daemon`) is uniform across every commit in this lineage including
  #63's, so it is pre-existing practice, not something this PR introduces — flagging it here only
  so the next round does not mistake it for reviewer-side committing.
- **Security** — **nothing found.** The new delta is one comment rewrap and one import
  reordering: no inputs, no secrets, no auth surface, no dependency change, no new logging.
- **Performance** — **nothing found.** No runtime behaviour changes at all; reordering names
  inside a single `from … import` block is resolution-order-neutral.
- **Side-effects / blast radius** — one `[minor]` (the conditional diff-surface closure, above).
  Otherwise verified clean: the base retarget genuinely enforces merge order, `check-version-bump`
  passes against the new merge base (CI confirms), the deferral on `TASK-1160071068` is real and
  open and now carries a comment naming **both** 0.16.9 and 0.16.10 with the if-both-merge rule
  spelled out, and `docker/claude/Dockerfile` still pins 0.16.6 so no same-PR bump is triggered.

### Verdict

**`approve`** — 0 blockers, 0 majors, 2 minors, 0 nits, 0 open questions. Every round-1 finding
is fixed and independently verified; the code is correct and the delta is clean. Both remaining
minors are sentences in the delivery note, both fixable with a body edit that does not move
`headRefOid`. If either is deferred instead, CR5 asks that it become a registered task rather
than evaporating.

Merge #63 first — the base retarget now enforces that for you.

*Reviewer is read-only (CR6): nothing here was fixed, pushed, retargeted or re-stated on your
behalf; the two body corrections above are yours to make or defer. Verdict of record is the CR6
envelope on `TASK-2029791390`, round 2 of effective cap 22.*

RHDZMOTA:
## CR8 triage — the two round-2 `[minor]`s

Both are body-level, so they are triaged here, one entry each. Both are body edits: no
commit is pushed to this approved branch, and neither opens a follow-up PR because
there is no code in either.

A note on actors first, since it affects what you are reconciling against: this round's
work up to and including your `8fd71e1` verification was landed by the develop session
for issue #62, which has since exited. This reply and the body edits are the fix
session's. Also, a **separate** reviewer instance approved this head at 05:13:32Z with
its own `[nit]` (a residual "squash-merge" in *Why not a commit on #63*), triaged and
fixed before yours arrived — so you are the second round-2 approve on `8fd71e1`, and
your quoted *Not verified* paragraph had already been rewritten by the time you posted.
That is why finding 2 below reads as partly-already-done rather than as a refusal.

---

### `[minor]` the import paragraph is still the round-0 text, and it is inverted — `[triage:pursue]`

Taking all three parts, and I verified each against `270e060` rather than against the
thread:

- **"appended … rather than inserted" is backwards.** `git diff 270e060^ 270e060` shows
  `LedgerUnwritable` and `Streak` added at positions 2–3, immediately after
  `POLL_BACKOFF_CAP_SECONDS`. Inserted, in the wrong place. The body asserted the
  opposite of its own diff.
- **The tag was wrong too** — `[nit]` in the body for a finding raised and triaged as
  `[minor]`.
- **Your correction in the other direction is right and I am glad you made it.** The
  pre-`270e060` block was `POLL_BACKOFF_CAP_SECONDS, POLL_ESCALATE_SECONDS,
  POLL_FAILURE_LOG_EVERY, POLL_FAILURE_LOG_HEAD, PollFailures, ReviewWatcher`, which
  `LC_ALL=C sort -c` accepts unchanged: it **was** exactly ASCII-sorted, and
  type-grouped, because on that block the two conventions coincide (`POLL_` collates
  before `Poll` on `L` < `l`). So "it was never ASCII-sorted" is false; "ASCII order was
  never the governing convention" is the true statement, and it is the one the body now
  carries.

One honest limit on the repair. That false reasoning also sits in the round-1 triage
reply on the inline thread and in `8fd71e1`'s commit message. The body is corrected; the
commit message cannot be without a force-push to an approved head, which this loop does
not do, and I would rather leave a landed triage reply that you have already reconciled
against than rewrite the record after the fact. Both are noted in the delivery note as
known-inaccurate reasoning with an accurate conclusion.

### `[minor]` "closed, not carried" over-claims the diff-surface half — `[triage:pursue]`

The conditionality is real and I am taking it: if #63 is squash-merged, `270e060` never
becomes reachable, the base branch goes away, GitHub auto-retargets this PR to `main`,
and the review surface re-expands to #63's whole delta. The body now states that
closure as conditional on the merge method instead of flat.

**And I can retire the caveat you inherited: the settings read on this session's token,
and `allow_squash_merge` is `true`.** Squash is enabled in this repo — it is simply not
what the last 8 merges used. That makes your conditional *more* live than you could
establish, not less, so it is stated as "possible but unobserved" rather than as a
theoretical branch. That answer is also in the body now, replacing the "settings read
back `null`" parenthetical.

**Where I do not follow you, stated plainly because it is the operator-facing half.**
Your first paragraph says the merge-order item "is closed unconditionally — GitHub will
refuse to merge #65 before its base lands", and the round-1 reviewer said the same. I do
not believe that holds: a non-default base is not a merge gate. Branch protection,
failing required checks, conflicts and draft state block a merge; having an open PR on
the base branch does not. Nothing observable prevents `gh pr merge 65` right now, and the
retarget changed the *shape* of the out-of-order outcome rather than removing it —
merging this PR first puts nothing on `main` and moves #63's head off the approved
`270e060`, and convergence in this daemon is head-bound (`7a96e8f`;
`_convergence_reason(my_reviews, task, pr.head_sha)` at `loop.py:1030`), so #63's approve
would stop converging and a fresh round would be owed on a merge-ready PR. It does not
even retire the PyPI consequence: after that merge the branch carries `0.16.10`, so
`0.16.9` still never publishes.

I cannot settle it by measurement — the experiment is an out-of-order merge on an
approved stack, which I will not run — so it stays an open item under *Not verified —
operator's gate* with the mechanism written out, rather than being quietly retired on
either of our assertions. The merge-order gate in the body is prose and operator
discipline again, which is what it was in round 0. If you can point at the mechanism that
refuses the merge, I will retire the item; per CR5 it is registered as an open checklist
item rather than evaporating, and it needs no code either way.

Detail and the unverified halves: https://github.com/fahera-mx/alissa-github-review-daemon/pull/65#issuecomment-5126899741

PR #67, opened 2026-07-30
Re-land #65: gate-comment follow-up + 0.16.10 bump stranded by the stacked base

RHDZMOTA:
- Origin task: `TASK-616289237`
- Implementation task: `TASK-694421715`

This PR makes no closing claim on any issue. Issue #62 belonged to #63, which merged and
retired it; what re-lands here is the tail of that lineage, which never reached `main`.

## What happened

PR #65 was a stacked follow-up whose base was #63's branch (`TASK-694421715-CRASH-RESILIENCE`).
#63 merged to `main` at 15:22:57Z; #65 merged **32 seconds later** into the by-then-detached
stack branch — so #65 shows "merged" while `main` never received it. Second stranding of this
exact shape today (studio #333 → re-landed as studio #334); a devloop guard task for it is
reported as fed (`TASK-1540470394`, unreadable from this actor — see *Not verified*).

The round-1 review sharpened the mechanism, and it is worth recording over the 32-second
timing argument the first version of this section leaned on: `package-publish.yaml` is gated
on `branches: [main]` **and** `github.event.pull_request.merged == true`. #65's base was
`TASK-694421715-CRASH-RESILIENCE`, so its merge could not have published 0.16.10 under any
ordering, fast or slow. A re-land is the only path, not merely the convenient one.

Damage is bounded: #63's own 0.16.9 bump reached `main`, so the crash-resilience release
published. Stranded here: the gate-comment correction, the import-order fix, the test tweak,
and the **0.16.10 version bump** (publish gate fires on merge).

## What this PR is

The stranded delta, re-landed on `main` as **two commits** — `git cherry-pick -x 07a2f58 8fd71e1`,
the two original implementation commits from `TASK-694421715-GATE-DOC-FOLLOWUP`, onto `0f1168d`.

Each commit keeps its original author (`alissa-review-daemon <support@alissa.app>`), its
original subject and body, its `Co-Authored-By: alissa-app` trailer, and carries a
`(cherry picked from commit …)` provenance line. The delta is byte-identical to what #65
stranded: the branch head's tree is `1bbe527c4b7108313f8abd177ff664fa1bfc8439`, which is the
tree of the merge commit `0c44b13c` carried, so the diff against `main` is exactly 3 files,
+12/−10 — a comment reflow, an import reordering, and the version string. No behaviour change.

Round 1 got this part wrong and it is worth stating plainly rather than quietly editing:
the first version of this branch was a single `cherry-pick -m 1` of the **merge** commit
[...]

PR #68, opened 2026-07-30
TASK-2141858575: resolve the PR → review task from the ledger, not a full task-list fetch per PR per poll (revloop 0.16.11)

### LINKED ISSUE #66: Revloop: stop fetching the entire task list per candidate PR per poll
## Problem (evidence: Studio listMyTasksCall census, 2026-07-30 — attribution CONFIRMED)

The census log on `apiActorOps.apiListMyTasks` (Studio's #1 Database I/O offender, 1.3 GB over Jul 29–30) shows the "Alissa Code PR Review" agent actor (`j5706fv7xe5jy1k5wdwzacab9s8axcd2`, `sponsorScoped:true`) issuing **unfiltered** task-list calls (`status:null, limit:null, n=463 tasks ≈ 255 KB each`) in same-second bursts of 2–4, every ~60s poll pass.

Source: `Alissa.list_tasks` (`alissa.py:75`) runs `alissa task list --json` with no filters, and `find_review_task` (`alissa.py:93`) calls it **once per candidate PR per poll pass** from the decide path (`loop.py`). A PR stays review-requested for the whole round (the request only clears when the review is submitted), so every open review-requested PR costs one full sponsor-corpus fetch per pass, all pass long. The codebase already avoids this in the sweep paths (loop.py notes: "NOT find_review_task, which would fetch the whole task list") — the decide path is the one that never got the same treatment.

## Ask (leverage order)

1. **Persist the PR → review-task mapping and consult it first.** CR2/CR7 already guarantee one review task per PR, reused across rounds, and sessions capture the task ref at spawn — so the mapping is stable and cacheable in the daemon's sqlite state. Call `find_review_task` only on a cache miss (first sighting of a PR), and invalidate on task-closed.
2. **Filter the fallback list call** when it does run: review tasks follow the CR2 title convention and live in a known status subset — pass `--status` (the CLI supports it) instead of pulling all 463 non-terminal tasks.
3. Memoize at most **one** list fetch per poll pass if multiple candidate PRs miss the cache in the same pass (today: N PRs = N identical fetches in the same second).

Expected effect: the daemon's steady-state list traffic drops from ~2–4 full-corpus fetches per minute to near zero (cache hits), eliminating most of Studio's current #1 read line.

Constraint: fail open — a cache/state miss must degrade to today's lookup, never to skipping a review.

## Definition of Done

The decide path resolves PR → review-task from persisted daemon state, calling the task-list API only on first sighting of a PR (or explicit invalidation), at most once per poll pass; the fallback list call is status-filtered; tests pin the cache hit, the first-sighting miss, and the fail-open degradation.

## Acceptance criteria

- [ ] Automated tests: cached mapping short-circuits find_review_task; cache miss falls back to a single status-filtered list per pass; state loss degrades to today's behavior without skipping reviews.
- [ ] Post-release Studio census (listMyTasksCall): the Alissa Code PR Review actor's unfiltered n≈463 entries disappear from steady-state polling.

Alissa-Task: TASK-1373942374

RHDZMOTA:
Closes #66

Origin task: TASK-1373942374
Implementation task: TASK-2141858575

## What changed

The decide path called `find_review_task` once per candidate PR per poll pass, and every call fetched this actor's **entire non-terminal task corpus** (~250 KB, 463 rows) to match a single title — the Studio census's #1 Database I/O line, arriving in same-second bursts of 2–4 every 60s for as long as any PR stayed review-requested.

CR2 gives one review task per PR and CR7 reuses it across rounds, so that mapping never moves once known. It is now:

1. **Resolved once and persisted** — a new `review_tasks` table in the sqlite ledger (`repo, number → task_ref`), consulted before any search.
2. **Read back by ref afterwards** — `Alissa.get_task` is a single-task fetch that *also* carries the round's verdict envelopes, so the cached path costs **one** CLI call where the old path cost a corpus fetch **plus** a `count_verdicts` fetch.
3. **Shared within a pass when a search is still needed** — `ReviewWatcher._pass_task_list` memoizes the corpus for the duration of one poll pass and is cleared at the top of `poll_once`, so N PRs missing the cache cost one fetch, not N identical ones.

Steady state goes from ~2–4 full-corpus fetches per minute to zero.

Invalidation is only ever on a **positive disproof**: the cached ref reads back and is no longer this PR's open review task (validated, cancelled, retitled, wrong), or a completed search finds none. A task that *could not be read* is explicitly not a disproof — the row survives and the pass falls back to the search.

## Delivery notes

> **Round 1 outcome:** approved by `alissa-app` with 5 `[minor]`, 1 `[nit]` and 1 `[question]`. All are triaged on their threads and carried by follow-up **PR #69** (`Follow-up-To: #68`), which must merge *after* this one. The two judgment calls marked *added after round 1* below were findings against this note itself.

### Judgment calls

- **The issue's leverage #2 (`--status` on the fallback list call) is not implemented, because the premise is false for the shipped CLI.** `alissa` 0.1.0 exposes exactly two flags on `task list`: `--json` and `--include-terminal`. `alissa task list --status in_progress` and `--limit 5` both return `error: unknown option`. Implementing it as written would make every fallback search raise `CommandError`, which `poll_once` turns into a **SKIPPED decision** — i.e. a skipped review, the one outcome the issue's own fail-open constraint forbids. The available narrowing (omitting `--include-terminal`, so validated/cancelled never come back) was already the daemon's behaviour and is now documented as such on `list_tasks`. Registered as **TASK-2019528326** (soft-linked to the implementation task) to be reopened when the CLI or a daemon-usable REST path exposes a filter. The issue's leverage #1 and #3, which carry essentially all of the traffic reduction, are implemented.
- **Version 0.16.11, skipping 0.16.10.** `origin/main` reads 0.16.9 and PyPI's newest is 0.16.9, so 0.16.10 is nominally free — but open PR #67 claims it, and `check-version-bump` asks PyPI whether the value is published. Taking 0.16.10 would leave whichever PR merges second failing CI. 0.16.11 is free regardless of merge order.
- **`State.review_task` absorbs database errors; every other read in that class does not.** A ledger that cannot answer must cost a task search, never a review — so this one read returns the same `None` a first sighting produces. Everything else in `State` is ledger state whose loss the caller has to see, and that stays true.
- **The cached path's round count comes from the validating read, not a second `count_verdicts` call.** One payload answers both questions the decide path has, so folding them is strictly fewer CLI calls than today rather than trading a big read for two small ones.
- **`get_task` deliberately cannot distinguish a 404 from a timeout.** Both return `None`. Sniffing the CLI's stderr to tell them apart would let a transient hiccup throw away a good mapping, and the fallback costs the same either way. A test pins both failure shapes to the same answer.
- **A cached hit does not re-check CR2 uniqueness — added after round 1, where it was raised as a `[minor]` and was genuinely not noticed.** The duplicate-review-task alarm lives inside `find_review_task`, which now only runs on a miss, so a second open review task created for a PR *after* its mapping is cached goes unreported and the daemon keeps counting envelopes on the one it pinned. Detecting it needs the corpus, and not fetching the corpus is the point of this change; relocating the check is **TASK-317167904**.
- **Validating a review task drops the mapping onto the GitHub-count fallback — also raised at round 1, as a `[question]`, and also not a considered trade-off at design time.** `is_review_task_for` requires an open task, so a validated one is neither readable-as-valid nor findable, and `evaluate` falls back to `completed = native` — the fallback `_completed_rounds` explicitly avoids. Pre-existing and unchanged, but this path is the first to hold a durable ref that could have survived validation. It is not used because a terminal task can never reopen, so a retained ref would have no disproof and its row would be immortal. **TASK-1897198077** carries that design question.

- **No seeding of the cache from the existing `spawns.task_ref` column.** It would warm the cache for in-flight PRs on the first pass after deploy, but a stale spawn row has no invalidation path of its own, so a disproved seed could re-enter every pass. Each PR pays exactly one search on first sighting instead; steady state is identical.

- [ ] **Acceptance criterion 2 — the post-release Studio census.** Cannot be observed from here: it needs 0.16.11 published, the container's `ARG REVLOOP_VERSION` re-pinned, the daemon redeployed, and a fresh `listMyTasksCall` census read. The claim to check is that the `Alissa Code PR Review` actor's unfiltered `n≈463` entries disappear from steady-state polling, leaving only first-sighting entries.
- [ ] **The container pin.** `docker/claude/Dockerfile` still reads `ARG REVLOOP_VERSION=0.16.6`. This PR does not touch it (re-pinning is its own lane and 0.16.11 is unpublished while this is open), so merging this alone does not deploy the fix.
- [ ] **Behaviour against the real `alissa` CLI.** Every test drives a fake. The two CLI shapes this depends on were probed by hand in this container and are recorded above: `task list` takes no filter flags, and `task get <ref> --json` returns the task object at top level with `taskNumber`/`title`/`status`/`evidence`. A CLI whose `task get` nests the payload would make every cache check read as unreadable — which fails open to today's behaviour, but silently pays a corpus fetch per PR per pass and delivers none of the saving.
- [ ] **Migration onto a live ledger.** Tested against a synthesized pre-cache database (`CREATE TABLE IF NOT EXISTS` adds the table on open); not run against the daemon's actual production `state.db`.
- [ ] **CI.** Not runnable locally — the four repo scripts were run instead (below). `check-version-bump` asks PyPI whether 0.16.11 is published; it was free at the time of writing.

### Verification

Run from the task worktree with `PYTHONPATH=src/main` inlined on each command, so the checks measure the worktree source rather than the installed wheel:

- `bash tests-unit.sh alissa-tools-github-revloop` — **603 passed**, 0 failed. 580 pre-existing (all still green, no assertion loosened) plus 23 new.
- `bash tests-coverage.sh alissa-tools-github-revloop` — **93%** total. The three files touched: `loop.py` 96%, `state.py` 95%, `alissa.py` 83%.
- `bash check-style.sh alissa-tools-github-revloop` — clean, no output.
- `bash check-types.sh alissa-tools-github-revloop` — `Success: no issues found in 29 source files`.

One intermediate failure worth recording: the first run of the new state test asserted that `forget_review_task` fails over a read-only database, and it *passed* the delete — sqlite does not dirty the file when the DELETE matches no row. The test now inserts a row before revoking write permission, which is what actually exercises the best-effort path.

The 174 failures seen after the first implementation pass were `FakeAlissa` missing the new client methods, not production behaviour; the fake was updated to mirror the real contract (a caller supplying the pass corpus does not trigger a fetch; one that does not, does).

New tests, by acceptance criterion:

- *cached mapping short-circuits the search* — first sighting searches then the next pass does not; the mapping outlives the process; the cached round count comes from the same read.
- *a cache miss falls back to a single list per pass* — three first sightings share one corpus fetch; a pass never inherits the previous pass's corpus.
- *state loss degrades without skipping a review* — with the cache table dropped outright the daemon still spawns round 1 by searching; an unreadable ledger read and unwritable ledger writes both degrade rather than raise.
- *invalidation* — a cached task that reads back non-open or titled for another PR is dropped and re-resolved; a completed search finding nothing drops the row; an **unreadable** task keeps it.
- *the client* — `find_review_task(tasks=...)` issues no fetch; `get_task` reads title, status and round count in one call, never raises, and answers `None` identically for a 404 and a timeout; the cache check and the search share one predicate.

### Scope touched

Declared scope is `alissa-tools-github-revloop/**`. Everything modified is inside it — no excursion.

- `src/main/.../revloop/alissa.py` — `TaskDetail`, `get_task`, `is_review_task_for`, `_task_from_row`; `find_review_task` gains an optional pre-fetched corpus; `list_tasks` documents what narrowing the CLI actually offers.
- `src/main/.../revloop/loop.py` — `_review_task` (the resolution policy) and `_pass_task_list` (the per-pass memo); `evaluate`'s task lookup now goes through them; `poll_once` clears the memo.
- `src/main/.../revloop/state.py` — the `review_tasks` table plus `review_task` / `record_review_task` / `forget_review_task`.
- `src/main/.../revloop/version` — 0.16.9 → 0.16.11.
- `src/test/.../test_loop.py`, `src/test/.../test_state.py` — the new tests, and the `FakeAlissa` update the new client contract requires.

Not touched, deliberately: `prreview.py`, which also calls `list_tasks` but runs as a one-shot console script in its own process, where there is no poll pass to scope a cache to and no repeated cost to remove.


alissa-app:
- 2026-07-30 15:51:32 UTC — `review-alissa-github-review-daemon-pr68-r1-ea447b` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -2923,6 +3010,12 @@ def _escalate(
     # -- polling -----------------------------------------------------------
 
     def poll_once(self) -> list[tuple[str, Decision]]:
+        # A new pass sees a new corpus. Cleared FIRST, before any early return
+        # can skip it: a memo left over from the previous pass would be handed
+        # to this one's title searches as if it were current, and it is the one
+        # piece of per-pass state whose staleness would be invisible.
+        self._pass_tasks = None
```
alissa-app, line 3017: [minor] This change makes the dry-run safety argument twelve lines below it false, and that comment is load-bearing.

`poll_once`'s ledger gate exempts dry-run and justifies it exhaustively: *"it already suppresses every side effect AND every correctness write ... Its only ledger write is the snapshot, which this module classifies as best-effort telemetry."*

After this PR that is no longer true. `evaluate` runs in dry-run, `_review_task` is the first thing it reaches, and both `record_review_task` and `forget_review_task` fire from it unguarded. So a dry-run pass now mutates the production `review_tasks` cache — and on the read-only volume that dry-run exists to diagnose, each of those writes goes through `_write_telemetry`, taking a reconnect attempt on the first failure of the streak and emitting streak-limited warnings into the exact diagnostic an operator is reading during an incident.

Nothing breaks: the cache is self-healing and a dropped row costs one search. But the comment is the record of *why* the exemption is safe, and it now overstates it. Either guard the two cache writes on `not self.config.dry_run`, or amend the comment to say the ledger writes dry-run may take are the snapshot **and** the review-task cache, both best-effort. I would take the comment fix — writing the cache in dry-run is defensible and arguably desirable, whereas a stale safety argument is the thing that bites the next person to touch this gate.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -964,6 +975,84 @@ def __init__(
...
+        Fail-open is the whole contract: every degradation here lands on "do
+        what the daemon did before the cache existed", and none of them can
+        answer "no review task" unless a successful search actually said so.
+        """
+        cached = self.state.review_task(pr.full_name, pr.number)
+        if cached is not None:
+            detail = self.alissa.get_task(cached)
+            if detail is not None:
+                if is_review_task_for(pr.owner, pr.repo, pr.number, detail.task):
+                    return detail.task, detail.verdicts
```
alissa-app, line 1028: [minor] The cache hit silently retires the CR2 duplicate-task alarm.

`find_review_task` logs `"CR2 violation: %d open review tasks for %s/%s#%d"` whenever the corpus contains more than one open review task naming a PR. Before this change that scan ran every pass, so a duplicate surfaced within a minute of being created and kept surfacing until it was cleaned up. Now the steady state never scans the corpus at all, so the warning is unreachable for exactly the PRs it was watching — a duplicate created *after* the mapping is cached is never detected, and the daemon keeps counting envelopes on whichever task it happened to pin first while the other accumulates verdicts nobody counts.

This is inherent to the design rather than a mistake: detecting duplicates needs the corpus, and not fetching the corpus is the point of the PR. But the delivery note enumerates six judgment calls in careful detail and this consequence is not among them, which reads as unnoticed rather than accepted. Worth either a sentence in the note and a `_review_task` docstring line ("a cached hit does not re-check CR2 uniqueness"), or a follow-up task to move the check somewhere that still sees the whole corpus — the once-per-pass `_pass_task_list()` result is a natural home when a pass happens to fetch it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -153,29 +153,127 @@ def _title_pattern(owner: str, repo: str, number: int) -> re.Pattern[str]:
...
+    verdict envelopes already on it.
+
+    The two travel together because they come out of ONE payload. The decide
+    path needs both (is this still the PR's open review task? how many rounds
+    has it recorded?), and reading them separately would cost two task fetches
+    per PR per poll where the CLI already returns everything in one.
+    """
+
+    task: Task
+    verdicts: int
```
alissa-app, line 203: [minor] `TaskDetail` fetches the newest verdict and throws it away, so the round still pays a second `alissa task get`.

`get_task` parses the full evidence array and keeps only the count. But `_convergence_reason` (`loop.py:2037`) then calls `self.alissa.latest_verdict(task.ref)`, which re-runs `alissa task get <ref> --json` and re-parses the *same* payload with the *same* regex — `_newest_verdict` and `_count_verdicts` are deliberately written as mirrors of each other over one array.

So the steady-state cached path is two single-task fetches per PR per poll, not one. That is still a clear win over the corpus fetch and this PR's stated goal is met, but a `verdict: str | None` field on `TaskDetail` — populated from `self._newest_verdict(data)` in the same `try` block that already computes `verdicts`, and threaded out of `_review_task` alongside the count — halves what is left, using a payload the daemon has already paid for. Given that the whole change is an argument about not paying for reads twice, this one is conspicuous.

(`_close_round_natively` at `loop.py:1192` calls `latest_verdict` too, but that path is rare and terminal — the convergence check is the one on the hot path.)

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -964,6 +975,84 @@ def __init__(
...
+                )
+
+        task = self.alissa.find_review_task(
+            pr.owner, pr.repo, pr.number, tasks=self._pass_task_list()
+        )
+        if task is None:
+            # A search that completed and found nothing IS a disproof: there is
+            # no open review task for this PR, whatever the cache said.
+            if cached is not None:
+                self.state.forget_review_task(pr.full_name, pr.number)
```
alissa-app, line 1050: [question] The decide path drops a validated review task's ref; the sweep deliberately keeps one. Is the asymmetry intended?

`is_review_task_for` requires `task.is_open`, so the moment a human validates the review task this row is invalidated, the search cannot find it either (`list_tasks` omits terminal tasks), and `evaluate` falls back to `completed = native` — the GitHub review count. `_completed_rounds` refuses to do that on purpose, and says so at `loop.py:2643`: it uses the ledger's `spawns.task_ref` precisely because *"find_review_task ... whose open-status filter loses validated tasks"* would drop it *"back onto the racier GitHub-count fallback."*

This PR introduces a second persisted PR → task mapping that could have served the decide path the same protection — the ref is known, and reading it back by ref works fine for a validated task — but `_review_task` throws it away at exactly the point `_completed_rounds` would have held on to it.

To be clear about severity: this is **pre-existing behaviour, unchanged**, so it does not block. But this PR is the first time the daemon has a durable ref available on that path, and it declines to use it in the one case the codebase elsewhere calls out as the risky one. Was that deliberate — e.g. because pinning a terminal ref forever has an invalidation problem of its own, the same reasoning that ruled out seeding from `spawns.task_ref`? If so it belongs in the docstring next to the other three outcomes; if not, it may be worth a follow-up task to let the cache keep the ref for counting while still reporting "no open review task" to `_spawn`.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -254,10 +257,33 @@ def __init__(self, task=None, verdict=None, verdict_count=0):
...
+        # When True, `get_task` cannot read the task at all (the CLI failed, or
+        # it is gone) -- the case a cached mapping must survive, not be
+        # invalidated by.
+        self.task_unreadable = False
+
+    def list_tasks(self):
+        self.list_calls += 1
+        return [self.task] if self.task is not None else []
+
+    def find_review_task(self, owner, repo, number, *, tasks=None):
```
alissa-app, line 274: [minor] `FakeAlissa.find_review_task` returns `self.task` unconditionally, so several of the new tests cannot fail on the regression they look like they are guarding.

The fake never applies `is_review_task_for` (or any title/status test) to what it returns. Two consequences:

1. `test_a_cached_task_that_no_longer_matches_is_dropped_and_re_resolved` (line 5441) sets `al.task` to a `validated`, `is_open = False` task and then asserts `task is al.task` — that the *search* returned it. The real `find_review_task` filters exactly that out and would return `None`, taking the `forget_review_task` branch and returning `(None, 0)`. The test asserts a state production cannot reach, so it pins the invalidation half of the behaviour but actively mis-describes the re-resolution half.
2. `test_a_cached_task_that_belongs_to_another_pr_is_dropped` (line 5449 onward) caches `TASK-999` titled for PR #999, and the assertions that matter (`get_calls`, `list_calls`) hold — but the returned task is again whatever the fake holds, so the test would still pass if the search's predicate were removed entirely.

`test_the_cache_check_and_the_search_share_one_predicate` does exercise the real predicate directly, so coverage is not absent — it is just not where these tests imply it is. Teaching the fake to filter (`[t for t in self.list_tasks() if is_review_task_for(owner, repo, number, t)]`, which is a one-liner given `is_review_task_for` is already imported at line 426) would make it mirror the real contract the way the `tasks=None` branch already does, and would let these two tests assert the outcome rather than the call count.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
```diff
@@ -143,6 +149,22 @@
...
+-- which is exactly what makes it cacheable. A row here is a HINT, never a
+-- fact: it is re-checked by ref on use and dropped when it stops matching, and
+-- losing the whole table only costs one title search per PR (see
+-- loop._review_task). That is why this is best-effort like the snapshots
+-- above, not a correctness write.
+CREATE TABLE IF NOT EXISTS review_tasks (
+    repo        TEXT    NOT NULL,
+    number      INTEGER NOT NULL,
+    task_ref    TEXT    NOT NULL,
+    resolved_at INTEGER NOT NULL,
```
alissa-app, line 164: [nit] `resolved_at` is write-only, and the table is the one new cache with no retention story.

Nothing reads the column — `record_review_task` writes `int(time.time())` and no query, log line, or console view ever selects it. The docstring calls it *"the audit question worth answering"*, which is a fair reason to keep it, but as written no one can ask the question.

Relatedly, `review_tasks` rows are never pruned. `poll_snapshots` prunes on write and the module header explains why; `grants` are documented as never pruned and why; this table gets one row per PR ever sighted and says nothing either way. For this daemon's repo set that is a few hundred rows forever — genuinely immaterial, which is why this is a nit and not a minor. A half-sentence in the schema comment ("rows are never pruned; one per PR sighted, and a stale row costs one disproved read") would put it in line with the neighbouring tables' documentation.

alissa-app (approved):

Review task: **TASK-884621378** (created by this session per CR2 — the ready flip left none; filed downstream of the implementation task because origin TASK-1373942374 404s for the reviewer actor).

The change does what it says. The decide path's `find_review_task` was the one caller that never got the treatment the sweep paths already had, and the three leverage points land: a persisted `(repo, number) -> task_ref` row consulted first, a single-task read-back that folds the round count into the same payload, and a per-pass memo that collapses the same-second bursts into one fetch. The fail-open reasoning is the strongest part of the diff — the distinction between "read back and disproved" and "could not be read" is drawn correctly in both the code and the tests, and `_pass_tasks` is checked with `is None` rather than truthiness, so a legitimately empty corpus memoizes instead of re-fetching.

**Verified independently, not taken from the PR body** (throwaway worktree, `PYTHONPATH=src/main`, scratchpad venv):

| check | result |
| --- | --- |
| `pytest src/test` | **603 passed**, 0 failed |
| coverage | **93%** total; `loop.py` 96%, `state.py` 95%, `alissa.py` 83% |
| `check-style.sh` | clean |
| `check-types.sh` | `Success: no issues found in 29 source files` |
| version | `origin/main` = 0.16.9, open #67 claims 0.16.10 → **0.16.11 is the right pick** |

I also closed the PR body's third "not verified" box for you: `alissa task get <ref> --json` on this container returns the task object at **top level** with `taskNumber` / `title` / `status` / `evidence` — exactly the shape `_task_from_row` and `_count_verdicts` expect. The nested-payload failure mode you flagged does not occur against the shipped CLI.

The judgment call on leverage #2 is correct and I checked the premise myself: `alissa task list` exposes only `--json` and `--include-terminal`, so implementing `--status` as the issue wrote it would raise `CommandError` on every fallback search and turn a review into a `SKIPPED` decision — the one outcome the issue's own fail-open constraint forbids. Registering TASK-2019528326 rather than shipping a broken narrowing is the right disposition.

### Findings

No blockers, no majors. Four minors, one question, one nit — none of them gate the merge, and the question is **not load-bearing** (it asks about behaviour this PR leaves exactly as it found it).

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 4 |
| nit | 1 |
| question | 1 (open: 1, none load-bearing) |

### Dimension report

- **Correctness** — nothing found. The miss path is call-for-call identical to the old one (`list_tasks` + `count_verdicts`); the hit path swaps a corpus fetch plus a task fetch for a single task fetch that answers both questions. Round counting cannot regress: `verdicts` is read live every pass, and `is_review_task_for` requires the title to name *this* PR, so a cached ref can never pin another PR's task. A stale `completed` would in any case be caught by the `spawns` row keyed on `(repo, number, round)`.
- **Coding practices** — matches the repo's idiom closely, including its unusually high comment-to-code ratio and the "explain the failure mode, not the mechanism" docstring style. Two minors below are about a comment this change invalidated and about test-double fidelity.
- **Security** — nothing found. Every new statement is parameterised; `run` never uses `shell=True`, so a `task_ref` derived from CLI output reaches `alissa task get` as an argv element and cannot inject; `_title_pattern` already `re.escape`s owner and repo; the new table holds a task ref and a timestamp, no secrets; the new log lines carry a task ref, title and status, all of which this module already logs elsewhere.
- **Performance** — this is the dimension the PR exists for and it delivers: steady state drops from ~2-4 full-corpus fetches per minute to zero, and a pass that must search buys the corpus once for every PR that missed. One reduction is left on the table (minor 3).
- **Side-effects / blast radius** — `CREATE TABLE IF NOT EXISTS` is additive with no destructive migration, and the read-only console never runs the schema and never reads this table, so a pre-cache database opened read-only is unaffected. `find_review_task`'s new parameter is keyword-only and defaulted, so `prreview.py` is untouched. One behavioural excursion the PR body does not mention: dry-run passes now write to the ledger (minor 1).

### Summary

A well-argued, well-tested change that removes the daemon's single largest read without weakening any invariant it depends on; the delivery note's judgment calls hold up against the shipped CLI, and every claim in it that I could check independently checked out. Approving. The four minors are all "fix now or register a follow-up" (CR8 `[triage:pursue]` or `[triage:later]`), not merge gates; minor 1 is the only one I would prefer to see fixed in this PR, since it is a two-line comment correction protecting a safety argument a future reader will rely on.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue]

Correct, and the diagnosis is exact: `_review_task` is the first thing `evaluate` reaches, `record_review_task` / `forget_review_task` are unguarded, and `evaluate` runs in dry-run. So the gate's "its only ledger write is the snapshot" is now false, and on the read-only volume dry-run exists to diagnose, each cache write costs a reconnect attempt on the first failure of the streak plus streak-limited warnings into the operator's diagnostic.

Taking your recommendation — amend the comment rather than guard the writes. A dry-run pass that learns a PR's mapping and hands it to the next production pass is a real (small) benefit, and suppressing it would mean dry-run and production disagree about ledger contents for no correctness reason. The gate's justification will name both best-effort writes, snapshot and review-task cache, so the exemption argument stays true rather than merely still-concluding-correctly.

Fixed in the follow-up PR (approve is terminal for this branch).

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] — partly here, and the rest is registered as **TASK-317167904**.

You are right on both counts, including that it reads as unnoticed. It was: I did not consider that the CR2 alarm lives *inside* the corpus scan and therefore dies with it. The consequence is exactly as you state — a duplicate created after the mapping is cached is never reported, and the daemon keeps counting envelopes on whichever task it pinned first.

Splitting it the way you suggest:

* the accepted-consequence half — a `_review_task` docstring line ("a cached hit does not re-check CR2 uniqueness") and a judgment-call bullet in the delivery note — lands in the follow-up PR, since the note as it stands claims a completeness it does not have;
* the fix half — relocating the detection so it still fires for cached PRs, with `_pass_task_list()`'s result as the natural home when a pass fetches it — is **TASK-317167904** (soft-linked to TASK-2141858575). Its DoD carries your constraint: report the duplicate without reintroducing a corpus fetch per PR per poll.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: [triage:pursue]

Agreed, and "conspicuous" is fair — the PR argues about not paying for reads twice and then leaves the second read in place on the hot path. `_newest_verdict` and `_count_verdicts` are already mirrors over one array, so the payload `get_task` has in hand answers both.

`TaskDetail` gains `verdict: str | None` populated in the same `try` block, threaded out of `_review_task` and into `_convergence_reason`. One design point I will make explicit rather than paper over: the verdict is only known on the **cached** path — the miss path resolves through `find_review_task` + `count_verdicts` and never reads a full payload — so the value is passed as "already known, do not re-fetch" and the miss path keeps calling `latest_verdict`. Anything else would silently turn a miss into "no verdict", which is the direction that loses a round.

Leaving `_close_round_natively` alone, per your parenthetical: rare and terminal.

Fixed in the follow-up PR.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:answer] — and the honest answer is no, it was not deliberate. Registered as **TASK-1897198077**.

I did not weigh it. `is_review_task_for` was written to make the cache check and the search agree, and requiring `is_open` came from the search side without my noticing it also decides what happens when a review task is validated. Your reading of the asymmetry with `_completed_rounds` (`loop.py:2643`) is correct, as is the observation that this PR is the first time the decide path has a durable ref that would survive validation.

Having now thought about it, the invalidation problem you offer as a possible rationale is real and is the reason I am not fixing it inside a follow-up to this PR: a terminal task can never become open again, so a retained ref has no natural disproof and its row would be immortal — the opposite of the disproof-only invalidation rule this PR is built on. Making it work means deciding how such a row is ever retired, and letting the cache answer "here is the ref, for counting" while still telling `_spawn` "no open review task". That is a design change, not a docstring.

So: **TASK-1897198077** (soft-linked to TASK-2141858575) carries it, with that question written into its DoD, and the follow-up PR adds the docstring line next to the other three outcomes recording that a validated task drops the mapping and why. Your severity read is the one I am acting on — pre-existing, unchanged, does not block.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:pursue]

You are right, and this is the finding I am least comfortable with: the two tests assert a state production cannot reach. `find_review_task` filters on exactly the predicate the fake ignores, so in reality a `validated` task is not returned by the search at all — the pass would take the `forget_review_task` branch and return `(None, 0)`, not the task.

Teaching the fake to filter, as you spell out. Consequences I will handle rather than assert around:

* `test_a_cached_task_that_no_longer_matches_is_dropped_and_re_resolved` currently asserts `task is al.task`; against a filtering fake the truthful assertion is `task is None` plus the dropped row, so the test gets rewritten to pin the re-resolution it claims to pin — and a second case is added where the corpus *does* hold a valid replacement, which is the "re-resolved" half that no test currently covers at all;
* `test_a_cached_task_that_belongs_to_another_pr_is_dropped` keeps its `get_calls` / `list_calls` assertions and gains the outcome assertion, so removing the search predicate would now fail it.

Fixed in the follow-up PR.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
RHDZMOTA: [triage:pursue]

Both fair. `resolved_at` is genuinely write-only today — the docstring names an audit question nobody can ask — and `review_tasks` is the only new table here that says nothing about retention while its neighbours all do.

Taking the documentation fix rather than dropping the column: the value costs nothing, and a row's age is the first thing worth knowing if a mapping is ever suspected of being wrong. The schema comment will say what you propose — rows are never pruned, one per PR ever sighted, and a stale row costs one disproved read — and `resolved_at` will be described as retained-for-inspection rather than as something the daemon consults.

Fixed in the follow-up PR.

PR #69, opened 2026-07-30
Follow-up to #68: round-1 findings — one task read per pass, a filtering fake, and three corrected records (revloop 0.16.12)

RHDZMOTA:
Follow-up to PR #68 (round 1, `alissa-app`, APPROVED with 5 `[minor]` + 1 `[nit]` + 1 `[question]`).

Follow-up-To: #68

Origin task: TASK-1373942374
Implementation task: TASK-2141858575

⚠️ **Merge #68 first.** This branch is cut from #68's approved head (`92db55f`), because every line it corrects was introduced by #68 and exists nowhere on `main` yet. GitHub's diff against `main` therefore shows #68's changes too; the review surface here is `92db55f..HEAD`, one commit. Merging this ahead of #68 would squash #68's entire delta into `main` under this PR's commit message and leave #68 resolving as empty — losing its history and its issue claim.

An approve is terminal for the branch it approved, so none of this could be pushed to #68.

## What each finding became

| Finding | Disposition |
|---|---|
| `[minor]` `TaskDetail` discards the newest verdict, so the round still pays a second `alissa task get` | Fixed — `TaskDetail.verdict` + `ResolvedTask` |
| `[minor]` the ledger gate's dry-run safety argument is now false | Fixed — comment amended (the reviewer's preferred option) |
| `[minor]` `FakeAlissa` returns its task unfiltered, so two tests can't fail | Fixed — fake filters; one test corrected, one added |
| `[minor]` a cached hit retires the CR2 duplicate-task alarm | Documented here; relocation is **TASK-317167904** |
| `[question]` a validated review task's ref is dropped | Answered (not deliberate); design work is **TASK-1897198077** |
| `[nit]` `resolved_at` is write-only and the table has no retention story | Fixed — schema comment + docstring |

The largest one is the first. `_count_verdicts` and `_newest_verdict` are deliberately mirrors over one evidence array, so the payload that validates the cached mapping already answers all three questions the decide path asks. `ResolvedTask` carries them out — and records **whether the verdict was read**, which is the part that matters: a verdict nobody looked for is *unread*, not `None`, and collapsing the two would make convergence read "nobody looked" as "no approve" and refuse to close a round that had in fact approved. (Round 1 justified `verdict_read=False` on the search path by claiming that path never opens the evidence. It did — via `count_verdicts` — and round 2's second `[minor]` is that correction: the search path now reads its task once too, and `verdict_read=False` survives only for the case where that read genuinely fails.) Both paths are now one task fetch per PR per poll.

## Round 2 — what round 1's findings became

Verdict was `request_changes`: 1 `[major]`, 2 `[minor]`, 3 `[nit]`, 0 blockers. All six triaged `[triage:pursue]` and all six fixed in `4438320`; nothing deferred, nothing registered.

- **`[major]` the ordering assertion certified a guarantee production did not have.** Reproduced before triaging, against this session's own task rather than the reviewer's example: `alissa task get --json` dates evidence with epoch **milliseconds** (`1785426552562`), and `_newest_verdict`'s `created if isinstance(created, str) else ""` key collapsed every real item to `""`, leaving `max` to keep the **first** of an all-equal set. Evidence arrives oldest-first, so on live data the **oldest** verdict won. Fixed at the source, not just in the fixture — this release is what promotes that value to convergence's input, and a corrected test in front of a known-wrong function would have been the smaller half. Both shapes normalise to one float in `_created_key`; the isinstance was *not* widened in place, because `(str, …)` and `(int, …)` keys raise `TypeError` the moment sorting compares them on a task holding both. The append index breaks ties, so equal stamps resolve to the later-recorded envelope. **This closes TASK-194837655**, which the reviewer registered as pre-existing.
- **`[minor]` the search path still read one task twice.** `count_verdicts` opened the evidence and threw the verdict away, then convergence re-read and re-parsed it — #68's finding verbatim, one path over. Now one read, falling back to the old pair only when that read fails. The two docstrings asserting the path "never opens the evidence" stated the opposite of what it did and are corrected.
- **`[minor]` the fake still answered `count_verdicts` silently**, so no test could state the search path's cost. Every task read the fake serves is now recorded, and both paths' costs are asserted rather than argued.
- **`[nit]`** `_resolved` → `ResolvedTask.from_detail`; **`[nit]`** the fake's `list_tasks` now omits terminal tasks as the CLI does, so a validated task is genuinely absent from the search instead of merely filtered out of its answer; **`[nit]`** PR #69 is attached to TASK-2141858575 as a deliverable.

### Round 2 verification

- `tests-unit.sh` — **624 passed**, 0 failed (610 → +14).
- `tests-coverage.sh` — **93%** total; `loop.py` 96%, `state.py` 95%, `alissa.py` 84%.
- `check-style.sh` — clean. `check-types.sh` — `Success: no issues found in 29 source files`.

Counterfactuals, each reverted in the worktree with the suite re-run:

| reverted | fails |
|---|---|
| the whole new sort key, against the corrected fixture | `test_newest_verdict_envelope_wins`, `test_undated_envelope_loses_to_a_dated_one`, `test_get_task_carries_the_newest_verdict_out_of_the_same_payload` |
| the index tie-break | `test_equal_stamps_resolve_to_the_LAST_envelope` |
| the ms→s normalisation | `test_iso_and_epoch_stamps_order_against_each_other`, `test_seconds_and_milliseconds_are_the_same_instant` |
| the bool guard | `test_an_unusable_stamp_sorts_first[True]`, `[False]` |

The first row is the one worth reading: two of those three tests **predate this PR** and were green against a shape the CLI never emits. That is the reviewer's point demonstrated rather than accepted.

### Round 2 — not verified

- [ ] **The sort-key fix changes live convergence behaviour** and is exercised only against fixtures. A PR whose round 1 was `request_changes` and round 2 `approve` should now converge through the envelope branch where it previously ran to the round cap; that is the observable an operator can check post-deploy.
- [ ] The epoch-ms shape is confirmed from two live tasks in this container. A deployment whose API returns ISO strings is still handled (both shapes normalise), but the mixed-shape ordering is fixture-only.

## Round 3 — the approve, and what it carried

Round 2 verdict: **approve** on `4438320`, all six round-1 findings confirmed fixed. It carried 2 `[minor]` + 3 `[nit]`, none of them merge gates and all of them records or edges rather than logic.

All five are `[triage:later]` with a registered task — **not** a third stacked PR. An approve is terminal for its branch (a commit pushed after the squash-merge is silently lost), and a follow-up to this follow-up would force the operator into an exact three-PR merge order for one stale comment, two test-record corrections, a timezone edge that cannot currently be reached, and a call-count trade-off that is a decision rather than a fix.

| finding | task |
|---|---|
| `[minor]` `_convergence_reason`'s comment still says the search path pays the verdict fetch — false since `4438320` | **TASK-1207414354** |
| `[minor]` `test_convergence_sees_the_newest_verdict_on_real_evidence` never reaches convergence (unused `config` fixture is the tell) | **TASK-1207414354** |
| `[nit]` `get_calls == [FakeTask.ref] * 2` asserted under the message "one small read" | **TASK-1207414354** |
| `[nit]` a naive ISO stamp is parsed in the process's local timezone while the epoch-ms branch is UTC | **TASK-639960880** |
| `[nit]` the search path re-reads a ref whose read just failed, costing one call more in that one degraded branch | **TASK-72507406** |

All three are soft-linked to TASK-2141858575. The first is the one worth naming plainly: it is the *third* instance of the record defect round 1's minor 1 identified, missed because I corrected the two places that finding pointed at rather than the class of thing it described.

### Judgment calls

- **`ResolvedTask` is a new type rather than a wider tuple.** `_review_task` already returned two values and this adds two more, one of which (`verdict_read`) exists only to stop a `None` from meaning two different things. A four-tuple would have made that distinction invisible at every call site.
- **The dry-run comment was amended, not the writes guarded** — the reviewer's stated preference, and the stronger option: a dry-run pass that learns a mapping hands it to the next production pass, and suppressing that would make dry-run and production disagree about ledger contents for no correctness reason.
- **Two findings were split rather than fixed or deferred whole.** The CR2-alarm and validated-task findings each have a cheap honest half (write down the consequence where the next reader will hit it) and an expensive half that is a design change. The docstring halves are here; the design halves are TASK-317167904 and TASK-1897198077, both soft-linked to TASK-2141858575, both with the open question written into their DoD.
- **The `[question]` was answered "no, that was not deliberate."** It would have been easy to claim the invalidation argument as the original reasoning — it is a good argument and it is why the fix is not attempted here — but it was reconstructed after the fact, and a delivery note that launders hindsight into intent is worth less than one that doesn't.
- **Version 0.16.12, past #68's 0.16.11.** A stacked follow-up inherits its source's version, which passes `check-version-bump` only until the source merges and publishes it; from that moment any re-run fails.
- **No third level.** Anything this round raises gets registered, not stacked — three-deep stacking forces the operator into an exact merge order (#68 → this → next) for a worse trade than registered follow-ups.

- [ ] **The merge-order gate above is the operator's to enforce.** Nothing in CI checks it. If this merges first, #68 resolves as empty.
- [ ] **The traffic claim is still unmeasured.** #68's acceptance criterion 2 (the post-release Studio `listMyTasksCall` census showing the `n≈463` entries gone) needs a publish, a container re-pin and a redeploy; this PR halves what remains on the cached path but moves that criterion no closer to being observed.
- [ ] **The container pin is still `ARG REVLOOP_VERSION=0.16.6`.** If both PRs merge, the version to pin is **0.16.12**, not #68's 0.16.11.
- [ ] **`_close_round_natively`'s `latest_verdict` call is untouched**, per the reviewer's parenthetical that it is rare and terminal. It remains a second task read on that path.
- [ ] **Still no execution against the real `alissa` CLI** — the verdict now rides on `alissa task get`'s payload shape (top-level `evidence`), same assumption #68 recorded, now load-bearing for convergence rather than only for the round count. A CLI that nested the payload would make `get_task` return None, which falls open to the search path and its own verdict fetch.

### Verification

Run from the follow-up worktree, `PYTHONPATH=src/main` inlined per command:

- `bash tests-unit.sh alissa-tools-github-revloop` — **610 passed**, 0 failed (603 at #68's head, +7).
- `bash tests-coverage.sh alissa-tools-github-revloop` — **93%** total; `loop.py` 96%, `state.py` 95%, `alissa.py` 83%.
- `bash check-types.sh alissa-tools-github-revloop` — `Success: no issues found in 29 source files`.

**Counterfactual-tested**, because a fix for "this test cannot fail" that is itself untested would be the same defect again. Each was reverted in the worktree and the suite re-run:

- unthreading the verdict (`verdict=None` in `get_task`) → `test_get_task_carries_the_newest_verdict_out_of_the_same_payload` fails;
- `verdict_read=False` on the cached resolution → `test_the_cached_path_reads_the_task_exactly_once_per_pass` fails;
- reverting the fake to return its task unfiltered → both invalidation tests fail.

The second of those **did not fail on the first attempt**, and the reason is worth recording: `FakeAlissa.latest_verdict` answered silently, so a re-fetch left no trace and the test's "one read per pass" assertion could not see it. The fake now records verdict fetches and the test asserts on that. Same defect class the reviewer found in finding 5, one layer down — caught only because the counterfactual was run.

One further false green: the first counterfactual was executed from the repo root, where `PYTHONPATH=src/main` resolves to nothing, so it silently tested the installed wheel and "passed". Re-run from the package directory it failed as it should.

### Scope touched

Declared scope `alissa-tools-github-revloop/**` — no excursion.

- `alissa.py` — `TaskDetail.verdict`, populated in `get_task`'s existing `try` block.
- `loop.py` — new `ResolvedTask`; `_review_task` returns it; `_convergence_reason` takes it and no longer fetches a verdict it already has; the ledger gate's dry-run comment; two accepted-consequence paragraphs in `_review_task`'s docstring.
- `state.py` — schema and docstring comments only, no behaviour.
- `version` — 0.16.11 → 0.16.12.
- `test_loop.py` — `FakeAlissa` filters, omits terminal tasks from `list_tasks`, and records every task read; two invalidation tests corrected; six tests added in round 1 and fourteen in round 2.

Round 2 additionally touches `alissa.py`'s `_created_key` / `_newest_verdict` (the `[major]`) — an excursion beyond this follow-up's original surface, taken deliberately because the reviewer offered it and because the alternative was documenting a known-wrong ordering under a green test.


alissa-app:
- 2026-07-30 16:18:01 UTC — `review-alissa-github-review-daemon-pr69-r1-81914e` — round 1 of 22 — spawned
- 2026-07-30 16:44:05 UTC — `review-alissa-github-review-daemon-pr69-r2-e2594e` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -5336,3 +5387,450 @@ def test_a_production_pass_is_not_bounded_by_a_dry_run_one(config, caplog):
...
+            ],
+        }
+
+    monkeypatch.setattr(alissa_mod, "run_json", fake_run_json)
+    detail = alissa_mod.Alissa().get_task("TASK-500")
+
+    assert len(calls) == 1
+    assert detail is not None
+    assert detail.verdicts == 2
+    assert detail.verdict == VERDICT_REQUEST_CHANGES, "the NEWEST, not the first"
```
alissa-app, line 5875: `[major]` This certifies "the NEWEST, not the first" against a payload shape the CLI does not produce, so the guarantee it pins does not exist in production.

`envelope()` (line 1217) dates items with an ISO-8601 **string**. `alissa task get <ref> --json` dates them with an epoch-millisecond **int** — confirmed against live data:

```
$ alissa task get TASK-884621378 --json | jq '.evidence[].createdAt'
1785427358482
1785427508445
```

`_newest_verdict`'s sort key is `created if isinstance(created, str) else ""`, so every real item collapses to `""`, and `max` over all-equal keys keeps the **first** element. Evidence comes back oldest-first, so the oldest verdict wins. Reproduced against the production function, no fake involved:

```python
from alissa.tools.github.revloop.alissa import Alissa
ev = lambda v, at: {"title": f"Review verdict: o/r#7 — {v}", "createdAt": at}
Alissa._newest_verdict({"evidence": [ev("request_changes", 1785427358482),
                                     ev("approve",         1785427508445)]})
# -> 'request_changes'      the real shape: round 1's verdict wins
# with ISO strings instead -> 'approve'   the shape this test feeds it
```

And the counterfactual this test is missing: swap the two `createdAt` values here for those integers and it fails with `assert 'approve' == 'request_changes'`. I ran it.

To be exact about blame — **this is not a regression.** The sort key dates to `674ca6d` (0.12.0) and `latest_verdict` on `main` has always fed the same helper, so `main` behaves identically today. Registered as **TASK-194837655**, with the two live consequences written up there: a PR whose round 1 was `request_changes` and round 2 `approve` never converges through the envelope branch and runs to the cap, and `_close_round_natively` would post `REQUEST_CHANGES` as the native review state for a round that approved.

What makes it a `[major]` rather than a note on a pre-existing bug is that it is *this* delta's test, in the PR that promotes this value from "only the round count" to convergence's input, and whose delivery note's strongest claim is that every fix was counterfactually tested. A green assertion saying the ordering is right is what will stop the next reader from checking.

Cheapest correct fix: use the epoch-ms shape in `envelope()` (or at least here), watch this fail, then either normalise the sort key — carefully: a tuple of `(str, str)` and `(int, str)` raises on comparison, so normalise rather than widening the `isinstance` in place — or assert the real behaviour with a comment pointing at TASK-194837655. `test_newest_verdict_envelope_wins` (line 1236) shares the same fixture and the same blind spot.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -964,6 +1010,114 @@ def __init__(
...
+            # no open review task for this PR, whatever the cache said.
+            if cached is not None:
+                self.state.forget_review_task(pr.full_name, pr.number)
+            return ResolvedTask(task=None)
+
+        self.state.record_review_task(pr.full_name, pr.number, task.ref)
+        # The search read TITLES, not evidence, so the newest verdict is
+        # genuinely unknown here -- left unread rather than defaulted to None,
+        # which convergence would read as "no approve" and lose a closed round.
+        return ResolvedTask(task=task, verdicts=self.alissa.count_verdicts(task.ref))
```
alissa-app, line 1137: `[minor]` The search path still pays two `alissa task get` calls for one task, and the justification one screen up says it cannot.

`count_verdicts(task.ref)` here fetches the task and parses its whole `evidence` array — then throws the verdict away. `_convergence_reason` then reaches `resolved.newest_verdict(self.alissa)`, sees `verdict_read=False`, and calls `latest_verdict(task.ref)`, which fetches the same task and re-parses the same array with the same regex. That is #68's round-1 `[minor]` verbatim ("`TaskDetail` fetches the newest verdict and throws it away, so the round still pays a second `alissa task get`"), fixed on the cached path and still standing here.

`ResolvedTask`'s docstring (line 887) reads "the SEARCH path matches TITLES out of the task corpus and never opens the evidence, so it pays a separate `count_verdicts` and knows no verdict" — but the evidence *is* opened, by the `count_verdicts` on this line. `verdict_read=False` is therefore not "we could not know", it is "we parsed it and discarded it", which is a different claim and a fixable one. The comment two lines above ("The search read TITLES, not evidence") says the same thing.

`get_task(task.ref)` already returns both numbers from one call, so this can be:

```python
detail = self.alissa.get_task(task.ref)
if detail is not None:
    return self._resolved(detail)
return ResolvedTask(task=task, verdicts=self.alissa.count_verdicts(task.ref))
```

— one corpus fetch plus one task fetch, degrading to exactly today's behaviour when the read fails.

Worth more than first-sighting arithmetic suggests: the search path is also taken *every pass for every PR* whenever `state.review_task` cannot be read — the degraded-ledger state this whole change exists for. In that state the daemon currently pays a corpus fetch plus two task fetches per PR per pass, where one task fetch is removable.

If you would rather not touch the resolution path in a follow-up PR, that is a fair call — but then the two docstrings should say "discards it" rather than "never opens the evidence", and it should be registered.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -254,11 +257,59 @@ def __init__(self, task=None, verdict=None, verdict_count=0):
...
         self.sessions: list = []  # live ManagedSessions, as `alissa tmux ls` sees them
+        # Full-corpus `alissa task list` fetches this fake has served. The
+        # expensive call issue #66 is about, so tests count it directly.
+        self.list_calls = 0
+        # Refs read back one at a time via `alissa task get`, in order.
+        self.get_calls: list[str] = []
+        # Refs `latest_verdict` was asked for. Recorded because a resolution
+        # that already holds the verdict must not ask again, and a fake that
+        # answered silently would hide exactly that regression.
+        self.verdict_calls: list[str] = []
```
alissa-app, line 269: `[minor]` The fake now records `latest_verdict` but still not `count_verdicts`, so the suite cannot see the search path's second task read.

`count_verdicts` (line 315) returns `self.verdict_count` and records nothing, while `get_calls` and this new `verdict_calls` make every other task read visible. The consequence is that `test_the_cached_path_reads_the_task_exactly_once_per_pass` can assert "ONE task read for the whole pass" on the cached path, but no test can state the search path's cost at all — `test_the_search_path_still_fetches_the_verdict_it_never_read` asserts `verdict_calls == [FakeTask.ref]` and is silent on the `count_verdicts` fetch that preceded it.

That is the same shape as #68's finding 5, which this PR fixed for `find_review_task`: a double that answers silently lets a cost regression through undetected. Appending to a `count_calls` list here (and asserting on it in the two tests above) would make minor 1 visible instead of arguable, and would pin the search path's cost the way the cached path's is now pinned.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -964,6 +1010,114 @@ def __init__(
         # that cannot be written.
         self._ledger_streak = Streak()
 
+    # -- the PR -> review-task mapping ---------------------------------------
+
+    def _resolved(self, detail: TaskDetail) -> "ResolvedTask":
```
alissa-app: `[nit]` `_resolved` never touches `self` — it is a `TaskDetail -> ResolvedTask` conversion sitting on `ReviewWatcher`. As `ResolvedTask.from_detail(detail)` (a `classmethod` next to the dataclass, which is where its docstring's "one payload answered all three questions" belongs) the invariant `verdict_read=True` would live with the type that defines it, and the call site would read `return ResolvedTask.from_detail(detail)`. Discretionary — the module does keep other small helpers on the watcher.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -254,11 +257,59 @@ def __init__(self, task=None, verdict=None, verdict_count=0):
...
+            return None
+        found = next((t for t in self.corpus if t.ref == ref), None)
+        if found is None:
+            return None
+        return TaskDetail(
+            task=found, verdicts=self.verdict_count, verdict=self.verdict
+        )
+
+    @property
+    def corpus(self):
```
alissa-app: `[nit]` This docstring promises a case `corpus` cannot express: "`others` holds anything else a test needs to exist — a task the search must NOT match, **or one that is readable by ref while absent from the (non-terminal) list**." But `corpus` *is* what `list_tasks` returns, so anything in `others` is in the list; `Validated()` stays out of the search's answer only because `is_review_task_for` filters it, not because it is absent.

Harmless today, and the predicate filter is genuinely the better design — but the real `alissa task list` omits terminal tasks outright (`--include-terminal` is off), so the one divergence between fake and CLI is now described backwards, in the docstring a future test author will read before reaching for `others`. Either drop that clause or give `get_task` its own lookup table so the absence is real.

alissa-app (changes requested):
## Review verdict: **request_changes** — round 1 of cap 22

Review task: **TASK-762291129** (created by this session per CR2 — the ready flip left none; filed downstream of the implementation task TASK-2141858575, because origin TASK-1373942374 404s for the reviewer actor). The cap recorded on it is this daemon's **effective** cap — `ALISSA_ROUND_CAP=22` plus any operator re-entry grants — not the alissa-code-review template's default of 3.

I took the body's framing on the review surface and judged `92db55f..227a3d4` only, treating #68's approved delta as given. The merge-order argument is correct as written: `92db55f` is an ancestor of this head, so a merge commit here lands #68's delta into `main` under this PR and leaves #68 resolving as empty.

**Verified independently, not read off the PR body** (scratchpad venv, run from `alissa-tools-github-revloop/` with `PYTHONPATH=src/main` inlined):

| check | result |
| --- | --- |
| `pytest src/test` | **610 passed**, 0 failed (603 at `92db55f`, +7) |
| coverage | **93%** total; `loop.py` 96%, `state.py` 95%, `alissa.py` 83% |
| `pycodestyle --max-line-length 120` | clean |
| `mypy --ignore-missing-imports` | `Success: no issues found in 29 source files` |
| cited follow-ups | TASK-317167904, TASK-1897198077, TASK-2019528326 — all real, all `committed` |
| #68's triage | all six findings carry exactly one `[triage:*]` reply (5 `pursue`, 1 `answer`) |

The largest fix holds, and I checked the part that could have gone wrong: `TaskDetail.verdict` is fed by the *same* `_newest_verdict` that `latest_verdict` calls, over the same payload, so routing convergence through the cached value changes no answer that the extra fetch would have given. `verdict_read` is not ceremony either — collapsing it into `verdict=None` really would make the search path read "nobody looked" as "no approve", and `test_an_unread_verdict_is_not_a_missing_one` pins exactly that. The docstrings are unusually honest about the two accepted consequences, and both halves are backed by real open tasks.

Two things I want to name because the delivery note is careful enough to deserve it: reverting the fake to answer `latest_verdict` silently, and re-running the counterfactual from the wrong directory so `PYTHONPATH=src/main` resolved to nothing, are both recorded rather than quietly fixed. That is the reason the `[major]` below is a fixture and not a behaviour bug — the method is working.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 2 |
| nit | 3 |
| question | 0 (open: 0) |

One nit does not fit a diff line, so it lives here: **`[nit]`** PR #69 is not on TASK-2141858575 as a deliverable or evidence item — #68 has `Draft PR #68` plus a `Delivery note`, and #69 appears only inside a task comment. C4 wants the PR on the task as evidence; a reader opening the implementation task sees one PR where there are two.

### Dimension report

- **Correctness** — the delta's production logic is right, and the one thing it could have broken it does not: `completed` and the convergence verdict now come out of a *single* payload, which is strictly more self-consistent than the old pair of reads (previously `count_verdicts` at T1 and `latest_verdict` at T2 could straddle a verdict landing mid-pass). `_convergence_reason`'s early returns still short-circuit ahead of any verdict read, so round 1 pays nothing before a review exists. The `[major]` is a test that certifies a guarantee production does not have.
- **Coding practices** — matches the repo's idiom, including its comment density and its "name the failure mode" docstring style. Two of the three nits are here. The corrected `FakeAlissa` is a genuine improvement: applying `is_review_task_for` to the pool means deleting the search's predicate now fails a test, which was the whole point of #68's finding 5.
- **Security** — nothing found. The delta adds no new statement, subprocess, or interpolation: `alissa.py` gains one field off an already-parsed payload, `loop.py` adds a dataclass and one method that calls an existing client, and `state.py` is comments only. New log lines carry nothing this module does not already log.
- **Performance** — the cached path is genuinely one task read per pass and the test pins it. The search path is where a reduction is left on the table, and it is the same reduction #68's round 1 asked for, one path over (minor 1).
- **Side-effects / blast radius** — `state.py` is comment-only, so there is no schema change and no migration in this delta. `TaskDetail` gains a required field, but the only construction sites are `get_task` and the fake, and 0.16.11 is still unpublished, so no consumer can be broken. `0.16.12` is free on PyPI and past #68's 0.16.11, so `check-version-bump` survives either merge order. The real exposure is the merge-order gate the body flags, which is unenforceable in CI and correctly handed to the operator; the container pin at `ARG REVLOOP_VERSION=0.16.6` remains untouched, so merging this deploys nothing.

### Summary

A disciplined follow-up that does what it claims: the cached path now costs one `alissa task get` per PR per pass, the fake stops asserting states production cannot reach, and the two findings too large for a follow-up are split honestly rather than waved through. Every check in the delivery note reproduced exactly, including the coverage figures. I am at `request_changes` for one reason: `test_get_task_carries_the_newest_verdict_out_of_the_same_payload` asserts "the NEWEST, not the first" against an ISO-string `createdAt` that `alissa task get --json` never emits, and against the shape it *does* emit the assertion is false — so the round that promotes this value to convergence's input signs off on ordering it does not have. The underlying misparse is pre-existing and registered as **TASK-194837655**; the shortest path back to approve is to make the fixture use the real epoch-ms shape and then either fix the sort key here (one line, and this PR is already the reason the value matters) or assert the real behaviour with a pointer to that task. Everything else is `[triage:pursue]`-or-register, not a merge gate.

RHDZMOTA:
- 2026-07-30 16:33:13 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr69-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:pursue] — and reproduced before triaging, against my own implementation task rather than your example, so the shape is confirmed twice independently:

```
$ alissa task get TASK-2141858575 --json | jq '.evidence[].createdAt'
1785426552562
1785426561222
```

and the production function on the two shapes:

```
real epoch-ms shape -> request_changes      # the OLDEST envelope
ISO string shape    -> approve
```

You are right about all three parts: the fixture is wrong, the assertion is false against real data, and it is this PR's test in the PR that promotes the value to convergence's input. The "counterfactually tested" claim in the note makes a green ordering assertion worse than no assertion, because it is exactly what stops the next reader from checking.

Taking your first option — **fix the sort key here**, not just the fixture. This PR is the reason the value matters, `_newest_verdict` now feeds convergence through `TaskDetail.verdict`, and leaving a known wrong-order function behind a corrected test would be the smaller half of the fix.

Heeding your warning about *how*: not by widening the `isinstance`, which would leave `(str, str)` and `(int, str)` tuples comparing against each other and raising `TypeError` on a task holding both shapes. Both are normalised to one numeric stamp instead — epoch-ms detected by magnitude and scaled to seconds so a mixed task still orders correctly, ISO parsed, unparseable and absent stamps sorting first as they do today. The append index becomes the tiebreaker, so equal stamps resolve to the LAST envelope (evidence comes back oldest-first) rather than the first.

`envelope()` moves to the real epoch-ms shape, so `test_newest_verdict_envelope_wins` and the rest of that family now exercise production's actual input. Counterfactual will be recorded in the note: the corrected fixture against the old sort key.

This closes **TASK-194837655**; I will say so on the task and in the PR body rather than leaving it registered against fixed code.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue]

Correct, and the docstring is the part that stings: "never opens the evidence" is false — `count_verdicts` opens it, parses it with the same regex, and discards the verdict. That is #68's finding verbatim, one path over, and I wrote a justification for the leftover instead of noticing it was the same defect.

Taking your patch shape:

```python
detail = self.alissa.get_task(task.ref)
if detail is not None:
    return self._resolved(detail)
return ResolvedTask(task=task, verdicts=self.alissa.count_verdicts(task.ref))
```

One task read on the search path too, degrading to exactly today's two-read behaviour if that read fails — so the fallback keeps the fail-open property rather than trading it for the saving.

Your point about the degraded-ledger state is the one that decided this: when `state.review_task` cannot be read, EVERY PR takes the search path EVERY pass, which is the state this whole change exists to serve, and it was paying a corpus fetch plus two task fetches per PR per pass. Both docstrings and the comment above the return are corrected to say what the path actually does.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:pursue]

Agreed — and it is the same lesson twice in two rounds, which is worth saying plainly: I fixed `find_review_task`'s silent double after #68 flagged it, added `verdict_calls` after my own counterfactual failed to bite, and still left `count_verdicts` answering silently. The rule I should have applied is that EVERY task read the fake serves must be observable, not the ones a finding happened to name.

`count_calls` added and asserted in both tests, so the search path's cost is pinned rather than argued: `test_the_cached_path_reads_the_task_exactly_once_per_pass` asserts it stays empty, and the search-path test asserts exactly one — which, with minor 1 fixed, is now `get_calls == [ref]` and `count_calls == []` on the happy path, with `count_calls` non-empty only on the degraded fallback.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue]

Fair, and I will take it rather than lean on the discretionary escape hatch: `verdict_read=True` is an invariant of the type, not of the watcher, and a conversion that never touches `self` sitting on `ReviewWatcher` is where the next such invariant goes to get lost.

Becomes `ResolvedTask.from_detail(detail)`, a classmethod beside the dataclass, carrying the "one payload answered all three questions" docstring with it. Call sites read `return ResolvedTask.from_detail(detail)` — both of them, now that the search path builds one too (minor 1).

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:pursue] — taking the second option, the one that makes the absence real.

You are right that I described the fake's single divergence from the CLI backwards. `--include-terminal` is off, so the real `alissa task list` omits terminal tasks outright; in my fake they were present in the list and merely filtered out by the predicate, which happens to give the same answer for the wrong reason — and the docstring told a future test author the opposite of what the code does.

Rather than dropping the clause, `list_tasks` now models the CLI: it returns only non-terminal tasks, while `get_task` resolves by ref across every task the fake knows, terminal or not — which is exactly the real asymmetry (`task get` reads a validated task fine; `task list` does not list it). `Validated()` then stays out of the search's answer because it is genuinely absent, and `is_review_task_for` filtering it becomes the belt-and-braces it is in production, rather than the only thing standing between the fake and a wrong answer.

RHDZMOTA:
[triage:pursue] — on the top-level `[nit]` from the verdict body (PR #69 is not on TASK-2141858575 as evidence).

Correct, and it is a C4 gap of exactly the kind that costs a reader the second half of the story: the task carries `Draft PR #68` and `Delivery note`, and #69 appears only in a comment, so an implementation task that produced two PRs looks like it produced one.

Attached as a deliverable on TASK-2141858575, naming the merge-order gate so the record is not just a URL — and I will keep the note in sync with this round's outcome rather than leaving the delivery note describing only round 1 of #68.

Also recorded there: this round closes **TASK-194837655** (the `[major]`), so that task does not sit open against code this PR fixes.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1946,9 +2117,14 @@ def _convergence_reason(
             return None
 
         # Only checkable once a review task exists; before that there is
-        # nowhere for a verdict to have been recorded.
-        if task is not None and self.alissa.latest_verdict(task.ref) == VERDICT_APPROVE:
-            return f"newest verdict envelope on {task.ref} reads approve"
+        # nowhere for a verdict to have been recorded. On the cached path the
+        # envelope was already read back with the task, so this costs nothing;
+        # on the search path it is the one fetch that still has to happen.
```
alissa-app, line 2122: `[minor]` This comment is now false, and it is the third instance of the record that round 1's minor 1 was about.

"on the search path it is the one fetch that still has to happen" described `227a3d4`. After this round's fix the search path reads the matched task at `loop.py:1130` and returns `ResolvedTask.from_detail`, so `verdict_read` is `True` and `newest_verdict` costs nothing here **on that path too**. The only path that still pays a fetch at this line is the degraded fallback — the matched task could not be read — which is the opposite of what "the search path" names.

The triage reply on minor 1 said "Both docstrings and the comment above the return are corrected", and they are: `ResolvedTask`'s docstring (line 886) and `_review_task`'s comment (line 1123) both now say what the code does. This one was missed because it lives in `_convergence_reason`, one screen away from the change.

Suggested: "...on the cached path *and on the search path* the envelope was already read back with the task, so this costs nothing; only a task that could not be read leaves the fetch to happen here."

Worth fixing rather than shrugging at, for the same reason round 1 gave: the next reader of this branch is deciding whether convergence is expensive, and this line answers that question wrongly.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -5336,3 +5417,596 @@ def test_a_production_pass_is_not_bounded_by_a_dry_run_one(config, caplog):
...
+
+def test_a_dated_envelope_beats_an_undated_one_either_way_round():
+    undated = envelope(VERDICT_REQUEST_CHANGES, 1, None)
+    dated = envelope(VERDICT_APPROVE, 2, "2026-07-18T10:00:00Z")
+
+    assert _verdicts(undated, dated) == VERDICT_APPROVE
+    assert _verdicts(dated, undated) == VERDICT_APPROVE
+
+
+def test_convergence_sees_the_newest_verdict_on_real_evidence(config, monkeypatch):
```
alissa-app, line 5988: `[minor]` This test never reaches convergence — it stops at `get_task`, so its name and its docstring both claim more than it asserts.

The docstring says "End to end, through the path this release created ... must converge. Before the fix this PR is the reason for, it read request_changes and ran to the round cap." What the body actually does is monkeypatch `run_json`, call `Alissa().get_task("TASK-500")`, and assert `detail.verdicts == 2` and `detail.verdict == VERDICT_APPROVE`. `_convergence_reason` is never called; no `Decision` is produced; nothing asserts `Action.CONVERGED`. The `config` fixture is requested and never used, which is the tell.

This is the same shape as round 1's `[major]`, one notch down: a green test whose name certifies a guarantee it does not exercise. The escape risk is genuinely low here — the remaining hop (`ResolvedTask.from_detail` → `newest_verdict` → the `VERDICT_APPROVE` comparison) is covered by `test_the_search_path_also_reads_its_task_exactly_once` and `test_the_cached_path_reads_the_task_exactly_once_per_pass`, both of which assert `Action.CONVERGED`. What is at stake is the record: this is the test a future reader will point at when asked "is the ordering fix covered end to end?"

Two honest ways out, both cheap: drive the payload through a watcher and assert `decision.action is Action.CONVERGED` (which is what the docstring promises, and which would then be the one test covering real-shaped evidence all the way to a decision), or rename it to what it does — `test_get_task_reads_the_newest_verdict_off_real_shaped_evidence` — drop the unused `config`, and trim the docstring to the claim it can support.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -218,6 +329,44 @@ def latest_verdict(self, task_ref: str) -> str | None:
...
+            seconds = float(value)
+            # Milliseconds, by magnitude: 1e11 seconds is the year 5138, while
+            # 1e11 milliseconds is 1973 -- so anything above it is ms, and a
+            # task holding both units still orders correctly.
+            if abs(seconds) > 1e11:
+                seconds /= 1000.0
+            return (1, seconds)
+        if isinstance(value, str) and value:
+            try:
+                return (1, datetime.fromisoformat(value.replace("Z", "+00:00")).timestamp())
```
alissa-app, line 365: `[nit]` A naive ISO stamp is interpreted in the process's **local** timezone here, so the docstring's "a task holding both units still orders correctly" is true only when that timezone is UTC (or when every ISO stamp carries an offset).

`datetime.fromisoformat("2026-07-18T10:00:00").timestamp()` uses the local zone; the epoch-ms branch is unambiguously UTC. A daemon running with `TZ` set anywhere else would order a naive ISO stamp against an epoch-ms one with an offset-sized skew. `test_iso_and_epoch_stamps_order_against_each_other` cannot see it — its fixture carries `Z`, and this container is UTC, so I could not make it bite either.

Low stakes today: the CLI emits ints, the ISO branch exists for fixtures and "the API's other surfaces". If those surfaces ever emit naive strings the skew becomes real, and the fix is one clause:

```python
parsed = datetime.fromisoformat(value.replace("Z", "+00:00"))
if parsed.tzinfo is None:
    parsed = parsed.replace(tzinfo=timezone.utc)
return (1, parsed.timestamp())
```

Discretionary — the current behaviour matches what the old lexicographic key did with mixed offsets, which is to say no worse.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -964,6 +1026,116 @@ def __init__(
...
+
+        self.state.record_review_task(pr.full_name, pr.number, task.ref)
+        # The search matched a TITLE out of the corpus; the round count and the
+        # verdict both live in the task's evidence, and ONE read returns both.
+        # Reading it here rather than paying `count_verdicts` now and
+        # `latest_verdict` again from convergence is the same saving the cached
+        # path makes -- and it matters most in the degraded state this whole
+        # change serves: an unreadable ledger sends EVERY PR down this path on
+        # EVERY pass (PR #69 round 1).
+        detail = self.alissa.get_task(task.ref)
```
alissa-app, line 1130: `[nit]` In one branch this re-reads the ref whose read just failed, so that path costs one call more than it did at `227a3d4`.

The branch is: a cached ref exists, `get_task(cached)` at line 1094 returns `None` (kept, correctly — an unreadable task is not a wrong one), the search runs, and it returns **the same task**. This line then reads that ref again. When the failure is persistent that second read fails too, and the pass ends up paying `get_task` ×2 + `count_verdicts` + convergence's `latest_verdict` = four task calls, where at `227a3d4` the same branch paid three.

Not raised higher because the re-read is not provably wasted: the first failure may have been transient, in which case this read succeeds and the pass gets its single-read resolution after all — which is a better outcome, not a worse one. And it only happens while the task CLI is failing, where the daemon is degraded anyway.

If you want it gone, the guard is small — skip the read when `task.ref == cached`, since that ref's read already failed this pass — and it is testable with the fake as it now stands (`get_calls` would show one entry rather than two). Equally fine to leave it and say so; it is a nit either way.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -5336,3 +5417,596 @@ def test_a_production_pass_is_not_bounded_by_a_dry_run_one(config, caplog):
...
+    w.evaluate(OWNER, REPO, NUMBER)
+    assert al.list_calls == 1, "first sighting has nothing cached; it searches"
+    assert al.get_calls == [FakeTask.ref], "then reads the task it matched"
+    assert state.review_task(SLUG, NUMBER) == FakeTask.ref
+
+    w._pass_tasks = None  # a fresh pass
+    w.evaluate(OWNER, REPO, NUMBER)
+
+    assert al.list_calls == 1, "the persisted mapping short-circuits the search"
+    assert al.get_calls == [FakeTask.ref] * 2, "resolved by ref, one small read"
```
alissa-app, line 5447: `[nit]` The message no longer matches the assertion: `[FakeTask.ref] * 2` is two entries, and the message says "one small read".

`get_calls` accumulates across both passes now that the first sighting reads the task it matched (line 5440), so the claim this line is making is "the *second* pass added exactly one read". The neighbouring test does the tidier thing for exactly this reason — `test_the_cached_round_count_comes_from_the_same_read` calls `al.get_calls.clear()  # the first sighting's read is not what is under test` before the cached pass.

Either clear before the second `evaluate` and keep the assertion at `[FakeTask.ref]`, or spell the accumulation out: `"the first sighting's read, then the cached pass's — one small read each"`.

alissa-app (approved):
## Review verdict: **approve** — round 2 of cap 22

Review task **TASK-762291129**. The cap recorded there is this daemon's **effective** cap — `ALISSA_ROUND_CAP=22` plus any operator re-entry grants — not the alissa-code-review template's default of 3; the description already carried 22, so nothing needed correcting.

Same review surface as round 1: `92db55f..HEAD`, with #68's approved delta taken as given. The new delta this round judges is `227a3d4..4438320`.

### Round-1 reconciliation

All five inline findings carry exactly one `[triage:*]` reply, all five are `pursue`, and all five landed in `4438320`. Nothing deferred, so no task references to verify.

| --- | --- | --- |
| `[major]` the ordering assertion certifies a guarantee production does not have | pursue | **fixed (`4438320`)** — at the source, not just the fixture |
| `[minor]` the search path still pays two `alissa task get` calls for one task | pursue | **fixed (`4438320`)** |
| `[minor]` the fake records `latest_verdict` but not `count_verdicts` | pursue | **fixed (`4438320`)** — `count_calls`, asserted on both paths |
| `[nit]` `_resolved` never touches `self` | pursue | **fixed (`4438320`)** — `ResolvedTask.from_detail` |
| `[nit]` the fake's `corpus` docstring describes its one divergence backwards | pursue | **fixed (`4438320`)** — `known` / `corpus` split |
| `[nit]` (non-inline) PR #69 is not on TASK-2141858575 as a deliverable (C4) | — (nits are triage-optional) | **fixed** — `Follow-up PR #69` is now a deliverable there |

The `[major]` was taken the harder way, and I checked that the harder way actually holds rather than reading the note:

- `_created_key` normalises both shapes to one float. Probed against the production function: epoch-ms `[request_changes@…482, approve@…445]` → `approve`; ISO → `approve`; a task mixing the two orders correctly and does not raise, which is the failure mode the round-1 comment warned about widening the `isinstance` into.
- The magnitude test is right: `1e11` seconds is the year 5138 and `1e11` ms is 1973, so no real epoch-second stamp is ever mistaken for ms.
- The `bool` guard is not ceremony — `True` is an `int`, and without it a boolean stamp would rank as epoch 1 ms rather than as unusable. Parametrised, including `False`.
- **Counterfactual, re-run by me** (sort key reverted to the old `str`-only behaviour in a copy of the tree): **8 tests fail** — `test_newest_verdict_envelope_wins`, `test_undated_envelope_loses_to_a_dated_one`, `test_get_task_carries_the_newest_verdict_out_of_the_same_payload`, and the five new ordering tests. The first three predate this round, which is the point the delivery note makes: they were green against a shape the CLI never emits, and they are not any more.
- **Counterfactual for minor 1** (the search path's `get_task` deleted): **5 tests fail**, including `test_an_unreadable_match_falls_back_to_the_two_reads` — so the fail-open fallback is pinned too, not just the saving.

`TASK-194837655` (the pre-existing misparse I registered in round 1) is fixed here and carries a comment saying so with the commit; it is still `committed` rather than closed, which is correct until this merges.

### Verified independently, not read off the PR body

Scratchpad venv, run from `alissa-tools-github-revloop/` with `PYTHONPATH=src/main`:

| check | result |
| --- | --- |
| `pytest src/test` | **624 passed**, 0 failed (610 at `227a3d4`, +14) |
| coverage | **93%** total; `loop.py` 96%, `state.py` 95%, `alissa.py` 84% |
| `bash check-style.sh` | clean |
| `bash check-types.sh` | `Success: no issues found in 29 source files` |
| CI on the head | all six checks green |
| PyPI | latest published is `0.16.9`, so `0.16.11` and `0.16.12` are both free |
| #68 | still **OPEN** — the merge-order gate is live |

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 2 |
| nit | 3 |
| question | 0 (open: 0) |

Both minors are record-drift rather than logic, and both are inside the delta this round judges. Neither gates the merge — but a deferred `[minor]` needs a registered task, not an evaporation.

### Dimension report

- **Correctness** — the `[major]` fix is real, correct at the edges I probed (ms/seconds, mixed shapes, `bool`, negative, unparseable), and pinned by tests that genuinely bite. The search path's single read degrades to exactly the old pair when the read fails, and that fallback has its own test. Both `[minor]`s below are about records that now describe the code wrongly, not about behaviour: `loop.py:2122` still says the search path pays the verdict fetch, and `test_convergence_sees_the_newest_verdict_on_real_evidence` stops one hop short of the convergence its name promises.
- **Coding practices** — matches the module's idiom and its "name the failure mode" comment style; `ResolvedTask.from_detail` is the right home for the `verdict_read=True` invariant, and the `known`/`corpus` split models the CLI's real asymmetry instead of describing it backwards. C1–C4 hold: co-author trailers on all three commits, branch `TASK-2141858575-CACHE-FOLLOWUP`, PR names the task, and #69 is now a deliverable on TASK-2141858575 (round 1's third nit). Two of the three nits are here.
- **Security** — nothing found. The delta adds one stdlib import (`datetime`) and no statement, subprocess, path, or interpolation. `datetime.fromisoformat` over an evidence field raises only `ValueError`, which is caught and degrades to "undated". No new log line carries anything this module does not already log.
- **Performance** — the dimension the round exists for, and it moves the right way: the search path drops from two task reads to one, and the fake now makes both paths' costs assertable rather than arguable. One nit: in the one branch where a *cached* ref was read and failed, the search's new read repeats that ref, so that already-degraded path costs one call more than it did at `227a3d4`.
- **Side-effects / blast radius** — `_newest_verdict` is shared, so this fix changes `latest_verdict` and therefore `_close_round_natively` and convergence for **every** PR in the loop once deployed; that is the intended fix (TASK-194837655) and the direction is strictly toward the newest envelope, but it is the delta's one live behaviour change and it is fixture-tested only, as the note's "not verified" box says. No schema change, no migration. `0.16.12` is free on PyPI (latest published: `0.16.9`), so `check-version-bump` survives either merge order. The container pin stays at `ARG REVLOOP_VERSION=0.16.6`, so merging deploys nothing. **The merge-order gate is still live: #68 is OPEN, and merging #69 first would land #68's delta into `main` under this PR and leave #68 resolving as empty.** Nothing in CI enforces it — it remains the operator's.

### Waived majors

None — there are none.

### Summary

Round 1's single `[major]` was fixed at the source rather than in the fixture, and the fix is the strongest part of the round: both `createdAt` shapes normalise to one scale, the tie-break resolves to the later-recorded envelope, the `bool`/undated edges are parametrised, and reverting the key fails eight tests — three of which predate this PR and were green against a shape the CLI never emits. The search path now reads its matched task once and falls back to exactly the old pair when that read fails, and the fake records every task read it serves, so both paths' costs are asserted instead of argued. All six round-1 findings are triaged `pursue` and all six hold; every published number reproduced (624 passed, 93%, style clean, mypy 29 files, CI green). I am at **approve**: what remains is two `[minor]` records that now describe the code wrongly — a convergence comment the round's own fix falsified, and a test whose name claims a convergence it never reaches — plus three nits. Fix them here or register them; neither is a merge gate. **Merge #68 first.**

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:later] — **TASK-1207414354**.

You are right, and the framing is the part I want to acknowledge: this is the third instance of the record defect round 1's minor 1 named. My triage reply that round said "Both docstrings and the comment above the return are corrected" — I corrected the two the finding pointed at and did not go looking for the third, which is exactly the failure mode of fixing what a finding names instead of what it means.

The suggested wording is right too: only a task that could not be read leaves the fetch to happen at that line, and calling that "the search path" names the opposite of the branch it describes.

Deferred rather than fixed only because round 2 is an **approve**: a commit pushed here now can be silently lost to a squash-merge, and a follow-up to this follow-up would be the third stacked level — which forces the operator into an exact three-PR merge order for what is a comment and two test-record corrections. **TASK-1207414354** carries all three (this comment, minor 2's test, and nit 3's assertion message), soft-linked to TASK-2141858575, with your suggested wording quoted in its DoD.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:later] — **TASK-1207414354** (same task as minor 1; all three record corrections travel together).

Correct on every point, including the tell: `config` is requested and never used, which is what a test that was going to drive a watcher and then didn't looks like. It asserts on a `TaskDetail` and calls the result convergence.

"The same shape as round 1's `[major]`, one notch down" is the right severity read and the reason I am not arguing it. The escape risk is low — as you say, both single-read tests assert `Action.CONVERGED` on the real-shaped path — but the record is what a future reader will point at for "is the ordering fix covered end to end?", and right now it answers yes when it means no.

Of your two ways out the DoD prefers the first (drive it through a watcher to `Action.CONVERGED`), because that would make it the one test carrying real-shaped evidence all the way to a decision, and falls back to the rename-and-trim if that turns out to need more fixture than it is worth. Deferred for the same approve/no-third-level reason as minor 1.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: [triage:later] — **TASK-639960880**.

Good catch, and I had not thought about it: the epoch-ms branch is UTC by construction while `fromisoformat` on a naive string silently adopts the process timezone, so the docstring's "a task holding both units still orders correctly" is conditional on `TZ` in a way nothing states. Your note that the existing test cannot bite — `Z` in the fixture, UTC container — is why I would not have found it either.

Registered rather than taken: it is a real (if currently unreachable) behaviour edge, the fix is your three-line clause plus a test that manipulates `TZ` to make it observable, and neither belongs in a post-approve push. **TASK-639960880** carries the clause and the test verbatim in its DoD.

Agreed on the stakes as stated — the CLI emits ints, the ISO branch is for fixtures and other surfaces, and the current behaviour is no worse than the lexicographic key it replaced.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:later] — **TASK-72507406**.

Accurate, including the reason it is only a nit: the re-read is not provably wasted. A transient first failure makes the second read succeed and the pass gets its single-read resolution after all, which is a better outcome than the guard would produce — so "skip when `task.ref == cached`" trades a worst case of four calls for giving up a real chance at one, and only inside a window where the task CLI is already failing.

That trade is a decision rather than a fix, which is why the task's DoD asks for it to be **decided and written down either way**, not just implemented. **TASK-72507406**, soft-linked to TASK-2141858575, with your `get_calls` observation recorded as the test hook.

Deferred for the approve/no-third-level reason above.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: [triage:later] — **TASK-1207414354** (bundled with the other two record corrections).

Right: `get_calls` now accumulates across both passes because the first sighting reads the task it matched, so the message describes a call count the assertion no longer makes. And the neighbouring test does do the tidier thing — I added that `clear()` there deliberately and did not carry it back one test.

The DoD names both of your options and prefers the `clear()`, so the two tests stop disagreeing about how they handle the first sighting's read.

PR #71, opened 2026-07-30
TASK-517630631: spawn back-pressure — gate reviewer spawns on max_concurrent_sessions, defer rounds under load (revloop 0.16.13)

### LINKED ISSUE #70: revloop: spawn back-pressure — max concurrent reviewer sessions, defer rounds under load (CPU-contention + capacity hygiene)
## Motivation (corrected from cancelled TASK-171166012 — read this framing, not that one)

The original task blamed the 2026-07-29 19:00Z container kill on OOM; Railway metrics falsified that (peak 3 GB of a 12 GB limit — the kill was platform-side). What the metrics DID show, and what stands as the motivation:

- **CPU contention is real:** during the 18:45–19:00Z review burst the service pegged its 2 vCPU ceiling with sustained spikes — 4+ concurrent reviewer sessions (each an interactive claude process) plus the poll loop time-slicing 2 vCPUs. Throttled sessions review slower, hold their round slots longer, widen the burst window, and inflate wall-clock per round (which also inflates grace/stale timers' effective meaning).
- **Capacity hygiene:** the 14:32Z sweep legitimately reported 10 live reviewer sessions against alarm cap 6 (5 busy + 5 in-grace). The 0.16.2 reaper handles leaks (terminal+idle), but NOTHING limits concurrency: `round_cap` is per-PR rounds, and `reap_session_cap` is explicitly "a threshold no healthy deployment reaches, not a capacity limit" (config.py). Spawn is ungated — a merge wave spawns unboundedly regardless of the container's actual CPU/memory budget.

## Fix — make spawn a gated edge

Before spawning a reviewer session for an enqueued round, count live sessions matching revloop's own grammar (both shapes the 0.16.2 matcher accepts — daemon `review-<repo>-pr<n>-r<k>-<nonce>` and hand-spawned `review-pr-<N>[-r<K>]`). At or above a new config key `max_concurrent_sessions` (suggest default 4; env pass-through `ALISSA_MAX_CONCURRENT_SESSIONS` like the reap knobs), **defer the spawn**: the round stays enqueued exactly as today (the "enqueued Ns ago" state already tolerates waiting) and retries on subsequent polls once a slot frees.

Requirements:
- **Fairness:** when a slot frees, the oldest enqueued round spawns first — no starvation. If the existing walk order is already FIFO, document that evidence in the Delivery-note instead of adding machinery.
- **Deferral is not failure:** a deferred round burns no round number and no attempt, must not trip the stale-round probe or respawn logic, and emits no operator escalation. One streak-limited INFO line per poll summarizing deferrals ("2 round(s) deferred — 4/4 sessions live"); no per-round spam.
- **Only own-grammar sessions count**; other lanes' sessions in the shared container are invisible to the gate (same ownership boundary as the reaper).
- **Alarm coherence:** config build rejects `reap_session_cap < max_concurrent_sessions` (an alarm below the spawn limit would page on healthy load). The cap ERROR still fires if hand-spawned sessions push the count past the alarm.
- **Re-entry (CR9):** re-entry-granted rounds queue through the same gate — delayed, never denied.
- Config validation, README settings table, `revloop.config.example.json`, container env pass-through (`revloop-config.sh` + Dockerfile ARG/ENV + entrypoint-config tests) follow the reap-knob pattern from PR #47 exactly.

## Guards / non-goals

- Do NOT touch reap logic (shipped, TASK-793080321), verdict/CI gating (TASK-738619486), identity enforcement, or crash resilience (TASK-1995856683) — compose, don't overlap.
- No CPU/memory measurement in v1 — count sessions, not resources.
- NEVER `tmux kill-server`; the gate defers spawns, it never kills anything.
- **Patch version bump mandatory** (publish gate); the deployed image's `REVLOOP_VERSION` pin must move past this release to take effect — name the required version in the Delivery-note (re-pin lane: TASK-1160071068).

## Delivery note

State the chosen default, FIFO/fairness evidence, deferral's interaction with round accounting and the stale-round probe, and the exact version the Dockerfile pin must reach.

## Definition of Done

Revloop gates reviewer-session spawns on a configurable max_concurrent_sessions (env-passable like the reap knobs): at the limit, enqueued rounds defer without burning rounds/attempts or tripping stale/respawn logic, spawn oldest-first when slots free, with streak-limited deferral logging; only own-grammar sessions count; config validates reap_session_cap >= max_concurrent_sessions; reap/verdict/identity/resilience lanes untouched; version patch-bumped and the required Dockerfile pin version named in the Delivery-note — so a merge wave queues instead of pegging the shared container's CPU with unbounded concurrent claude sessions.

## Acceptance criteria

- [ ] With max_concurrent_sessions=N and N own-grammar sessions live, an enqueued round does not spawn; it spawns on a later poll after a slot frees. Deferral burns no round number/attempt, triggers no stale-round respawn and no escalation. Evidence: tests.
- [ ] With multiple rounds deferred, the oldest enqueued round spawns first when one slot frees; no starvation across ≥3 queued rounds in tests (or documented evidence the existing walk order is already FIFO).
- [ ] Sessions not matching revloop's own grammar do not count toward the limit; hand-spawned review-pr-<N> sessions DO count. Evidence: grammar-parametrized tests.
- [ ] New key validated (>=1; reap_session_cap < max_concurrent_sessions rejected), env pass-through wired in revloop-config.sh + Dockerfile with entrypoint-config tests green, README + example config updated — matching the reap-knob pattern.
- [ ] Version file patch-bumped in the same PR; Delivery-note names the version the deployed Dockerfile pin must reach for the gate to be live.

Alissa-Task: TASK-1751196185

RHDZMOTA:
Closes #70

Origin task: TASK-1751196185
Implementation task: TASK-517630631

Nothing bounded how many reviewer sessions ran at once. `round_cap` bounds *rounds
per PR*; `reap_session_cap` is an alarm the loop logs and cannot act on. So a merge
wave spawned one interactive agent per PR, simultaneously, against a fixed container
— on 2026-07-29 the 18:45–19:00Z burst pegged the deployment's 2 vCPU ceiling with
4+ concurrent reviewers plus the poll loop, and throttled sessions hold their round
slots longer, widening the very burst that starves them.

The spawn is now a **gated edge**. Immediately before spawning — past every branch
that decides *whether* a round is owed — the daemon counts live sessions matching
its own grammar and, at or above the new `max_concurrent_sessions` (default 4),
defers: the round stays enqueued, writes no ledger row, and is retried next poll.

- **Deferral is not failure.** No spawn-ledger row ⇒ the round number is untouched
  (next poll computes the same `completed + 1`), no attempt is spent, and there is
  no `spawn_age`, so the stale-round probe and its respawn branch cannot see the
  round at all. Nothing is posted: no comment, no page, no escalation.
- **Oldest waiter first.** `review_requests` is a `search/issues` query sorted by
  *relevance*, not stable between calls — so the walk is re-ordered: already-deferred
  rounds first in the order they were first deferred, then the rest in the search's
  own order.
- **Own grammar only.** Both shapes count (`review-<repo>-pr<n>-r<k>-<nonce>` and
  hand-spawned `review-pr-<n>`); other lanes in the shared container are invisible,
  the same ownership boundary the reaper draws.
- **One INFO line per pass**, streak-limited, never escalating — a full container is
  the gate working.
- **Alarm coherence.** `Config.build` refuses `reap_session_cap < max_concurrent_sessions`.
- Env pass-through `ALISSA_MAX_CONCURRENT_SESSIONS` on the reap-knob contract
  (`revloop-config.sh` + Dockerfile ARG/ENV + entrypoint-config tests), README
  settings table + example config, version `0.16.12 → 0.16.13`.

**Round 2** answers all three round-1 findings — see *Judgment calls* for the one
place the implementation is narrower than the suggested shape, and the round-2
entries in *Verification*.

### Judgment calls

- **[round 2, `[major]`] The gate escalates on a STALL, not on a deferral streak.**
  The finding is right that a closed gate needs a channel an operator sees: busy
  sessions are never reaped, a hand-spawned `review-pr-<n>` on an open PR is outside
  the reaper's scope by design, undecidable sessions are spared every poll — four of
  those against the shipped defaults (limit 4, alarm 6) is a permanently shut gate
  under a silent alarm. Implemented with the machinery the review pointed at
  (`Streak`'s `crossing`/`escalated`, mirroring `_note_ledger_unwritable`), but on a
  narrower predicate: a second streak counts only passes that deferred **and spawned
  nothing**, cleared by any spawn. Escalating on the deferral streak itself would
  page forever on a wave permanently larger than the gate — a daemon reviewing
  everything, just serially — which is the same "an alarm that fires in the steady
  state trains people to ignore it" argument the alarm-coherence validation rests on.
  Reasoning posted on the thread and not contested at push time.
- **[round 2] An escalated stall announces its own recovery**, past the streak limit
  (`spawn gate: spawning again after N pass(es) …`). Not asked for; found while
  testing: the limiter suppressed the draining pass, so the operator's last word on a
  recovered gate was the WARNING. That is exactly what `_note_ledger_writable` exists
  to prevent for the ledger gate.
- **[round 2, `[minor]`] The dry-run census fix is scoped to the census, not to the
  cap alarm.** The sweep now tracks dry-run "would reap" names and `_check_session_cap`
  subtracts them when seeding the gate's count; production is unchanged (the list is
  empty). The alarm's own count still reads "sessions this pass could not free", which
  in dry-run is all of them — that reading predates this PR and each survivor carries a
  `dry-run: would have reaped (…)` holdout reason. Flagged on the thread as available
  to widen if the reviewer reads the alarm the other way.

- **FIFO needed machinery; the existing walk order is not FIFO.** The issue offered
  "document the evidence instead of adding machinery" as an out. It does not apply:
  `GitHub.review_requests` is a `search/issues` query with no `sort`, i.e. GitHub's
  *best-match relevance* ranking — unrelated to waiting time and not stable between
  calls. Under a gate handing out one slot per freed session that starves whichever
  PR keeps losing. So an in-memory `_waiting` map stamps each round with a monotonic
  **counter** (not a clock — same-pass deferrals are microseconds apart and a tie
  would fall back to the search's order) and `_gate_order` sorts waiters ahead of
  newcomers. A pass with no deferrals is byte-for-byte the old walk.
- **The queue is in memory, not in the ledger.** It is a fairness *order*, not a
  decision the daemon must remember: a restart costs one pass of ordering (every
  waiter is re-stamped on its next deferral) and nothing else — not worth a table on
  the poll path. It is pruned each pass against the live candidate set so a merged
  or converged PR cannot anchor the order forever.
- **An unreadable session list fails OPEN, loudly.** The issue did not say. Chosen
  because `alissa tmux ls` is also the reaper's only input: a CLI that cannot answer
  it means nothing is being reaped either, and refusing every spawn would turn one
  broken subprocess into a fleet-wide review outage with the container idle. Failing
  open restores exactly pre-gate behaviour for the outage's duration, warned once
  per pass.
- **This pass's own spawns count against the gate.** The census is read once per
  pass and a just-enqueued session takes a moment to appear in `tmux ls`, so `_spawn`
  increments it. Without that, one pass would hand the same free slot to every PR in
  a wave — the unbounded burst the gate exists to close, reintroduced by a caching
  detail. Incremented in dry-run too (the diagnostic reports what production would do).
- **A new `QUEUED` action and a new `queued` snapshot column, not a share of the
  existing `deferred`.** The issue asked only for deferral. `deferred` names a round
  whose *live session* is still working, and `webui.sources` counts `in_flight +
  deferred` as **active sessions** — folding gated rounds in would inflate exactly
  the number this gate holds down. Cost: one schema column (additive, with the
  existing `_ADDED_COLUMNS` migration) and one CSS pill selector.
- **The stale-round "re-enqueuing" warning moved below the gate.** It previously
  fired in the same branch that decided a respawn was owed; a gated respawn would
  have made it claim a session the gate never started.
- **`max_concurrent_sessions` default 4, deliberately below the reap alarm (6)** —
  the issue's suggestion, adopted: two vCPUs and a reviewer session is a full
  interactive agent. Upgrade hazard, documented in the README: a deployment that
  already sets `reap_session_cap` below 4 now **fails to load** until it lowers
  `max_concurrent_sessions` too.

- [ ] **The gate is inert until the image pin moves.** `docker/claude/Dockerfile`
      pins `ARG REVLOOP_VERSION=0.16.6`; the pass-through key is rejected as unknown
      by anything older than this release, so the pin must reach **`0.16.13`** (the
      version this PR ships) for the gate to be live. The open re-pin PR #64
      currently proposes `0.16.12` — it needs 0.16.13, or a further re-pin does.
      Re-pin lane: TASK-1160071068.
- [ ] **PyPI publish of 0.16.13** happens on the repo's publish workflow at merge;
      not exercised here.
- [ ] **No container was built or booted** — there is no docker in this runner. The
      Dockerfile `ARG`/`ENV` pass-through is verified only by sourcing
      `revloop-config.sh` directly in `tests-entrypoint-config.sh`; the ARG→ENV wiring
      itself is unbuilt (identical in shape to the four knobs already shipped this way).
- [ ] **No real tmux sessions, no real `claude`.** The census, its seeding from the
      reap sweep, and the grammar filter are exercised against a fake `Alissa`. The
      gate's behaviour against a genuinely full container is unobserved.
- [ ] **The CPU premise is not measured.** Per the issue's own non-goal, v1 counts
      sessions and not resources — that 4 concurrent reviewers is the honest ceiling
      for 2 vCPU is an operator judgement, not a measurement in this PR. If the
      deployment is resized, the knob is the tuning surface.
- [ ] **GitHub Actions were not run here** — only their local equivalents (below).

### Verification

Run from the task worktree, `PYTHONPATH=src/main` inline on each script (a bare
`export` does not survive between shell invocations and would silently test the
installed wheel instead of these edits):

- `bash tests-unit.sh alissa-tools-github-revloop` → **658 passed** (655 at round 1;
  round 2 adds the stall escalation, the draining-gate non-escalation, the stall
  recovery line, and the dry-run census).
- `bash tests-coverage.sh alissa-tools-github-revloop` → **93%** total.
- `bash check-style.sh alissa-tools-github-revloop` (pycodestyle, 120 cols) → clean.
- `bash check-types.sh alissa-tools-github-revloop` (mypy) → **Success: no issues
  found in 29 source files**.
- `bash docker/claude/tests-entrypoint-config.sh` → **ALL PASS**, including the new
  `max_concurrent_sessions` override assertion and its paired `reap_session_cap`.
  Note the cross-check block unsets `ALISSA_MAX_CONCURRENT_SESSIONS`: that block
  imports the **Dockerfile-pinned** library (0.16.6), which predates the key and
  would reject it — a version skew, not a default drift.
- Version bump 0.16.12 → 0.16.13 confirmed against both the version file on `main`
  and the newest PyPI release (0.16.12).
- Gate-specific tests added in `test_loop.py`: deferral at the limit, no ledger row /
  no round number / no attempt burned, spawn on a freed slot, no escalation and no
  comment, one summary line per pass (not per round), no escalation however long the
  streak runs, grammar-parametrized counting (own shapes count, other lanes do not),
  oldest-waiter-first across 3 queued rounds with the search order reversing and then
  pinned adversarially under it, a new arrival not jumping the queue, a departed PR
  releasing its place, one pass handing out one slot rather than one per PR, CR9
  re-entry queuing through the gate, a capped PR still escalating under a full gate,
  a gated stale-round respawn claiming nothing, fail-open on an unreadable list (both
  entry paths), one `tmux ls` per pass, snapshot visibility, and the config-validation
  set (default/layering, `>= 1`, alarm-below-limit rejected, equal accepted, shipped
  defaults coherent, cap alarm still firing on hand-spawns).
- **Provenance:** the round-1 commit was authored by attempt 1 of this session lane,
  which died before opening a PR. Attempt 2 re-ran every check above against that tree
  and published it; the code was not re-written. The round-2 commit is attempt 2's.
- **Round-2 caveat, honestly:** the first local run of the new gate tests was green
  against the *installed* wheel rather than the worktree (`PYTHONPATH=src/main` is
  relative to the package dir, and the run was from the repo root). Caught because
  three tests failed on the correct path; every figure above is from a re-run with the
  path right, and the whole suite was re-run through `tests-unit.sh` itself.

### Scope touched

Against the issue's declared scope (spawn gate + config + env pass-through + docs):

- `…/revloop/loop.py` — the gate, the FIFO order, the census, the summary line, the
  `QUEUED` action, the moved stale-round warning; **round 2:** the stall streak and
  its WARNING/recovery lines, and the dry-run census correction in `sweep_sessions` /
  `_check_session_cap`.
- `…/revloop/config.py` — `max_concurrent_sessions`, default, both validations.
- `…/revloop/__main__.py` — `--max-concurrent-sessions` override flag.
- `…/revloop/version` — 0.16.12 → 0.16.13.
- `…/test_revloop/test_loop.py` — the tests above.
- `docker/claude/{Dockerfile,revloop-config.sh,tests-entrypoint-config.sh,README.md}`,
  `revloop.config.example.json`, `README.md` — the pass-through and the docs.
- **Excursions, called out:** `…/revloop/state.py` (new `queued` snapshot column +
  migration entry) and `…/revloop/webui/page.py` (one CSS selector, `.pill.queued`
  sharing the `.pill.deferred` styling). Neither is in the issue's list; both follow
  from the judgment call not to fold gated rounds into `deferred`, which the console
  reads as active sessions.
- **Untouched, as guarded:** reap logic, verdict/CI gating, identity enforcement,
  crash resilience. No `tmux kill-server` anywhere — the gate defers, it never kills.


alissa-app:
- 2026-07-30 17:46:59 UTC — `review-alissa-github-review-daemon-pr71-r1-2547b6` — round 1 of 22 — spawned
- 2026-07-30 18:13:58 UTC — `review-alissa-github-review-daemon-pr71-r2-f93e24` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1251,6 +1347,168 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
+        """
+        held = [(slug, d) for slug, d in results if d.action is Action.QUEUED]
+        if not held:
+            ended = self._gate_streak.resolve(time.monotonic())
+            if ended is not None:
+                passes, seconds = ended
+                log.info(DEFERRAL_CLEARED, passes, seconds / 60)
+            return
+
+        should_log, _ = self._gate_streak.record(time.monotonic())
```
alissa-app: [major] **A closed gate never escalates, and it can be held closed indefinitely by sessions the reaper is documented never to free.**

The docstring's reasoning — "a full container is the gate WORKING", page on the reap alarm instead — holds only while the queue *drains*. It does not hold for a gate that has been shut for hours, and three session classes can shut it permanently while `reap_session_cap` stays silent:

1. **Busy sessions are never reaped.** `sweep_sessions` (`loop.py:2691`): *"A busy session is still doing something ... never yank the slot from under it, whatever its PR's state."* A wedged agent that never goes idle holds its slot forever.
2. **Hand-spawned `review-pr-<n>` on an OPEN PR is never reaped**, by design — *"Those are reaped on a TERMINAL PR only (issue #46) ... Superseded rounds of an OPEN PR are deliberately out of scope"*. These count against the gate (correctly, per this PR) but the reaper will not free them.
3. **Undecidable sessions** (`_resolve_pr` returns None) are spared every poll.

Four such sessions and the daemon spawns nothing, forever. `reap_session_cap` cannot catch it: the new validation guarantees `reap_session_cap >= max_concurrent_sessions`, and the shipped defaults are 6 and 4 — so 4 or 5 stuck sessions is a **fully closed gate with a silent alarm**. Nor does the stale-round ping help: class 2 has no spawn-ledger row (so no `spawn_age`, so no probe), and the gated rounds themselves are deliberately invisible to it. The result is a fleet-wide review outage whose only trace is a streak-limited `INFO` line.

That is precisely the state this module elsewhere calls the worst one to be in. From `_note_ledger_unwritable`:

> *"A daemon that is up, polling, and deciding NOTHING is the most misleading state it can be in, so once the condition outlasts POLL_ESCALATE_SECONDS every logged line says so at page-worthy level."*

The machinery is already here and already wired — this line calls `record()`, which returns `(should_log, crossing)`, and `Streak` carries `.escalated`. The PR discards the second value into `_`.

Suggested shape, mirroring `_note_ledger_unwritable` exactly:

```python
should_log, _crossing = self._gate_streak.record(time.monotonic())
if not should_log:
    return
...
log.log(
    logging.WARNING if self._gate_streak.escalated else logging.INFO,
    DEFERRAL_SUMMARY, ...
)
```

This does not conflict with the DoD's *"emits no operator escalation"*: that governs a **deferred round** (no CR9 escalation, no PR comment, no page for the round itself), which stays true. This is about the **daemon-level condition** of having spawned nothing for an extended period — a different subject, and the one thing about back-pressure an operator genuinely must find out about.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -2673,6 +2935,11 @@ def _check_session_cap(
         """
         killed = set(reaped)
         remaining = sorted(s.name for s in sessions if s.name not in killed)
+        # The spawn gate's census for this pass, taken from the same post-sweep
+        # list the alarm counts: what the sweep could not free is exactly what
+        # is still holding CPU when the gate decides. Seeded here so the gate
+        # never runs its own `alissa tmux ls`.
+        self._census_probed, self._session_census = True, len(remaining)
```
alissa-app: [minor] **The dry-run census counts sessions production would have reaped, so `--dry-run` reports `QUEUED` where production spawns.**

This seeds the census from `remaining` = sessions minus `reaped`. But in dry-run the sweep never appends to `reaped` — it logs `"[dry-run] would reap ..."`, records a holdout, and `continue`s (`loop.py:2731-2734`). So `killed` is empty and `remaining` still contains every session production would have freed this pass.

Verified against the repo's own fixtures — identical container state, one reapable session, one PR owing a round:

```
production: census after sweep = 0
dry-run   : census after sweep = 1
```

This is the exact mirror of the case the PR *did* handle, and it contradicts the principle stated for it in `_spawn`:

> *"Incremented in dry-run too: the diagnostic's job is to report the decisions production would take."*

It only ever errs conservative (dry-run over-reports the gate), and no production behaviour changes — hence `[minor]`, not `[major]`. But it degrades the one tool the module says is reached for during exactly this incident class: *"the one tool that answers 'what would you do right now' -- asked, precisely, during the substrate incident this whole change is about"* (`poll_once`). An operator dry-running a loaded container would be told rounds are gated that production would in fact spawn.

Fix is local — subtract what dry-run would have reaped, e.g. seed from `len(remaining) - len(would_reap)` by tracking dry-run holdouts, or have the sweep append to `reaped` in dry-run and let the caller's `return len(reaped)` stay honest by other means. If deferred, this needs a registered task rather than evaporating.

docker/claude/revloop-config.sh
```diff
@@ -82,6 +83,7 @@ render_revloop_config() {
      + (if $cap   == "" then {} else { round_cap:          ($cap   | tonumber) } end)
      + (if $grace == "" then {} else { reap_grace_seconds: ($grace | tonumber) } end)
      + (if $scap  == "" then {} else { reap_session_cap:   ($scap  | tonumber) } end)
+     + (if $maxses == "" then {} else { max_concurrent_sessions: ($maxses | tonumber) } end)
```
alissa-app: [nit] Whitespace drift against the block's existing alignment. The surrounding `--arg` names are padded to a common column (`cap   `, `grace `, `scap  `, `cwait `) and the `if` conditions line up on `==`; `maxses` is a character wider and pushes both this line and its `--arg` at line 77 out of the column, so the block no longer reads as a table.

Same class, in `README.md`: the reflowed snapshot-columns paragraph leaves `` `converged`, `capped`, `escalated`, `skipped`) and the reap count, plus a JSON column of the `` noticeably longer than its neighbours. Neither is caught by `check-style.sh` (Python only). Purely cosmetic — take it or leave it.

alissa-app (changes requested):
## Round 1 review — `request_changes`

Effective round cap for this loop: **22** (`ALISSA_ROUND_CAP`, plus operator re-entry grants — none acked on this PR). Review task: **TASK-1407877081**, created downstream of the implementation task because origin `TASK-1751196185` 404s for the reviewer actor; its DoD was read through `TASK-517630631.upstreamTasks` and the diff was judged against that, not against the PR description.

This is a strong, unusually well-reasoned change — the judgment calls in the PR body are the real ones, and the tests pin behaviour rather than implementation. I reproduced everything claimed and mutation-tested the two load-bearing mechanisms.

**Verified locally**, from a throwaway worktree at `6a32cb1`:

- `pytest src/test` → **655 passed** (matches the claim).
- `check-style.sh` (pycodestyle) → clean. `check-types.sh` (mypy) → **Success: no issues found in 29 source files**.
- `docker/claude/tests-entrypoint-config.sh` → **ALL PASS**, new assertions included.
- Mutation test — reverting `self._gate_order(requests)` to `requests` fails exactly 3 tests (`test_the_oldest_waiter_takes_the_next_free_slot`, `test_no_round_starves_behind_a_reshuffling_search`, `test_a_new_arrival_does_not_jump_the_queue`). The FIFO order is genuinely pinned, not decorative.
- Mutation test — removing the `_spawn` census self-increment fails 2 tests. The one-slot-per-pass guarantee is pinned too.
- The upgrade hazard does **not** bite this deployment: `ALISSA_REAP_SESSION_CAP` is empty in the running container's env, so the library default 6 applies and `6 >= 4` holds. The hazard is real for any deployment that sets it below 4, and the README documents it.
- Rollback is safe: `queued` is `NOT NULL DEFAULT 0`, and 0.16.12's `_insert_snapshot` names its columns explicitly, so a downgrade against a migrated DB still inserts.
- `Config.build` does reject unknown keys (`config.py:315`), so the README's `REVLOOP_VERSION >= 0.16.13` requirement and the entrypoint cross-check's version-skew comment are both accurate.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 |
| minor | 1 |
| nit | 1 |
| question | 0 |

### Dimension report

- **Correctness:** the gate satisfies the DoD — deferral writes no ledger row, burns no round number or attempt, is invisible to the stale-round probe, posts nothing, and re-entry-granted rounds queue rather than being denied. Gate placement past every "is a round owed" branch is right. One `[major]` on the gate's *liveness*: a closed gate never escalates and can be held closed indefinitely by session classes the reaper is documented never to free.
- **Coding practices:** matches repo idiom closely — comment density, the `Streak` reuse, the `_ADDED_COLUMNS` migration pattern, the reap-knob pass-through contract. Commit carries the `Co-Authored-By: alissa-app` trailer (C1), branch is `TASK-517630631-SPAWN-BACKPRESSURE` (C2), PR names the task and the issue (C3), PR is on the task as evidence (C4). No dead code, no untracked TODOs. One `[nit]` on whitespace alignment.
- **Security:** nothing found. No new injection surface (the one new subprocess read, `list_review_sessions`, is pre-existing and argv-based, not shell); no secrets in the diff, config, or fixtures; new log lines interpolate only PR slugs, integer counts, and a `CommandError` message; no auth/permission surface touched; no dependencies added.
- **Performance:** nothing found. `_live_session_count` memoizes per pass and is seeded from the sweep's existing `alissa tmux ls`, so the gate adds **zero** subprocess and zero GitHub calls — pinned by `test_the_gate_costs_no_extra_session_list_in_a_poll_pass`. `_gate_order` and the summary sort are O(n log n) on a candidate list bounded by the review-request search. A queued PR still pays its full per-pass evaluation, but that is inherent to deciding a round is owed, not a regression.
- **Side-effects / blast radius:** the `queued` snapshot column is additive with a migration entry and rolls back safely; keeping it out of `deferred` is the right call and is what stops `webui.sources.active_sessions` (`in_flight + deferred`) from inflating. The `.pill.queued` CSS selector is the only console change needed since `read_snapshots` is `SELECT *` and stage records pass through generically. The `reap_session_cap >= max_concurrent_sessions` validation is a documented boot-time hard failure for deployments that set the alarm below 4 — verified not to affect this one. One `[minor]` on dry-run fidelity.

### Waived majors

None.

Address the one `[major]` — surface a sustained gate-closed condition at a level an operator sees, reusing the `crossing`/`escalated` signal `_gate_streak.record()` already returns and `_note_ledger_unwritable` already consumes. Triage the `[minor]` and the `[nit]` per CR8 (`[triage:pursue]` / `[triage:ignore]` with reasoning / `[triage:later]` citing a task). Nothing here requires a redesign.

RHDZMOTA:
- 2026-07-30 18:02:02 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr71-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]`

Accepted, and the reading of the failure is right: busy sessions are never reaped, hand-spawned `review-pr-<n>` on an OPEN PR is never reaped by design, undecidable sessions are spared every poll — so 4 of them against the shipped defaults (limit 4, alarm 6) is a permanently closed gate under a silent alarm, and the gated rounds are deliberately invisible to the stale probe. A streak-limited INFO line is not a channel for that.

One deliberate refinement to the suggested shape, because escalating on the *deferral* streak would page on a condition this PR argues elsewhere is healthy: a pass that defers 5 rounds **and spawns one** is a queue draining, not an outage — and with a wave permanently larger than the gate that streak never ends, so it would cross 30 min and then page forever on a daemon that is reviewing everything, just serially. That is the "an alarm that fires in the steady state trains people to ignore it" argument the alarm-coherence validation is built on.

So the escalation is armed on the **stall**, not on the deferral:

- `_gate_streak` keeps its current job — the log-frequency limiter, unchanged.
- a second `Streak` (`_gate_stall`) records only passes that deferred **and spawned nothing**, and is cleared by any spawn. Its `crossing` bypasses the streak limit, and once `escalated` the summary line switches template and level: `logging.WARNING`, naming how long nothing has spawned, how many rounds wait, and that the sweep cannot free a busy or hand-spawned session — the survivors an operator has to go look at.

Same machinery you pointed at (`record()`'s second value and `Streak.escalated`, mirroring `_note_ledger_unwritable`), same `POLL_ESCALATE_SECONDS` window; only the predicate is narrower. Your reading of the DoD's "emits no operator escalation" as governing the deferred *round* and not the daemon-level condition is the one I am implementing against.

Tests: a gate held closed past `POLL_ESCALATE_SECONDS` logs at WARNING and says nothing has spawned; a gate that drains one round per pass never escalates however long it holds (the existing "never escalates" test is rewritten to that predicate rather than deleted).

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]`

Confirmed — I reproduced it. In dry-run the reap branch logs, records a holdout and `continue`s without appending to `reaped`, so `killed` is empty and the census counts sessions production would have freed. Fixed in this round: the sweep now tracks the dry-run "would reap" names and `_check_session_cap` subtracts them from the **census** only. In production the list is empty, so the seeded count is byte-for-byte what it is today.

Scoped deliberately to the census and not to the alarm: `_check_session_cap`'s own count is documented as "sessions this pass could not free", and in dry-run the sweep genuinely freed nothing — its survivors carry the `dry-run: would have reaped (...)` holdout reason that says so. That dry-run reading of the alarm predates this PR, and widening the fix into it would be an unasked-for change to a shipped alarm's semantics. Flagged in the delivery note rather than silently left; happy to take it in this round instead if you read the alarm the other way.

Test added: identical container state with one reapable session and one PR owing a round produces the same gate decision under `--dry-run` as in production.

docker/claude/revloop-config.sh
RHDZMOTA: `[triage:pursue]`

Taking it — the block reads as a table and a six-character variable name is not worth breaking that for. The jq binding is renamed so it pads into the existing column, and the `if` conditions line back up on `==`. The reflowed snapshot-columns paragraph in `README.md` is re-wrapped to its neighbours' width too.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1251,6 +1385,206 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
+            ended = self._gate_stall.resolve(now)
+            if escalated and ended is not None:
+                passes, seconds = ended
+                log.info(GATE_STALL_CLEARED, passes, seconds / 60, len(held))
+        else:
+            _, crossing = self._gate_stall.record(now)
+            # The crossing BYPASSES the streak limit, on the same reasoning as
+            # the ledger gate's: it is a state change, and suppressing it would
+            # hide the one transition the line exists to show.
+            should_log = should_log or crossing
```
alissa-app, line 1555: [minor] **The crossing bypass is the one new mechanism no test pins — and dropping it silently is a regression this repo has already had once.**

This line is what stops the *deferral* limiter from swallowing the *stall* escalation, and it is a hand-rolled bypass rather than the one inside `Streak.record` (necessarily so — it ORs `crossing` from `_gate_stall` into `should_log` from `_gate_streak`). It survives mutation: replacing it with `should_log = should_log` leaves **all 658 tests green**.

The existing tests miss it because they start both streaks on the same pass, so the crossing lands where the limiter would have logged anyway. Decouple them — a gate that drains for a while and *then* wedges — and the bypass is load-bearing. Driving `_note_deferrals` directly, 25 draining passes (deferring **and** spawning) then wedging, at a 60s tick:

```
as shipped                 : first WARNING at wedge pass 30   (= POLL_ESCALATE_SECONDS)
should_log = should_log    : first WARNING at wedge pass 34
```

Four passes late here, and bounded by `POLL_FAILURE_LOG_EVERY` (10) in general — so up to nine polls of an outage that is already declared page-worthy, silent. Drain-then-wedge is not a contrived shape either: it is precisely a merge wave that drains normally until a session goes busy-forever, which is the scenario the `[major]` was about.

What makes this worth a comment rather than a shrug is that `Streak`'s own docstring records the same failure already happening in this module:

> *"Extracted so the two callers that need 'streak, limit, escalate, recover' ... cannot disagree about it (PR #63 round-2 nit; the gate's hand-rolled copy dropped the crossing bypass, which made the parameters table's 'the escalation crossing is never suppressed' false at any poll interval other than 60s)."*

The extraction protects callers that take `Streak`'s `should_log` verbatim. This caller can't, so it re-implements the bypass by hand — the exact thing that regressed before — and this time nothing would catch it.

A test in the shape of the drain-then-wedge above pins it in a few lines. If you'd rather defer, this needs a registered task rather than evaporating (CR8) — it is a `[minor]`, not a `[nit]`, because the property it protects is one the module has already documented itself losing.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -141,6 +141,81 @@
...
+# for half an hour is not that. Three session classes can hold it shut with the
+# reap alarm silent: a BUSY session is never reaped whatever its PR's state, a
+# hand-spawned `review-pr-<n>` on an OPEN PR is out of the reaper's scope by
+# design (issue #46), and an undecidable session is spared every poll. Four of
+# those against the shipped defaults (limit 4, alarm 6) is a fleet-wide review
+# outage that no other channel reports: the gated rounds write no ledger row,
+# so the stale-round probe cannot see them either.
+DEFERRAL_STALLED = (
+    "spawn gate: %d round(s) deferred — %d/%d reviewer sessions live, and "
+    "NOTHING has spawned for %.0f min. %s This is no longer back-pressure "
```
alissa-app, line 171: [nit] `"This is no longer back-pressure doing its job"` asserts more than the predicate can know.

The predicate is "deferred and spawned nothing for 30 minutes", which is satisfied by the wedged-session case this line describes — and also by four *legitimately slow* reviews finishing nothing inside the window. In that second case the sentence is false: back-pressure is doing exactly its job, the container is simply saturated by long-running work.

I don't think the predicate is wrong — nothing completing for half an hour at the limit is worth an operator's eyes either way, and the line does send them to the survivors, where a busy-and-healthy session reads differently from a wedged one. It's the diagnosis being stated as fact that I'd soften, in a module whose log prose is otherwise careful to describe the observation and let the operator draw the conclusion (`DEFERRAL_STALLED`'s own neighbours do this well). Something like *"nothing has started for long enough that this may no longer be back-pressure doing its job"* costs nothing and stays true in both cases.

Purely discretionary — the mechanism is right and the pointer to the survivors is the part that matters.

alissa-app (approved):
## Round 2 review — `approve`

Effective round cap for this loop: **22** (`ALISSA_ROUND_CAP`, plus operator re-entry grants — none acked on this PR). Review task: **TASK-1407877081**. Judged head: `5623806d`.

All three round-1 findings were triaged `[triage:pursue]` on their threads before the push — triage is complete, so the round did not bounce — and all three are **delivered**. The `[major]` came back with a *better* predicate than I proposed, and I verified the improvement rather than just the fix.

| --- | --- | --- |
| `[major]` a closed gate never escalates | pursue | **fixed** (`5623806d`) |
| `[minor]` dry-run census counts slots production would free | pursue | **fixed** (`5623806d`) |
| `[nit]` `--arg` block alignment / README reflow | pursue | **fixed** (`5623806d`) |

**`[major]` — verified, and the narrowing is right.** I proposed escalating on the deferral streak; that would have paged on a healthy serial drain, and the implementer was correct to reject it. The shipped predicate is a second `Streak` (`_gate_stall`) recording only passes that deferred **and spawned nothing**, cleared by any spawn. The two streaks are genuinely independent — `_gate_stall.clear()` on a pass with no deferrals, `resolve()` on a pass that spawns — and an escalated stall that recovers logs `GATE_STALL_CLEARED` outside the streak limit, so the operator's last word is never the WARNING. I also checked the path I was most suspicious of: when the queue empties entirely, `_gate_stall.clear()` is silent, but `_gate_streak.resolve()` on that same pass still emits `DEFERRAL_CLEARED`, so there is no way for an escalated stall to end with no recovery line at all.

The rewritten "never escalates" test now pins **draining**, not merely not-escalating — `assert _spawned_numbers(results)` inside the loop is the assertion that was missing.

**`[minor]` — verified.** `would_reap` is threaded through the sweep and subtracted from the census seed only, not from `remaining`. That scoping is exactly right, and it is the answer to the open question in the triage reply: the alarm's count is documented as "sessions this pass could not free", a dry-run pass genuinely freed none of them, and widening the fix into `_check_session_cap`'s alarm would have changed a shipped alarm's meaning unasked. **Census-only was the correct call — leave the alarm alone.**

**`[nit]` — verified.** `maxses` → `gate` pads to the block's 7-character column and the `==` conditions line back up; no stale `maxses` reference anywhere in the tree; the README paragraph is re-wrapped.

### Verified locally

From a throwaway worktree at `5623806d`:

- `pytest src/test` → **658 passed** (655 + 4 new − 1 rewritten).
- `check-style.sh` → clean. `check-types.sh` → **Success: no issues found**.
- `docker/claude/tests-entrypoint-config.sh` → **ALL PASS**.
- CI on the head: **all 6 checks SUCCESS**.
- Commit carries `Co-Authored-By: alissa-app` (C1); branch, PR-names-task and PR-on-task all still hold (C2–C4).

**Mutation tests on the new mechanisms** — 4 of 5 killed:

| mutation | result |
| --- | --- |
| revert the census correction to `len(remaining)` | **killed** — `test_the_dry_run_census_counts_the_slots_production_would_free` |
| drop the `Action.SPAWNED` guard (stall records every deferral) | **killed** — 2 tests, incl. the draining test |
| never escalate (`_gate_stall.escalated` → warning branch dead) | **killed** — 2 tests |
| drop the `GATE_STALL_CLEARED` recovery line | **killed** — `test_one_spawn_clears_the_stall_escalation` |
| drop the crossing bypass (`should_log = should_log or crossing`) | **SURVIVES** — see the `[minor]` below |

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 1 |
| nit | 1 |
| question | 0 (open: 0) |

### Dimension report

- **Correctness:** the delta closes both prior findings against the DoD, and the DoD's "emits no operator escalation" still holds — that governs the deferred *round* (no `Action.ESCALATED`, no round number burnt, no PR comment), and `_note_deferrals` only logs. Format strings and their argument tuples match on both new templates; `escalated` is captured before `resolve()` clears it. One `[minor]`, on test coverage rather than behaviour.
- **Coding practices:** matches repo idiom — the `Streak` reuse, the comment density the module holds itself to, the constant-per-log-template convention. C1–C4 hold. No dead code. `would_reap`'s `| None = None` default is unexercised (one caller, always passes it), which is defensible as signature shape and not worth a finding.
- **Security:** nothing found. The new log templates interpolate only integers, a float and the same PR-slug string the existing summary already carried; no new subprocess, no new input surface, no dependency change.
- **Performance:** nothing found. The `waiting` string is built once, still after the `should_log` early return, so no pass pays for a line it does not log. The census seed goes from `len()` to a set-membership `sum()` over the same post-sweep list — O(n) on a list bounded by the session cap. No new subprocess and no new GitHub call.
- **Side-effects / blast radius:** contained. `_check_session_cap` has exactly one caller; `_session_census` is read only by the gate and `_spawn`'s self-increment, so the correction cannot reach the `REVIEWER SESSION CAP EXCEEDED` alarm or the snapshot counts. No config key added, so `0.16.13` remains the right version and the entrypoint cross-check is unaffected. The new WARNING creates no decision and no `Action.ESCALATED`, so the snapshot's `escalated` bucket and `webui.sources.active_sessions` are untouched.

### Verdict

**`approve`** — round 2 of cap 22. No blockers, no majors, clean reconciliation. The one `[minor]` is a test-coverage gap on a bypass that is *correctly implemented here*; per CR8 it wants a `[triage:pursue]` in a follow-up push or a `[triage:later]` with a registered task, but it does not gate the merge. The `[nit]` is discretionary.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — in a **follow-up PR**, not a push to this branch.

The finding is correct and the mutation evidence settles it: the bypass is hand-rolled *because* it ORs `_gate_stall`'s crossing into `_gate_streak`'s `should_log`, which is precisely the shape `Streak`'s docstring records this module already regressing once (PR #63 round-2). A property that has been lost before, with nothing pinning it, is a `[minor]` and not a `[nit]` — agreed.

The reason it is not another commit here: this round's verdict is an **approve**, so this head is merge-racy — the operator may merge at any moment and a commit pushed after the squash-merge is silently lost. So the drain-then-wedge test lands as a follow-up PR carrying `Follow-up-To: #71`, together with the `[nit]` below.

The test will be the shape you drove: N draining passes (deferring **and** spawning) so the two streaks decouple, then a wedge, asserting the first WARNING lands on the pass that crosses `POLL_ESCALATE_SECONDS` and not up to `POLL_FAILURE_LOG_EVERY` passes later. Mutating `should_log = should_log` must fail it — I will state that counterfactual result in the follow-up's delivery note rather than assert it in prose.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]` — same follow-up PR.

Taking it, and the reasoning is the part I agree with rather than the wording: the predicate is "deferred and spawned nothing for 30 minutes", which four legitimately slow reviews satisfy just as well as a wedged session. Stating the diagnosis as fact is the line doing something the observation cannot support, in a module whose log prose otherwise describes what it saw and leaves the conclusion to the reader — and an operator who has been told the wrong cause reads the survivor list looking for the wrong thing.

Your phrasing, adopted: *"nothing has started for long enough that this may no longer be back-pressure doing its job"*. The pointer to the sweep's survivors — the part that actually moves an operator — is unchanged.

Not pushed here for the same reason as the `[minor]`: an approve is terminal for this branch.

PR #75, opened 2026-08-10
TASK-1369970340: reviewer console container memory tile (cgroup split + top processes by RSS)

### LINKED ISSUE #74: Reviewer console: container-wide memory tile (cgroup anon/file/slab breakdown + top processes by RSS)
## Context (audited 2026-08-09, Railway service `alissa-studio-github-review-daemon`)

Railway showed a flat ~6 GB memory plateau with zero reviewer sessions. Answering "is that real?" required a manual shell into the container: `ps` (total RSS 176 MB) plus `/sys/fs/cgroup/memory.current` (5.98 GB) and `memory.stat` (`anon 71.6 MB`, `file 2.20 GB`, `slab_reclaimable 3.71 GB`) — i.e. ~5.9 GB was reclaimable kernel cache, not a leak (root cause: the entrypoint's boot-time volume walk; fixed separately in task 357393641).

The reviewer console (`alissa-revloop-ui`) currently cannot answer this question: `webui/sysinfo.py` accounts CPU%/RSS only per managed tmux-session pane tree, and the only container-wide stat is the workspace `disk_usage` tile. An operator staring at a platform memory graph has no in-console way to distinguish real process memory from cache the kernel would drop under pressure.

## Scope

Extend `alissa-tools-github-revloop` (python package in this repo) so the console renders a **container memory tile** next to the existing disk tile:

1. **`sysinfo.cgroup_memory()`** — read cgroup v2 `memory.current` plus a small fixed set of `memory.stat` keys: `anon`, `file`, `inactive_file`, `shmem`, `slab_reclaimable`, `slab_unreclaimable`. Follow the module's existing conventions: the cgroup root path is a parameter (default `/sys/fs/cgroup`) so it is testable against a crafted fixture directory; absent/unreadable/malformed files return `None` fields, never raise (the console must render fine on a host without cgroup v2, e.g. a dev laptop or macOS). Derive and expose the operator-facing summary numbers: `charged` (current), `resident` (anon), `reclaimable` (file + slab_reclaimable — the "not really in use" bucket).
2. **`sysinfo.top_procs(n)`** — top-N processes by RSS across the WHOLE `/proc` (not just session pane trees), reusing `build_index`'s single-scan snapshot; each row: pid, comm (from the stat line's `(comm)` field — note the existing last-`)` parse), rss_bytes. Same vanished-PID tolerance and `proc_root` parameterization as the rest of the module.
3. **Console surface** (`webui/sources.py` + `webui/page.py`): a memory stat tile showing charged vs resident vs reclaimable (so a Railway-style "6 GB" reads as "72 MB real + 5.9 GB cache" at a glance), and a compact top-5 process list (comm + RSS). Follow the existing panel/build patterns; no new endpoints beyond what the current page-build flow needs; no extra sampling loops (single snapshot per page build, consistent with the module's sample-free doctrine).
4. **Tests** mirroring the existing sysinfo test style: fixture cgroup dir (happy path, missing dir, malformed values) and top_procs over the existing crafted `/proc` fixtures (including a vanished PID). Version bump per the repo's release flow (`check-version-bump` workflow).

## Out of scope

- The entrypoint chown/cache fix (task 357393641, runs in a parallel lane).
- Bumping `REVLOOP_VERSION` in `docker/claude/Dockerfile` — that pin follows the PyPI release this PR triggers on merge, as a separate deploy step.
- Alerting/thresholds; this is a read-only observability tile.

## Acceptance detail

- On a cgroup-v2 Linux host the tile shows non-None charged/resident/reclaimable that sum sensibly against `memory.stat`; on a host without cgroup v2 the tile renders a graceful "unavailable" state instead of erroring.
- Top-process list matches `ps aux --sort=-rss` head within snapshot noise.

## Definition of Done

PR on fahera-mx/alissa-github-review-daemon flipped to ready-for-review (per alissa-code-git conventions, PR URL attached as task evidence) adding sysinfo.cgroup_memory() + sysinfo.top_procs() with fixture-driven tests, the console memory tile + top-5 process list wired through sources/page, and the package version bumped for release.

## Acceptance criteria

- [ ] Unit tests cover cgroup_memory (fixture dir happy path, missing cgroup root, malformed values -> None fields without raising) and top_procs (ordering by RSS, vanished-PID tolerance, proc_root parameterization); full test suite and CI checks green on the PR.
- [ ] After the release is deployed, the reviewer console on Railway shows the memory tile with charged vs resident vs reclaimable and a top-5 process list consistent with an in-container ps/memory.stat spot check.

Alissa-Task: TASK-1937423814

RHDZMOTA:
Closes #74

**Origin task:** TASK-1937423814
**Implementation task:** TASK-1369970340

## What this adds

The console could not answer the question a Railway memory graph raises. `webui/sysinfo.py` accounted CPU%/RSS per managed tmux pane tree only, and the workspace disk tile was the sole container-wide stat — so a flat ~6 GB plateau with zero reviewer sessions had to be settled by shelling into the container for `memory.current`, `memory.stat` and `ps`.

Two container-wide readers in `sysinfo`, both following the module's existing conventions (root is a parameter, every read degrades to `None` instead of raising, no sampling loop):

- **`cgroup_memory(cgroup_root="/sys/fs/cgroup")`** — cgroup v2 `memory.current` plus a fixed six-key slice of `memory.stat` (`anon`, `file`, `inactive_file`, `shmem`, `slab_reclaimable`, `slab_unreclaimable`), and the three operator-facing numbers derived from them: `charged` (current), `resident` (anon), `reclaimable` (file + slab_reclaimable). Raw keys ride along for an in-container spot check.
- **`top_procs(n)`** — top-N by RSS across the whole `/proc` (`pid`, `comm`, `rss_bytes`), off the single-scan `build_index` snapshot, same vanished-PID tolerance and `proc_root` parameterization as the rest of the module.

Console surface: a **Container Memory** tile whose headline is the charge and whose sub splits it (`6.0 GB` / `71.6 MB resident · 5.9 GB reclaimable`), plus a read-only **top-5 process** panel. `dashboard()` now builds the `/proc` index once and hands it to both the session table and the process list, so the page build walks `/proc` once, not twice.

## Delivery notes

**Judgment calls**

- The issue names the console surface but not the tile's meter. It is the **resident share of the charge** (`resident/charged`), so the shared warn/crit thresholds keep the same meaning as on the Volume and Rate tiles: the resident part is the only part a memory limit can actually kill for. On the audited plateau this reads 1% (green), which is the correct signal.
- `reclaimable` is `None` unless **both** `file` and `slab_reclaimable` parsed. A partial sum would publish a smaller cache than the container holds — the wrong direction for "is that plateau a leak?" — so an unknown stays unknown rather than becoming a wrong number.
- `read_stat` now also returns `comm` (first `(` to the existing last `)`). The issue asks for comm in each `top_procs` row and specifies it comes from that parse; carrying it on the shared row is what lets `top_procs` reuse `build_index` instead of re-reading every `stat`. This changed one existing exact-dict assertion in `test_webui_sysinfo.py`.
- `sessions()` gained an optional `index=` parameter and `dashboard()` builds the `/proc` index eagerly. The issue asks for one snapshot per page build; the memory panel needs the scan unconditionally, so the old lazy build survives only for callers that pass nothing (a table with no live pane still never scans `/proc`).
- The tile is the 5th; `.tiles` went from a 4- to a 5-column grid with a new 1100px breakpoint inserted **before** the existing 900px one so the narrow case still wins. No other design tokens touched, no new accent colour.
- The top-process panel is read-only (no kill button). Those PIDs are host-wide and unmanaged; the sessions table stays the only place a process is killed from.
- `memory.max` is not read. The issue's key set does not include it and the tile needs no limit to answer its question.

- [ ] Acceptance criterion 2 is post-deploy by construction: the Railway console showing this tile requires the PyPI release this PR triggers **and** the `REVLOOP_VERSION` re-pin in `docker/claude/Dockerfile`, which the issue puts out of scope for this PR. Nothing here is live until that pin lands.
- [ ] The in-container `ps` / `memory.stat` spot check on the Railway service itself. It was verified on this runner (also cgroup v2 Linux) but not in `alissa-studio-github-review-daemon`, whose cgroup is what the audited 6 GB plateau came from.
- [ ] The rendered tile was never opened in a browser — no browser in this environment. The page JS was syntax-checked with `node --check` and the tile's formatter output was executed headlessly against the audited numbers, which is not the same as seeing the grid at five columns.
- [ ] The graceful "no cgroup v2" path was exercised against a **missing directory**, not against a real macOS or cgroup-v1 host.
- [ ] CI itself (this environment runs the four check scripts directly, not the workflows). `check-version-bump` in particular: the file moves 0.16.14 → 0.16.15, which was free on PyPI at the time of writing.

**Verification**

- `tests-unit.sh` — **681 passed** (was 656; 25 new).
- `tests-coverage.sh` — 94% total; `sysinfo.py` 99% (its one uncovered line pre-dates this change), `sources.py` 95%, `page.py` 100%.
- `check-style.sh` (pycodestyle) — clean.
- `check-types.sh` (mypy) — `Success: no issues found in 29 source files`.
- All four run from the package dir with `PYTHONPATH=src/main` inline, so they measure the worktree source and not the installed wheel.
- Live-host checks, not just fixtures: `cgroup_memory()` against this runner's real `/sys/fs/cgroup` (charged 2.22 GB = resident 1.20 GB + reclaimable 1.01 GB, consistent with `memory.stat`), and `top_procs(5)` against `ps aux --sort=-rss | head` — same processes in the same order, RSS within snapshot noise (~1%). The full `dashboard()` payload was built on this host and JSON-serialized.
- Counterfactual: three defects injected one at a time (partial `reclaimable` sum; `top_procs` sorted ascending without the PID tie-break; `dashboard()` rescanning `/proc` per panel) produced 9 failures across 5 distinct tests. The suite fails when the implementation is wrong.

**Scope touched**

Inside the issue's declared scope (`alissa-tools-github-revloop/**`), 7 files:

- `webui/sysinfo.py` — `cgroup_memory`, `top_procs`, `comm` on `read_stat`, module docstring.
- `webui/sources.py` — `cgroup_root` ctor param, `TOP_PROCS`, `sessions(index=)`, `tiles.memory` + `top_procs` in the payload, one `/proc` scan per build.
- `webui/page.py` — the tile, the top-process panel, the 5-column grid.
- `.../revloop/version` — 0.16.14 → 0.16.15.
- three `test_webui_*.py` files.

Excursions: none outside the declared scope. Nothing in `docker/`, `.github/`, or the daemon's loop/state layers was touched. Two pre-existing test assertions were edited rather than added to — the exact-dict `read_stat` check and the `panel_id` list — both mechanical consequences of the new fields.


alissa-app:
- 2026-08-10 05:48:40 UTC — `review-alissa-github-review-daemon-pr75-r1-c91ab5` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py
```diff
@@ -158,6 +197,121 @@ def tree_usage(
...
+    cgroup tree -- every field is None and the console renders "unavailable";
+    the root is a parameter so all of that is testable off a fixture dir.
+    """
+    root = Path(cgroup_root)
+    charged = _read_int_file(root / _CGROUP_CURRENT)
+    stat = _read_stat_keys(root / _CGROUP_STAT, _CGROUP_STAT_KEYS)
+    out: "dict[str, int | None]" = {
+        "charged": charged,
+        "resident": stat["anon"],
+        "reclaimable": _sum_or_none(stat["file"], stat["slab_reclaimable"]),
```
alissa-app, line 309: [minor] `reclaimable` counts `shmem`, which is not reclaimable without swap.

In cgroup v2 `memory.stat`, `shmem` (tmpfs, shm segments, shared anonymous mmaps) is a **subset of `file`**. It is swap-backed, not drop-backed: with swap disabled or tightly capped — the normal container case — those pages cannot be reclaimed under pressure at all, and they are exactly what the container gets OOM-killed for.

So on a tmpfs-heavy container the tile answers the operator's question backwards: a 6 GB charge that is 5 GB tmpfs reads as "5.9 GB reclaimable, no leak", when nothing there is reclaimable. That is the one way this tile can actively mislead rather than merely under-report — and it is invisible from the console, because `shmem` rides in the payload but is never rendered.

The key is already parsed and sitting right there, so the honest derivation is `(file - shmem) + slab_reclaimable`, with `shmem` folded into `resident` or shown as its own magnitude.

Calling this **[minor]** rather than [major] on two counts: the audited plateau this tile was built for has `shmem 0`, and the issue's own scope text prescribes `file + slab_reclaimable` verbatim — so the implementation is faithful to the spec, and it is the spec that is imprecise. Changing it is a follow-up decision, not a fix to this PR. Tracked on TASK-680186617.

The docstring does acknowledge the overlap ("`shmem` is counted inside `file` while behaving like anon memory"), which is why this is a finding about the *rendered number* and not about the author having missed it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
```diff
@@ -422,6 +430,21 @@
...
     } else { out += tile('Volume', '--', 'unavailable'); }
+    // The tile a platform memory graph sends you here to read: the headline is
+    // what the container is CHARGED, the sub splits it into what is really
+    // process memory and what is cache the kernel would drop under pressure --
+    // so a "6 GB" plateau reads as "72 MB real + 5.9 GB cache" at a glance.
+    // The meter is the RESIDENT share of the charge, deliberately: it is the
+    // only part of a rising charge that a limit can actually kill for, so the
+    // shared warn/crit thresholds mean the same thing here as on the others.
+    var mem = t.memory;
+    if (mem && mem.charged != null) {
```
alissa-app, line 441: [minor] The tile discards the partial cgroup read that `sysinfo` deliberately preserves.

Gating the whole tile on `mem.charged != null` means a payload with `charged: null` but a good `resident` / `reclaimable` renders as `'--' / 'no cgroup v2'` — the breakdown is thrown away.

That directly contradicts the contract the reader layer went out of its way to establish. `cgroup_memory` reads `memory.current` and `memory.stat` through separate helpers precisely so each degrades on its own, and there is a test pinning it whose docstring says so in as many words:

> `test_cgroup_memory_current_missing_keeps_the_stat_split` — "Each file degrades on its own: an unreadable `memory.current` does not blank the breakdown **the console can still show**."

The console cannot, in fact, still show it. The test asserts a data-layer property and reads as if it asserted a rendering property, so nothing catches the gap.

`charged` is only needed for the headline and the meter, both of which already have their own null handling (`bytes()` returns `'--'`, and `mpct` is null-guarded on `!mem.charged`). Gating on "any of the three is non-null" and letting the headline fall back to `'--'` would make the tile match its own reader.

**[minor]** and not higher because the condition is genuinely rare — on a real cgroup v2 tree `memory.current` and `memory.stat` appear and disappear together — and because the failure is under-reporting, never a wrong number. Tracked on TASK-680186617.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py
```diff
@@ -40,24 +54,49 @@
...
+    operator nothing about what is holding the memory.
+    """
     try:
         data = (Path(proc_root) / str(pid) / "stat").read_text()
     except (OSError, ValueError):
         return None
     rparen = data.rfind(")")
-    if rparen == -1:
+    lparen = data.find("(")
+    if rparen == -1 or lparen == -1 or lparen > rparen:
```
alissa-app, line 91: [minor] This new guard is not exercised by any test — mutation-verified.

I reverted the added clause to the original `if rparen == -1:` and re-ran the full suite at this head: **681 passed, zero failures.** No fixture writes a `stat` line without a `(`, or with a `)` before its `(`, so the branch this PR added is invisible to the bar.

It is not dead code — without it, `data[lparen + 1:rparen]` on a line with no `(` silently evaluates to `data[0:rparen]`, so `comm` would come back carrying the PID and whatever else preceded it, and the top-process panel would render that garbage as a process name rather than skipping the row. That is the behaviour the guard exists to prevent, and it is precisely the behaviour nothing would notice regressing.

Worth noting that `sysinfo.py`'s 99% statement coverage does not contradict this: the three new conditions live on one already-covered `if` statement, so line coverage cannot see the gap. This is the kind of hole only a mutation check finds — and the PR's own counterfactual exercise (three injected defects, 9 failures) is a good instinct that simply did not reach this line.

A sibling of the existing `test_read_stat_comm_with_parens_and_spaces` asserting `read_stat` returns `None` for `"42 no-parens-here …"` closes it in one assertion. **[minor]**, tracked on TASK-680186617.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
```diff
@@ -422,6 +430,21 @@
...
+      var msub = (mem.resident == null ? '--' : bytes(mem.resident)) + ' resident · ' +
+        (mem.reclaimable == null ? '--' : bytes(mem.reclaimable)) + ' reclaimable';
+      var mpct = (mem.resident == null || !mem.charged) ? null :
+        Math.round(100 * mem.resident / mem.charged);
+      out += tile('Container Memory', bytes(mem.charged), msub, mpct);
+    } else { out += tile('Container Memory', '--', 'no cgroup v2'); }
```
alissa-app, line 447: [nit] `'no cgroup v2'` names one of the three conditions that reach it.

This branch is also what renders when:

1. the host **has** cgroup v2 but the daemon is not in a private cgroup namespace — on the true v2 root cgroup the memory controller interface files do not exist, so a bare-metal or `--cgroupns=host` deployment lands here and is told it has no cgroup v2 when it does; and
2. the read was partial (see the comment above).

I confirmed the happy path is real for the deployment target: this runner reports `0::/` for `/proc/self/cgroup` with `memory.current` present at the mount root, i.e. a private cgroup namespace, which is the Docker default on v2 hosts and what Railway gives you — so `cgroup_memory()` genuinely reads *this container's* charge, not the host's. Reading `/sys/fs/cgroup` directly rather than resolving the process's own cgroup path is the right call at this size; the point is only that when the assumption does not hold, the failure is silent and the label misattributes it.

`'unavailable'` — the word the origin task's acceptance detail actually uses, and the word the neighbouring Volume and GH Rate tiles already use for the same situation — would be both accurate and more consistent than a diagnosis the console cannot make. Implementer's discretion.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -479,9 +497,18 @@ def dashboard(self) -> dict:
         snaps = self.snapshots(SPARK_POINTS)
         latest = snaps[0] if snaps else None
         ledgers = self.ledgers()
-        sessions = self.sessions()
+        # ONE /proc scan for this whole build: the session table walks pane
+        # trees out of it and the top-process list ranks the same snapshot, so
+        # the two panels can never disagree about a process that exited between
+        # them -- and the host pays for one walk per poll, not two.
+        proc_index = sysinfo.build_index(self._proc_root)
```
alissa-app, line 504: [nit] Hoisting the scan here also moves it *earlier* than the pane PIDs it serves.

The snapshot is now taken before `sessions()` shells out to `alissa tmux ls` and `tmux list-panes`, where previously the lazy build happened after the pane PID was in hand. A session whose pane starts inside that window is absent from `stats`, so `tree_usage` returns the empty row and the table shows `--` for CPU% and RSS for one poll.

Self-healing within ~10s and strictly cosmetic — and building the index once is clearly the right trade against walking `/proc` twice per build, which is what this hoist buys. Flagging it only because the comment above sells the change purely as "the two panels can never disagree", and the freshness cost on the other side of that trade is worth a clause. Implementer's discretion.

alissa-app (approved):

Reviewed as a fresh instance against issue #74 / origin task TASK-1937423814, not against the PR body's claims. Review task: **TASK-1247640091**.

The diff does what the origin task asks and does it in the module's own idiom: roots are parameters, every read degrades to `None` instead of raising, nothing samples, and the `/proc` walk stays a single scan per page build. Findings are **3 × [minor] + 2 × [nit]**, no blockers, no majors — so this approves.

### Independent verification (re-run here, not taken from the PR body)

- `pytest src/test` at `5351ee4` — **681 passed**.
- `coverage report` — `sysinfo.py` 99%, `sources.py` 95%, `page.py` 100%.
- `pycodestyle --max-line-length 120 src` — clean. `mypy` — clean. All six GitHub checks green, including `check-version-bump` on 0.16.14 → 0.16.15.
- **Live host, not fixtures.** `cgroup_memory()` against this runner's real `/sys/fs/cgroup`: charged 6.71 GB, resident 5.58 GB, reclaimable 1.08 GB, consistent with `memory.stat`. `top_procs(5)` vs `ps -eo pid,comm,rss --sort=-rss | head`: **same five processes, same order, RSS within ~2%** — the origin task's second acceptance detail holds.
- The first acceptance detail (graceful degradation without cgroup v2) holds at the data layer and at the page layer; see the two comments on the tile's fallback for the edges it does *not* cover.

### Dimension report

- **Correctness** — the definition-of-done is satisfied end to end (`cgroup_memory`, `top_procs`, fixture tests, tile + top-5 panel through `sources.py`/`page.py`, version bump, suites green, PR carries `Closes #74` and both task refs and is flipped ready). Three minors, all on edges rather than the main path: the `reclaimable` derivation counts `shmem`, the tile discards a partial cgroup read that `sysinfo` deliberately preserves, and the new `read_stat` paren guard is unexercised. No scope excursions — every file is inside the issue's declared `alissa-tools-github-revloop/**`, and the two edited pre-existing assertions are mechanical consequences of the new `comm`/`panel_id` fields.
- **Coding practices** — matches the repo idiom closely: comment density, the "root is a parameter" convention, the `_read_int_file` / `_read_stat_keys` / `_sum_or_none` private-helper split, and the JS render functions built the same way as `renderSessions`. Branch `TASK-1369970340-MEMORY-TILE`, PR names the task, PR attached to the task as evidence, commit carries the `alissa-app <support@alissa.app>` co-author trailer. No dead code, no debug leftovers, no untracked TODOs, no new dependencies. One documentation nit: `webui/__init__.py` still maps `sysinfo.py` as "/proc process-tree CPU%/RSS" and still describes the console's live signals as "`alissa tmux ls`, a `/proc` walk of each session's pane-PID tree" — the package docstring is now the only one in the module that has not learned about the container-wide readers, while `sysinfo.py` and `sources.py` were both updated.
- **Security** — nothing found. The new payload is auth-gated behind the existing fail-closed passcode/session/CSRF layer; the process panel publishes `comm` only, never `cmdline` or `environ`, so no argv secrets leak; `comm` is rendered through the existing `esc()` (which `String()`-coerces, so the numeric `pid` is safe too), closing the obvious XSS path from a process named ``; the panel is deliberately read-only, so no new privileged action reaches unmanaged host PIDs; no path/shell/SQL interpolation, no new dependency, no secret in diff or fixtures.
- **Performance** — nothing found that gates merge. The change is net-positive on the loaded path: `dashboard()` hoists `build_index` so one `/proc` walk feeds both the session table and the process list, where a naive wiring would have walked twice. The idle path pays a new cost — a console with zero live sessions used to skip `/proc` entirely and now scans it every poll — but the panel cannot exist without it, and the lazy build is correctly preserved for callers that pass no index. `sorted()` over the full stats dict rather than `heapq.nlargest` is O(N log N) on a few hundred PIDs: not worth changing.
- **Side-effects / blast radius** — nothing found. `read_stat` gains a key (additive; the only in-repo consumers are `tree_usage` and the new `top_procs`, both key-addressed), `sessions()` gains an optional keyword (existing callers unchanged), `Sources.__init__` gains a defaulted `cgroup_root`. `tiles.memory` and `top_procs` are additive JSON fields on `/api/state`, a page-scoped contract with no cross-repo consumer. Nothing irreversible, no migration, no external side effect. Rollout is the ordinary one the issue already scopes out: this PR triggers the PyPI release and the `REVLOOP_VERSION` re-pin in `docker/claude/Dockerfile` lands separately, so nothing is live on Railway until that pin moves. A failure in either reader degrades to `None` and renders "unavailable" — it cannot take the page down.

### Waived majors

None — there are no majors.

### Minors, and where they go

All three minors are follow-up material rather than merge blockers, and per CR5 a deferred minor must be carried by a registered task. They are on **TASK-680186617**. The one worth a second look before merging is the `shmem` caveat: it is the only finding that can make the tile *mislead* rather than merely under-report, on a container with a large tmpfs.

Nothing here needs a re-review round — the loop can converge at round 1 of 22.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py
RHDZMOTA: [triage:pursue] — and you are right that it needs a decision, not a patch, so here is the decision and its reasoning.

The issue's scope text does say `file + slab_reclaimable` verbatim, but that formula is a shorthand for the bucket it names in the same sentence: "the 'not really in use' bucket". Where those two disagree, the words are the requirement and the formula is the approximation — a tile whose whole purpose is to stop an operator misreading a plateau must not itself misread one. Counting swap-backed tmpfs as droppable is the one failure mode that makes the number wrong in the reassuring direction, which is worse than any amount of under-reporting.

So: `reclaimable` becomes `(file - shmem) + slab_reclaimable`, clamped at zero, and `shmem` stops being invisible — it gets its own magnitude in the tile rather than being folded silently into `resident`, because "5 GB of tmpfs" and "5 GB of process heap" are different operator problems even though neither is droppable.

This lands on the follow-up PR, not here: the approve is terminal for this head. Carried by TASK-680186617 (yours) alongside the other two minors, and the deviation from the issue's literal formula will be stated as a judgment call in that PR's delivery note so the origin-side validator sees it rather than discovering it in a diff.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: [triage:pursue] — accepted in full, including the sharper half of the finding.

The gap you actually found is not the `&&` in the tile; it is that `test_cgroup_memory_current_missing_keeps_the_stat_split` asserts a data-layer property while its docstring claims a rendering one. That docstring is mine and it was wrong about the console, so the fix is two-sided: gate the tile on "any of the three is non-null" and let the headline fall back through `bytes()`'s own null handling, AND put a page-level assertion behind the sentence so the claim is pinned where it is made. A test whose prose over-claims is worse than a missing test, because it reads as covered.

Follow-up PR, not this head. TASK-680186617.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py
RHDZMOTA: [triage:pursue] — mutation-verified beats my counterfactual, and this is exactly the hole it missed.

My three injected defects all landed on code paths a fixture already drove; the guard I added is on a path no fixture reaches, so the exercise could not have caught it by construction. Worth recording as the general lesson: a counterfactual only proves the tests that exist are load-bearing, never that the new branch has one.

Adding the sibling assertion you describe — `read_stat` returns None for a paren-less line, and for a `)` preceding its `(` — and I will re-run it as a mutation (revert the clause, confirm the new test fails) rather than just confirming it passes, since passing is not the property in question.

Follow-up PR. TASK-680186617.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: [triage:pursue] — taking `'unavailable'`.

Two reasons beyond consistency with the Volume and GH Rate tiles. First, the origin task's acceptance detail uses that exact word, so the console will now say what the criterion says. Second, and the part that decided it: `'no cgroup v2'` is a *diagnosis*, and the console has no evidence for it. It cannot distinguish an absent controller from a host-namespace deployment where the controller exists but its interface files do not appear at the v2 root — and a wrong diagnosis is worse than no diagnosis, because it sends an operator to check the wrong thing.

Thank you for confirming the namespace assumption against `/proc/self/cgroup` on the runner; that is the check I did not make, and it is what makes the happy path trustworthy rather than merely working here.

Follow-up PR.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: [triage:pursue] — the comment is selling one side of a trade as if it had no other side, and that is a fair thing to flag even at nit weight.

The hoist is still right, but "the two panels can never disagree" is true only about the snapshot's internal consistency; it says nothing about the snapshot's age relative to the pane PIDs resolved after it. A session whose pane starts inside that window shows `--` for one poll. Self-healing in ~10s and cosmetic, as you say — but a comment that omits the cost is how the next person to touch this concludes the ordering is free and moves something else across it.

Adding the clause on the follow-up PR. Code unchanged; the ordering stays as it is.

PR #76, opened 2026-08-10
TASK-479723890: bridge-executor role — run `alissa bridge start` as a second service from the revloop image (issue #73)

### LINKED ISSUE #73: Bridge executor sidecar: run `alissa bridge start` as a second service from the revloop image (env-gated)
Twin of TASK-602054139 in the devloop repo, for the review daemon's image: give Alissa Studio queue-mode agents an always-on executor on this infrastructure. Contract and operator docs live in the studio repo: `docs/design/local-bridge-queue-mode.md` and `docs/howto/queue-agents.md`. The executor daemon is `alissa bridge start` (already in the alissa CLI): registers a `bridgeExecutors` row, polls `/v1/bridge/jobs`, runs each claimed job as an `alissa code --handoff claude` session in its own tmux session, fulfills from a result file.

**PINNED DECISION — sidecar service, not the daemon's start script.** The executor runs as its OWN container: a second Railway service deployed from this same image with an executor role switch. It must NOT join `docker/claude/entrypoint.sh`'s process chain. Rationale (recorded so it survives review rounds):
1. Queue jobs are tmux sessions inside the executor's container and run for hours (default 6h budget). Revloop redeploys/restarts would kill mid-run sessions and burn a retry attempt each (`reconcileResumed` hands back, attempt+1). Daemon and executor must not share a restart domain.
2. A separate service = a separate environment, which is a security control: the executor resolves job-spec env NAMES from its own process env (`resolveJobEnv`), so anything the service holds is nameable by any queue agent of the token's user. Run the executor service with a minimal credential set — NOT the revloop's full env — and document exactly which vars it exposes.
3. An in-container background sidecar (the devloop repo's UI-console pattern) was considered and REJECTED: fine for a stateless console, wrong for hour-long stateful jobs.

Implementation sketch (same image, no second Dockerfile):
- Role switch in `docker/claude/entrypoint.sh` (e.g. `CONTAINER_ROLE=executor`) or a sibling `entrypoint-executor.sh` — implementer's call; executor mode skips the revloop daemon entirely.
- Executor mode: bootstrap the workspace exactly as the daemon does (this repo's bootstrap-from-manifest flow), then exec `alissa bridge start --executor-id "$ALISSA_BRIDGE_EXECUTOR_ID" --label "$ALISSA_BRIDGE_LABEL" --workspace-root "$WORKSPACE_ROOT" [--max-concurrent/--interval/--handoff from env]`.
- Env contract (all `ALISSA_BRIDGE_*`): `ALISSA_BRIDGE_EXECUTOR=0|1` master gate, default 0 (executor mode refuses to start when 0); `ALISSA_BRIDGE_EXECUTOR_ID` required, default `revloop-executor` — MUST differ from the devloop executor's id (same user + same id = mutual takeover via registration takeover); `ALISSA_BRIDGE_LABEL`, `ALISSA_BRIDGE_MAX_CONCURRENT`, `ALISSA_BRIDGE_POLL_SECONDS`, `ALISSA_BRIDGE_HANDOFF` (default claude).
- Identity persistence: the CLI's executor identity file and `CLAUDE_CONFIG_DIR` on the persisted volume, so a restart is a clean re-register rather than a takeover that strands sticky claims. NOTE this repo's 2026-07-29 volume-failure lesson: volume unavailability must degrade non-fatally (log + retry), reusing the entrypoint's existing auth-triage/backoff posture rather than crash-looping.
- `agents.yaml`: keep the claude profile WITHOUT `disable_alissa_code` so job sessions get codeSessions registration + 10-min log checkpoints.
- Firewall: verify `init-firewall.sh` in executor role admits what job sessions need (Alissa API, api.anthropic.com, github.com); assert, don't assume.
- Tests: this repo's `tests-entrypoint-*.sh` suites are the precedent — add executor-mode coverage there (config rendering, gate refusal, identity requirement).

Out of scope: studio-repo changes; auto-creating queue agents; the devloop twin (TASK-602054139).

## Definition of Done

The revloop image can be deployed as a second, executor-role service: role switch + env contract (ALISSA_BRIDGE_EXECUTOR gate default-off, required distinct ALISSA_BRIDGE_EXECUTOR_ID defaulting to revloop-executor, label/concurrency/interval/handoff knobs), workspace bootstrap reused, executor identity and CLAUDE_CONFIG_DIR on the persisted volume with non-fatal volume degradation, agents.yaml still routing through alissa code, firewall verified, and the sidecar decision plus exposed-env security note documented — with executor-mode covered in the tests-entrypoint suites.

## Acceptance criteria

- [ ] Executor role starts alissa bridge start only when ALISSA_BRIDGE_EXECUTOR=1 (default off refuses cleanly), never inside the revloop daemon's process chain; daemon mode is unaffected when the gate is off.
- [ ] ALISSA_BRIDGE_EXECUTOR_ID is required, defaults to revloop-executor (distinct from devloop), identity + CLAUDE_CONFIG_DIR persist on the volume, and volume/auth failures back off non-fatally per this repo's entrypoint posture — covered in tests-entrypoint suites.
- [ ] The docker README documents the sidecar decision, the exact env vars exposed to job specs, minimal-credential guidance, and verified firewall allowances for job sessions.

Autonomous-Scope: docker/claude/*, README.md
Alissa-Task: TASK-34600665

RHDZMOTA:
Closes #73

`CONTAINER_ROLE=executor` boots this same image as an Alissa Studio queue
executor (`alissa bridge start`) instead of the review daemon — deployed as a
**second service**, never as a sidecar of the first.

The split is enforced by construction rather than by convention: the executor
role `exec`s the CLI, so the container's only process is the executor (no
`alissa worker`, no console, no `alissa-revloop`), and the daemon role never
starts a bridge however the `ALISSA_BRIDGE_*` env is set. Both directions are
asserted in the new suite.

**What landed**

* **Role switch + gates**, resolved before any bootstrap so a deploy mistake
  fails at the top of the log. `ALISSA_BRIDGE_EXECUTOR` defaults **off** and
  refuses cleanly; `ALISSA_BRIDGE_EXECUTOR_ID` is required, defaults to
  `revloop-executor` (distinct from the devloop's — identity is
  `(userId, executorId)` and registration *takes over*, so a shared id means two
  services evicting each other) and is slug-validated at boot rather than at
  register time. `--handoff` is pinned structurally to `claude`; label /
  concurrency / interval are pass-through, on this repo's usual
  env-&gt;CLI-default contract.
* **Identity persistence.** `ALISSA_CONFIG_DIR` defaults onto the volume in this
  role, so `bridge-executor.json` (id + fingerprint) survives a redeploy instead
  of re-registering as a changed machine. That choice also makes an unavailable
  volume land in the `alissa auth login` preflight's existing "config dir
  unwritable" class — capped backoff forever, never a crash loop.
  `CLAUDE_CONFIG_DIR` is verified under the workspace root, warn-only.
* **Bootstrap reuse.** Same manifest generation, same claude first-run seeding,
  same `alissa code workspace sync`. Only `revloop.config.json` and the spawn
  ledger reset are skipped — both belong to a daemon this service never starts.
* **`agents.yaml`** keeps routing through `alissa code` (no
  `disable_alissa_code`), and the resolved file — model pin included — is copied
  into the executor's config dir on every boot, so the image stays the source of
  truth for the profile.
* **Firewall verified, not assumed.** The existing allowlist already admits
  everything a job session needs; `init-firewall.sh` is now sourceable so the
  test asserts that membership against the shipped list instead of a second copy.
* **Docs.** `docker/claude/README.md` gains a "Bridge executor role" section with
  the sidecar decision and its rationale, the deploy recipe, the env table, the
  exposed-env security note, and the persistence/degradation contract.

No dist files were touched, so no version bump (the guard's own scope note
exempts `docker/`, `README.md` and `.github/`).

## Delivery notes

**Judgment calls**

* Role switch lives **inside** `entrypoint.sh` rather than a sibling
  `entrypoint-executor.sh` (the issue left this to the implementer). Steps 0–3
  are shared verbatim — privilege drop, auth triage, model pin, manifest, hub
  sync — and a sibling script would have duplicated ~250 lines of hard-won
  crash-resilience logic that then drifts.
* Two knobs to arm, not one (`CONTAINER_ROLE=executor` **and**
  `ALISSA_BRIDGE_EXECUTOR=1`), because the issue specifies both. The role
  selects; the gate consents. A service with only the gate set stays the daemon.
* "Refuses cleanly" is implemented as `die` (non-zero exit + a message naming the
  remedy), matching this entrypoint's existing posture for misconfiguration (the
  console passcode gate). It is not exit 0.
* `ALISSA_BRIDGE_POLL_SECONDS` maps to the CLI's `-i/--interval`; there is no
  `--poll-seconds` flag. The env name from the issue is kept and the mapping is
  documented in the entrypoint and the README.
* Persisting the identity was done by relocating the **whole** `ALISSA_CONFIG_DIR`
  onto the volume, not by symlinking `bridge-executor.json`. A symlink write
  failure would throw inside `alissa bridge start` (that write is not guarded);
  relocating the dir puts volume failure in the entrypoint's existing retry class
  instead. The cost is that `agents.yaml` must be copied across (step 2e-ii) —
  **and, as round 1 found, that the CLI's executor lockfile and the stored API
  token come along too.** The lock is now swept before `exec` (a fresh container
  has no executor running, so a lock found there is stale by construction) and
  the token at rest is documented. See the round-1 note below.
* No build ARGs for any bridge knob — they are runtime-only, like the console
  knobs. Baking `CONTAINER_ROLE` would make the artifact role-specific, and an
  ARG default for the arming gate is a value every deploy inherits. Rationale is
  in the Dockerfile next to the console's equivalent note.
* Label is pass-through (omitted when unset) rather than defaulted to something
  friendlier. That follows the repo's "no hidden entrypoint layer" precedence
  contract; the consequence is that Studio shows the container hostname unless
  `ALISSA_BRIDGE_LABEL` is set, and the README says so.
* `ALISSA_REVIEW_SKILLS` is reused unchanged for the executor's manifest (it
  still installs the review skills into the workspace). The issue asked for the
  bootstrap "exactly as the daemon does", so no new skills knob was invented.

- [ ] **No container was built or run.** There is no docker in this environment;
      every entrypoint assertion comes from booting the real `entrypoint.sh`
      with stubbed CLIs, which is this repo's established precedent. The image
      layers themselves are unexercised.
- [ ] **`alissa bridge start` was never executed.** Its flag set was read off the
      installed CLI (`alissa bridge start --help` and the option registration in
      `cli.mjs`), and the entrypoint's argv is asserted against a stub. Nothing
      registered a real `bridgeExecutors` row, claimed a job, or ran one.
      (A `--help` probe does **not** validate flags — commander short-circuits
      before parsing — so the help listing is the evidence, not an exit code.)
- [ ] **Identity persistence across a real restart is unproven.** The tests
      assert the identity path lands on the volume and that the config dir is
      honoured; they do not restart a container and confirm the CLI re-registers
      without a takeover or a fingerprint change.
- [ ] **The firewall was never raised.** `init-firewall.sh` needs root and
      `NET_ADMIN`. The test asserts the allowlist's *contents*; it does not
      confirm a job session actually reaches those hosts through live iptables
      rules. Worth one `ALISSA_ENABLE_FIREWALL=1` smoke run on the first deploy.
- [ ] **Executor-id collision with the devloop's executor** is documented and
      defaulted apart, not detected. Nothing at boot can see the other service's
      id; if an operator sets them equal, the takeover loop is what will surface
      it.
- [ ] **The `ALISSA_CONFIG_DIR` relocation moves the stored API token onto the
      volume** (`config.json`). Now stated in the operator docs too, not only
      here. It is a real change in where a secret rests, and accepting it is the
      operator's call — nothing about the executor needs the token persisted; it
      is the price of persisting the identity file beside it.
- [ ] **tmux env inheritance for job sessions** was reasoned about, not measured:
      `agents.yaml` is written to *both* the baked home and the relocated config
      dir so the profile resolves either way, rather than proving which one a
      spawned session reads.

**Verification**

* `bash docker/claude/tests-entrypoint-executor.sh` — **PASS** (new suite, 11
  cases / 66 assertions: both roles, gate refusal, empty + malformed + over-long
  id, pass-through knobs, extra-CMD-args, unwritable volume with mid-flight heal,
  stale executor lock, `CLAUDE_CONFIG_DIR` on and off the volume, firewall
  allowlist membership).
* **Counterfactual on the round-1 `[major]`**: with the stale-lock sweep line
  deleted and nothing else changed, case 7b fails three ways — the boot exits 1,
  the log carries the CLI's `is already running on this machine`, and
  `bridge-argv` is never written. Restored, it passes. The `alissa` stub models
  the CLI's own `acquireExecutorLock` / `runningExecutorPid` so that case
  exercises the real refusal rather than a stand-in.
* `bash docker/claude/tests-entrypoint-config.sh` — **PASS** (incl. the live
  library-default cross-check against the pinned 0.16.14).
* `bash docker/claude/tests-entrypoint-identity.sh` — **PASS**.
* `bash docker/claude/tests-entrypoint-auth.sh` — **PASS**.
* `bash docker/claude/tests-entrypoint-ui.sh` — **PASS** (against the real
  installed sidecar).
* `bash check-style.sh alissa-tools-github-revloop` — **PASS**.
* `bash check-types.sh alissa-tools-github-revloop` — **PASS** (29 files).
* `bash tests-unit.sh alissa-tools-github-revloop` — **659 passed, 1 failed**.
  The failure is `test_webui_sysinfo.py::test_read_stat_basic`, which parses this
  machine's `/proc`; it fails identically on an untouched checkout of this repo
  and no Python file is in this diff. Pre-existing and environmental, not
  introduced here — but flagging it rather than calling the run green.
* `bash -n` on every modified shell file; `.github/workflows/check-entrypoint.yaml`
  parsed with pyyaml and its step list confirmed.
* `firewall_domains` sourced and its output diffed against the previous inline
  array, including the `ALISSA_FIREWALL_EXTRA` extension path.

**Scope touched**

Declared scope was `docker/claude/*, README.md`.

* `docker/claude/entrypoint.sh` — role switch, gates, identity/config-dir
  placement, executor `exec` point, daemon-only guards.
* `docker/claude/init-firewall.sh` — allowlist extracted into a sourceable
  `firewall_domains`, body moved into `init_firewall` behind the
  direct-execution guard this repo already uses in `revloop-config.sh`.
  Behaviour is unchanged; the refactor exists so the allowlist can be asserted.
* `docker/claude/tests-entrypoint-executor.sh` — new.
* `docker/claude/Dockerfile` — comments only (the two-role note and the
  "deliberately not an ARG" rationale). No layer, pin or ENV changed.
* `docker/claude/README.md` — the new section plus the entrypoint-steps list.
* `README.md` — pointer to the new role, and the entrypoint-suite list corrected
  (it claimed "two shell suites"; there were four before this PR, five now).
* **Excursion:** `.github/workflows/check-entrypoint.yaml` — outside the declared
  scope. Adding the suite without wiring it into CI would ship a test nothing
  runs; it is one step plus a comment, and no other workflow was touched.

### Round 1 (request_changes) — what changed

All five inline findings and both body-level findings triaged `[pursue]`; replies
are on their threads and in
[this comment](https://github.com/fahera-mx/alissa-github-review-daemon/pull/76#issuecomment-5236566907).

* **[major] stale executor lock.** Real, and mine. Persisting `ALISSA_CONFIG_DIR`
  also persisted the CLI's `bridge/executor-<id>.lock`, whose staleness is decided
  by a bare `kill(pid, 0)` — no boot id, no start time. On a volume that survives
  an ungraceful stop, the next boot's fresh PID namespace can make the recorded
  PID look live, the CLI refuses to start, and the platform restart-loops it.
  Swept just before `exec`, scoped to this executor's id (an `executor-*.lock`
  glob would delete a live peer's lock if a config dir were ever shared, while a
  stale lock under another id is never consulted for ours).
* **[minor]** the boot-time id check now carries the CLI's 64-character bound;
  **[minor]** the API token at rest on the volume is documented for operators;
  **[nit]** README step label `3b`→`3d`; **[nit]** the extra-CMD-args
  pass-through the README promotes as the smoke test is now covered by a test.
* **[minor] commit trailers** — the base commit was amended rather than topped
  with a trailer-only commit (a trailer describes the commit it sits on), so this
  round is a force-push: `24c2aa6` → `770744e` + `691c191`.
* **[nit] `is_truthy` vs `ALISSA_ENABLE_FIREWALL`** — comment softened rather than
  the gate widened, with the underlying footgun registered as **TASK-112733143**
  (`committed`, soft-linked to TASK-479723890). Folding that gate in would make a
  deploy that today sets `true` — and therefore never needed `--cap-add=NET_ADMIN`
  — start attempting the firewall init and die on it, turning a silently-degraded
  container into a crash-looping one. That wants a deliberate rollout, not a
  drive-by inside a fix round. Happy to be overruled and do it here.

### Round 2 (approve) — one deferred nit

Approved. One new `[nit]`: in `tests-entrypoint-executor.sh`'s header case map the
`4b` entry is filed between `6` and `7` while the suite runs it between `4` and
`5`, and its arrow column is one space out. Both true, and my error from the
round-1 fix.

**Deferred, not fixed here — `[triage:later]`, tracked as TASK-2137886541**
(`committed`, soft-linked to TASK-479723890).

The branch carries an approve, so it takes no further commits: an approved branch
is merge-racy and anything pushed after the squash-merge is silently lost. The
normal path — a follow-up PR off `main` — does not work for *this* finding:
`docker/claude/tests-entrypoint-executor.sh` does not exist on `main`, it arrives
with this PR. A branch cut from the default branch could therefore only re-add
the whole file, which add/add-conflicts with this PR whichever merges first, and
basing it on this branch is the stranding hazard that makes "off the default
branch" a rule. Risking a merge conflict on an approved PR to reorder a comment
block is the wrong trade, so the task carries the sequencing constraint instead:
branch off `main` **after** this merges.

Nothing functional is deferred — the suite's execution order was always correct.

Origin task: TASK-34600665
Implementation task: TASK-479723890

### DIFF
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -554,7 +554,8 @@ What it must **not** carry:
 
 ### Persistence, and what happens when the volume is not there
 
-Two things have to outlive a redeploy, and both live on the `/workspace` volume:
+Two things have to outlive a redeploy, and both live on the `/workspace` volume —
+plus a third that lands there as a consequence, and that you should know about:
 
 * **the executor identity** — `${ALISSA_CONFIG_DIR}/bridge-executor.json`, the id
   plus the fingerprint Studio displays. In this role `ALISSA_CONFIG_DIR` defaults
@@ -566,9 +567,34 @@ Two things have to outlive a redeploy, and both live on the `/workspace` volume:
   if a deploy moved it off; a queue job is long and unattended, and re-logging in
   is manual.
 
+* **the Alissa API token — a consequence, not a goal.** `alissa auth login` (the
+  same preflight the daemon role runs) writes the *verified* `ALISSA_API_TOKEN`
+  in cleartext to `${ALISSA_CONFIG_DIR}/config.json` with default permissions.
+  Relocating that directory onto the volume therefore moves that secret from the
+  ephemeral home to persistent storage. Nothing about the executor needs it
+  persisted — it is the price of persisting the identity file that sits beside
+  it — but the volume now holds a live credential, so give it the same handling
+  you would give any credential store: no snapshots into shared buckets, no
+  attaching it to a second service to "have a look", and rotate the token if the
+  volume is ever exposed. This is the same class of exposure as
+  [Exactly what a job spec can name](#exactly-what-a-job-spec-can-name) above,
+  one layer down: that section is about what a job can *read from the process*,
+  this is about what sits *on the disk*.
+
 The resolved `agents.yaml` is copied into the executor's config dir on **every**
 boot, so the image — not a file left behind by an older image — is always the
 source of truth for the profile.
+
+The CLI's executor **lockfile** (`…/bridge/executor-<id>.lock`) also lands in
+that directory, and it is deliberately *not* treated as persistent state: the
+entrypoint deletes this executor's lock immediately before starting. A lock is a
+claim that a process on this machine is already running, decided by a bare
+`kill(pid, 0)`; a container that died ungracefully leaves one behind, and the
+next boot's fresh PID namespace can easily make that dead PID look alive —
+refusing to start, forever. A fresh container has no executor running by
+definition, so the lock is stale by construction. Only *this* executor's lock is
+removed, never every `executor-*.lock`, so a lock belonging to some other id is
+left alone rather than pulled out from under whoever owns it.
 
 Volume unavailability is **not** fatal. Because the identity dir is the same
 directory the `alissa auth login` preflight probes, a missing or root-owned mount
@@ -642,7 +668,8 @@ volumes:
    daemon's on-demand `alissa code workspace add` no-ops on a repo already listed
    in the manifest, leaving an empty folder and looping forever hub-ifying a hub
    that never completes.
-3b. **Executor role only, and the end of the shared path**: `exec alissa bridge
+3d. **Executor role only, and the end of the shared path**: drop this executor's
+   stale lockfile, then `exec alissa bridge
    start …`. `exec`, so the container's only process is the executor — steps 4
    and 5 below never run in this role, and the daemon role never reaches this
    step. Nothing after this point applies to an executor service.
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -9,6 +9,11 @@ and enqueues sessions; the worker is what drains the queue and spawns reviewers,
 so the image bundles all three tiers (see the top-of-file comment in
 [`Dockerfile`](./Dockerfile)).
 
+The same image also serves a **second, separate service**: with
+`CONTAINER_ROLE=executor` it runs `alissa bridge start` as an Alissa Studio queue
+executor instead of the review loop. See
+[Bridge executor role](#bridge-executor-role-a-second-service-from-this-same-image).
+
 ## Build
 
 ```sh
@@ -437,6 +442,156 @@ docker run -d --name alissa-review \
 
 See [`init-firewall.sh`](./init-firewall.sh) for the allowlist.
 
+## Bridge executor role (a SECOND service from this same image)
+
+`CONTAINER_ROLE=executor` boots this image as an **Alissa Studio queue
+executor** instead of the review daemon. The executor is `alissa bridge start`:
+it registers a `bridgeExecutors` row, polls `/v1/bridge/jobs`, and runs each
+claimed job as an `alissa code --handoff claude` session in its own tmux session
+inside the workspace hubs.
+
+### Why a separate service, and not a sidecar of the daemon
+
+This is a pinned decision, recorded here so it survives the next round of
+refactoring:
+
+1. **Different restart domains.** A queue job is a tmux session that runs for
+   hours (6h default budget). Every revloop redeploy or restart would kill the
+   in-flight ones, and each hand-back costs the job a retry attempt
+   (`reconcileResumed`, attempt + 1). The review daemon redeploys often; the
+   executor must not.
+2. **A separate service is a separate environment, and that is a security
+   control.** The executor resolves a job spec's env **names** out of its own
+   process environment (`resolveJobEnv`), so *every variable this service holds
+   is nameable by any queue agent belonging to the token's user*. Running it as
+   its own service is what lets you give it a minimal credential set instead of
+   the review daemon's full one. See [Exactly what a job spec can
+   name](#exactly-what-a-job-spec-can-name) below.
+3. **An in-container background sidecar was considered and rejected.** That
+   pattern (the devloop repo's UI console) is fine for a stateless console; it is
+   wrong for hour-long stateful jobs, for reason 1.
+
+The entrypoint enforces the split by construction rather than by convention: the
+executor role `exec`s the CLI, so the container's only process is the executor —
+no `alissa worker`, no `alissa-revloop`, no console. And the daemon role never
+starts a bridge, however the `ALISSA_BRIDGE_*` variables are set. Both directions
+are asserted in
+[`tests-entrypoint-executor.sh`](./tests-entrypoint-executor.sh).
+
+### Deploying it
+
+Same image, same Dockerfile, no second build. On Railway: add a second service
+from this repo, attach **its own** volume at `/workspace`, and set the runtime
+variables below. Locally:
+
+```sh
+docker run -d --name alissa-bridge-executor \
+  -e CONTAINER_ROLE=executor \
+  -e ALISSA_BRIDGE_EXECUTOR=1 \
+  -e ALISSA_BRIDGE_EXECUTOR_ID=revloop-executor \
+  -e ALISSA_BRIDGE_LABEL="Revloop executor" \
+  -e ALISSA_API_TOKEN=alissa_… \
+  -e GH_TOKEN=ghp_… \
+  -e ALISSA_REVIEW_REPOS="fahera-mx/studio.alissa.app" \
+  -v alissa-executor-workspace:/workspace \
+  alissa-review-daemon
+```
+
+Extra CMD args pass through to `alissa bridge start`, so
+`docker run … alissa-review-daemon --once` is a one-poll smoke test.
+
+### Configuration (runtime env only — no build ARGs)
+
+Unlike the daemon knobs, none of these is a build ARG. `CONTAINER_ROLE` is what
+makes one image serve two services, so baking it would make the artifact
+role-specific; `ALISSA_BRIDGE_EXECUTOR` is an arming gate, so it ships unset; and
+the rest only mean anything on a service that already set those two.
+
+| Variable | Default | Meaning |
+| --- | --- | --- |
+| `CONTAINER_ROLE` | `daemon` | `executor` selects this role. Any other value **dies at boot** naming the two the image ships |
+| `ALISSA_BRIDGE_EXECUTOR` | *(unset ⇒ off)* | the arming gate: `1`/`true`/`yes`/`on`. With the role set and this off, the container **refuses cleanly** and registers nothing — selecting the role is not consent to claim jobs |
+| `ALISSA_BRIDGE_EXECUTOR_ID` | `revloop-executor` | **required** (an explicitly empty value is fatal), validated as a slug at boot. It **must differ from every other executor of the same Alissa user** — the devloop image's executor above all. Identity is `(userId, executorId)` and registration *takes over* an existing row, so a shared id means two services evicting each other in a loop and stranding every sticky claim pinned to it. Never left to the CLI's own fallback (a slugified hostname, which on a platform is a fresh string per deploy) |
+| `ALISSA_BRIDGE_HANDOFF` | `claude` | agent used when a job spec names none. Structural, like `ALISSA_AGENT_PROFILE`: it must name a profile in the baked [`agents.yaml`](./agents.yaml) |
+| `ALISSA_BRIDGE_LABEL` | *(CLI default: the hostname)* | human name shown in Studio; **pass-through** — unset ⇒ the CLI decides |
+| `ALISSA_BRIDGE_MAX_CONCURRENT` | *(CLI default)* | jobs to run at once (the CLI clamps to 1–16); **pass-through** |
+| `ALISSA_BRIDGE_POLL_SECONDS` | *(CLI default: 15)* | seconds between queue polls; maps to the CLI's `--interval`; **pass-through** |
+
+The model pin works exactly as it does for the daemon: `ALISSA_AGENT_MODEL`
+(default `opus`) is rewritten into the `claude` profile's `command:` at boot, and
+job sessions inherit it. The profile deliberately carries **no**
+`disable_alissa_code`, which is what makes the CLI launch it via `alissa code -y
+--handoff claude` — that wrapper is what registers the codeSession and its
+10-minute log checkpoints. Adding the flag would launch a bare `claude` and lose
+both.
+
+### Exactly what a job spec can name
+
+A job spec carries environment variable **names**, never values; the executor
+resolves each name against its own process environment and fails the job
+(`spec_rejected`) if one is missing. So the set of variables a queue agent can
+pull into a job session is *exactly the set this service holds*. Keep it minimal.
+
+What the executor genuinely needs:
+
+| Variable | Why |
+| --- | --- |
+| `ALISSA_API_TOKEN` | polls the job queue and reports results — without it there is no executor |
+| `GH_TOKEN` (or `GITHUB_TOKEN`) | the workspace bootstrap clones the hubs, and job sessions do the repo work |
+| `CLAUDE_CODE_OAUTH_TOKEN` / `ANTHROPIC_API_KEY` | only if you are not using the persisted `claude /login` on the volume (which is preferred — see [claude auth](#claude-auth-log-in-once-persisted-on-the-volume-recommended)) |
+| `ALISSA_REVIEW_REPOS` | the repo allowlist the manifest (and therefore the hubs) is built from |
+
+What it must **not** carry:
+
+* **the reviewer's GitHub token.** An executor posts no verdicts. If
+  `ALISSA_REVIEWER_TOKEN_ENV` is set on an executor service the entrypoint warns
+  and ignores it — but the *token itself* being present is the problem, because a
+  job spec can name it.
+* **`ALISSA_UI_PASSCODE`.** The console is a review-loop surface and is never
+  started in this role.
+* anything else the review daemon happens to hold. Copying the daemon service's
+  variable set across is the mistake this whole split exists to prevent.
+
+### Persistence, and what happens when the volume is not there
+
+Two things have to outlive a redeploy, and both live on the `/workspace` volume:
+
+* **the executor identity** — `${ALISSA_CONFIG_DIR}/bridge-executor.json`, the id
+  plus the fingerprint Studio displays. In this role `ALISSA_CONFIG_DIR` defaults
+  to `/workspace/.alissa-config` (the CLI's own default is the *ephemeral* home,
+  which would mint a new fingerprint on every boot and re-register as a changed
+  machine). Set it explicitly only if you move the volume.
+* **the claude credential** — `CLAUDE_CONFIG_DIR`, already `/workspace/.claude-config`
+  in this image. The entrypoint verifies it is under the workspace root and warns
+  if a deploy moved it off; a queue job is long and unattended, and re-logging in
+  is manual.
+
+The resolved `agents.yaml` is copied into the executor's config dir on **every**
+boot, so the image — not a file left behind by an older image — is always the
+source of truth for the profile.
+
+Volume unavailability is **not** fatal. Because the identity dir is the same
+directory the `alissa auth login` preflight probes, a missing or root-owned mount
+lands in that gate's "config dir unwritable" class: it logs what it saw and
+retries with capped backoff, forever, and proceeds the moment the mount appears —
+[the same posture](#alissa-auth-failures-are-triaged-not-guessed) the rest of the
+entrypoint takes, and deliberately not a crash loop.
+
+### Firewall
+
+The egress allowlist is shared by both roles and **verified**, not assumed: a job
+session needs the Alissa API (`api.alissa.app`), `api.anthropic.com`, GitHub
+(`github.com`, `api.github.com`, `codeload.github.com`,
+`objects.githubusercontent.com`) and the skill/installer hosts
+(`skills.alissa.app`, `share.alissa.app`), and every one of those is already in
+[`init-firewall.sh`](./init-firewall.sh). Nothing had to be added for this role.
+`firewall_domains` there is sourceable precisely so
+[`tests-entrypoint-executor.sh`](./tests-entrypoint-executor.sh) asserts that
+membership against the shipped list instead of a second copy of it. Package
+registries (`registry.npmjs.org`, `pypi.org`, `files.pythonhosted.org`) are
+allowed too, which a job that installs dependencies will want; extend with
+`ALISSA_FIREWALL_EXTRA` for anything else.
+
 ## docker-compose
 
 ```yaml
@@ -465,7 +620,11 @@ volumes:
 
 ## What the entrypoint does
 
-0. As root: `chown` the `/workspace` mount to `alissa`, (optionally) raise the
+0. Resolve `CONTAINER_ROLE` (`daemon` by default) and, in the `executor` role,
+   its gates — the arming flag, the executor id, and where the identity is
+   persisted. A bad role or an unarmed executor **dies here**, before any
+   bootstrap. See [Bridge executor role](#bridge-executor-role-a-second-service-from-this-same-image).
+0b. As root: `chown` the `/workspace` mount to `alissa`, (optionally) raise the
    egress firewall, then drop to `alissa` via `gosu`. Everything below runs
    unprivileged.
 1. Preflight + onboard the identities: validate `gh` (fatal if missing) and run
@@ -483,6 +642,10 @@ volumes:
    daemon's on-demand `alissa code workspace add` no-ops on a repo already listed
    in the manifest, leaving an empty folder and looping forever hub-ifying a hub
    that never completes.
+3b. **Executor role only, and the end of the shared path**: `exec alissa bridge
+   start …`. `exec`, so the container's only process is the executor — steps 4
+   and 5 below never run in this role, and the daemon role never reaches this
+   step. Nothing after this point applies to an executor service.
 4. Start `alissa worker --daemon`, wait until it reports running (the daemon only
    *warns* if the worker is absent, so ordering matters).
 4b. When `ALISSA_UI_ENABLED` is set, start the reviewer console
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -11,6 +11,24 @@
 #
 # The daemon is a thin poller; the worker is what actually spawns reviewers, so
 # the worker MUST be running first — the daemon only warns if it isn't.
+#
+# TWO ROLES, ONE IMAGE (CONTAINER_ROLE, issue #73)
+# ------------------------------------------------
+# `CONTAINER_ROLE=executor` boots this same image as an Alissa Studio queue
+# EXECUTOR instead: it shares steps 0-3 (identities, workspace bootstrap, hub
+# sync) and then `exec`s `alissa bridge start` — no worker, no console, no
+# `alissa-revloop`. The executor is a SEPARATE SERVICE, not a sidecar of the
+# daemon, and the split is the point:
+#
+#   * queue jobs are hours-long tmux sessions; a revloop redeploy would kill
+#     them mid-run and burn a retry attempt each. Different restart domain.
+#   * the executor resolves a job spec's env NAMES out of its OWN process env,
+#     so every variable this service holds is nameable by any queue agent of the
+#     token's user. A separate service is what keeps that set minimal — see
+#     "Bridge executor role" in docker/claude/README.md.
+#
+# So the two roles never share a process chain: whichever one is selected, the
+# other's supervisor is never started.
 # =============================================================================
 set -euo pipefail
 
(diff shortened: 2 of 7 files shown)


alissa-app:
- 2026-08-10 05:57:23 UTC — `review-alissa-github-review-daemon-pr76-r1-292523` — round 1 of 22 — spawned
- 2026-08-10 06:23:35 UTC — `review-alissa-github-review-daemon-pr76-r2-f32e14` — round 2 of 22 — spawned

docker/claude/entrypoint.sh
```diff
@@ -28,6 +46,81 @@ WORKSPACE_ROOT="${ALISSA_WORKSPACE_ROOT:-/workspace}"
...
+  #     fingerprint and re-register as a changed machine. Point the whole config
+  #     dir at the persisted volume instead (it also holds the verified API token,
+  #     same posture as the claude credential already kept there), and EXPORT it
+  #     so the CLI actually sees it.
+  #
+  #     Volume unavailability is handled, not crashed on: this is the same dir
+  #     the auth preflight's writability probe (class 2 below) guards, so a mount
+  #     that is missing or root-owned degrades to that capped backoff — log and
+  #     retry forever, never a crash loop (2026-07-29 lesson).
+  export ALISSA_CONFIG_DIR="${ALISSA_CONFIG_DIR:-${WORKSPACE_ROOT}/.alissa-config}"
```
alissa-app, line 128: [major] Moving the whole `ALISSA_CONFIG_DIR` onto the volume also moves the CLI's **executor lockfile** there, and that lock is not safe to persist across containers.

In the CLI, `executorLockPath()` is `configDir()/bridge/executor-<id>.lock`, and `runningExecutorPid()` decides whether it is stale with a bare `process.kill(pid, 0)` — no boot id, no start time, no cmdline check. Before this line, `configDir()` was the ephemeral home, so a lock could never outlive its container and the staleness check could never be wrong. After it, the lock lives on `/workspace`.

Failure path: any ungraceful stop (OOM kill, platform hard-restart, SIGKILL after the grace period — the SIGTERM handler waits out the current poll *and* tears down in-flight job sessions before exiting) leaves the lockfile behind. On the next boot the recorded PID is evaluated against a **fresh PID namespace that restarts at 1 and replays the same deterministic boot**. I measured this: booting the real entrypoint three times with stubbed CLIs, the `exec`'d bridge process landed on a constant 52-PID stride each time, i.e. the boot consumes a fixed number of PIDs. In a container that means the same absolute PID every boot — so the stale lock's PID is very likely live, and can be the executor's own process (`exec` gives the CLI the entrypoint shell's PID). `acquireExecutorLock` then throws `A bridge executor "revloop-executor" is already running on this machine (pid N)`, the CLI exits non-zero, the platform restarts it, and it repeats.

That is precisely the crash-loop this file's 2026-07-29 lesson exists to prevent, and it is not in the delivery note's risk list.

Cheapest fix, and it already has a precedent two steps up: step 3b clears the stale spawn ledger on the reasoning that *"a fresh container has no reviewer running by definition"*. The identical argument applies here — a fresh container has no executor running by definition — so drop the stale lock in the 3d block just before `exec`:

```sh
rm -f "${ALISSA_CONFIG_DIR}"/bridge/executor-*.lock
```

(Alternatively, persist only `bridge-executor.json` rather than the whole dir — but the delivery note already explains why the symlink route was rejected, so the sweep is the better fit.) Worth a test case alongside case 7: a lock file left behind with a live PID must not stop the next boot.

docker/claude/entrypoint.sh
```diff
@@ -28,6 +46,81 @@ WORKSPACE_ROOT="${ALISSA_WORKSPACE_ROOT:-/workspace}"
...
+  #     defaulted (never the CLI's slugified-hostname fallback, which on a
+  #     platform is a random per-deploy string), required non-empty, and
+  #     validated here so a bad slug fails at boot instead of at register time.
+  #     `-` not `:-`: an explicitly EMPTY value is an error, not a request for
+  #     the default — it is exactly how a deploy silently falls back to the
+  #     hostname slug.
+  BRIDGE_EXECUTOR_ID="${ALISSA_BRIDGE_EXECUTOR_ID-revloop-executor}"
+  [ -n "${BRIDGE_EXECUTOR_ID}" ] \
+    || die "ALISSA_BRIDGE_EXECUTOR_ID is set but EMPTY — an executor id is required. Leave it unset for the default 'revloop-executor', or give this service its own id; it MUST differ from every other executor of the same Alissa user (the devloop image's executor included), because registering the same id takes the other one over."
+  printf '%s' "${BRIDGE_EXECUTOR_ID}" | grep -qE '^[a-z0-9][a-z0-9-]*$' \
```
alissa-app: [minor] This slug check is looser than the CLI's, so it does not fully deliver the "fails at boot instead of at register time" property the comment above it promises.

The CLI uses `EXECUTOR_ID_RE = /^[a-z0-9][a-z0-9-]{0,63}$/` — a **64-character cap**. This pattern is unbounded, so a 65+ character `ALISSA_BRIDGE_EXECUTOR_ID` sails through the boot gate and then dies inside `alissa bridge start` with `Invalid executor id "…"` — after the whole bootstrap has run, and as a restart loop rather than a boot-time refusal.

One character fixes it:

```sh
grep -qE '^[a-z0-9][a-z0-9-]{0,63}$'
```

and the `die` message below could name the length bound alongside the character set.

docker/claude/README.md
```diff
@@ -437,6 +442,156 @@ docker run -d --name alissa-review \
...
+* **the reviewer's GitHub token.** An executor posts no verdicts. If
+  `ALISSA_REVIEWER_TOKEN_ENV` is set on an executor service the entrypoint warns
+  and ignores it — but the *token itself* being present is the problem, because a
+  job spec can name it.
+* **`ALISSA_UI_PASSCODE`.** The console is a review-loop surface and is never
+  started in this role.
+* anything else the review daemon happens to hold. Copying the daemon service's
+  variable set across is the mistake this whole split exists to prevent.
+
+### Persistence, and what happens when the volume is not there
```
alissa-app, line 555: [minor] This section lists two things that outlive a redeploy — the executor identity and the claude credential — but not the third one this PR actually relocated: **the Alissa API token**.

`alissa auth login` (step 2c, which runs in this role too) writes the verified token in cleartext to `${ALISSA_CONFIG_DIR}/config.json`, with default file permissions. Pointing `ALISSA_CONFIG_DIR` at `/workspace/.alissa-config` therefore moves that secret from the ephemeral home onto the persisted volume. The entrypoint comment at 2d(d) acknowledges it ("it also holds the verified API token"); the operator-facing doc does not, and this is the section an operator reads when deciding what the volume needs to be.

Given the issue's DoD calls for an exposed-env security note specifically, add a third bullet here — the token is now at rest on the volume, so the volume inherits the same handling expectations as the credential store — and it pairs naturally with the [Exactly what a job spec can name](#exactly-what-a-job-spec-can-name) section above.

docker/claude/README.md
```diff
@@ -483,6 +642,10 @@ volumes:
    daemon's on-demand `alissa code workspace add` no-ops on a repo already listed
    in the manifest, leaving an empty folder and looping forever hub-ifying a hub
    that never completes.
+3b. **Executor role only, and the end of the shared path**: `exec alissa bridge
```
alissa-app: [nit] The label `3b` collides with a different step: `entrypoint.sh` already has its own `3b` (the stale spawn-ledger reset), and the executor `exec` is labelled `3d` there. A maintainer cross-referencing the two files lands on the wrong block.

Renaming this to `3d` would match the source and cost nothing.

docker/claude/tests-entrypoint-executor.sh
```diff
@@ -0,0 +1,394 @@
...
+#   5. daemon role, bridge env set        -> still the daemon; bridge never runs
+#   6. unknown CONTAINER_ROLE             -> dies naming the two it ships
+#   7. volume unwritable                  -> capped backoff, NOT a crash loop,
+#                                            and it self-heals mid-flight
+#   8. CLAUDE_CONFIG_DIR off the volume   -> WARNs, still boots
+#   9. firewall + agents.yaml             -> the allowlist really does admit what
+#                                            a job session needs; the profile
+#                                            really does still route via alissa code
+#
+# Usage: bash docker/claude/tests-entrypoint-executor.sh
```
alissa-app, line 36: [nit] The suite covers the flag mapping thoroughly but not the **extra-CMD-args pass-through**, which the README promotes as the smoke test (`docker run … alissa-review-daemon --once`). Since `exec alissa bridge start "${BRIDGE_ARGS[@]}" "$@"` is what makes that work, a one-line case — boot with a trailing `--once` and assert it reaches `bridge-argv` after the structural flags — would pin the documented behaviour.

Worth noting the suite is otherwise excellent: I ran it against this branch and all 57 assertions pass, and the negative assertions (case 2's "no worker, no revloop", case 5's "the gate alone never starts an executor") are the right ones to have picked.

alissa-app (changes requested):
## Review verdict: **request_changes** — round 1 of cap 22

Review task: **TASK-235063392** · Origin task: **TASK-479723890** · Issue #73

Solid, unusually well-reasoned change: the role split is enforced by `exec` rather than by convention, both directions are asserted, and the delivery note's "not verified" list is honest. I booted the real entrypoint myself — `tests-entrypoint-executor.sh` passes (57 assertions), and every CLI flag the entrypoint emits exists on the installed `alissa bridge start` (`--executor-id`, `--label`, `-i/--interval`, `--max-concurrent`, `--workspace-root`, `--handoff`, `--once`). I also confirmed against the CLI source that `configDir()` honours `ALISSA_CONFIG_DIR` for **both** `agents.yaml` and `bridge-executor.json`, so step 2e-ii's copy is genuinely load-bearing and not defensive. The Dockerfile declares no `CMD`, so the `"$@"` pass-through in 3d is safe and the README's `docker run … --once` claim holds.

One **[major]** blocks approval: relocating the whole `ALISSA_CONFIG_DIR` onto the volume also relocated the CLI's *executor lockfile* there, which reintroduces the crash-loop class this entrypoint is built to avoid. Details inline.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 3 |
| nit | 3 |
| question | 0 |

### Dimension report

- **Correctness** — the DoD is met: role switch, both gates, defaulted+validated id, bootstrap reuse, identity on the volume, non-fatal volume degradation, `agents.yaml` unchanged for `alissa code` routing, firewall asserted, docs, and a new suite wired into CI (all checks green). Two findings: the **[major]** stale-lock hazard and a **[minor]** gap between the boot-time id validation and the CLI's own.
- **Coding practices** — matches this entrypoint's idiom closely (die-at-boot for misconfiguration, pass-through contract, comment density). One **[minor]**: the commit carries no `Co-authored-by:` trailers, which every other non-merge commit in recent history does (C1). Two **[nit]**s below.
- **Security** — the strongest part of the PR: the `resolveJobEnv` exposure is documented with a concrete allow/deny list, `ALISSA_REVIEWER_TOKEN_ENV` warns rather than silently ignores, the console is hard-off in this role, no secrets in the diff, no new dependency. One **[minor]**: the persistence section omits that the *Alissa API token itself* now rests on the volume.
- **Performance** — nothing found. The executor path is strictly less work than the daemon path (skips `revloop.config.json` and the ledger); the `agents.yaml` copy is one small file per boot; the firewall refactor resolves the same host list exactly once, as before.
- **Side-effects / blast radius** — `init-firewall.sh` is the only shared-path file touched, and the refactor is behaviour-preserving: I ran it both ways (direct execution still raises the firewall; sourcing only defines `firewall_domains`). No `CMD` in the image, so `"$@"` cannot smuggle daemon flags into `alissa bridge start`. All four README anchors added here resolve. The `.github/workflows/check-entrypoint.yaml` excursion is outside the declared `Autonomous-Scope` but is justified in the delivery note and is the right call — a suite nothing runs is not coverage.

### Not waived

The `[major]` is not waived. Shortest path back to approve: sweep the stale executor lock before `exec` (or scope the persisted dir to the identity file), plus the two one-line fixes. The minors and nits are individually fine to `[triage:later]` with a task.

### Findings with no anchorable diff line

- **[minor] The commit carries no `Co-authored-by:` trailers.** `24c2aa6` has an empty trailer block, while every other non-merge commit in recent history carries `Co-authored-by: alissa-app <support@alissa.app>` (and some also the agent trailer) — `a5752df`, `5623806`, `6a32cb1`, `09d2c4e`. This is the alissa-code-git C1 convention and the repo's own established practice; it needs an amend (or a follow-up commit on the branch) before merge.
- **[nit] The new `is_truthy` helper does not cover `ALISSA_ENABLE_FIREWALL`.** Its comment says *"One definition, so the console gate (2d) and the bridge-executor gate below can never drift apart"* — but `entrypoint.sh:144` still tests `[ "${ALISSA_ENABLE_FIREWALL:-0}" = "1" ]`, so `ALISSA_ENABLE_FIREWALL=true` silently leaves egress wide open while reading as enabled. The behaviour is pre-existing and out of this PR's scope, so either fold that third gate in while you are here, or soften the comment so it does not claim coverage it does not have.

RHDZMOTA:
- 2026-08-10 06:12:21 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr76-r1-a1` — fix edge — spawned

docker/claude/entrypoint.sh
RHDZMOTA: [pursue] Confirmed, and thank you — I re-read the CLI and it is exactly as you describe. `executorLockPath()` is `configDir()/bridge/executor-<id>.lock`, and `runningExecutorPid()` decides staleness with a bare `process.kill(pid, 0)` and nothing else: no boot id, no start time, no cmdline. So once the lock lives on the volume, a live-but-unrelated PID in a fresh namespace reads as "still running", `acquireExecutorLock` throws, the CLI exits non-zero, and the platform restarts it into the same state. That is a crash loop I introduced, in the one file whose whole posture is not to have any, and you are right that it was not in my risk list.

I am taking the sweep, with one deliberate narrowing: I remove **this executor's** lock rather than globbing `executor-*.lock`.

```sh
rm -f "${ALISSA_CONFIG_DIR}/bridge/executor-${BRIDGE_EXECUTOR_ID}.lock"
```

The "a fresh container has no executor running by definition" argument is what licenses the sweep, and it holds for every id in the container — but the glob has a failure mode the narrow form does not: if a config dir is ever shared (not how Railway volumes work, but nothing in the entrypoint enforces it), the glob deletes a *live* peer's lock and lets two executors run under different ids with one believing it is alone. A stale lock belonging to some other id is never consulted for ours, so leaving it is harmless cruft, and I would rather leave cruft than delete a lock I cannot prove is dead. The comment records that reasoning.

Test case added alongside case 7, built as the counterfactual you asked for: it plants a lock file containing a **live** PID (the test's own `$$`) and asserts the boot still reaches `alissa bridge start`. I checked that it fails without the sweep line, so it pins the fix rather than merely passing next to it.

docker/claude/entrypoint.sh
RHDZMOTA: [pursue] Correct — `EXECUTOR_ID_RE = /^[a-z0-9][a-z0-9-]{0,63}$/` at cli.mjs:26669, and my pattern was unbounded, so a 65+ character id passed the boot gate and then died inside the CLI after the whole bootstrap had run. That is precisely the "fails at boot instead of at register time" property the comment above it claims, not delivered.

Taking `{0,63}` verbatim, and the `die` message now names the length bound alongside the character set. Added boot cases for a 64-character id (accepted, reaches `bridge start`) and a 65-character one (refused at boot), so the boundary itself is pinned rather than just the regex being swapped.

docker/claude/README.md
RHDZMOTA: [pursue] Agreed, and this is the more important half of the finding: the entrypoint comment admits the token moves, the operator-facing doc did not, and this section is exactly where an operator decides what the volume is.

Added a third bullet stating that `alissa auth login` writes the verified `ALISSA_API_TOKEN` in cleartext to `${ALISSA_CONFIG_DIR}/config.json`, that pointing the config dir at the volume therefore puts that secret at rest there, and that the volume consequently inherits the same handling expectations as the credential store — cross-linked to the job-spec-env section as you suggest.

It was in my delivery note's "not verified" list as an operator call, which was the wrong place for it: that list is what could not be checked, not what the operator needs to know.

docker/claude/README.md
RHDZMOTA: [pursue] Real collision, and a maintainer cross-referencing the two files lands on the stale-ledger reset instead. Renamed to `3d` to match `entrypoint.sh`.

docker/claude/tests-entrypoint-executor.sh
RHDZMOTA: [pursue] Fair — the README promotes `docker run … --once` as the smoke test and nothing pinned it, so `"$@"` could be dropped from the `exec` line and every existing assertion would still pass.

Added a case that boots with a trailing `--once` and asserts it arrives in `bridge-argv` **after** the structural flags, so both its presence and its position are covered.

RHDZMOTA:
## Round 1 triage — the two findings with no anchorable diff line

The five inline findings are answered on their own threads; all five are `[pursue]`. These two are here because they have nowhere to hang.

**[minor] The commit carries no `Co-authored-by:` trailers — `[pursue]`.**

Verified: `24c2aa6` has an empty trailer block, and `a5752df`, `5623806`, `6a32cb1` and `09d2c4e` all carry `Co-Authored-By: alissa-app <support@alissa.app>` (with `09d2c4e` also carrying the agent trailer). I amended the original commit rather than adding a trailer-only commit on top — a trailer describes the commit it sits on, so a separate commit saying "these people co-authored the previous one" records the convention without satisfying it. That makes this round a force-push; the round-1 fixes ride on a second commit above it, and both carry the trailers.

**[nit] `is_truthy` does not cover `ALISSA_ENABLE_FIREWALL` — `[pursue]`, taking your second option, with a task behind the first.**

You are right that the comment overclaims, and you are also right that the behaviour is pre-existing and out of scope. I have softened the comment to name the two gates it actually covers instead of implying it is the container's one definition.

I deliberately did **not** fold the third gate in, and the reason is this file's own posture rather than scope discipline. Folding it in changes what arms a root-privileged control: a deploy currently setting `ALISSA_ENABLE_FIREWALL=true` today gets no firewall and therefore never needed `--cap-add=NET_ADMIN`. After the fold, that same unchanged deploy would attempt the firewall init and hit `die "firewall init failed — did you pass --cap-add=NET_ADMIN?"` — turning a silently-degraded container into a crash-looping one, which is the exact failure class this entrypoint is built to avoid, arriving on a PR that never mentioned the firewall. That wants a deliberate rollout, not a drive-by inside a round-1 fix.

Registered as **TASK-112733143** ("Entrypoint flag truthiness: ALISSA_ENABLE_FIREWALL only accepts 1, unlike every other gate"), status `committed`, soft-linked to TASK-479723890. Its definition of done carries the migration hazard above, so whoever picks it up does not rediscover it the hard way. Happy to be overruled and fold it in here if you would rather have it now.

The `[major]` is fixed rather than waived, and the delivery note in the PR body is updated: the stale-lock hazard is recorded as a defect this round found (not a residual risk), and the API-token-at-rest bullet has moved out of "not verified" into the operator docs where it belongs.

docker/claude/tests-entrypoint-executor.sh
```diff
@@ -0,0 +1,492 @@
...
+#   2. role=executor, gate on             -> execs `alissa bridge start` with the
+#                                            structural flags; no worker, no
+#                                            daemon, no revloop.config.json;
+#                                            identity + agent profile on the volume
+#   3. executor id required + validated   -> empty dies, bad slug dies
+#   4. the pass-through knobs             -> label / concurrency / interval only
+#                                            appear when they are set
+#   5. daemon role, bridge env set        -> still the daemon; bridge never runs
+#   6. unknown CONTAINER_ROLE             -> dies naming the two it ships
+#   4b. the id length bound              -> matches the CLI's own 64, so a long
```
alissa-app, line 25: [nit] The new `4b` entry is filed out of order in this map: it sits after `5` and `6`, but case 4b actually runs between case 4 and case 5 in the body. The `7b` entry just below is placed correctly, so this one reads as the odd one out.

The `->` column is also one space left of every other row (`the id length bound` needs one more space of padding), which is what drew my eye to it.

Both are one-line fixes and purely cosmetic — the suite itself is correctly ordered.

alissa-app (approved):
## Round 2 of 22 — approve

All seven round-1 findings were triaged (`[pursue]` on every one) and all seven hold up. I re-verified each against the branch rather than against the triage replies, and swept the new delta with the full rubric. One new `[nit]`, inline. No blockers, no majors, nothing waived.

### Reconciliation of round 1

| finding | triage | disposition |
| --- | --- | --- |
| `[major]` persisted executor lockfile → restart loop | pursue | **fixed** (`691c191`) — verified below |
| `[minor]` boot-time id regex unbounded vs the CLI's 64 | pursue | **fixed** (`691c191`) |
| `[minor]` API token at rest on the volume undocumented | pursue | **fixed** (`691c191`) |
| `[minor]` no `Co-authored-by:` trailers | pursue | **fixed** (`770744e`, amend) |
| `[nit]` README step `3b` collides with entrypoint's `3b` | pursue | **fixed** (`691c191`) |
| `[nit]` extra-CMD-args pass-through untested | pursue | **fixed** (`691c191`) |
| `[nit]` `is_truthy` comment overclaims | pursue (partial) | **comment fixed**; gate widening **deferred** to TASK-112733143 (open, `committed`) |

**The `[major]`.** The fix is `rm -f "${ALISSA_CONFIG_DIR}/bridge/executor-${BRIDGE_EXECUTOR_ID}.lock"` at `entrypoint.sh:817`, and I checked it three ways:

- **The counterfactual is real.** Deleting line 817 and changing nothing else makes case 7b fail exactly three ways — `bridge-argv` never written, the log carries `is already running on this machine`, and the boot exits 1. Restored, it passes. (My first attempt at this reproduced *nothing*; that was my own pattern bug, not the test's. The line-numbered deletion is the valid run.)
- **The swept path is the CLI's path.** In the installed CLI, `configDir()` returns `process.env.ALISSA_CONFIG_DIR` when set (`cli.mjs:21261`), `executorLockPath(id)` is `configDir()/bridge/executor-<id>.lock` (`:26803`), and `assertExecutorId` validates without normalizing — so the entrypoint's string and the CLI's string are identical, character for character.
- **The narrow scoping is provably safe, not just prudent.** `bridge start` calls `acquireExecutorLock(identity.executorId)` (`:28202`) where `identity` comes from `resolveExecutorIdentity(opts.executorId, …)` — and `--executor-id` is unconditional in `BRIDGE_ARGS`, so the override always wins and the id the CLI locks is exactly the id the entrypoint sweeps. Declining the `executor-*.lock` glob costs nothing and avoids deleting a live peer's lock. I think the narrowing was the better call than my original suggestion.

I also checked the sweep is not a deletion primitive: `BRIDGE_EXECUTOR_ID` is slug-validated at line 107, well before it is interpolated into a path at 817, so a hostile `ALISSA_BRIDGE_EXECUTOR_ID` cannot traverse out of `bridge/`.

**The id bound** is now `^[a-z0-9][a-z0-9-]{0,63}$` — `EXECUTOR_ID_RE` verbatim (`cli.mjs:26669`) — with 64-accepted / 65-refused pinned at the boundary.

**The trailers.** `770744e`'s tree is byte-identical to `24c2aa6`'s (`769faa4c…`), so the force-push was trailer-only and round 1's reading of the base still stands unchanged. Both commits now carry both trailers.

**The deferred nit.** TASK-112733143 exists and is open, and its DoD carries the migration hazard. Folding `ALISSA_ENABLE_FIREWALL` into `is_truthy` would turn a deploy that sets `true` today — no firewall, so no `--cap-add=NET_ADMIN` — into one that attempts the init and dies on it. Converting a silently-degraded container into a crash-looping one, inside a fix round, on a PR that never mentioned the firewall, is the wrong trade. The deferral is right and I am not re-raising it.

### Delta sweep

- **Correctness:** nothing found beyond the nit. All four code fixes verified; `tests-entrypoint-executor.sh` passes (66 assertions, 11 cases), as do the `config`, `identity` and `auth` suites.
- **Coding practices:** nothing found. The `[ … ] && pass || bad` form in the new case 4b matches case 3's existing idiom, and the comment density around the sweep matches this file's norm.
- **Security:** nothing found, and the dimension improved — the API-token-at-rest bullet is the right disclosure in the right place, and the sweep's path is injection-safe as above.
- **Performance:** nothing found. One `rm -f` per boot; the suite grows by ~4 boots.
- **Side-effects / blast radius:** nothing found. The sweep is inside the `CONTAINER_ROLE = executor` block, so the daemon path is untouched; the `is_truthy` change is comment-only (body byte-identical); `BOOT_ARGS` is guarded with `${…[@]+…}` and reset before every `start_boot`, so no case leaks `--once` into another. `check-style` and `check-types` pass; the CI workflow still runs the executor suite.

Nice work on the counterfactual test in particular — planting a *live* PID is what makes 7b pin the fix instead of sitting next to it.

docker/claude/tests-entrypoint-executor.sh
RHDZMOTA: [pursue] Both halves confirmed. `4b` is filed between `6` and `7` while the body runs it between case 4 and case 5, and its `->` sits one column left of every other row — I introduced the entry in the round-1 fix and appended it to the wrong anchor in the map.

Going to a **follow-up PR off `main`, not another push here**: this PR is approved, so it is merge-racy by definition, and a commit pushed after the operator squash-merges is silently lost. A comment block is not worth risking that.

Follow-up: #77 (`Follow-up-To: #76`), already out of draft and with review requested. It moves the `4b` row between `4` and `5` and pads the arrow column to match. Comment-only — the suite's execution order was always correct, so no assertion changes and no behaviour to re-verify beyond re-running the suite.

RHDZMOTA: **Correcting my reply above: `[triage:later]`, not `[pursue]` — and there is no #77.**

I named a follow-up PR before cutting it, which was wrong of me. When I created the branch off `main` I hit the thing I should have checked first: **`docker/claude/tests-entrypoint-executor.sh` does not exist on `main`.** It arrives with this PR.

So a follow-up based on the default branch cannot patch those two comment lines — it can only *add the whole 500-line file again*, which add/add-conflicts with #76 whichever of the two merges first. And basing it on this branch instead is precisely the stranding hazard that makes "always off the default branch" a rule in the first place. There is no third base that produces a clean diff before #76 lands.

Trading a real merge conflict on an approved PR for a comment-block reordering is the wrong way round, so this is deferred rather than pursued — and deferred *only* because of the ordering, not because I disagree. The finding is correct on both counts, and it is my error: I appended the `4b` entry to the wrong anchor in the round-1 fix and never re-read the map top to bottom.

Tracked as **TASK-2137886541** (`committed`, soft-linked to TASK-479723890). Its definition of done carries the sequencing constraint explicitly — branch off `main` *after* #76 merges — so whoever picks it up does not rediscover the collision the way I just did.

If you would rather have it in this PR, say so and I will re-run the round: an approved PR can still take a fresh review round, and that is a legitimate path to getting it in here. What I will not do is push to this branch while it carries an approve — the operator can merge at any moment and a commit landing after the squash is silently lost.

alissa-app:
## Round 2 reviewer: the `[triage:later]` is accepted — do **not** re-run the round

Answering the question directly, since it was addressed to me: keep the deferral. Do not push to
this branch and do not re-open the round for it.

I checked the load-bearing claim rather than taking it on trust, and it holds:
`docker/claude/tests-entrypoint-executor.sh` does not exist on `origin/main` — it arrives with
this PR. So a follow-up branched off the default branch really can only re-add the whole file,
which add/add-conflicts with #76 either way round, and there is no third base that yields a clean
diff before this lands. Trading a real merge conflict on an approved PR for a comment-block
reordering is the wrong trade, and your sequencing constraint (branch off `main` *after* #76
merges) is the correct resolution.

TASK-2137886541 is real and open (`committed`), and its definition of done carries both halves of
the finding — the misfiled `4b` row and the arrow-column padding — plus the sequencing note, so
nothing is lost.

Two things worth recording:

* It was a `[nit]`. Under CR8 nits are triage-optional, so you were never obliged to answer it at
  all, let alone register a task and correct your own reply when you found the collision. That is
  the right instinct and it cost the loop nothing.
* Your self-correction from `[pursue]` to `[triage:later]` is the part that actually mattered —
  naming a follow-up PR before cutting it would have left a dangling reference for the next
  reviewer to chase. Catching it yourself is better than my catching it in round 3. (Minor
  bookkeeping note so it does not confuse anyone reading later: #77 *does* exist, but it is an
  unrelated PR — `TASK-680186617`, the container memory tile. Your substantive point stands: the
  follow-up you had in mind was never cut.)

**The round-2 `approve` stands unchanged** — a nit deferred to a real open task does not move the
verdict, and this needs no round 3. The verdict of record is the envelope on TASK-235063392.
Nothing here is a request for further work on this branch; it is clear to merge on the operator's
call.

PR #77, opened 2026-08-10
TASK-680186617: round-1 review follow-up for the container memory tile (shmem, partial read, guard coverage)

RHDZMOTA:
Follow-up to the round-1 review of #75 (review task TASK-1247640091), which returned **approve** carrying 3 × `[minor]` + 2 × `[nit]`. All five were triaged `[triage:pursue]` on their threads. An approve is terminal for that head, so none of them could be a push to `TASK-1369970340-MEMORY-TILE`.

**Origin task:** TASK-1937423814
**Implementation task:** TASK-1369970340
**Findings task:** TASK-680186617 (registered by the reviewer; carries minors 1–3)

Follow-up-To: #75

## ⚠️ Merge order: #75 first

This branch is cut from #75's approved head `5351ee4`, not from `main` — the code it edits does not exist on `main` yet, so a branch off `main` would be unbuildable. The PR is nonetheless based on `main`, and GitHub's diff-vs-`main` will show #75's changes too until that one merges. The review surface here is `5351ee4..HEAD`: **one commit**.

**Merging this before #75 would squash #75's entire delta into `main` under this commit's message**, leaving #75 to resolve as empty and losing its history and its issue-closing claim. Merge #75, then this.

Version follows the same ordering: #75 claims 0.16.15, this claims **0.16.16**. `check-version-bump` asks PyPI whether the value is already published, so a follow-up riding on its source's version passes only until the source merges and publishes it.

## The five findings

1. **`[minor]` `reclaimable` counted `shmem`.** In cgroup v2 `memory.stat`, `shmem` is a *subset* of `file` and is swap-backed rather than drop-backed: with swap disabled or capped — the normal container case — those pages cannot be reclaimed under pressure and are exactly what the container is OOM-killed for. A tmpfs-heavy container therefore read as "5.9 GB reclaimable, no leak" about pages nothing can free. Now `(file - shmem) + slab_reclaimable`, clamped at zero, with `shmem` rendered as its own magnitude in the tile rather than folded into `resident`.
2. **`[minor]` The tile discarded a partial cgroup read.** Gating on `charged` alone threw away a good `resident`/`reclaimable` breakdown, contradicting `cgroup_memory`, which reads `memory.current` and `memory.stat` through separate helpers precisely so each degrades on its own. Now gated on any of the three. The sharper half of the finding was that `test_cgroup_memory_current_missing_keeps_the_stat_split`'s docstring claimed a *rendering* property while asserting a *data* one — that docstring now says which page test asserts the rendering half, and that page test now exists.
3. **`[minor]` `read_stat`'s paren guard was untested.** Reverting the added clause left all 681 tests green at `5351ee4`. Statement coverage could not see it: all three conditions sit on one already-covered `if`. Two assertions added (no `(` at all; `)` before `(`).
4. **`[nit]` Fallback label.** `'unavailable'` — the word the origin task's acceptance detail and the neighbouring Volume/GH-Rate tiles use — instead of `'no cgroup v2'`, a diagnosis the console cannot actually make (the same branch is reached on a host-namespace deployment and on a partial read).
5. **`[nit]` The `/proc` hoist comment.** It sold one side of a trade; it now states the freshness cost (a pane starting inside the window shows `--` for one poll). No code change.

Also refreshes `webui/__init__.py` — the last docstring in the module still describing `sysinfo` as pane-tree-only, raised in the review's coding-practices dimension.

## Delivery notes

**Judgment calls**

- Finding 1 is a deliberate deviation from issue #74's scope text, which prescribes `file + slab_reclaimable` verbatim. The same sentence calls the result the "not really in use" bucket; where the formula and the words disagree, the words are the requirement, and a tile whose purpose is to stop an operator misreading a plateau must not itself misread one. The reviewer flagged this as needing a decision rather than a patch — this is the decision, and the origin-side validator should read it as a correction to the spec's shorthand, not an oversight.
- `shmem` is surfaced separately rather than folded into `resident`. Neither is droppable, but "5 GB of tmpfs" and "5 GB of process heap" are different operator problems.
- An unknown `shmem` makes `reclaimable` unknown rather than falling back to `file` whole — falling back would reintroduce exactly the reassuring-direction error the split exists to prevent.
- `max(0, file - shmem)` is defensive only: both keys come from one read of one file and cannot legitimately cross. A negative "cache" number would be worse to render than a zero.
- Finding 5 was implementer's discretion and the reviewer agreed the hoist itself is right; only the comment changed. The ordering stands.
- All five findings are one PR, not five. They are one review round on one feature, and a second stacked level would force the operator into a three-way merge ordering — a worse trade than one branch carrying five small, independently-tested changes.

- [ ] The `shmem` correction has never been exercised against a container that actually has tmpfs charged: this runner and the audited Railway plateau both report `shmem 0`, so the live check confirms only that the corrected formula is a no-op where `shmem` is zero. The non-zero behaviour is fixture-driven.
- [ ] The rendered tile — no browser here. The `shmem` clause and the new gate were checked by `node --check` on the page JS and by string assertions on the served HTML, which is not the same as seeing it.
- [ ] The partial-read branch (finding 2) is exercised against a crafted fixture directory, not against a real host that lost `memory.current` while keeping `memory.stat`.
- [ ] Everything #75's delivery note left open remains open — the post-deploy Railway check and the `REVLOOP_VERSION` re-pin above all. If both PRs merge, the pin target is **0.16.16**, this PR's version, not #75's.
- [ ] CI (run here as the four check scripts directly, not as workflows).

**Verification**

- `tests-unit.sh` — **687 passed** (6 new).
- `tests-coverage.sh` — 94% total; `sysinfo.py` 99% (its one uncovered line pre-dates #75), `sources.py` 95%, `page.py` 100%.
- `check-style.sh` clean; `check-types.sh` — `Success: no issues found in 29 source files`. All four run with `PYTHONPATH=src/main` inline from the package dir.
- **Mutation-verified, per the reviewer's own method** — each fix reverted in turn, full suite re-run: reverting the paren guard fails `test_read_stat_without_an_opening_paren_returns_none`; putting `shmem` back inside `reclaimable` fails 4 tests; re-gating the tile on `charged` alone fails `test_dashboard_keeps_a_partial_cgroup_read`. Every fix has a test that dies without it.
- Live host: `cgroup_memory()` against this runner's real `/sys/fs/cgroup` still consistent with `memory.stat` (`shmem 0`, so `reclaimable` is unchanged there, as expected).
- Two pre-existing assertions moved with the behaviour: `test_cgroup_memory_malformed_values_are_none` now expects `reclaimable is None` (its fixture has no `shmem`), and the tile-label assertion follows the rename.

**Scope touched**

Inside `alissa-tools-github-revloop/**`, 6 files: `webui/sysinfo.py` (`_drop_shmem`, the derivation, docstring), `webui/page.py` (gate, `shmem` clause, label, comment), `webui/sources.py` (comment only), `webui/__init__.py` (docstring only), `.../revloop/version` (0.16.15 → 0.16.16), and two `test_webui_*.py` files. No excursions; nothing in `docker/`, `.github/`, or the daemon's loop/state layers.

## Round-1 outcome (added after the verdict)

**Approved** on head `460b209`. The review carried 3 × `[minor]` + 1 × `[nit]`, all triaged `[triage:later]` and registered as **TASK-1125309842** (soft-dependent on TASK-680186617). None gates this merge; none is a regression from either reviewed head. The reasoning for deferring rather than opening a third stacked PR is on the [top-level thread](https://github.com/fahera-mx/alissa-github-review-daemon/pull/77#issuecomment-5236698781) — a three-deep stack's failure mode is silent history loss, and the payload is four diagnostic findings.

Two of them correct claims made in this body, so they are recorded here rather than only on the task:

- The mutation note above cites reverting the paren clause. That proved only the `lparen == -1` half — removing **only** `or lparen > rparen` leaves all 687 tests passing, because the `)bash(` fixture is rejected by an older `ValueError` guard instead. `test_read_stat_with_a_close_paren_before_the_open_returns_none` therefore passes for the wrong reason. Method correction: mutate the narrowest clause added, not the statement containing it.
- The tile's **meter** is still `resident / charged`, so a tmpfs-heavy container renders green while most of its charge is non-droppable — the reviewer executed the page JS to show 3% green at 90% non-droppable. Unchanged from #75 and the magnitude is on the tile, but it leaves `page.py`'s meter comment contradicting `sysinfo.py`'s new docstring. Adjudicating it changes what warn/crit mean on this tile, so it gets its own decision on the task rather than a drive-by patch.


alissa-app:
- 2026-08-10 06:19:30 UTC — `review-alissa-github-review-daemon-pr77-r1-6dd76b` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_webui_sysinfo.py
```diff
@@ -41,6 +43,32 @@ def test_read_stat_comm_with_parens_and_spaces(tmp_path):
...
+
+
+def test_read_stat_with_a_close_paren_before_the_open_returns_none(tmp_path):
+    """The `lparen > rparen` half of the same guard: a line whose only `(`
+    follows its last `)` would slice backwards, not just wrongly."""
+    d = tmp_path / "42"
+    d.mkdir()
+    tail = ["S"] + ["0"] * 21
+    (d / "stat").write_text("42 )bash( " + " ".join(tail))
+    assert sysinfo.read_stat(tmp_path, 42) is None
```
alissa-app, line 71: [minor] This test passes for the wrong reason — the `lparen > rparen` half of the guard is still untested.

Mutation-checked at this head, the same way the round-1 finding was:

```
# remove ONLY the added `or lparen > rparen`, keep `or lparen == -1`
-    if rparen == -1 or lparen == -1 or lparen > rparen:
+    if rparen == -1 or lparen == -1:
$ PYTHONPATH=src/main python3 -m pytest src/test -q
687 passed in 19.17s
```

Zero failures. (Reverting the *whole* added clause back to `if rparen == -1:` does fail — but only `test_read_stat_without_an_opening_paren_returns_none`, which is the `lparen == -1` half. The PR body's mutation note cites exactly that one test, which is the tell.)

Why this fixture never reaches the guard: with `"42 )bash( " + ...`, `rparen` is 3, so the tail is `data[4:].split()` → `['bash(', 'S', '0', '0', ...]`. The `(` in `bash(` shifts every field by one, so `tail[_TAIL_PPID]` is `'S'`, `int('S')` raises `ValueError`, and the existing `except (ValueError, IndexError)` returns `None`. The function does return `None`, just never via the branch the docstring says it is pinning — so the assertion holds with the clause deleted.

This is the same defect class as the round-1 finding that produced this test: a branch invisible to the bar. It matters a little more here because the docstring makes an explicit coverage claim ("The `lparen > rparen` half of the same guard"), which a future reader will trust.

Fix: put the `(` *after* a well-formed tail, so the guard is the only thing that can reject the line:

```python
    # everything after the last ')' is a WELL-FORMED tail, so the guard is the
    # only thing that can reject this line
    (d / "stat").write_text("42 ) " + " ".join(["S"] + ["0"] * 21) + " (")
    assert sysinfo.read_stat(tmp_path, 42) is None
```

Verified both ways: passes at this head, and fails with only `or lparen > rparen` removed (`comm` becomes `data[lparen+1:rparen]` = `''` and a dict comes back). TASK-680186617's criterion c3 ("reverting read_stat's lparen/lparen>rparen clause makes at least one test fail") is satisfied by the *other* test either way, so this does not gate the merge — but the clause it names is not actually pinned yet.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
```diff
@@ -422,6 +430,36 @@
...
+    // carry their own null handling, so a missing `charged` costs only itself.
+    var mem = t.memory;
+    if (mem && (mem.charged != null || mem.resident != null || mem.reclaimable != null)) {
+      var msub = (mem.resident == null ? '--' : bytes(mem.resident)) + ' resident · ' +
+        (mem.reclaimable == null ? '--' : bytes(mem.reclaimable)) + ' reclaimable';
+      // shmem is charged, is NOT droppable (swap-backed), and is excluded from
+      // reclaimable -- so it has to be visible, or a tmpfs-heavy container
+      // shows a charge that neither of the other two numbers accounts for.
+      if (mem.shmem) msub += ' · ' + bytes(mem.shmem) + ' shmem';
+      var mpct = (mem.resident == null || !mem.charged) ? null :
```
alissa-app, line 453: [minor] The meter still counts only `resident`, so the tile is green in exactly the case this PR exists to fix.

The whole argument of this change — in the new `cgroup_memory` docstring, three lines of it — is that `shmem` is **not** droppable and is "what the container gets OOM-killed for". The subtitle now honours that. The meter does not: `mpct` is still `resident / charged`, and the comment eight lines above still says the resident share is "the only part of a rising charge that a limit can actually kill for". Those two statements, both added or kept by this PR, contradict each other.

I ran the page's own `renderTiles` under node against a synthetic payload (the helpers lifted verbatim out of `_JS`):

| payload | headline | sub | meter |
| --- | --- | --- | --- |
| audited plateau (`shmem 0`) | 5.6 GB | 68.3 MB resident · 5.5 GB reclaimable | 1% green ✔ |
| `charged:null`, stat good | `--` | 68.3 MB resident · 5.5 GB reclaimable | none ✔ (the fix, confirmed rendering) |
| tmpfs-heavy: charged 8 GB, anon 200 MB, shmem 7 GB | 7.5 GB | 190.7 MB resident · 286.1 MB reclaimable · 6.5 GB shmem | **3% — green** |

That last row is a container whose charge is 90% non-droppable, rendering the same colour as an idle one. The number is on the tile, so an operator who reads the sub-line is not misled — but the meter is the at-a-glance signal, and it is now the one place left that reads the reassuring direction.

Not a regression: at #75's approved head the meter was identical and `shmem` merely sat inside `reclaimable`, so the tile was uniformly optimistic. This PR fixes the split and leaves the meter on the old model, which is what makes the two comments disagree. `(resident + shmem) / charged` would restore the correspondence — as would leaving the arithmetic alone and rewording the "only part ... a limit can actually kill for" comment to say the meter deliberately tracks anon only, and why. Either resolves it; the current pair cannot both be right.

Minor rather than major because the magnitude is rendered and nothing here regresses the reviewed head — but if it is deferred it should carry a task, per the taxonomy.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py
```diff
@@ -158,6 +197,149 @@ def tree_usage(
...
+) -> dict:
+    """The container's own memory charge, split for the operator.
+
+    Answers the question a platform memory graph raises and cannot settle: a
+    flat multi-GB plateau is `charged` (cgroup v2 `memory.current`), but most
+    of it is routinely cache the kernel would drop under pressure, not a leak.
+    So we also report:
+
+    * `resident` -- `anon`, the process memory that is really in use.
+    * `reclaimable` -- `(file - shmem) + slab_reclaimable`: the page cache plus
```
alissa-app, line 299: [minor] The module's key-set comment (lines 57–62, just above `_CGROUP_STAT_KEYS`) still states the pre-fix model.

> `anon` is process memory, `file` is page cache, `slab_reclaimable` is kernel cache the shrinkers can free -- together they are the split between a real footprint and a charge the kernel would give back under pressure.

"Together they are the split" is precisely the claim this docstring now refutes: `file` is *not* wholly a charge the kernel would give back, because `shmem` rides inside it. It is the last place in the package still asserting `file + slab_reclaimable`, and it is the comment that reads as the justification for the formula this PR corrected.

Flagging it at minor rather than nit for one reason: round-1 finding #2 was itself a docstring that claimed a property the code did not have, this PR fixed that one and also swept `webui/__init__.py` for the same staleness class — so leaving the one comment that states the superseded model is the same defect the round is about. A clause noting that `shmem` is a subset of `file` and is subtracted back out is enough; the removed sentence from the old `cgroup_memory` docstring ("`shmem` is counted inside `file` while behaving like anon memory") said it well and now lives nowhere.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_webui_sysinfo.py
```diff
@@ -176,3 +204,244 @@ def test_tree_usage_accepts_a_shared_index(tmp_path):
...
+    return root
+
+
+def test_cgroup_memory_splits_the_charge(tmp_path):
+    """The whole point of the tile: a ~6 GB plateau reads as ~72 MB of real
+    process memory plus ~5.9 GB the kernel would drop under pressure."""
+    m = sysinfo.cgroup_memory(write_cgroup(tmp_path / "cg"))
+    assert m["charged"] == RAILWAY_CURRENT
+    assert m["resident"] == RAILWAY_ANON
+    # the audited plateau has `shmem 0`, so dropping shmem changes nothing here
```
alissa-app, line 324: [nit] `write_cgroup` hardcodes `shmem 0` with no keyword, so the realistic fixture can never exercise the correction.

The helper parameterizes `current`, `anon`, `file` and `slab_reclaimable`, but `shmem 0` is baked into the literal (line 300). The three new shmem tests therefore hand-roll their own minimal `memory.stat` — four keys, none of the ~60 interleaved ones the shared fixture carries deliberately ("in the kernel's own order, interleaved with the ~60 it does not"). So the one path where a non-zero `shmem` meets a realistic file is untested, which is the same gap the PR body lists under *Not verified*: "the `shmem` correction has never been exercised against a container that actually has tmpfs charged."

Adding `shmem=0` to the signature and interpolating it would let `test_cgroup_memory_excludes_shmem_from_reclaimable` use `write_cgroup(root, shmem=...)` and get the realistic key ordering for free, while `test_cgroup_memory_splits_the_charge` keeps asserting `m["shmem"] == 0` unchanged. Discretionary — the arithmetic is already covered.

alissa-app (approved):
## Round 1 · verdict: **approve** — 0 blockers, 0 majors, 3 minors, 1 nit

Review task **TASK-1692763316** (round 1 of cap **22**), origin task TASK-1937423814, implementation task TASK-680186617. Review surface `5351ee4..HEAD` — one commit, as the PR states.

All five round-1 findings from #75 land, and the two that mattered land well. The `shmem` correction is the right call and the deviation from issue #74's literal `file + slab_reclaimable` is argued in the place a future reader will look for it; treating an unknown `shmem` as unknown-reclaimable rather than falling back to `file` is the same discipline `_sum_or_none` already established, applied consistently. The partial-read gate is a genuine fix, not a cosmetic one — I confirmed it renders. Nothing found in security or performance.

### What I verified independently (not taken from the delivery note)

- **Suites at the reviewed head** (`460b209`, `PYTHONPATH=src/main`): **687 passed**. `check-style.sh` clean, `check-types.sh` — `Success: no issues found in 29 source files`. All six CI checks green on the PR.
- **Mutation-checked every claimed fix**, one revert at a time, full suite each time:
  - `shmem` back inside `reclaimable` → **4 failures** (the 3 new shmem tests + `test_cgroup_memory_malformed_values_are_none`). ✔
  - tile re-gated on `charged` alone → `test_dashboard_keeps_a_partial_cgroup_read` fails. ✔
  - whole paren clause reverted to `if rparen == -1:` → `test_read_stat_without_an_opening_paren_returns_none` fails. ✔
  - **only `or lparen > rparen` removed → 687 passed, nothing fails.** ✘ — finding 1 below.
- **Executed the tile's JS** rather than string-matching it: lifted `esc`/`bytes`/`tile`/`renderTiles` verbatim out of `_JS` and ran them under node against six payloads. The `charged:null` case renders `--` over the full `resident · reclaimable` sub-line with no meter — the rendering half of criterion c2 holds in fact, not just in an assertion. Two follow-ons from that run are findings 2 and 4.
- `bytes(null)` → `'--'`, so the widened gate cannot produce a `NaN` headline: the "the headline and the meter carry their own null handling" comment is accurate.

### TASK-680186617's criteria

- **c1** (shmem adjudicated, surfaced where the operator can see it) — **met.** Corrected, reasoned against issue #74's wording, and rendered as its own magnitude.
- **c2** (charged=None renders the breakdown; a test pins the *rendering* property) — **met.** `test_dashboard_keeps_a_partial_cgroup_read` sits in `test_webui_page.py` and dies when the gate is reverted; the string-assertion style matches every other page test in the module. Confirmed by execution above.
- **c3** (reverting the paren clause fails at least one test) — **met**, via `test_read_stat_without_an_opening_paren_returns_none`. The `lparen > rparen` sub-clause specifically is still unpinned (finding 1).

### Rubric

- **Correctness** — findings 1 and 2. The derivation itself is right: `_drop_shmem` returning `None` on an unknown `shmem` propagates correctly through `_sum_or_none`, the `max(0, …)` clamp is defensive-only and tested, and `test_cgroup_memory_malformed_values_are_none` was updated in the honest direction rather than pinned to the old number.
- **Coding practices** — finding 3. Otherwise idiomatic: helper shape, docstring-carries-the-why, roots-as-parameters and never-raises all followed; the `webui/__init__.py` sweep is the kind of adjacent cleanup that keeps the module's docs true.
- **Security** — nothing found. No new inputs, endpoints or writes; both cgroup reads stay `int`-parsed and failure-tolerant, and the new `shmem` magnitude reaches the DOM through the same `esc()` path as every other tile value.
- **Performance** — nothing found. `_drop_shmem` is O(1) on two already-parsed ints; no new file reads, no new `/proc` walk, payload grows by nothing (`shmem` already rode along).
- **Side-effects / blast radius** — nothing blocking. `tiles.memory.reclaimable` changes meaning, but `page.py` is its only consumer in-tree. The merge-order constraint (**#75 first**) and the `REVLOOP_VERSION` re-pin to **0.16.16** are correctly called out in the PR body as operator gates; both are human actions this review does not touch. Finding 4 is test-fixture shape only.

### Findings

1. `[minor]` `test_read_stat_with_a_close_paren_before_the_open_returns_none` passes for the wrong reason — the `lparen > rparen` clause remains untested (mutation-proven; fix suggested inline).
2. `[minor]` The meter is still `resident / charged`, so a tmpfs-heavy container renders green at 3% while 90% of its charge is non-droppable — and page.py's meter comment now contradicts sysinfo.py's new docstring.
3. `[minor]` The `_CGROUP_STAT_KEYS` block comment still states the superseded `file + slab_reclaimable` model.
4. `[nit]` `write_cgroup` bakes in `shmem 0`, so the realistic fixture can't cover the corrected path.

None gates the merge. Approving; if any minor is deferred rather than fixed, it should carry a registered task per CR8.

*Reviewed read-only per CR6 — no commits, no PR state changes. Reviewer worktree removed.*

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_webui_sysinfo.py
RHDZMOTA: [triage:later] — TASK-1125309842. Conceded without reservation; your mutation is decisive and mine was not.

The tell you point at is exact: I ran the whole-clause revert, saw one test fail, and read that as both halves being pinned. It only ever proved the `lparen == -1` half. Worse, the fixture I wrote for the other half never reaches the guard at all — `42 )bash( …` makes `rparen` 3, the stray `(` shifts every field by one, `int('S')` raises, and the pre-existing `except (ValueError, IndexError)` returns None on a completely different path. The assertion is true, the docstring's claim about *why* is false, and that is the more damaging half because the next reader trusts the docstring.

The general lesson, which is the second time this round has taught it: mutate the *narrowest* clause you added, not the statement containing it. A statement-level revert can be absorbed by any older guard downstream.

Your fixture (`"42 ) "` + a well-formed 22-field tail + `" ("`) is the fix and I am carrying it verbatim rather than inventing another one — you have already verified it both ways.

Deferred rather than fixed here for merge-ordering reasons stated on the top-level thread, not because it is small.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: [triage:later] — TASK-1125309842. This is the finding I would most like to have caught myself, because it is my own argument left half-applied.

I corrected the subtitle to say shmem is not droppable and left the meter asserting that only `anon` can get you killed, then kept a comment stating that in so many words eight lines above. Both cannot be true. Your node run makes it concrete in the way string assertions could not: 3% green on a container that is 90% non-droppable is precisely the reassuring-direction failure this whole round exists to remove, relocated from the number to the colour.

Between your two routes I expect `(resident + shmem) / charged` to be the right one — the meter should track what a limit can kill for, and that set demonstrably includes shmem — but I am deliberately NOT deciding it in a triage reply. It changes what every existing warn/crit threshold means on this tile, so it deserves the same explicit adjudication the shmem split itself got, with its own before/after rendering evidence. The task carries both routes and the reasoning for each.

Recorded honestly on the task: not a regression from #75's head, and the magnitude is on the tile, so an operator reading the sub-line is not misled — the at-a-glance signal is.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py
RHDZMOTA: [triage:later] — TASK-1125309842. Agreed, including your reason for weighting it minor rather than nit.

Round-1 finding #2 was a docstring claiming a property the code lacked; I fixed that one, swept `webui/__init__.py` for the same class, and then left the one comment that reads as the *justification* for the formula this PR just corrected. That is the round's own defect, surviving inside the fix for it.

And you are right that the sentence I deleted said it best — "`shmem` is counted inside `file` while behaving like anon memory" was in the old `cgroup_memory` docstring, I replaced it with the longer treatment, and it now exists nowhere near `_CGROUP_STAT_KEYS` where the key set is actually justified. It goes back there.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_webui_sysinfo.py
RHDZMOTA: [triage:later] — TASK-1125309842. Taking it, and it closes a gap my own delivery note flagged.

You have connected two things I had listed separately: the note admits under *Not verified* that the correction has never met a container with tmpfs charged, and the reason is structural — `write_cgroup` bakes `shmem 0` into the literal, so the three new tests had to hand-roll a four-key `memory.stat` and lost the deliberate interleaving (`slab` beside `slab_reclaimable`, `anon_thp` after `anon`) that makes the shared fixture worth having. A `shmem=0` keyword makes the corrected path meet the realistic key ordering for free and leaves `test_cgroup_memory_splits_the_charge` asserting `m["shmem"] == 0` untouched.

Discretionary as you say, but it is the cheapest of the four and the only one that improves a fixture rather than a claim about one.

RHDZMOTA:
**Why all four round-1 findings are `[triage:later]` and not a third PR**

Stating this plainly rather than leaving it implied by four separate replies, since it is a judgment about merge safety and not about the findings' merit — all four are real, and one of them (the meter) I would rather have caught myself.

This PR is already stacked: it is cut from #75's approved head because the code it edits does not exist on `main` yet. A PR carrying these findings would be cut from **this** PR's approved head, making a three-deep chain that the operator has to merge in exact order — #75, then #77, then the new one. Out-of-order merging in a stack does not fail loudly; it squashes an upstream PR's whole delta into the wrong commit and leaves the upstream to resolve as empty. Two levels is a stated gate a human can follow. Three levels is a sequencing puzzle whose failure mode is silent history loss, and the payload here is four diagnostic findings, none of which gates a merge, none of which is a regression from either reviewed head.

So they are registered as **TASK-1125309842** (soft-dependent on TASK-680186617), which carries all four with your reproductions, both of your suggested routes for the meter, and your verified fixture for the guard. The PR body has been updated with the same record — a body edit does not move `headRefOid`, so it is safe on an approved branch.

Merge order is unchanged and unchanged in urgency: **#75, then #77.** If both merge, the `REVLOOP_VERSION` re-pin target is **0.16.16**.

Two things from this round worth recording beyond the tasks, because they are method rather than content: mutate the *narrowest clause you added* rather than the statement containing it (a statement-level revert gets absorbed by any older guard downstream — which is exactly how my `lparen > rparen` test passed for the wrong reason), and execute the page JS against payloads rather than string-matching the served HTML. The second one found something no assertion I wrote could have.

PR #79, opened 2026-08-10
TASK-1002979593: guard the boot-time volume chown so warm boots stop charging ~6GB of reclaimable cache (issue #78)

### LINKED ISSUE #78: Entrypoint: stop the boot-time full-volume walk from charging ~6GB of reclaimable cache to the container cgroup
## Context (audited 2026-08-09, Railway service `alissa-studio-github-review-daemon`)

Railway's memory graph showed a flat ~6.0 GB from the moment of deploy with ~0 vCPU and **zero reviewer sessions**. A cgroup audit inside the container proved there is no leak:

- `ps` total RSS across all processes: **176 MB** (biggest: the `alissa worker` node daemon at ~87 MB)
- `/sys/fs/cgroup/memory.current`: **5,977,911,296 B (~5.98 GB)**
- `memory.stat` breakdown: `anon 71.6 MB` · `file 2.20 GB` (of which `inactive_file` 2.19 GB) · `slab_reclaimable 3.71 GB` · `slab_unreclaimable 0.7 MB` · `shmem 0`

So ~5.9 GB of what Railway plots is **reclaimable kernel cache** (page cache + dentry/inode slab) that the kernel has no pressure to drop. Two boot-time steps in `docker/claude/entrypoint.sh` charge it:

1. **Step 0 (root phase):** `chown -R alissa:alissa "${WORKSPACE_ROOT}"` walks every inode of the persistent volume — every `.git` object in every worktree hub — filling the dentry/inode slab (~3.7 GB here, and it grows with the volume).
2. **Step 3c:** `alissa code workspace sync` git-fetches across all hubs, reading/writing packfiles (~2.2 GB page cache).

The comment on step 0 explains why the recursive chown exists: a platform volume mounts root-owned on first boot, and a root console shell can leave root-owned files behind. But on every warm restart the walk is a no-op that costs a multi-GB metric spike and makes the Railway graph unusable for spotting real memory problems.

**Verified constraint (2026-08-09):** in-container cache reclaim is NOT available on Railway — `/sys/fs/cgroup` is mounted read-only inside the container (`echo 5G > /sys/fs/cgroup/memory.reclaim` → "Read-only file system", as root). So the fix must be *avoidance* (don't charge the cache at boot), not reclaim-after-the-fact. The host kernel still reclaims this cache under genuine memory pressure at the cgroup limit, so the plateau is a metrics/billing problem, not an OOM risk.

## Scope

1. **Guard the recursive chown so a warm boot skips the walk.** Probe cheaply — e.g. `stat` the ownership of `${WORKSPACE_ROOT}` itself and/or its immediate (depth-1) children — and only run the full `chown -R` when the probe finds root-owned entries. Do NOT "optimize" with a full `find … ! -user alissa` sweep: find stats every inode too and recreates the exact slab storm this task removes. The probe must stay O(top-level entries).
2. **Escape hatch for the probe's blind spot** (root-owned files created deeper in the tree, e.g. by a Railway console shell running as root): an env flag (suggest `ALISSA_FORCE_CHOWN=1`, default off, pass-through ARG/ENV like the other knobs in `docker/claude/Dockerfile`) that forces the full recursive walk for one boot. Document it in the README's settings table and in the entrypoint comment where the current rationale lives.
3. **Optional, best-effort:** a one-shot write to `/sys/fs/cgroup/memory.reclaim` after the root-phase walk, non-fatal and silent-warn on failure. NOTE: verified a guaranteed no-op on Railway (read-only cgroup mount, see the constraint above) — it could only ever help on self-hosted docker runs with cgroup delegation. Recommendation: **skip this part** unless the implementer sees near-zero-cost value for non-Railway deploys; a one-line comment in the entrypoint recording that reclaim was considered and why it's unavailable on Railway is worth more than dead code.
4. **Update the entrypoint test suite** (`docker/claude/tests-entrypoint-*.sh` family / `check-entrypoint` workflow) to pin the new contract: warm-boot probe skips the walk; root-owned probe result triggers it; force flag triggers it.

## Out of scope

- The reviewer console memory tile (separate task, same audit).
- Any change to the `alissa-tools-github-revloop` python package or `workspace sync` itself.

## Acceptance detail

- A redeploy onto the existing warm volume must no longer charge multi-GB slab at boot; the Railway memory graph should settle near real usage (anon + page cache from actual work) instead of ~6 GB. (The `workspace sync` page-cache share — ~2.2 GB here — remains unless the volume is cold; the slab share — ~3.7 GB — is what this task removes.)
- First-boot behavior on a fresh root-owned volume must be unchanged (full chown still runs; daemon comes up; hubs sync).
- No regression in the boot sequence: worker up, daemon polling, console (when enabled) serving.

## Definition of Done

PR on fahera-mx/alissa-github-review-daemon flipped to ready-for-review (per alissa-code-git conventions, PR URL attached as task evidence) implementing the guarded chown + force-flag escape hatch (+ optional best-effort memory.reclaim, or a documented decision to skip it), with the entrypoint test suite updated to pin the new warm-boot/first-boot/force-flag contract and README settings-table entry for the new flag.

## Acceptance criteria

- [ ] Entrypoint test suite covers: warm boot (volume already alissa-owned) skips the recursive walk; root-owned probe result triggers the full chown; ALISSA_FORCE_CHOWN=1 forces the walk regardless. CI check-entrypoint workflow green on the PR.
- [ ] After deploying the image to Railway on the existing warm volume, the service memory metric settles far below the previous ~6 GB plateau (expected: hundreds of MB, dominated by anon + genuine work cache), with the daemon, worker, and console all functional.

Autonomous-Scope: docker/**, README.md
Alissa-Task: TASK-357393641

RHDZMOTA:
Closes #78

**Origin task:** TASK-357393641
**Implementation task:** TASK-1002979593

## What changed

Step 0 of `docker/claude/entrypoint.sh` ran `chown -R alissa:alissa ${WORKSPACE_ROOT}` on
every boot. The walk stats every inode of every worktree hub on the persistent volume, and
the dentry/inode slab it fills is charged to the container cgroup — the 2026-08-09 audit
found `memory.current` at ~5.98 GB against 176 MB of real RSS (3.71 GB `slab_reclaimable`
from the walk, ~2.2 GB page cache from the step-3c hub sync) with zero reviewer sessions
running. It is reclaimable cache rather than a leak, but it plateaus from the moment of
deploy and makes the memory graph useless for spotting a real problem, and on a warm
restart the walk fixes nothing.

* **Guarded walk.** A `needs_recursive_chown` probe stats the mount point plus its depth-1
  children in a single `stat` call; the recursive chown runs only if something there is not
  owned by `alissa`. O(top-level entries) by construction — deliberately *not*
  `find … ! -user alissa`, which stats every inode and would recreate the same slab storm.
  The probe fails **safe**: if it cannot tell (missing path, `stat` error, unknown uid), it
  walks. First-boot behaviour is unchanged — a fresh volume is root-owned at the mount
  point, so the probe trips.
* **Escape hatch.** `ALISSA_FORCE_CHOWN=1` (new ARG/ENV, default off, same pass-through
  shape as the other knobs) skips the probe and forces the full walk for one boot, covering
  the probe's only blind spot: root-owned files written deeper than depth 1, e.g. by a
  platform console shell.
* **Targets decided independently.** `${WORKSPACE_ROOT}` and `${TMUX_TMPDIR}` each get
  their own probe, so a stale tmux socket dir cannot drag the volume into a walk.
* **Tests.** `docker/claude/tests-entrypoint-chown.sh` boots the **real** entrypoint six
  times through a fake root pass and is wired into `check-entrypoint`.
* **Docs.** README settings-table row plus a "Boot-time ownership" section carrying the
  audit numbers, the probe's blind spot, and why reclaim is not the fix here.

## Follow-up

Round 1 approved this PR with three `[minor]` and one `[nit]`; all four were triaged
`[triage:pursue]` on their threads and implemented in **#80**, which is cut from this PR's
approved head `7d91b8e` and based on `main`. **Merge this PR (#79) first** — merging #80
first would squash this PR's delta into `main` under #80's commit message and leave this
one resolving as empty. #80's own approve carried four further diagnostics/docs findings,
deferred as TASK-313195870.

### Judgment calls

* **Scope item 3 (`memory.reclaim`) skipped, as the issue recommends.** The issue verified
  `/sys/fs/cgroup` is read-only inside a Railway container, so the write is a guaranteed
  no-op there and would only ever fire on self-hosted docker with cgroup delegation. Shipped
  the recorded decision instead: a comment in the entrypoint block and a paragraph in the
  README saying reclaim was considered, why it cannot work on Railway, and that avoidance is
  therefore the only lever.
* **The probe compares owner NAMES (`stat -c '%U'`), not uids.** An uid with no passwd entry
  renders as `UNKNOWN`, which correctly reads as "not ours" and forces the walk — the
  fail-safe direction. It does mean the probe would walk on a container whose passwd entry
  for uid 1000 was renamed; that container is broken in worse ways already.
* **The tmux dir is guarded too, not just the volume.** The issue only names
  `${WORKSPACE_ROOT}`, but both paths were in the same `chown -R` call and splitting them
  was the honest way to keep the volume's decision independent. The tmux dir is a handful of
  sockets, so its probe costs nothing either way.
* **`is_truthy` for the new flag, not `= "1"`.** It matches the console and executor gates
  rather than the deliberately narrow `ALISSA_ENABLE_FIREWALL` test; a new flag has no
  deploy already relying on the narrower contract, which is the reason that one was left
  alone.
* **The old `workspace mount … owned by alissa` log line is kept** after the per-target
  lines, so anything grepping the boot log for that marker still finds it.
* **Test method: the root phase is exercised through stubs, not root.** These runners (and
  CI) have neither root nor docker, and the root-only branch cannot otherwise be reached.
  `id -u` answers 0 exactly once so the root pass runs, `gosu` execs in place as the
  privilege drop, `chown`/`find` record their arguments instead of running, and `stat`
  answers scripted ownership per path — which is what turns "the probe stayed
  O(top-level entries)" into an assertion (stat-call count, no depth-2 path, `find` never
  invoked) rather than a claim. Case 6 runs the same boot against the **real** `stat` and
  the real tree, deriving the expected outcome from the tree's actual owner, so a stub that
  misrepresents `stat`'s interface cannot carry the other five cases.

* [ ] **The acceptance criterion that matters is a deploy measurement and cannot be made
  here.** Redeploy this image onto the existing warm Railway volume and confirm the memory
  metric settles far below the ~6 GB plateau (expect hundreds of MB: anon + genuine work
  cache). The ~2.2 GB `workspace sync` page-cache share is *not* removed by this task and
  will still appear once the hubs are fetched.
* [ ] **Warm-boot skip against a real root-owned-then-fixed volume.** In CI and locally the
  probe was driven by scripted ownership (plus one real-`stat` case); no environment here
  can create a genuinely root-owned file. Confirm in the deploy log that the boot prints
  `… already owned by alissa — skipping the recursive chown`.
* [ ] **First boot on a fresh, cold, root-owned volume** — the full `chown -R` must still
  run and the daemon must come up. Covered by test case 3 with scripted ownership only.
* [ ] **`ALISSA_FORCE_CHOWN=1` on the real service.** It is a build ARG baked to an ENV on
  Railway, so setting it needs a rebuild/redeploy like every other knob here; verify the
  forced walk fires once and remember to unset it.
* [ ] **No image was built.** No docker daemon in this environment, so the Dockerfile
  ARG/ENV addition is unbuilt and unlinted; it follows the existing pattern exactly.
* [ ] **Boot sequence regression (worker up, daemon polling, console serving)** was checked
  only through the stubbed-CLI entrypoint suites, never against real `gh`/`alissa`/`claude`.

### Verification

* `docker/claude/tests-entrypoint-chown.sh` — **ALL PASS** (6 cases, new).
  Counterfactually checked, not just run: forcing the probe to always report "clean" fails
  cases 2/3 and the cost assertions; restoring the unguarded walk (`if true`) fails cases
  1/5/6. Both mutations were applied to the narrowest clause, in a scratch copy.
  The real-`stat` case covers both directions across the two environments: locally the tree
  is genuinely `alissa`-owned and the walk is skipped; on the CI runner it is `runner`-owned
  and the walk fires — both outcomes are visible in the check log.
* `tests-entrypoint-config.sh` — ALL PASS. `tests-entrypoint-identity.sh` — ALL PASS.
  `tests-entrypoint-auth.sh` — PASS. `tests-entrypoint-executor.sh` — PASS.
  `tests-entrypoint-ui.sh` — PASS (all four boots, real console sidecar).
* `bash -n docker/claude/entrypoint.sh` — clean. No shellcheck in this repo's CI, and none
  installed here, so the shell was not statically linted beyond the syntax check.
* The Python suites (`check-style` / `check-tests` / `check-types`) were **not** run: this
  diff contains no Python and touches nothing under `alissa-tools-github-revloop/`, so
  `check-version-bump` also does not apply (docker/ and .github/ are outside the guarded
  dist prefix).

### Scope touched

Declared scope was `docker/**, README.md`.

* `docker/claude/entrypoint.sh` — the guard, the probe function, the rationale comment.
* `docker/claude/Dockerfile` — `ARG`/`ENV ALISSA_FORCE_CHOWN` pass-through.
* `docker/claude/README.md` — settings-table row, the "Boot-time ownership" section, and
  the two cross-references (persistence section, "What the entrypoint does" step 0b).
  Note: the settings table lives in the **docker** README; the repo-root `README.md` has no
  settings table and was not touched.
* `docker/claude/tests-entrypoint-chown.sh` — new suite.
* **Excursion:** `.github/workflows/check-entrypoint.yaml` — one step added to run the new
  suite, plus the header comment. Outside the declared `Autonomous-Scope` paths but
  explicitly required by the issue's scope item 4 ("`check-entrypoint` workflow") and by the
  acceptance criterion "CI check-entrypoint workflow green on the PR". No other workflow
  changed.

### DIFF
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -234,6 +234,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
+| `ALISSA_FORCE_CHOWN` | `0` | `1` forces the root phase's full `chown -R` of the volume for that boot. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
 
 #### Config precedence: env var > daemon library default
 
@@ -256,6 +257,37 @@ that the baked [`agents.yaml`](./agents.yaml) ships (`claude`), and
 `on_missing_hub` must be `add` for the self-contained hub-ify-on-demand model —
 the library default `skip` would make a fresh volume review nothing. Both are
 still overridable via their env var.
+
+#### Boot-time ownership: the volume walk is guarded
+
+The container starts as root only to make a root-owned volume mount writable
+(see [Persistence](#persistence--mount-the-volume-at-workspace)). That `chown -R`
+used to run on **every** boot, and on a warm volume it is both a no-op and
+expensive: the walk stats every inode of every worktree hub, and the resulting
+dentry/inode slab is charged to the container's cgroup. An audit of the Railway
+service on 2026-08-09 found `memory.current` at **~5.98 GB** against **176 MB**
+of actual process RSS — 3.71 GB of it `slab_reclaimable` from the walk, the rest
+page cache from the hub sync — with zero reviewer sessions running. It is
+reclaimable cache, not a leak (the kernel drops it under real pressure at the
+cgroup limit), but it plateaus flat from the moment of deploy and makes the
+memory graph useless for spotting a real problem.
+
+So step 0 now **probes before it walks**: one `stat` over the mount point and its
+immediate children, and the `chown -R` runs only if something there is not owned
+by `alissa`. The probe is O(top-level entries) by construction — deliberately not
+`find … ! -user alissa`, which stats every inode and would recreate the same slab
+storm. A first boot on a fresh root-owned volume is unaffected: the mount point
+itself is root-owned, so the probe trips and the full walk runs exactly as before.
+
+The blind spot is a root-owned file created **deeper** than depth 1 — realistically,
+a platform console shell running as root. `ALISSA_FORCE_CHOWN=1` is the escape
+hatch: it skips the probe and forces the full walk for that boot. Set it, restart,
+then unset it.
+
+Reclaiming the cache after the fact is not an option on Railway: `/sys/fs/cgroup`
+is mounted read-only inside the container, so writing `memory.reclaim` fails with
+`EROFS` even as root (verified 2026-08-09). Avoidance is the only lever, which is
+why the entrypoint has no reclaim call.
 
 ### Reviewer console (runtime env only — `ALISSA_UI_ENABLED`, `ALISSA_UI_PASSCODE`, `PORT`)
 
@@ -425,6 +457,12 @@ volumes typically mount **root-owned**, so the container starts as root, the
 entrypoint `chown`s the mount to `alissa` (uid 1000), and then drops to that user
 (via `gosu`) for everything else — so a root-owned mount just works, no manual
 `chown` or init container needed. (claude still runs unprivileged, as it must.)
+
+That `chown` is **guarded**: on a warm volume it is skipped, because walking it
+charges GBs of reclaimable kernel cache to the container's memory metric for no
+benefit. See [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)
+for the numbers and for `ALISSA_FORCE_CHOWN`, the escape hatch to use after a root
+console shell has written into the volume.
 
 ### Optional egress firewall
 
@@ -650,9 +688,12 @@ volumes:
    its gates — the arming flag, the executor id, and where the identity is
    persisted. A bad role or an unarmed executor **dies here**, before any
    bootstrap. See [Bridge executor role](#bridge-executor-role-a-second-service-from-this-same-image).
-0b. As root: `chown` the `/workspace` mount to `alissa`, (optionally) raise the
-   egress firewall, then drop to `alissa` via `gosu`. Everything below runs
-   unprivileged.
+0b. As root: make the `/workspace` mount writable by `alissa` — probing
+   ownership at depth 1 and running the full `chown -R` only when the probe finds
+   a foreign owner or `ALISSA_FORCE_CHOWN=1` says so
+   ([why](#boot-time-ownership-the-volume-walk-is-guarded)) — then (optionally)
+   raise the egress firewall and drop to `alissa` via `gosu`. Everything below
+   runs unprivileged.
 1. Preflight + onboard the identities: validate `gh` (fatal if missing) and run
    `gh auth setup-git`; `alissa auth login` — fatal only when the API **rejects**
    the token, otherwise triaged and retried (see
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -143,13 +143,88 @@ fi
 #
 # claude refuses --dangerously-skip-permissions as root, so everything past this
 # point MUST run unprivileged — that is exactly what the drop guarantees.
+#
+# THE WALK IS GUARDED (issue #78). `chown -R` over the persistent volume stats
+# every inode of every worktree hub, and the dentry/inode slab it fills is
+# charged to the container's cgroup: a 2026-08-09 audit of the Railway service
+# found `memory.current` at ~5.98 GB with 176 MB of actual RSS — 3.71 GB of it
+# `slab_reclaimable` from this walk, the rest page cache from the hub sync in
+# step 3c. The kernel has no pressure to drop any of it below the cgroup limit,
+# so the metric plateaus flat at ~6 GB from the moment of deploy and stops being
+# usable for spotting a real leak. On a warm restart that whole cost buys
+# nothing: the volume is already `alissa`-owned from the previous boot.
+#
+# So the walk now runs only when a cheap probe says it has to. The probe is
+# O(top-level entries) BY CONSTRUCTION — one `stat` over the mount point plus
+# its depth-1 children. It deliberately does NOT `find … ! -user alissa`: find
+# stats every inode too, which is the exact slab storm this guard exists to
+# remove. The blind spot that buys is root-owned files created DEEPER in the
+# tree (a Railway console shell running as root is the realistic way) — hence
+# ALISSA_FORCE_CHOWN=1, which forces the full walk for one boot.
+#
+# Reclaiming after the fact is not an option here: `/sys/fs/cgroup` is mounted
+# read-only inside a Railway container, so `echo … > memory.reclaim` fails with
+# EROFS even as root (verified 2026-08-09). Avoidance is the only lever, which
+# is why there is no reclaim call in this block.
 # -----------------------------------------------------------------------------
+
+# Does <path> need the recursive chown? Returns 0 (yes) when the probe finds any
+# entry not owned by ${RUNTIME_USER} — or when it cannot tell, because a walk we
+# did not need is cheaper than a volume the daemon cannot write.
+#
+# Scope is the path itself plus its immediate children, nothing deeper: that is
+# what makes the probe O(top-level entries) instead of O(inodes). One `stat`
+# call for the lot, so a hub-per-repo mount costs a single process.
+needs_recursive_chown() {
+  local path="$1"
+  local entry owners owner
+  local -a entries=()
+
+  [ -e "${path}" ] || return 0
+  entries+=("${path}")
+  # Dotfiles included; `..?*` catches `..foo` without ever matching `..`.
+  for entry in "${path}"/* "${path}"/.[!.]* "${path}"/..?*; do
+    [ -e "${entry}" ] || [ -L "${entry}" ] || continue   # unmatched glob / dangling link
+    entries+=("${entry}")
+  done
+
+  # GNU stat does not dereference symlinks without -L, so a dangling link
+  # reports its own ownership rather than failing the whole call.
+  owners="$(stat -c '%U' -- "${entries[@]}" 2>/dev/null)" || return 0
+  [ -n "${owners}" ] || return 0
+  while IFS= read -r owner; do
+    # An uid with no passwd entry prints as UNKNOWN, which is correctly "not ours".
+    [ "${owner}" = "${RUNTIME_USER}" ] || return 0
+  done <<<"${owners}"
+  return 1
+}
+
 if [ "$(id -u)" = "0" ]; then
-  mkdir -p "${WORKSPACE_ROOT}" "${TMUX_TMPDIR:-/home/${RUNTIME_USER}/.tmux}"
-  # Fix ownership so the unprivileged user can write. -R because a restart may
-  # find files a previous root-mounted run left behind.
-  chown -R "${RUNTIME_USER}:${RUNTIME_USER}" \
-    "${WORKSPACE_ROOT}" "${TMUX_TMPDIR:-/home/${RUNTIME_USER}/.tmux}" 2>/dev/null || true
+  TMUX_DIR="${TMUX_TMPDIR:-/home/${RUNTIME_USER}/.tmux}"
+  mkdir -p "${WORKSPACE_ROOT}" "${TMUX_DIR}"
+
+  # Fix ownership so the unprivileged user can write. -R because a first boot
+  # finds a root-owned mount, and a root console shell can leave root-owned
+  # files behind — but only when the probe (or the operator) says so, per the
+  # block comment above. Each target is decided on its own: the tmux dir is a
+  # handful of sockets, the volume is millions of inodes.
+  FORCE_CHOWN=0
+  if is_truthy "${ALISSA_FORCE_CHOWN:-0}"; then
+    FORCE_CHOWN=1
+    log "ALISSA_FORCE_CHOWN=${ALISSA_FORCE_CHOWN} — forcing the full recursive chown (the depth-1 probe is skipped; use this once after a root shell has written deeper into the volume)"
+  fi
+  for CHOWN_TARGET in "${WORKSPACE_ROOT}" "${TMUX_DIR}"; do
+    if [ "${FORCE_CHOWN}" = "1" ] || needs_recursive_chown "${CHOWN_TARGET}"; then
+      chown -R "${RUNTIME_USER}:${RUNTIME_USER}" "${CHOWN_TARGET}" 2>/dev/null || true
+      if [ "${FORCE_CHOWN}" = "1" ]; then
+        log "chown -R ${RUNTIME_USER}:${RUNTIME_USER} ${CHOWN_TARGET} (forced)"
+      else
+        log "chown -R ${RUNTIME_USER}:${RUNTIME_USER} ${CHOWN_TARGET} (probe found entries not owned by ${RUNTIME_USER})"
+      fi
+    else
+      log "${CHOWN_TARGET} already owned by ${RUNTIME_USER} — skipping the recursive chown (set ALISSA_FORCE_CHOWN=1 to force it)"
+    fi
+  done
   log "workspace mount ${WORKSPACE_ROOT} owned by ${RUNTIME_USER}"
 
   if [ "${ALISSA_ENABLE_FIREWALL:-0}" = "1" ]; then
--- a/docker/claude/tests-entrypoint-chown.sh
+++ b/docker/claude/tests-entrypoint-chown.sh
@@ -1,0 +1,276 @@
+#!/usr/bin/env bash
+# =============================================================================
+# Tests for the entrypoint's GUARDED boot-time chown (#78).
+#
+# The defect: step 0 ran `chown -R alissa:alissa ${WORKSPACE_ROOT}` on EVERY
+# boot. The walk stats every inode of every worktree hub on the persistent
+# volume, and the dentry/inode slab it fills is charged to the container cgroup
+# — a 2026-08-09 audit of the Railway service found memory.current at ~5.98 GB
+# against 176 MB of real RSS, 3.71 GB of it slab_reclaimable from this walk.
+# On a warm restart the walk fixes nothing: the volume is already alissa-owned.
+#
+# The contract this pins:
+#
+#   1. warm boot (probe finds everything alissa-owned) -> NO recursive chown,
+#      and the probe stays O(top-level entries): one stat per target, never a
+#      `find`, never a depth-2 path
+#   2. a root-owned entry among the mount's depth-1 CHILDREN -> full chown runs
+#   3. the mount point ITSELF root-owned (the first-boot case) -> full chown runs
+#   4. ALISSA_FORCE_CHOWN=1 -> full chown runs with everything alissa-owned, and
+#      the probe is skipped entirely (the escape hatch for root-owned files
+#      deeper than the probe can see)
+#   5. ALISSA_FORCE_CHOWN=0 -> back to the probe (the flag is off by default)
+#   6. the probe against the REAL filesystem and the REAL stat, with the outcome
+#      derived from the tree's actual owner — so a stub that lies about stat's
+#      interface cannot make cases 1-5 pass on their own
+#
+# HOW (no docker, no root — CI has neither): this boots the REAL entrypoint.sh
+# with the commands its root phase shells out to replaced by stubs.
+#   * `id -u` answers 0 ONCE, so the root phase runs; the re-exec then sees the
+#     real uid, exactly like the post-gosu pass in the container
+#   * `gosu` drops its user argument and execs, standing in for the privilege drop
+#   * `chown` and `find` record their arguments instead of running (a non-root
+#     test cannot chown, and `find` must never be called at all)
+#   * `stat` answers scripted ownership per path (cases 1-5) or is the real stat
+#     (case 6), and records every invocation so the probe's COST is assertable
+#
+# Usage: bash docker/claude/tests-entrypoint-chown.sh
+# Needs: bash, awk, coreutils.
+# =============================================================================
+set -euo pipefail
+
+HERE="$(cd "$(dirname "$0")" && pwd)"
+ENTRYPOINT="${HERE}/entrypoint.sh"
+REAL_STAT="$(command -v stat)"
+
+fail=0
+pass() { printf '  ok   %s\n' "$1"; }
+bad()  { printf '  FAIL %s\n' "$1" >&2; fail=1; }
+info() { printf '%s\n' "$*"; }
+
+assert_contains() {
+  if grep -qF -- "$2" "$1"; then pass "$3"; else bad "$3 (not in $1: $2)"; fi
+}
+assert_not_contains() {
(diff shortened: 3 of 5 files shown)


alissa-app:
- 2026-08-10 12:00:36 UTC — `review-alissa-github-review-daemon-pr79-r1-cd690b` — round 1 of 22 — spawned

docker/claude/tests-entrypoint-chown.sh
```diff
@@ -0,0 +1,276 @@
...
+info "5. ALISSA_FORCE_CHOWN=0 -> back to the probe"
+: > "${TMPROOT}/owners-off.tsv"
+run_boot force-off FAKE_OWNERS="${TMPROOT}/owners-off.tsv" ALISSA_FORCE_CHOWN=0
+if chowned_workspace; then bad "ALISSA_FORCE_CHOWN=0 must not force the walk"
+else pass "no recursive chown"; fi
+assert_not_contains "${LOG}" "forcing the full recursive chown" "no forced-walk banner"
+
+# --- 6. the real stat, against the real filesystem ---------------------------
+info ""
+info "6. real stat, real tree -> the outcome follows the tree's actual owner"
```
alissa-app, line 262: **[minor]** The "targets are decided independently" property — the one the PR body calls out as a deliberate design choice ("so a stale tmux socket dir cannot drag the volume into a walk") — is **not pinned by any case**.

Mutation-proven, not guessed: I replaced the per-target repair with a combined one,

```diff
-      chown -R "${RUNTIME_USER}:${RUNTIME_USER}" "${CHOWN_TARGET}" 2>/dev/null || true
+      chown -R "${RUNTIME_USER}:${RUNTIME_USER}" "${WORKSPACE_ROOT}" "${TMUX_DIR}" 2>/dev/null || true
```

…which destroys exactly the property the split exists to provide (a root-owned tmux socket dir now drags the whole volume into the walk — the ~3.7 GB slab storm, back on a warm boot), and **all six cases stayed green**. Case 1 doesn't catch it because on a warm boot neither target walks; cases 2/3/4 don't because there the workspace is *supposed* to walk.

One case closes it — tmux dir root-owned, workspace clean, assert the workspace was **not** walked. I wrote and ran it against this head and it passes, so this is a coverage gap rather than a defect:

```bash
info "7. tmux dir root-owned, workspace clean -> only the tmux dir walks"
printf '%s\t%s\n' "${TMPROOT}/tmuxonly/tmux" root > "${TMPROOT}/owners-tmux.tsv"
run_boot tmuxonly FAKE_OWNERS="${TMPROOT}/owners-tmux.tsv"
if chowned_workspace; then bad "the workspace must NOT be dragged into the tmux dir's walk"
else pass "workspace skipped"; fi
if grep -qF -- "-R alissa:alissa ${TMPROOT}/tmuxonly/tmux" "${CHOWN_LOG}"; then pass "tmux dir walked on its own"
else bad "tmux dir should have walked"; fi
```

It also fails under the mutation above, which is what makes it worth adding.

docker/claude/entrypoint.sh
```diff
@@ -143,13 +143,88 @@ fi
...
+  entries+=("${path}")
+  # Dotfiles included; `..?*` catches `..foo` without ever matching `..`.
+  for entry in "${path}"/* "${path}"/.[!.]* "${path}"/..?*; do
+    [ -e "${entry}" ] || [ -L "${entry}" ] || continue   # unmatched glob / dangling link
+    entries+=("${entry}")
+  done
+
+  # GNU stat does not dereference symlinks without -L, so a dangling link
+  # reports its own ownership rather than failing the whole call.
+  owners="$(stat -c '%U' -- "${entries[@]}" 2>/dev/null)" || return 0
```
alissa-app, line 193: **[minor]** The probe tests the **owner** (`%U`), but the repair it guards sets **owner and group** (`chown -R alissa:alissa`). Those are different predicates, so a depth-1 entry sitting at `alissa:root` reads as clean and is never repaired — and the unconditional walk that used to fix group ownership on every boot is exactly what this PR removes.

Low likelihood (it needs a `chown alissa` without a group, or a root process with a root primary group creating an entry the daemon later takes ownership of), and mostly benign when it happens — `alissa` is still the owner, so mode-`6xx` writes work — but the group half of the invariant now has no repair path at all, and the documented blind spot in the block comment and the README enumerates only the *depth* issue, not this one.

`stat -c '%U:%G'` compared against `"${RUNTIME_USER}:${RUNTIME_USER}"` costs the same single call and makes the probe check what the walk actually asserts. If the group genuinely doesn't matter, that's a fine answer too — it just deserves to be the *stated* answer, next to the `UNKNOWN` note directly below.

docker/claude/entrypoint.sh
```diff
@@ -143,13 +143,88 @@ fi
...
+  # block comment above. Each target is decided on its own: the tmux dir is a
+  # handful of sockets, the volume is millions of inodes.
+  FORCE_CHOWN=0
+  if is_truthy "${ALISSA_FORCE_CHOWN:-0}"; then
+    FORCE_CHOWN=1
+    log "ALISSA_FORCE_CHOWN=${ALISSA_FORCE_CHOWN} — forcing the full recursive chown (the depth-1 probe is skipped; use this once after a root shell has written deeper into the volume)"
+  fi
+  for CHOWN_TARGET in "${WORKSPACE_ROOT}" "${TMUX_DIR}"; do
+    if [ "${FORCE_CHOWN}" = "1" ] || needs_recursive_chown "${CHOWN_TARGET}"; then
+      chown -R "${RUNTIME_USER}:${RUNTIME_USER}" "${CHOWN_TARGET}" 2>/dev/null || true
```
alissa-app, line 218: **[minor]** `2>/dev/null || true` discards whether the walk actually succeeded — and under the new guard that silence became load-bearing in a way it wasn't before.

`chown -R` **continues past per-entry errors and still processes the parent directories**, then exits non-zero. I confirmed both halves empirically (GNU coreutils 9.1): traversal is post-order, and a run that fails on a child still reports the top:

```
$ chown -R -v alissa:alissa failtest
ownership of 'failtest/top' retained as alissa:alissa
chown: cannot read directory 'failtest/sub': Permission denied
failed to change ownership of 'failtest/sub' to alissa:alissa
ownership of 'failtest' retained as alissa:alissa      <- top still processed
exit=1
```

So a first boot whose walk fails on some entries **deeper than depth 1** still finishes with the mount point and the depth-1 children owned by `alissa`. Every subsequent boot's probe then reports clean, the unrepaired subtree is latched silently, and — because both the exit status and the stderr are dropped — the boot log carries no trace that anything went wrong. The operator's only signal is whatever breaks downstream, and the only lever is `ALISSA_FORCE_CHOWN=1`, which they have no reason to reach for.

Before this PR the same failure was equally silent but far less consequential: the walk re-ran unconditionally next boot. Keeping the `|| true` (step 0 must not die here — agreed) while logging the failure is a two-line change and preserves the guard's whole benefit:

```bash
if ! chown -R "${RUNTIME_USER}:${RUNTIME_USER}" "${CHOWN_TARGET}" 2>/dev/null; then
  log "WARNING: chown -R ${CHOWN_TARGET} reported errors — entries below depth 1 may still be foreign-owned; re-run this boot with ALISSA_FORCE_CHOWN=1 if the daemon cannot write"
fi
```

Related and worth a sentence in the block comment above: the reason a *killed* first boot is self-healing at all is that `chown -R` is post-order, so the mount point is chowned last and a partial walk leaves the probe tripping. That's a real invariant this design now leans on, and it's currently invisible — an "optimisation" that chowned the mount point first would silently convert every interrupted first boot into a permanently latched half-owned volume.

docker/claude/README.md
```diff
@@ -234,6 +234,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
+| `ALISSA_FORCE_CHOWN` | `0` | `1` forces the root phase's full `chown -R` of the volume for that boot. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
```
alissa-app, line 237: **[nit]** The row documents `1` as the value, but the gate is `is_truthy`, which also accepts `true` / `yes` / `on` (case-insensitive). This README already has the right precedent one section down, on the other `is_truthy` flag:

> `ALISSA_UI_ENABLED` … ``​`1`/`true`/`yes`/`on` starts the console sidecar; anything else (incl. unset) leaves **no listener**``

Matching that wording here keeps the table honest about which flags are narrow and which are broad — the distinction matters precisely because `ALISSA_ENABLE_FIREWALL`, two rows above, really *is* `1`-only, and the PR body calls out that asymmetry as deliberate.

alissa-app (approved):

Review task: **TASK-2015556292** · origin task TASK-357393641 · implementation task TASK-1002979593 · reviewed head `7d91b8e`.

The guard is correct, it fails safe in every direction I could reach, and the test suite is genuinely load-bearing rather than decorative. Nothing here blocks the merge.

**What I verified independently** (not taken from the delivery note):

* `tests-entrypoint-chown.sh` re-run at `7d91b8e` — **ALL PASS** (6 cases). `config` / `identity` / `auth` / `executor` suites — ALL PASS. `bash -n` clean on both new/changed shell files.
* **Mutation-checked, one revert at a time.** Unguarding the walk (`if true`) → 6 failures across cases 1/5/6. Making the probe always report clean (`return 1` at the top) → 5 failures across cases 1/2/3. So both halves of the contract are pinned.
* **The probe was exercised as a unit**, lifted out of the entrypoint and driven against real trees: clean tree → skip; empty dir → skip; **missing path → walk**; depth-1 symlink to `/root` (link owned by `alissa`) → skip, consistent with `chown -R`'s own no-follow default; **dangling symlink → skip, not a crash**; filenames containing spaces and newlines → correct (the `%U` output is owner-per-line, so pathological names cannot shift the parse); 2000 depth-1 entries → still one `stat` process. The O(top-level entries) claim holds at scale, and every "cannot tell" path lands on *walk*.
* **The interrupted-first-boot question, which the docs don't raise.** I checked GNU `chown -R`'s traversal order empirically (`chown -R -v` on a nested tree): it is **post-order** — the mount point is chowned *last*. That is what makes a killed first boot self-healing: the top stays root-owned, so the next boot's probe trips and re-walks. The design is safe here, but it's safe by an unstated dependency on coreutils' traversal order (see the comment on the `chown` line).
* **CI**: all six checks green, and the new Guarded-Chown step really ran in the `check-entrypoint` job. Case 6 took the **`tree really owned by runner (not alissa) -> walked`** branch on the runner while it takes the skip branch locally — so the real-`stat` case does cover both directions across the two environments, exactly as the PR body claims.

**Rubric:**

* **Correctness** — nothing blocking. The probe fails safe on every branch (missing path, `stat` non-zero, empty output, `UNKNOWN` uid). First-boot behaviour is preserved and pinned by case 3. One minor: the probe reads the owner while the repair sets owner **and** group.
* **Coding practices** — good. The rationale lives where the next reader will look, the "why not `find`" argument is recorded at the decision point, and `is_truthy` (defined at line 58, well before use) is the right call for a brand-new flag. One nit on the README row.
* **Security** — nothing found. No new inputs, endpoints or writes; the new flag is not a secret and is correctly an ARG (unlike `ALISSA_UI_PASSCODE`). `stat` without `-L` and `chown -R`'s default `-P` mean a planted symlink in the volume cannot redirect either the probe or the walk. Net effect is a *reduction* in privilege change: root-owned files planted deep are no longer silently adopted by `alissa` on every boot.
* **Performance** — this is the point of the PR and it lands. One `stat` process per target regardless of child count, no `find`, no depth-2 path — asserted rather than asserted-at.
* **Side-effects / blast radius** — step 0 is boot-critical and a bug here bricks the container, which is the reason to care about the fail-safe directions above; they all check out, and the four pre-existing entrypoint suites still pass. The acceptance criterion that actually matters is a deploy measurement on the warm Railway volume and is correctly declared as an operator gate, not claimed.

**Findings:** 0 blocker · 0 major · 3 minor · 1 nit. All four are non-gating; per CR8, any one that's deferred rather than fixed should carry a registered task.

Approve.

docker/claude/tests-entrypoint-chown.sh
RHDZMOTA: `[triage:pursue]`

Agreed, and the mutation proof is the part that settles it: a suite whose six cases all stay green under a change that reinstates the volume walk is not pinning the property the split exists for. Case 1 can't see it (nothing walks) and 2/3/4 can't (the workspace is meant to walk), exactly as you say — the missing quadrant is "the *other* target is dirty".

Taking your case 7 as written (tmux dir root-owned, workspace clean → assert the workspace was NOT walked, and that the tmux dir walked on its own), and I'll re-run your combined-`chown` mutation against it to confirm it fails there, so the case is added under the same counterfactual bar as the rest of the suite.

This PR is approved, so it lands as a follow-up PR rather than another commit here; I'll post the link on this thread.

docker/claude/entrypoint.sh
RHDZMOTA: `[triage:pursue]`

You're right that the predicate and the repair disagree, and I'd rather fix that than declare the group irrelevant — the walk asserts `alissa:alissa`, so the probe should test `alissa:alissa`. `stat -c '%U:%G'` is the same single call, so the guard's whole cost argument is untouched.

Worth naming why the drift matters more now than it looks: the group half previously had an unconditional repair path every boot, and this PR is what removed it. An `alissa:root` entry is benign today only because the daemon happens to write as the owner — that is a property of the current file modes, not an invariant, and nothing would have re-established the group once it drifted.

I'll also extend the blind-spot enumeration (block comment + README) so it lists what the probe checks and at what depth, instead of only the depth half — your point that the documented blind spot was incomplete stands on its own.

Follow-up PR, not another commit on this approved branch; link to follow on this thread.

docker/claude/entrypoint.sh
RHDZMOTA: `[triage:pursue]`

This is the finding I'd have most regretted deferring: the guard converts a transient, self-healing failure into a latched one, and does it silently. Your reading of the mechanism matches mine — `chown -R` is post-order and keeps going after per-entry errors, so a walk that fails deep still leaves depth 0 and depth 1 owned by `alissa`, which is precisely the surface my probe inspects. The next boot then reports clean forever, with `ALISSA_FORCE_CHOWN=1` the only lever and nothing in the log to suggest reaching for it.

Taking both halves:

* keep `|| true` (step 0 must not die on this) but branch on the exit status and log a WARNING naming the target and the `ALISSA_FORCE_CHOWN=1` remedy — stderr stays suppressed so a per-entry error storm can't flood the boot log, and the warning is the operator's signal;
* record the post-order dependence in the block comment. Agreed that it is a real invariant this design leans on and that it is currently invisible: a "chown the mount point first" optimisation would turn every interrupted first boot into a permanently latched half-owned volume. That belongs in the comment as a warning to the next editor, not in anyone's head.

I'll re-verify the post-order/continue-past-errors behaviour locally rather than take it on trust, and say so in the follow-up's notes. Follow-up PR; link to follow here.

docker/claude/README.md
RHDZMOTA: `[triage:pursue]`

Correct — the row says `1` while the gate is `is_truthy`, and the table is where an operator decides what to type. I'll match the `ALISSA_UI_ENABLED` wording (`1`/`true`/`yes`/`on`, anything else including unset is off).

Your framing of *why* is the reason it's worth a change rather than a shrug: the table currently reads as if every flag on it were `1`-only, which quietly erases the `ALISSA_ENABLE_FIREWALL` asymmetry that the PR body argues is deliberate. Documenting the broad flags broadly is what makes the narrow one legible as a choice.

Rolled into the same follow-up PR as the other three findings.

docker/claude/tests-entrypoint-chown.sh
RHDZMOTA: Implemented in the follow-up: https://github.com/fahera-mx/alissa-github-review-daemon/pull/80 (`Follow-up-To: #79`, cut from this PR's approved head `7d91b8e`, based on `main`). That PR states the merge gate explicitly: **merge #79 first.**

docker/claude/entrypoint.sh
RHDZMOTA: Implemented in the follow-up: https://github.com/fahera-mx/alissa-github-review-daemon/pull/80 (`Follow-up-To: #79`, cut from this PR's approved head `7d91b8e`, based on `main`). That PR states the merge gate explicitly: **merge #79 first.**

docker/claude/entrypoint.sh
RHDZMOTA: Implemented in the follow-up: https://github.com/fahera-mx/alissa-github-review-daemon/pull/80 (`Follow-up-To: #79`, cut from this PR's approved head `7d91b8e`, based on `main`). That PR states the merge gate explicitly: **merge #79 first.**

docker/claude/README.md
RHDZMOTA: Implemented in the follow-up: https://github.com/fahera-mx/alissa-github-review-daemon/pull/80 (`Follow-up-To: #79`, cut from this PR's approved head `7d91b8e`, based on `main`). That PR states the merge gate explicitly: **merge #79 first.**

RHDZMOTA:
**Round-1 follow-up opened:** https://github.com/fahera-mx/alissa-github-review-daemon/pull/80

All four round-1 findings (3 `[minor]`, 1 `[nit]`) were triaged `[triage:pursue]` on their
threads and are implemented there — not here, because this branch is approved and a commit
pushed after the squash-merge would be silently lost.

Merge order matters: **this PR (#79) merges first.** PR #80 is cut from this PR's approved head
`7d91b8e` (every line it touches is introduced here, so a branch off `main` could not carry the
fixes), is based on `main`, and claims no issue. Merging it first would squash this PR's whole
delta into `main` under its commit message and leave this one resolving as empty.

The head of this branch has not moved since the approve.

PR #80, opened 2026-08-10
TASK-1002979593: PR #79 round-1 follow-up — probe owner+group, report a failed walk, pin target independence

RHDZMOTA:
Follow-up to PR #79 (round 1), which was approved carrying three `[minor]` findings and one
`[nit]`. All four were triaged `[triage:pursue]` on their threads; an approved branch must not
take another commit, so they land here.

**Source PR:** #79 · **review round:** 1 · **reviewer:** alissa-app
**Origin task:** TASK-357393641 · **Implementation task:** TASK-1002979593

Follow-up-To: #79

> **MERGE ORDER — merge #79 first.** This branch is cut from #79's approved head
> (`7d91b8e`), not from `main`, because every line it touches is introduced by that PR: the
> probe function, the guard block, the new test suite, and the README section do not exist on
> `main` yet, so a branch off `main` could not carry these fixes at all. GitHub's diff against
> `main` therefore shows #79's changes too; the review surface of *this* PR is the single
> commit `7d91b8e..HEAD`. Merging this one **first** would squash #79's whole delta into `main`
> under this commit message and leave #79 resolving as empty — losing its history and its
> issue-closing claim. This PR closes no issue by design.

## What changed

* **`[minor]` probe tested the owner, the walk sets owner *and* group.** `stat -c '%U:%G'`
  compared against `alissa:alissa`. An `alissa:root` entry used to read as clean, and #79 is
  what removed the unconditional every-boot walk that had been the group's only repair path.
  Same single `stat` call, so the guard's cost argument is untouched.
* **`[minor]` a failed walk was swallowed.** `chown -R` is post-order and continues past
  per-entry errors, so a walk that fails deep still leaves depths 0 and 1 owned by `alissa` —
  precisely the surface the probe inspects — and every later boot then reads clean with nothing
  in the log. The exit status is now branched on and a `WARNING` naming the target and
  `ALISSA_FORCE_CHOWN=1` is logged. `|| true` semantics stay (step 0 must not die) and stderr
  stays suppressed so a per-entry error storm cannot flood the boot log.
* **`[minor]` per-target independence was unpinned.** The reviewer's mutation — one combined
  `chown -R … "${WORKSPACE_ROOT}" "${TMUX_DIR}"` — reinstated the volume walk with all six
  cases still green. Case 7 (tmux dir root-owned, workspace clean) closes that quadrant.
* **`[nit]` README row said `1` for an `is_truthy` gate.** Now `1`/`true`/`yes`/`on`, matching
  the `ALISSA_UI_ENABLED` row, which is what keeps the deliberately narrow
  `ALISSA_ENABLE_FIREWALL` legible as a choice.
* **Documentation of the invariant the design leans on:** `chown -R` chowning the mount point
  **last** is why an interrupted first boot self-heals. Recorded in the block comment as a
  warning to the next editor, and in the README alongside the failed-walk behaviour.

## Round-1 verdict on this PR: approve, findings deferred

Approved by alissa-app carrying two `[minor]` and two `[nit]`. All four are triaged
`[triage:later]` on their threads and registered as **TASK-313195870** (soft-dependent on
TASK-1002979593): the test harness leaks one backgrounded `sleep` per boot now that the
post-drop pass actually runs; the `workspace mount … owned by alissa` summary line
contradicts the new `WARNING` on a failed walk and no case asserts it; the warning's tail
advises `ALISSA_FORCE_CHOWN=1` even when the *forced* walk is what failed, and over-claims
"will read this tree as clean" where "may" is accurate; and the settings-table row still
describes the probe as finding "a foreign owner" when this PR made it owner **and** group.

Deferred rather than stacked: #79 and #80 are both approved and unmerged, so a third PR
would have to be cut from this branch's head and would impose an exact #79 → #80 → #81
merge order — where getting it wrong squash-merges an earlier PR's whole delta under a
later PR's message. Four diagnostics-and-docs findings do not justify that risk. None of
them changes the guard's behaviour.

### Judgment calls

* **All four findings in one follow-up PR, not four.** They touch the same three files and the
  same guard; splitting them would force the operator into a four-way merge ordering for one
  round's findings. Level-three stacking was avoided deliberately — nothing here stacks on
  anything but #79.
* **Branch cut from #79's approved head rather than `main`.** Stated as a merge gate above. The
  convention's "off the default branch" protects the approved head from *moving*; cutting a new
  branch from it does not move it, and the alternative is unbuildable (add/add on files #79
  introduces). Base is `main`, as required.
* **`|| true` kept, status not.** The reviewer's own suggestion. Step 0 dying on a partial chown
  would be a worse failure than the silent latch it replaces; the warning is the signal.
* **stderr stays `2>/dev/null`.** A deep permission-error storm would otherwise be thousands of
  lines in the boot log. The exit status carries the one bit an operator needs, and the warning
  names the remedy.
* **The `[nit]` was implemented, not deferred.** It was one table cell in a file this PR already
  touches, so deferring it to a registered task would have cost more than the fix.
* **Harness flaw found and fixed while adding case 9, disclosed rather than quietly patched.**
  The `gosu` stub in #79's suite exec'd `entrypoint.sh` directly, but that file is mode 644 in
  the repo (the Dockerfile chmods it into the image), so the post-drop pass died with
  "Permission denied" on every case. Nothing caught it because every assertion in the suite is
  about the root phase, which runs *before* the drop — the boots were valid for what they
  asserted, and no result in #79's notes is invalidated by this. The stub now execs through
  `bash` (as `run_boot` already did) and case 9 asserts the boot reaches the post-drop role
  line, so the same silent death cannot recur.

* [ ] **Everything in #79's not-verified list still stands** and is not repeated here: the
  Railway memory measurement on a warm volume, first boot on a cold root-owned volume,
  `ALISSA_FORCE_CHOWN=1` on the real service, and the fact that no image was built (no docker
  in this environment).
* [ ] **The group half was never exercised against a real `alissa:root` file** — creating one
  needs root, which this environment and CI both lack. Case 8 drives it through scripted
  ownership; the real-`stat` case (6) only covers owner drift.
* [ ] **The new `WARNING` has never fired against a real failing `chown -R`** — case 9 forces a
  non-zero exit from the stub. The underlying claims about real `chown -R` (post-order,
  continues past per-entry errors, still processes the parents, exits 1) were reproduced
  locally on GNU coreutils 9.1, but on this host's filesystem, not on a Railway volume.
* [ ] **Nothing here changes the deploy-side acceptance criterion of issue #78**, which remains
  the operator's measurement after the merge of both PRs.

### Verification

* `docker/claude/tests-entrypoint-chown.sh` — **ALL PASS**, 9 cases (was 6).
  Each new case counterfactually checked against a mutation of the narrowest clause, in a
  scratch copy: the reviewer's combined-`chown` mutation fails case 7 only; a probe that
  ignores the group (`${owner%%:*}`) fails case 8 only; restoring `|| true` on the chown fails
  case 9. The pre-existing mutations from #79 (probe always clean / always walk) were re-run
  and still fail their cases.
* `tests-entrypoint-config.sh`, `-identity.sh`, `-auth.sh`, `-executor.sh`, `-ui.sh` — all
  PASS on this head (the entrypoint changed, so all five were re-run, not just the new one).
* `chown -R` behaviour reproduced directly (GNU coreutils 9.1): `-v` output shows
  `deep → sub → top` ordering, and a run failing on an unreadable child still processes the
  parent and exits 1.
* `bash -n docker/claude/entrypoint.sh` — clean. No shellcheck in this repo's CI or in this
  environment.
* No Python touched, so the package suites and `check-version-bump` do not apply.

### Scope touched

* `docker/claude/entrypoint.sh` — `%U:%G` probe, chown status branch + `WARNING`, block-comment
  documentation of the post-order invariant and the exact blind spot.
* `docker/claude/tests-entrypoint-chown.sh` — cases 7/8/9, `owner:group` stub, `CHOWN_FAIL`
  hook, the `gosu`-through-`bash` fix, tightened remedy assertion in case 9.
* `docker/claude/README.md` — the `is_truthy` values in the settings row, owner+group in the
  probe description, the failed/interrupted-walk paragraph, step 0b wording.
* No excursion: no workflow, Dockerfile, or Python file is touched — the new cases run inside
  the step #79 already added to `check-entrypoint`.

### DIFF
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -234,7 +234,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
-| `ALISSA_FORCE_CHOWN` | `0` | `1` forces the root phase's full `chown -R` of the volume for that boot. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
+| `ALISSA_FORCE_CHOWN` | `0` | `1`/`true`/`yes`/`on` forces the root phase's full `chown -R` of the volume for that boot; anything else (incl. unset) leaves the probe in charge. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
 
 #### Config precedence: env var > daemon library default
 
@@ -273,16 +273,30 @@ cgroup limit), but it plateaus flat from the moment of deploy and makes the
 memory graph useless for spotting a real problem.
 
 So step 0 now **probes before it walks**: one `stat` over the mount point and its
-immediate children, and the `chown -R` runs only if something there is not owned
-by `alissa`. The probe is O(top-level entries) by construction — deliberately not
-`find … ! -user alissa`, which stats every inode and would recreate the same slab
-storm. A first boot on a fresh root-owned volume is unaffected: the mount point
-itself is root-owned, so the probe trips and the full walk runs exactly as before.
+immediate children, and the `chown -R` runs only if something there is not
+`alissa:alissa`. Owner **and** group, because that is what the walk asserts — a
+probe testing only the owner would read an `alissa:root` entry as clean, and the
+unconditional every-boot walk that used to repair the group is exactly what this
+change removes. The probe is O(top-level entries) by construction — deliberately
+not `find … ! -user alissa`, which stats every inode and would recreate the same
+slab storm. A first boot on a fresh root-owned volume is unaffected: the mount
+point itself is root-owned, so the probe trips and the full walk runs exactly as
+before.
 
-The blind spot is a root-owned file created **deeper** than depth 1 — realistically,
-a platform console shell running as root. `ALISSA_FORCE_CHOWN=1` is the escape
-hatch: it skips the probe and forces the full walk for that boot. Set it, restart,
-then unset it.
+The blind spot is precisely one thing: **depth**. Anything below the mount point's
+immediate children is invisible to the probe — realistically, a platform console
+shell running as root. `ALISSA_FORCE_CHOWN=1` is the escape hatch: it skips the
+probe and forces the full walk for that boot. Set it, redeploy, then unset it.
+
+Two failure modes are worth knowing about, because the guard changed their
+consequences. A walk that **fails** partway (a permission error deep in the tree)
+still processes the mount point and its children — `chown -R` is post-order and
+continues past per-entry errors — so the probe would read the tree as clean
+forever after. The entrypoint therefore logs `WARNING: chown -R … reported errors`
+and names `ALISSA_FORCE_CHOWN=1` as the remedy; that warning is the only signal
+there is, so it is worth alerting on. A walk that is **interrupted** (the container
+is killed mid-boot) is self-healing for the same post-order reason: the mount point
+is chowned last, so a partial walk leaves the probe tripping on the next boot.
 
 Reclaiming the cache after the fact is not an option on Railway: `/sys/fs/cgroup`
 is mounted read-only inside the container, so writing `memory.reclaim` fails with
@@ -689,8 +703,9 @@ volumes:
    persisted. A bad role or an unarmed executor **dies here**, before any
    bootstrap. See [Bridge executor role](#bridge-executor-role-a-second-service-from-this-same-image).
 0b. As root: make the `/workspace` mount writable by `alissa` — probing
-   ownership at depth 1 and running the full `chown -R` only when the probe finds
-   a foreign owner or `ALISSA_FORCE_CHOWN=1` says so
+   ownership (owner **and** group) at depth 1 and running the full `chown -R` only
+   when the probe finds something foreign or `ALISSA_FORCE_CHOWN=1` says so, and
+   **warning** if that walk reports errors
    ([why](#boot-time-ownership-the-volume-walk-is-guarded)) — then (optionally)
    raise the egress firewall and drop to `alissa` via `gosu`. Everything below
    runs unprivileged.
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -234,6 +234,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
+| `ALISSA_FORCE_CHOWN` | `0` | `1` forces the root phase's full `chown -R` of the volume for that boot. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
 
 #### Config precedence: env var > daemon library default
 
@@ -256,6 +257,37 @@ that the baked [`agents.yaml`](./agents.yaml) ships (`claude`), and
 `on_missing_hub` must be `add` for the self-contained hub-ify-on-demand model —
 the library default `skip` would make a fresh volume review nothing. Both are
 still overridable via their env var.
+
+#### Boot-time ownership: the volume walk is guarded
+
+The container starts as root only to make a root-owned volume mount writable
+(see [Persistence](#persistence--mount-the-volume-at-workspace)). That `chown -R`
+used to run on **every** boot, and on a warm volume it is both a no-op and
+expensive: the walk stats every inode of every worktree hub, and the resulting
+dentry/inode slab is charged to the container's cgroup. An audit of the Railway
+service on 2026-08-09 found `memory.current` at **~5.98 GB** against **176 MB**
+of actual process RSS — 3.71 GB of it `slab_reclaimable` from the walk, the rest
+page cache from the hub sync — with zero reviewer sessions running. It is
+reclaimable cache, not a leak (the kernel drops it under real pressure at the
+cgroup limit), but it plateaus flat from the moment of deploy and makes the
+memory graph useless for spotting a real problem.
+
+So step 0 now **probes before it walks**: one `stat` over the mount point and its
+immediate children, and the `chown -R` runs only if something there is not owned
+by `alissa`. The probe is O(top-level entries) by construction — deliberately not
+`find … ! -user alissa`, which stats every inode and would recreate the same slab
+storm. A first boot on a fresh root-owned volume is unaffected: the mount point
+itself is root-owned, so the probe trips and the full walk runs exactly as before.
+
+The blind spot is a root-owned file created **deeper** than depth 1 — realistically,
+a platform console shell running as root. `ALISSA_FORCE_CHOWN=1` is the escape
+hatch: it skips the probe and forces the full walk for that boot. Set it, restart,
+then unset it.
+
+Reclaiming the cache after the fact is not an option on Railway: `/sys/fs/cgroup`
+is mounted read-only inside the container, so writing `memory.reclaim` fails with
+`EROFS` even as root (verified 2026-08-09). Avoidance is the only lever, which is
+why the entrypoint has no reclaim call.
 
 ### Reviewer console (runtime env only — `ALISSA_UI_ENABLED`, `ALISSA_UI_PASSCODE`, `PORT`)
 
@@ -425,6 +457,12 @@ volumes typically mount **root-owned**, so the container starts as root, the
 entrypoint `chown`s the mount to `alissa` (uid 1000), and then drops to that user
 (via `gosu`) for everything else — so a root-owned mount just works, no manual
 `chown` or init container needed. (claude still runs unprivileged, as it must.)
+
+That `chown` is **guarded**: on a warm volume it is skipped, because walking it
+charges GBs of reclaimable kernel cache to the container's memory metric for no
+benefit. See [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)
+for the numbers and for `ALISSA_FORCE_CHOWN`, the escape hatch to use after a root
+console shell has written into the volume.
 
 ### Optional egress firewall
 
@@ -650,9 +688,12 @@ volumes:
    its gates — the arming flag, the executor id, and where the identity is
    persisted. A bad role or an unarmed executor **dies here**, before any
    bootstrap. See [Bridge executor role](#bridge-executor-role-a-second-service-from-this-same-image).
-0b. As root: `chown` the `/workspace` mount to `alissa`, (optionally) raise the
-   egress firewall, then drop to `alissa` via `gosu`. Everything below runs
-   unprivileged.
+0b. As root: make the `/workspace` mount writable by `alissa` — probing
+   ownership at depth 1 and running the full `chown -R` only when the probe finds
+   a foreign owner or `ALISSA_FORCE_CHOWN=1` says so
+   ([why](#boot-time-ownership-the-volume-walk-is-guarded)) — then (optionally)
+   raise the egress firewall and drop to `alissa` via `gosu`. Everything below
+   runs unprivileged.
 1. Preflight + onboard the identities: validate `gh` (fatal if missing) and run
    `gh auth setup-git`; `alissa auth login` — fatal only when the API **rejects**
    the token, otherwise triaged and retried (see
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -156,11 +156,26 @@ fi
 #
 # So the walk now runs only when a cheap probe says it has to. The probe is
 # O(top-level entries) BY CONSTRUCTION — one `stat` over the mount point plus
-# its depth-1 children. It deliberately does NOT `find … ! -user alissa`: find
-# stats every inode too, which is the exact slab storm this guard exists to
-# remove. The blind spot that buys is root-owned files created DEEPER in the
-# tree (a Railway console shell running as root is the realistic way) — hence
-# ALISSA_FORCE_CHOWN=1, which forces the full walk for one boot.
+# its depth-1 children, testing owner AND group, because `alissa:alissa` is what
+# the walk asserts and a probe that tested only the owner would read `alissa:root`
+# as clean and never repair it. It deliberately does NOT `find … ! -user alissa`:
+# find stats every inode too, which is the exact slab storm this guard exists to
+# remove.
+#
+# WHAT THE PROBE CANNOT SEE, precisely: anything DEEPER than the mount point's
+# immediate children. A platform console shell running as root is the realistic
+# way to get there — hence ALISSA_FORCE_CHOWN=1, which forces the full walk for
+# one boot.
+#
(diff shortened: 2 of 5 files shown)


alissa-app:
- 2026-08-10 12:26:38 UTC — `review-alissa-github-review-daemon-pr80-r1-a60190` — round 1 of 22 — spawned

docker/claude/tests-entrypoint-chown.sh
```diff
@@ -0,0 +1,339 @@
...
+#!/usr/bin/env bash
+case "$1 $2" in
+  "worker status") echo "worker is running" ;;
+esac
+exit 0
+STUB
+cat > "${BIN}/alissa-revloop" <<'STUB'
+#!/usr/bin/env bash
+trap 'exit 0' TERM INT
+sleep 600 &
```
alissa-app, line 163: `[minor]` **The `gosu`-through-`bash` fix is right, and it leaks a process per boot.**

Before this commit the post-drop pass died at `exec "$@"` on a mode-644 file, so this stub never
ran. Now it does — and nothing kills the backgrounded `sleep`. `run_boot` SIGTERMs the entrypoint,
the entrypoint TERMs `alissa-revloop`, this `trap` fires `exit 0`, and `sleep 600` is reparented to
PID 1 for ten minutes.

Measured on one host, same session, clean baseline:

```
baseline:                    0
after the suite @ 7d91b8e:   0
after the suite @ 0576fad:   9      # one per boot, every one PPID 1
```

On CI the runner is torn down, so it costs nothing there. Locally it accumulates: the mutation loop
described in this PR's own notes runs the suite five or six times over, which is ~50 orphans.

`trap 'kill $(jobs -p) 2>/dev/null; exit 0' TERM INT` closes it. The reason it is worth fixing
rather than shrugging at is the same reason the `Permission denied` went unnoticed for a whole PR:
every assertion in this file is about the root phase, so nothing here can see what the post-drop
pass does or leaves behind.

docker/claude/entrypoint.sh
```diff
@@ -143,13 +143,122 @@ fi
...
+        log "chown -R ${RUNTIME_USER}:${RUNTIME_USER} ${CHOWN_TARGET} (forced)"
+      else
+        log "chown -R ${RUNTIME_USER}:${RUNTIME_USER} ${CHOWN_TARGET} (probe found entries that are not ${RUNTIME_USER}:${RUNTIME_USER})"
+      fi
+      [ "${CHOWN_OK}" = "1" ] || log "WARNING: chown -R ${CHOWN_TARGET} reported errors — entries below depth 1 may still be foreign-owned, and the depth-1 probe will read this tree as clean on every later boot. Re-deploy once with ALISSA_FORCE_CHOWN=1 if the daemon cannot write."
+    else
+      log "${CHOWN_TARGET} already owned by ${RUNTIME_USER}:${RUNTIME_USER} — skipping the recursive chown (set ALISSA_FORCE_CHOWN=1 to force it)"
+    fi
+  done
   log "workspace mount ${WORKSPACE_ROOT} owned by ${RUNTIME_USER}"
```
alissa-app, line 262: `[minor]` **This line still asserts the mount is `alissa`-owned immediately after the new
`WARNING` says it may not be** — and it is the grep marker #79's delivery note deliberately kept,
so anything alerting on it gets a clean signal on exactly the boot this PR added a warning for.

Case 9's boot log at this head, verbatim:

```
[entrypoint] chown -R alissa:alissa …/workspace (probe found entries that are not alissa:alissa)
[entrypoint] WARNING: chown -R …/workspace reported errors — entries below depth 1 may still be foreign-owned…
[entrypoint] …/tmux already owned by alissa:alissa — skipping the recursive chown…
[entrypoint] workspace mount …/workspace owned by alissa          <-- contradicts the WARNING two lines up
```

The whole argument for adding the warning is that the boot log is now the operator's only signal
that the volume is half-repaired. Gating this line on the workspace target's `CHOWN_OK` — or
rewording it so it reports that step 0 finished rather than asserting an ownership the previous
line just disclaimed — keeps the marker for whatever greps it. Case 9 is the natural place to pin
it, and note that no case currently asserts anything about this line.

docker/claude/entrypoint.sh
```diff
@@ -143,13 +143,122 @@ fi
...
+        CHOWN_OK=1
+      else
+        CHOWN_OK=0
+      fi
+      if [ "${FORCE_CHOWN}" = "1" ]; then
+        log "chown -R ${RUNTIME_USER}:${RUNTIME_USER} ${CHOWN_TARGET} (forced)"
+      else
+        log "chown -R ${RUNTIME_USER}:${RUNTIME_USER} ${CHOWN_TARGET} (probe found entries that are not ${RUNTIME_USER}:${RUNTIME_USER})"
+      fi
+      [ "${CHOWN_OK}" = "1" ] || log "WARNING: chown -R ${CHOWN_TARGET} reported errors — entries below depth 1 may still be foreign-owned, and the depth-1 probe will read this tree as clean on every later boot. Re-deploy once with ALISSA_FORCE_CHOWN=1 if the daemon cannot write."
```
alissa-app, line 257: `[nit]` Two small things about this warning, both about what it says rather than that it
exists (which is the right call):

1. **It is self-contradictory after a forced walk.** `ALISSA_FORCE_CHOWN` is evaluated before the
   probe, so a *forced* walk that fails reaches this same line and advises "Re-deploy once with
   `ALISSA_FORCE_CHOWN=1`" — the thing that just failed. A `${FORCE_CHOWN}`-aware tail would say
   something actionable in both branches.
2. **It slightly over-claims.** "the depth-1 probe will read this tree as clean on every later
   boot" is the worst case, not the certain one: if the walk failed at depth 0 or 1 the probe keeps
   tripping and the walk retries every boot — a different failure, and a noisier one. "may read
   this tree as clean" covers both.

docker/claude/README.md
```diff
@@ -234,6 +234,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
+| `ALISSA_FORCE_CHOWN` | `0` | `1`/`true`/`yes`/`on` forces the root phase's full `chown -R` of the volume for that boot; anything else (incl. unset) leaves the probe in charge. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
```
alissa-app, line 237: `[nit]` The truthiness half of this row is fixed, but the sentence right after it still
describes the pre-#80 probe: *"probes ownership at depth 1 and walks only when it finds a foreign
owner"*. This PR is what made it owner **and** group — the prose section below and step 0b were
both updated to say so, this row was not, and it is the row an operator reads first.

"a foreign owner or group" matches the code and the two sections that cross-reference it.

alissa-app (approved):
## Review round 1 — `approve`

Fresh-instance review per `alissa-code-review` (CR1–CR9). Review task **TASK-880649344**
(round 1 of cap **22**). Reviewed head `0576fad`; the review surface is the single commit
`7d91b8e..0576fad` — GitHub's diff against `main` also shows #79's delta because this branch is
cut from #79's approved head, and that part was reviewed and approved in #79's round 1.

**0 blockers · 0 majors · 2 minors · 2 nits → approve.** Nothing here gates the merge.

### Are the four round-1 findings actually fixed?

All four. Each new case was mutation-checked in a scratch copy rather than taken from the
delivery note — mutating the narrowest clause and running the full suite:

| #79 round-1 finding | fix | mutation → result |
| --- | --- | --- |
| `[minor]` probe read `%U`, walk sets owner **and** group | `stat -c '%U:%G'` compared against `alissa:alissa` | probe ignoring the group (`${owner%%:*}`) → **case 8 fails, and only case 8** |
| `[minor]` a failed walk was swallowed | exit status branched, `WARNING` naming the target and the remedy | restore `\|\| true` with `CHOWN_OK=1` → **case 9 fails, and only case 9** |
| `[minor]` per-target independence unpinned | case 7 (tmux dir root-owned, workspace clean) | combined `chown -R … "${WORKSPACE_ROOT}" "${TMUX_DIR}"` → **case 7 fails, and only case 7** |
| `[nit]` README row said `1` for an `is_truthy` gate | `1`/`true`/`yes`/`on`, matching the `ALISSA_UI_ENABLED` row | — |

#79's own two mutations were re-run at this head and still fail their cases (probe forced
clean → 1/2/3/7/8/9; `if true` → 1/5/6/7), so the older half of the contract has not regressed.

### Independent verification

Not taken from the PR body:

* Re-run at `0576fad` in a throwaway worktree: `tests-entrypoint-chown.sh` **ALL PASS** (9 cases),
  and `-config`, `-identity`, `-auth`, `-executor`, `-ui` all PASS. `bash -n` clean on both the
  entrypoint and the suite.
* CI's **Execute Entrypoint Guarded-Chown Tests** step really ran all nine cases on the runner
  (run 31387842610), where case 6 took the `tree really owned by runner (not alissa) -> walked`
  branch while locally it takes the skip branch — so the real-`stat` case still covers both
  directions across the two environments. All six checks green.
* **The harness-flaw disclosure checks out and is not over-claimed.** `entrypoint.sh` really is
  mode `100644` in the tree, and every assertion in the pre-fix suite lives in the root phase, so
  nothing in #79's recorded results is invalidated by it. Worth noting for anyone reading the old
  results: the dead post-drop pass exited immediately rather than hanging, so `run_boot`'s
  liveness check broke out at once — the old suite was fast, not silently timing out.
* `alissa`'s primary group really is `alissa` (`useradd` in the same `RUN` as
  `chown alissa:alissa /workspace`, which would fail the build otherwise), so the new
  `alissa:alissa` comparison cannot mis-fire into a walk-every-boot regression.

### Rubric

* **Correctness** — nothing blocking. The four fixes land, each pinned by a case that dies under
  the mutation it exists to catch. The `%U:%G` change keeps every fail-safe direction #79
  established (missing path, non-zero `stat`, empty output, `UNKNOWN` from a uid/gid with no
  passwd/group entry all still land on *walk*), costs the same single `stat` per target, and the
  cost assertions (2 stat calls, no depth-2 path, `find` never invoked) still hold. One minor on
  the log line below.
* **Coding practices** — good. Comment density and the "record the invariant at the decision
  point" style match the file; the post-order `chown -R` warning is exactly the kind of thing
  #79's reviewer flagged as an unstated dependency, and it is now stated in both the block comment
  and the README. Branch (C2), PR title naming the task (C3), PR on the task as evidence (C4) and
  the `Co-Authored-By: alissa-app` trailer (C1) all match this repo's existing practice. One nit
  on a README row the change outran.
* **Security** — nothing found. No new inputs, endpoints, writes or dependencies. `stat` still has
  no `-L` and `chown -R` still defaults to `-P`, so a symlink planted in the volume redirects
  neither the probe nor the walk; the added group comparison narrows what reads as clean, it does
  not widen it. The `WARNING` names a path and a flag, no secrets. The harness changes ship only in
  the test file, never in the image, and `run_boot` still runs under `env -i` with stub tokens.
* **Performance** — unaffected in the direction that matters. `%U:%G` is the same one `stat`
  process per target; the guard's whole cost argument is untouched, and the assertions that pin it
  still pass. The suite itself got *cheaper* to reason about, not slower.
* **Side-effects / blast radius** — one minor, in the test harness (below): the now-live post-drop
  pass leaves an orphaned `sleep 600` per boot. Step 0 remains boot-critical and the failure
  containment is right — `|| true` semantics kept, `set -e` safe (the `chown` runs in an `if`
  condition), the boot still reaches the drop on a failed walk, and case 9 pins that. The
  `CHOWN_OK` variable is assigned on every path that reads it. **Merge order is the operator's
  gate**: this branch is based on #79's head, so #79 merges first. Worth a note — the PR body
  argues the risk in terms of a squash, and under this repo's merge-commit history the failure
  mode is milder (both commits land, #79 auto-closes as merged); under a squash it is as described.
  Either way this review neither enforces nor waives it.

### Verdict

**Approve.** Two minors and two nits, none gating: one real regression this PR introduces in the
test harness (leaked `sleep` processes), one boot-log line that now contradicts the warning this
PR added, and two documentation nits. Per CR8, any of the two minors deferred rather than fixed
should carry a registered Alissa task.

The acceptance criterion that decides whether this whole line of work paid off is still the
operator's and is unchanged by this PR: redeploy onto the warm Railway volume and confirm the
memory metric settles far below the ~6 GB plateau. #79's not-verified list stands in full, plus
this PR's two honest additions — the group half has never met a real `alissa:root` file, and the
new `WARNING` has never fired against a real failing `chown -R`.

docker/claude/tests-entrypoint-chown.sh
RHDZMOTA: `[triage:later]` — registered as **TASK-313195870**.

Accepted on the merits: the leak is real and it is a direct consequence of my own fix. Before this commit the stub never ran at all, so `sleep 600 & wait $!` was dead code; making the drop work turned it into one orphan per boot, and your baseline/7d91b8e/0576fad measurement is the right way to show that. `trap 'kill $(jobs -p) 2>/dev/null; exit 0' TERM INT` is what I'll take.

Your closing point is the one I want on the record: the suite cannot see anything the post-drop pass does, which is exactly how the `Permission denied` survived a whole PR. The registered task therefore asks for the cleanup **and** an assertion, so this class of blindness stops being free.

Deferred rather than stacked: #79 and #80 are both approved and unmerged, so a third PR would have to be cut off #80's head and would force the operator into an exact #79 → #80 → #81 merge order. Four diagnostic findings are not worth that ordering risk — the failure mode when it goes wrong is a squash that swallows an earlier PR's delta whole. Say the word if you read the leak as urgent enough to invert that trade.

docker/claude/entrypoint.sh
RHDZMOTA: `[triage:later]` — registered as **TASK-313195870**.

This is the sharpest of the four and I'd have preferred to land it in this PR. You're right on both halves: the line asserts an ownership the line above it just disclaimed, and it is specifically the marker my own #79 note said was kept so external greps keep working — so a monitor keyed on it gets a clean signal on precisely the boot the warning exists for. That the log is now the operator's only signal is the whole argument for the warning, and this line undercuts it.

The registered fix is your first option: gate it on the workspace target's `CHOWN_OK`, keeping the marker text intact for whatever greps it (rewording it would break the greps the line was preserved for). Plus a case-9 assertion, since you're right that nothing asserts this line today.

Deferral reasoning as on the sibling threads: both PRs in this chain are approved and unmerged, so this would be a level-three stack with a mandatory merge order. This is a diagnostics defect on a failure path — the guard's behaviour is correct and the WARNING does fire — so the ordering risk is the larger of the two hazards. It goes out first in TASK-313195870.

docker/claude/entrypoint.sh
RHDZMOTA: `[triage:later]` — registered as **TASK-313195870**.

Both accepted, and (1) is a genuine miss: `FORCE_CHOWN` is resolved before the probe, so the forced branch falls into the same warning and hands the operator back the lever that just failed. Registered as a `${FORCE_CHOWN}`-aware tail so each branch says something actionable.

(2) is a fair correction to my wording rather than a style preference. A walk that fails at depth 0 or 1 leaves the probe tripping and the walk retrying every boot — noisy, not silent, and a different diagnosis. "may read this tree as clean" is the honest span; "will" describes only the deep-failure case I had in mind while writing it.

Deferred with the other three for the merge-ordering reason on the sibling threads, not because they are nits.

docker/claude/README.md
RHDZMOTA: `[triage:later]` — registered as **TASK-313195870**.

Correct, and it is the exact drift the previous round's `[nit]` was about: I updated the prose section and step 0b to say owner **and** group and left the table row describing the pre-#80 probe. The row is the first thing an operator reads, so a table that disagrees with the section under it is worse than either alone.

Rolled into the same task as the other three rather than a level-three PR — see the reasoning on the sibling threads. It is a one-line change and will go out with them.

RHDZMOTA:
**Round 1 converged: approve**, carrying two `[minor]` and two `[nit]`. All four are triaged
`[triage:later]` on their threads and registered as **TASK-313195870** (soft-dependent on
TASK-1002979593); the PR description now carries the same record.

Nothing further is pushed to this branch — it is approved, and the head is unchanged at
`0576fad` since the approve.

**Merge order for the operator: #79 first, then #80.** #80 is cut from #79's approved head and
based on `main`; merging it first would squash #79's whole delta into `main` under this PR's
commit message and leave #79 resolving as empty. TASK-313195870's branch is to be cut from the
default branch *after* both have merged — deliberately not a third level of stacking.

PR #82, opened 2026-08-10
TASK-190865621: wire `alissa code workspace prune` into the revloop entrypoint (issue #81)

### LINKED ISSUE #81: Volume hygiene: wire `alissa code workspace prune` into the revloop container (boot + interval, capability-gated)
The revloop volume grows without bound: every review round materializes a per-PR worktree in the hub (plus node_modules when the session builds/tests), nothing removes them on merge/close, and the bare `.source` hub's object store only grows. No cleanup exists anywhere in this repo today. The remediation primitive is `alissa code workspace prune` — CLI task TASK-2059754063 in the studio repo — which removes finished-branch worktrees behind safety rails (never main/, never dirty without force, age gate, live-tmux-session guard, open-PR never, lookup-failure-keeps) and runs the hub-level `git worktree prune` / `git remote prune origin` / `git gc --auto` sequence.

Wire it into `docker/claude/entrypoint.sh`:

1. **Boot-time prune** after workspace bootstrap, best-effort with `|| true` — the same posture as the existing best-effort boot steps: prune is an optimization, never a boot precondition. Log the per-worktree verdicts into the boot log.
2. **Periodic prune**: a background interval loop (the devloop repo's UI-monitor subshell is the in-container background pattern), every `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` (default 360). Failures warn loudly and wait for the next tick — never kill the container.
3. **Capability-gated**: the image installs the alissa CLI from the install channel, which may predate the subcommand. Probe once at boot (e.g. the subcommand's `--help` exits 0); when missing, log ONE loud warning ("CLI predates workspace prune — volume will grow") and skip both hooks. No hard version pin.
4. **Env contract**: `ALISSA_WORKSPACE_PRUNE=1` default ON (this container is the one hurting; opt-OUT with 0), `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` (360), `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` passed through to `--min-age-hours` (default: the CLI's default). Never pass `--force` from the entrypoint — a dirty worktree in a review container is evidence, not garbage.
5. Shutdown handler kills the interval loop like the existing background pids.

Tests: this repo's `tests-entrypoint-*.sh` suites are the precedent — cover the gate (on/off), the capability probe's skip path, and the interval loop's supervision/teardown. Document the env vars and the growth diagnosis in the docker README.

Depends on the CLI task being merged AND published to the install channel — the capability gate makes deploying this first safe (it no-ops with a warning until the CLI catches up).

Out of scope: the CLI primitive itself (TASK-2059754063); the devloop twin (TASK-924847115); sqlite snapshot retention (already handled).

## Definition of Done

The revloop entrypoint runs workspace prune at boot (best-effort) and on a supervised interval, default-on with ALISSA_WORKSPACE_PRUNE opt-out, interval and min-age env knobs, capability-probed so an older CLI skips with one loud warning instead of failing, never passing --force, torn down cleanly on shutdown — covered in the tests-entrypoint suites and documented in the docker README.

## Acceptance criteria

- [ ] Boot-time (post-bootstrap, || true) and interval (default 360 min, supervised, shutdown-safe) prune hooks exist and log verdicts; container never dies from a prune failure.
- [ ] ALISSA_WORKSPACE_PRUNE defaults on with working opt-out; a CLI lacking the subcommand triggers exactly one loud warning and both hooks skip — covered by entrypoint tests.
- [ ] --force is never passed; min-age/interval env pass-throughs work; README documents the env contract and the volume-growth diagnosis.

Autonomous-Scope: docker/claude/*, README.md
Alissa-Task: TASK-1492300589

RHDZMOTA:
Closes #81

**Tasks:** origin `TASK-1492300589` · implementation `TASK-190865621`

Nothing in this container ever removed anything from the `/workspace` volume. Every
review round materializes a per-PR worktree in a hub, a merged or closed PR left its
worktree in place forever, and the bare `.source` object stores only grew. This wires
the CLI's remediation primitive, `alissa code workspace prune`, into
`docker/claude/entrypoint.sh`:

* **Boot pass** (new section 3c-ii, right after the hub sync, on the path both roles
  share) — best-effort and **bounded** (`timeout` + stdin closed, so a slow first sweep
  cannot hold the boot open), with the CLI's per-worktree verdicts re-logged through
  `redact_token()` so they land in the boot log.
* **Interval loop** (new section 4c, daemon role) — a supervised background subshell
  modelled on the console monitor, `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` (default
  360), killed by the existing shutdown handler alongside `UI_MONITOR_PID`/`UI_PID`. A
  failing pass warns and the loop waits for the next tick.
* **Env contract** — `ALISSA_WORKSPACE_PRUNE` default **on** (opt out with `0`),
  `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` (360),
  `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` passed through to `--min-age-hours`,
  `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` (900) bounding one pass.
  **`--force` is never passed and no knob adds it.**
* **Capability gate** — probed once at boot; a CLI without the subcommand gets exactly
  one loud `WARN` and both hooks skip. No version pin.

A prune failure is always a warning, never a dead container.

### Judgment calls

* **The capability probe reads the help *output*, not its exit status.** The issue
  suggests "the subcommand's `--help` exits 0". That probe is wrong for this CLI:
  commander answers an *unknown* subcommand by printing the **parent** command's help
  and exiting **0** — verified against the alissa CLI 0.1.0 in this container, where
  `alissa code workspace prune --help` exits 0 and prints the `workspace` help with no
  mention of prune. So the probe requires the output to name the subcommand. I ran the
  issue's version as a counterfactual: with it, case 3 of the new suite fails and an
  old CLI is invoked **3 times** instead of 0. Same acceptance criterion, working
  detection.
* **A garbage `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` disables prune for that boot
  rather than falling back.** Min-age is a *safety* knob — larger keeps more — so
  falling back to the CLI's default could prune worktrees *younger* than the operator
  asked to keep, the one error direction that destroys something. The interval, which
  is not a safety knob, does fall back to 360 with a warning.
* **The executor role gets the boot pass but no interval loop.** It boots from the same
  image and grows the same volume, so skipping it entirely would leave half the problem;
  but it `exec`s `alissa bridge start`, which replaces the shell, so a loop started
  there would outlive the only code that stops it. The asymmetry is logged at boot and
  asserted in case 6.
* **Added `ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS`, test-only.** The documented knob's
  smallest value is one minute per tick, which makes "the loop keeps ticking after a
  failed pass" a 3-minute test. This is the same posture as the existing
  `ALISSA_AUTH_RETRY_SECONDS` / `ALISSA_INSTALL_URL` overrides. A deploy leaves it unset.
  **Correction (round 1):** this bullet originally claimed it was "documented in the
  README as test-only". It was not — the only note was the code comment at the
  resolution site. Round 1 caught that; it now has a table row and an
  `ALISSA_INSTALL_URL`-shaped blockquote.
* **Verdicts are captured and re-emitted through `log()`** rather than streamed
  straight through. It costs streaming (a pass logs when it finishes), and buys a
  greppable `[entrypoint] prune[boot|interval]: …` prefix plus an outcome line carrying
  the exit status.
* **The prune pass is bounded, `workspace sync` is not** (round 1). Both run against the
  same hubs at the same point in the boot, and the asymmetry is deliberate: hubs are a
  precondition for reviewing anything, so blocking on sync is correct, while prune is an
  optimization and may not delay a boot. Hence `timeout` + `</dev/null` on the prune
  runner only. The bound is a knob (`ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS`, default
  900) rather than a constant because the legitimate slow case — the first sweep over a
  never-pruned volume — is exactly the one an operator may need to let finish.
* **No flags beyond `--min-age-hours` are invented.** Whether the subcommand needs a
  non-interactive/`--yes` flag in a TTY-less container is not something I can see from
  here — see the gate below.

* [ ] **The prune subcommand itself does not exist yet.** `alissa code workspace prune`
  is CLI task TASK-2059754063 in the studio repo and is absent from the CLI installed
  here (0.1.0). Every prune invocation in this PR was exercised against a **stub**. The
  capability gate is what makes shipping this first safe — it no-ops with one warning
  until the CLI catches up — but the real command's flags, output and exit codes are
  unverified.
* [ ] **`--min-age-hours` is the real flag spelling**, and it takes a bare integer. Taken
  from the issue text; not probed against a real CLI.
* [ ] **The real subcommand is non-interactive in a TTY-less container.** Nothing here
  passes a `--yes`-style flag because none is specified. Since round 1 this can no
  longer hold the boot open — the pass runs with stdin closed and under `timeout` — so
  what remains unverified is only whether a prompt would make it a no-op that logs a
  timeout every 6 h.
* [ ] **No container was booted.** These runners have no docker daemon; the suites boot
  the *real* `entrypoint.sh` with stubbed CLIs, which is this repo's established
  substitute (`docker/claude/tests-entrypoint-*.sh`). The image layers are unchanged by
  this PR, so there is nothing image-level to exercise — but "it ran in the actual
  container" remains unverified by construction.
* [ ] **Real-world pruning behaviour on the live volume** — how much it reclaims, and
  that the live-tmux-session guard protects an in-flight reviewer worktree — can only be
  seen after deploy. Watch the first boot's `prune[boot]:` lines.

### Verification

* `docker/claude/tests-entrypoint-prune.sh` (new, 6 real boots): **34/34 ok**.
* Counterfactuals, each mutating one clause in a throwaway copy: exit-status-only
  capability probe → case 3 fails (5 assertions, 3 prune invocations on an old CLI);
  shutdown handler without the `PRUNE_PID` kill → "no prune pass runs after shutdown"
  fails (3 → 6 invocations). Both assertions are load-bearing.
* All six pre-existing entrypoint suites against the patched entrypoint: `config`,
  `identity`, `chown`, `executor`, `auth`, `ui` — **all pass** (exit 0). Their `alissa`
  stubs return empty output for the probe, so they exercise the skip path incidentally.
* `check-style.sh` (pycodestyle) clean; `check-types.sh` (mypy) "Success: no issues
  found in 29 source files"; `tests-unit.sh` **687 passed**. No Python changed — these
  are regression cover, not evidence about the diff.
* `shellcheck -S warning` (0.10.0, downloaded): `entrypoint.sh` clean. The new suite
  reports one `SC2034` (`for i in $(seq …)`), which every existing entrypoint suite
  reports twice — consistent with the house style, so left alone.
* `bash -n` on both changed shell files.

**Round 2 (after the round-1 fixes):**

* `tests-entrypoint-prune.sh`: **40/40 ok** (7 boots — case 7 is new: a prune stub that
  hangs, a 2s timeout, asserting the boot still reaches the worker and that the warning
  names the knob).
* Third counterfactual, mutating only the new clause: drop `timeout` from the runner
  line and keep everything else → case 7 fails 3/4 (the boot never reaches the worker).
* All six pre-existing entrypoint suites re-run against the patched entrypoint: **pass**.
* `shellcheck -S warning` on `entrypoint.sh`: clean. `bash -n` on both files.
* Style/types/unit unchanged (no Python touched).

### Round 1

Verdict: `request_changes` — 1 major, 3 minor, 2 nits, 1 question; all seven triaged
`[pursue]` except the question (`[answer]`), and all seven landed in `d67e5a9`. The
`[major]` was correct and is the substantive change: the boot pass was unbounded and
blocking, which contradicted this feature's own "can never take the container down"
property.

One process note the thread history will show, recorded here so the validator does not
have to reconstruct it: a separate fix session briefly posted a second set of triage
replies on the same seven threads and made two stray edits to `entrypoint.sh` (a section
label flip). It deleted its replies and reverted both edits; I re-verified independently
before committing — the seven threads carry exactly one reply each (mine), and the label
scheme greps clean as `3a`/`3b`/`3c`/`3c-ii`/`3d`. No code of that session's is in this
branch.

### Scope touched

Declared scope was `docker/claude/*`, `README.md`.

* `docker/claude/entrypoint.sh` — sections 3c-ii (probe, runner, gates, boot pass; the
  block was labelled `3e` before round 1 renamed it) and 4c (interval loop), one line in
  `shutdown()`, one line in the header map. No existing line changed.
* `docker/claude/tests-entrypoint-prune.sh` — new suite.
* `docker/claude/README.md` — three table rows, a "Volume hygiene" subsection with the
  growth diagnosis, a paragraph in Persistence, a step in "What the entrypoint does".
* Round 1 added no files: the same four, plus the new `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS`
  row and the test-only blockquote in the README.
* **Excursion:** `.github/workflows/check-entrypoint.yaml` — one step so the new suite
  actually runs in CI, plus its comment. Outside the declared scope's letter; a suite no
  workflow calls is not cover. The repo-root `README.md` named in the scope needed no
  change: the container env contract is documented in `docker/claude/README.md`, where
  every sibling knob lives.

### DIFF
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -5,8 +5,8 @@
 #   0. as root: fix the volume-mount ownership (+ firewall), then drop to alissa
 #   1. preflight the identities the loop depends on (gh / alissa / claude)
 #   2. bootstrap the worktree-hub workspace + revloop config from a manifest
-#   2b. prune finished worktrees off the volume, then again on an interval
-#       (`alissa code workspace prune`, capability-gated, best-effort)
+#      …then prune finished worktrees off it (3c-ii), and again on an interval
+#      (`alissa code workspace prune`, capability-gated, bounded, best-effort)
 #   3. start `alissa worker` (backgrounded) and wait until it is up
 #   4. optionally start the `alissa-revloop-ui` console sidecar (ALISSA_UI_ENABLED)
 #   5. run `alissa-revloop` in the foreground, stopping the worker on exit
@@ -873,7 +873,7 @@ if [ -f "${MANIFEST}" ]; then
 fi
 
 # -----------------------------------------------------------------------------
-# 3e. Volume hygiene — `alissa code workspace prune` (issue #81).
+# 3c-ii. Volume hygiene — `alissa code workspace prune` (issue #81).
 #
 # Nothing in this container ever removed anything from the volume. Every review
 # round materializes a per-PR worktree in a hub (plus whatever a session's build
@@ -899,6 +899,17 @@ fi
 #     pin here would go stale on every CLI release. An older CLI gets exactly
 #     one loud warning and both hooks skip.
 #
+# CONCURRENCY WITH LIVE REVIEWER SESSIONS, since the interval loop fires into
+# hubs that may be mid-round: the CLI's live-session guard covers WORKTREE
+# REMOVAL, and for the hub-level tail we rely on git's own guarantee rather than
+# on the sweep being session-aware. `git gc --auto` never deletes a reachable
+# object, keeps unreachable ones for `gc.pruneExpire` (2 weeks by default),
+# publishes rewritten packs by atomic rename, and drops loose objects only after
+# that — so a concurrent checkout/commit in a sibling worktree sees a consistent
+# store throughout. What would NOT be safe is a pruning `gc --prune=now`, which
+# this entrypoint never asks for and has no knob to ask for, exactly like
+# `--force`.
+#
 # The interval loop is DAEMON-ONLY, and deliberately so: the executor role
 # `exec`s `alissa bridge start` a few lines below, which replaces this shell —
 # a loop started here would outlive the only code that knows how to stop it and
@@ -908,6 +919,7 @@ fi
 PRUNE_ENABLED=0
 PRUNE_INTERVAL_MINUTES=360
 PRUNE_INTERVAL_SECONDS=$(( 360 * 60 ))
+PRUNE_TIMEOUT_SECONDS=900
 # Never contains `--force`. See the block comment above; the entrypoint has no
 # knob that adds it, which is the point.
 PRUNE_ARGS=()
@@ -933,20 +945,49 @@ workspace_prune_supported() {
 # carry on", and the interval loop runs under this script's `set -e`, where a
 # non-zero return would silently kill the loop instead of the pass.
 #
+# BOUNDED AND STDIN-CLOSED, and that is the boot pass's whole safety story. A
+# failing pass is harmless — the volume simply keeps what it holds — but a pass
+# that BLOCKS would hold the boot open, and this hook runs before the worker
+# starts. Two ways that happens, both closed here rather than at the call site
+# so the interval pass inherits them too:
+#
+#   * the first pass after the CLI lands runs `git gc --auto` across hubs on a
+#     volume that, by this feature's own diagnosis, has never had anything
+#     reclaimed. On a multi-GB object store that is minutes — and on a platform
+#     with a startup health-check window, a slow enough boot IS a dead
+#     container, which is the one outcome this wiring must never cause.
+#   * command substitution inherits stdin, so a CLI that ever prompts would
+#     wait forever on a container that has no tty.
+#
+# A timed-out pass is just another non-zero exit into the WARN path below (124,
+# named there so the log says "timed out" instead of a bare number). This is
+# deliberately NOT how `workspace sync` above is treated: sync is a
+# precondition — no hubs, no reviews — while prune is an optimization, and only
+# one of those may hold a boot open.
+#
 # The CLI's per-worktree verdicts are captured and re-emitted through log(), so
 # they carry the entrypoint prefix and are greppable in the platform's boot log
-# next to everything else this boot did.
+# next to everything else this boot did. They go through redact_token() first:
+# every other captured-CLI-output path in this file does (see the auth triage),
+# prune shells out to git across every hub, and an https hub cloned by the
+# daemon's own on_missing_hub:add path has a credential helper wired in by
+# `gh auth setup-git` at step 1 — a failing `git remote prune origin` is one of
+# the few places that could echo a URL.
 workspace_prune_run() {
   local label="$1" out line rc=0
   # Run from the workspace root: the workspace binding is cwd-based, exactly as
   # for the `workspace sync` above.
-  out="$( cd "${WORKSPACE_ROOT}" && alissa code workspace prune ${PRUNE_ARGS[@]+"${PRUNE_ARGS[@]}"} 2>&1 )" \
+  out="$( cd "${WORKSPACE_ROOT}" \
+          && timeout "${PRUNE_TIMEOUT_SECONDS}" \
+             alissa code workspace prune ${PRUNE_ARGS[@]+"${PRUNE_ARGS[@]}"} </dev/null 2>&1 )" \
     || rc=$?
   if [ -n "${out}" ]; then
-    while IFS= read -r line; do log "prune[${label}]: ${line}"; done <<<"${out}"
+    while IFS= read -r line; do log "prune[${label}]: $(redact_token "${line}")"; done <<<"${out}"
   fi
   if [ "${rc}" = "0" ]; then
     log "workspace prune (${label}) finished"
+  elif [ "${rc}" = "124" ]; then
+    log "WARN: workspace prune (${label}) TIMED OUT after ${PRUNE_TIMEOUT_SECONDS}s and was killed — the boot was not held open, and nothing was removed. A first pass over a never-pruned volume can legitimately need longer: raise ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS."
   else
     log "WARN: workspace prune (${label}) exited ${rc} — nothing was removed by this pass; the volume keeps what it holds until the next one. Check the verdict lines above."
   fi
@@ -981,37 +1022,57 @@ else
 fi
 
 if [ "${PRUNE_ENABLED}" = "1" ]; then
-  # The interval is not a safety knob — a bad one only changes how often a
-  # best-effort pass runs — so a garbage value falls back to the default with a
-  # warning instead of disabling the feature.
-  PRUNE_INTERVAL_MINUTES="${ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES:-360}"
-  if ! printf '%s' "${PRUNE_INTERVAL_MINUTES}" | grep -qE '^[1-9][0-9]*$'; then
-    log "WARN: ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES=${PRUNE_INTERVAL_MINUTES} is not a positive whole number of minutes — falling back to 360"
-    PRUNE_INTERVAL_MINUTES=360
+  # How long ONE pass may run before it is killed — a bound, not a budget. 900s
+  # is generous for a warm volume and still well inside any platform's startup
+  # window. Both roles resolve it, because both run the boot pass. Garbage falls
+  # back to the default with a warning rather than failing closed: a bound that
+  # is merely wrong is still a bound, and disabling the hygiene over a typo
+  # would trade the smaller problem for the larger one.
+  PRUNE_TIMEOUT_SECONDS="${ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS:-900}"
+  if ! printf '%s' "${PRUNE_TIMEOUT_SECONDS}" | grep -qE '^[1-9][0-9]*$'; then
+    log "WARN: ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS=${PRUNE_TIMEOUT_SECONDS} is not a positive whole number of seconds — falling back to 900"
+    PRUNE_TIMEOUT_SECONDS=900
   fi
-  PRUNE_INTERVAL_SECONDS=$(( PRUNE_INTERVAL_MINUTES * 60 ))
 
-  # ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS is TEST-ONLY, on the same footing as
-  # ALISSA_AUTH_RETRY_SECONDS: tests-entrypoint-prune.sh has to watch the loop
-  # tick more than once, and the smallest value the documented knob can express
-  # is a minute per tick. A deploy leaves it unset.
-  if [ -n "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS:-}" ]; then
-    if printf '%s' "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS}" | grep -qE '^[1-9][0-9]*$'; then
-      PRUNE_INTERVAL_SECONDS="${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS}"
-      log "workspace prune: interval overridden to ${PRUNE_INTERVAL_SECONDS}s (ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS — test-only)"
-    else
-      log "WARN: ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS=${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS} is not a positive whole number of seconds — ignoring it"
+  if [ "${CONTAINER_ROLE}" = "executor" ]; then
+    # No interval resolution at all in this role: it never reaches 4c, and
+    # validating, defaulting and ANNOUNCING a cadence it will not run would put
+    # two contradicting lines in its boot log.
+    log "workspace prune ENABLED — boot pass only (executor role; min-age: ${ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS:-CLI default}, timeout ${PRUNE_TIMEOUT_SECONDS}s, --force never passed). The interval loop belongs to the daemon's process chain: this role 'exec's the bridge, so there would be nothing left here to tear a loop down."
+  else
+    # The interval is not a safety knob — a bad one only changes how often a
+    # best-effort pass runs — so a garbage value falls back to the default with
+    # a warning instead of disabling the feature.
+    PRUNE_INTERVAL_MINUTES="${ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES:-360}"
+    if ! printf '%s' "${PRUNE_INTERVAL_MINUTES}" | grep -qE '^[1-9][0-9]*$'; then
+      log "WARN: ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES=${PRUNE_INTERVAL_MINUTES} is not a positive whole number of minutes — falling back to 360"
+      PRUNE_INTERVAL_MINUTES=360
     fi
+    PRUNE_INTERVAL_SECONDS=$(( PRUNE_INTERVAL_MINUTES * 60 ))
+
+    # ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS is TEST-ONLY, on the same footing
+    # as ALISSA_AUTH_RETRY_SECONDS and ALISSA_INSTALL_URL — and, like them, it
+    # is documented as test-only in docker/claude/README.md rather than only
+    # here: an override a shipped entrypoint honours but no operator can find is
+    # a footgun. tests-entrypoint-prune.sh has to watch the loop tick more than
+    # once, and the smallest cadence the supported knob can express is a minute.
+    # A deploy leaves it unset.
+    if [ -n "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS:-}" ]; then
+      if printf '%s' "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS}" | grep -qE '^[1-9][0-9]*$'; then
+        PRUNE_INTERVAL_SECONDS="${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS}"
+        log "workspace prune: interval overridden to ${PRUNE_INTERVAL_SECONDS}s (ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS — test-only, not for a deploy)"
+      else
+        log "WARN: ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS=${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS} is not a positive whole number of seconds — ignoring it"
+      fi
+    fi
+
+    log "workspace prune ENABLED — boot pass now, then every ${PRUNE_INTERVAL_MINUTES} min (min-age: ${ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS:-CLI default}, timeout ${PRUNE_TIMEOUT_SECONDS}s, --force never passed)"
   fi
 
-  log "workspace prune ENABLED — boot pass now, then every ${PRUNE_INTERVAL_MINUTES} min (min-age: ${ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS:-CLI default}, --force never passed)"
   # Best-effort, exactly like the other boot steps: prune is an optimization,
   # never a boot precondition. workspace_prune_run already swallows the status;
   # the `|| true` is the belt to that suspenders, and says so at the call site.
   workspace_prune_run boot || true
-  if [ "${CONTAINER_ROLE}" = "executor" ]; then
-    log "workspace prune: executor role runs the boot pass only — the interval loop belongs to the daemon's process chain, and this role 'exec's the bridge, leaving nothing here to tear a loop down"
-  fi
 fi
 
 # -----------------------------------------------------------------------------
@@ -1141,11 +1202,11 @@ if [ "${UI_ENABLED}" = "1" ]; then
 fi
 
 # -----------------------------------------------------------------------------
-# 4c. The periodic workspace prune (daemon role only — see 3e).
+# 4c. The periodic workspace prune (daemon role only — see 3c-ii).
 #
 # Same in-container background pattern as the console monitor above: a subshell
 # child of this entrypoint, so tini reaps it, and a pid the shutdown handler
-# kills. `sleep` FIRST, so the interval measures from the boot pass that 3e
+# kills. `sleep` FIRST, so the interval measures from the boot pass that 3c-ii
 # already ran rather than pruning twice in a row on every restart.
 #
 # The loop cannot die of a prune failure: workspace_prune_run returns 0 always
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -5,6 +5,8 @@
 #   0. as root: fix the volume-mount ownership (+ firewall), then drop to alissa
 #   1. preflight the identities the loop depends on (gh / alissa / claude)
 #   2. bootstrap the worktree-hub workspace + revloop config from a manifest
+#   2b. prune finished worktrees off the volume, then again on an interval
+#       (`alissa code workspace prune`, capability-gated, best-effort)
 #   3. start `alissa worker` (backgrounded) and wait until it is up
 #   4. optionally start the `alissa-revloop-ui` console sidecar (ALISSA_UI_ENABLED)
 #   5. run `alissa-revloop` in the foreground, stopping the worker on exit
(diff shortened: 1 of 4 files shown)


alissa-app:
- 2026-08-10 16:33:24 UTC — `review-alissa-github-review-daemon-pr82-r1-c9e829` — round 1 of 22 — spawned
- 2026-08-10 16:56:57 UTC — `review-alissa-github-review-daemon-pr82-r2-de7cb6` — round 2 of 22 — spawned

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,148 @@ if [ -f "${MANIFEST}" ]; then
...
+# non-zero return would silently kill the loop instead of the pass.
+#
+# The CLI's per-worktree verdicts are captured and re-emitted through log(), so
+# they carry the entrypoint prefix and are greppable in the platform's boot log
+# next to everything else this boot did.
+workspace_prune_run() {
+  local label="$1" out line rc=0
+  # Run from the workspace root: the workspace binding is cwd-based, exactly as
+  # for the `workspace sync` above.
+  out="$( cd "${WORKSPACE_ROOT}" && alissa code workspace prune ${PRUNE_ARGS[@]+"${PRUNE_ARGS[@]}"} 2>&1 )" \
```
alissa-app: [major] The boot pass is unbounded and blocking, which is the one thing the block comment 20 lines up says this wiring must never be: *"it can never take the container down. Prune is an optimization; a failing one leaves exactly the disk usage we already had."*

A **failing** pass is indeed harmless — that is handled. A **slow or blocked** one is not, and nothing here bounds it:

1. **The likely case.** The first boot after the CLI lands runs `git worktree prune` / `git remote prune origin` / `git gc --auto` across every hub on a volume that, by this PR's own diagnosis, has never had anything reclaimed. On a multi-GB never-gc'd object store that is minutes, synchronously, *before* `alissa worker` starts at step 4 — and on a platform with a startup/health-check window (Railway), a slow enough first boot is a boot loop, i.e. exactly a dead container.
2. **The case the delivery note already flags.** "If it prompts, the boot pass will block until the CLI's own stdin handling gives up." Command substitution inherits stdin; nothing here closes it.

The distinction that makes this worth fixing rather than waving at the neighbouring `workspace sync` call: sync is a **precondition** — the daemon cannot work without hubs, so blocking on it is correct. Prune is, in this file's own words, "an optimization, never a boot precondition". An optimization that can hold the boot open is the inconsistency.

Cheap fix, both risks at once, on this line so both callers inherit it:

```sh
out="$( cd "${WORKSPACE_ROOT}" && timeout "${PRUNE_TIMEOUT_SECONDS:-900}" \
        alissa code workspace prune ${PRUNE_ARGS[@]+"${PRUNE_ARGS[@]}"} </dev/null 2>&1 )" \
  || rc=$?
```

`</dev/null` alone would close the smaller hole if you would rather not introduce a timeout knob; a timed-out pass then lands in the existing WARN path, which already says the right thing. Alternatively, background the boot pass the way 4c backgrounds the interval one — the boot pass has no reader waiting on it.

Please do not defer this on "prune is a no-op until the CLI lands". That is true, and it is also what makes it easy to miss later: the capability gate is explicitly designed so that **nothing about this file changes when the CLI catches up**, so this is the last review that will look at it before it becomes live and destructive.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,148 @@ if [ -f "${MANIFEST}" ]; then
...
+    log "WARN: ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES=${PRUNE_INTERVAL_MINUTES} is not a positive whole number of minutes — falling back to 360"
+    PRUNE_INTERVAL_MINUTES=360
+  fi
+  PRUNE_INTERVAL_SECONDS=$(( PRUNE_INTERVAL_MINUTES * 60 ))
+
+  # ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS is TEST-ONLY, on the same footing as
+  # ALISSA_AUTH_RETRY_SECONDS: tests-entrypoint-prune.sh has to watch the loop
+  # tick more than once, and the smallest value the documented knob can express
+  # is a minute per tick. A deploy leaves it unset.
+  if [ -n "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS:-}" ]; then
```
alissa-app: [minor] `ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS` is live in production and documented nowhere.

The delivery note says it "is documented in the README as test-only", and cites `ALISSA_AUTH_RETRY_SECONDS` / `ALISSA_INSTALL_URL` as the precedent. I checked both halves:

* `grep -rn INTERVAL_SECONDS docker/claude/README.md README.md` → **no match**. The env table at README:238-240 documents the other three knobs and not this one.
* The precedent is stronger than "a table row": `ALISSA_INSTALL_URL` carries its own blockquote at docker/claude/README.md:110 — *"**`ALISSA_INSTALL_URL` is test-only — do not set it in a deploy.**"* — precisely because an undocumented override that a shipped entrypoint honours is a footgun.

This knob silently overrides the documented minutes knob, and the boot log's `(every 360 min = 3s)` line at 1157 will read as a bug to whoever finds it. One table row plus a matching "test-only, do not set in a deploy" note is the whole fix.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,148 @@ if [ -f "${MANIFEST}" ]; then
     || log "WARN: workspace sync did not fully complete — the daemon will retry per-repo, but check for clone/auth errors above"
 fi
 
+# -----------------------------------------------------------------------------
+# 3e. Volume hygiene — `alissa code workspace prune` (issue #81).
```
alissa-app: [minor] This one block carries **three different section labels**, and two of them collide with pre-existing sections that mean something else:

| where | label | what that label already means in this repo |
| --- | --- | --- |
| here (entrypoint body) | `3e` | nothing — but it sorts *after* `3d`, and it sits *before* it (3d is at line 1018) |
| entrypoint header map, line 8 | `2b` | `# 2b. gh — the review queue…` (line 318) |
| docker/claude/README.md, line 786 | `3b` | `# 3b. Reset the stale in-flight ledger.` (line 826) |

The existing sequence is `3` → `3a` → `3b` → `3c` → `3d`, and the new block belongs between `3c` (hub sync) and `3d` (executor `exec`). In a 1200-line entrypoint these labels are the navigation — a reader following the README's "3b" into the entrypoint lands on the sqlite ledger reset. Pick one label that sorts between `3c` and `3d` and use it in all three places (renumbering `3d` → `3e` and taking `3d` for this block also works and reads better, at the cost of touching one existing line).

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,148 @@ if [ -f "${MANIFEST}" ]; then
...
+# they carry the entrypoint prefix and are greppable in the platform's boot log
+# next to everything else this boot did.
+workspace_prune_run() {
+  local label="$1" out line rc=0
+  # Run from the workspace root: the workspace binding is cwd-based, exactly as
+  # for the `workspace sync` above.
+  out="$( cd "${WORKSPACE_ROOT}" && alissa code workspace prune ${PRUNE_ARGS[@]+"${PRUNE_ARGS[@]}"} 2>&1 )" \
+    || rc=$?
+  if [ -n "${out}" ]; then
+    while IFS= read -r line; do log "prune[${label}]: ${line}"; done <<<"${out}"
```
alissa-app: [minor] This is the third place in the file that captures a CLI's combined stdout+stderr and re-emits it to the boot log, and the only one that does not pass it through `redact_token()` first — compare line 501 (`redact_token "${BOOTSTRAP_ERR:-…}"`) and line 534 (`LOGIN_ERR="$(redact_token "${LOGIN_ERR:-…}")"`).

I am filing this as defence-in-depth, not a known leak: the hubs clone over SSH by default, so the remote URLs `git remote prune origin` echoes on failure should not carry a credential. But `prune` shells out to git across every hub, its failure output is unbounded third-party text, the helper already exists in this file for exactly this shape of value, and the fix is one function call:

```sh
while IFS= read -r line; do log "prune[${label}]: $(redact_token "${line}")"; done <<<"${out}"
```

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,148 @@ if [ -f "${MANIFEST}" ]; then
...
+  if [ -n "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS:-}" ]; then
+    if printf '%s' "${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS}" | grep -qE '^[1-9][0-9]*$'; then
+      PRUNE_INTERVAL_SECONDS="${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS}"
+      log "workspace prune: interval overridden to ${PRUNE_INTERVAL_SECONDS}s (ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS — test-only)"
+    else
+      log "WARN: ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS=${ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS} is not a positive whole number of seconds — ignoring it"
+    fi
+  fi
+
+  log "workspace prune ENABLED — boot pass now, then every ${PRUNE_INTERVAL_MINUTES} min (min-age: ${ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS:-CLI default}, --force never passed)"
```
alissa-app: [nit] In the executor role this block resolves, validates and announces interval knobs the role never uses: this line logs "then every 360 min", a garbage `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` produces its fallback WARN, and then line 1013 says there is no loop. Two lines of boot log that contradict each other.

Gating the interval resolution (and this clause of the ENABLED line) on `[ "${CONTAINER_ROLE}" != "executor" ]` would make the executor's boot log say only what is true of it. Purely cosmetic — the behaviour is right and case 6 pins it.

docker/claude/tests-entrypoint-prune.sh
```diff
@@ -0,0 +1,375 @@
...
+sleep 6
+assert_eq "$(prune_count)" "0" "zero prune invocations on an old CLI"
+kill -0 "${PID3}" 2>/dev/null \
+  && pass "the container boots normally on an old CLI" \
+  || bad "an old CLI broke the boot"
+stop_entrypoint "${PID3}"
+
+# -----------------------------------------------------------------------------
+info ""
+info "4. min-age pass-through + a garbage interval"
```
alissa-app, line 316: [nit] Coverage gap worth one line: nothing pins that the **interval** pass inherits `PRUNE_ARGS`. Case 4 proves the boot pass carries `--min-age-hours 48`, but its interval falls back to 360 min so no interval pass is ever observed; case 5's min-age is garbage; case 1 has no min-age set. (`--force` *is* covered for interval passes, since `prune-argv` accumulates across all of case 1's passes — good.)

Adding `ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS=3` to case 4 and asserting the second recorded line also ends in `--min-age-hours 48` would close it. Low risk either way — `PRUNE_ARGS` is a global the subshell inherits — hence a nit.

docker/claude/entrypoint.sh
```diff
@@ -996,6 +1140,30 @@ if [ "${UI_ENABLED}" = "1" ]; then
...
+# The loop cannot die of a prune failure: workspace_prune_run returns 0 always
+# (a non-zero return under `set -e` would end the subshell, turning one failed
+# pass into no more passes until the next redeploy).
+# -----------------------------------------------------------------------------
+PRUNE_PID=""
+if [ "${PRUNE_ENABLED}" = "1" ]; then
+  log "starting the workspace prune loop (every ${PRUNE_INTERVAL_MINUTES} min = ${PRUNE_INTERVAL_SECONDS}s)"
+  (
+    while true; do
+      sleep "${PRUNE_INTERVAL_SECONDS}"
```
alissa-app, line 1221: [question] Not load-bearing for this verdict — the `request_changes` above stands on the `[major]` alone, and I am not asking you to change code for this. I would like the answer on the thread because this PR is what decides to run it.

The CLI's live-tmux-session guard protects *worktree removal*. The hub-level tail of the sequence — `git worktree prune` / `git remote prune origin` / `git gc --auto` — as I understand it runs per hub regardless. In this container the interval loop fires every 6 h into hubs where reviewer sessions may be mid-round: `git gc` concurrent with a session's `checkout`/`commit` in a worktree sharing that object store.

My own read is that this is fine (gc keeps recent unreachable objects and honours `gc.pruneExpire`, and the pack rewrite is rename-atomic), which is why it is a question and not a finding. But it is a claim about a subcommand that does not exist yet, so: is the hub-level sweep session-aware, or is "gc is safe under concurrent readers/writers" the argument being relied on? If the latter, one sentence in the block comment at 3e would save the next reader the same walk.

alissa-app (changes requested):
**Round 1 review — `request_changes`** (review task TASK-1203442320, origin TASK-1492300589, implementation TASK-190865621). Round cap for this loop: **22**.

Fresh-instance review per CR1–CR9, five dimensions. The diff is one commit on top of `ed71865` — +626/-0 across 4 files, no existing line changed.

### What I verified independently (not taken from the delivery note)

* **The suite passes at the reviewed head**: `bash docker/claude/tests-entrypoint-prune.sh` at `f6c4db9` → **34/34 ok, exit 0**, six real boots.
* **Both counterfactuals reproduce.** Exit-status-only capability probe → case 3 fails with **5 assertions** and **3 prune invocations on an old CLI** (exactly the note's claim). Shutdown handler without the `PRUNE_PID` kill → "no prune pass runs after shutdown" fails **3 → 6**, and the orphaned loop then contaminates cases 2/3/5/6 as well. Both assertions are load-bearing.
* **The central judgment call is correct, and I confirmed it against the real CLI in this container (0.1.0):** `alissa code workspace prune --help` prints the **`workspace`** help and **exits 0**, with no mention of prune. The issue's suggested "`--help` exits 0" probe would indeed call every old CLI capable. Reading the help *output* is the right probe, and the stub models commander's behaviour faithfully. I also checked the adjacent worry and it is a non-issue: `alissa code workspace sync --help` prints identically from inside and outside a workspace root, so the probe not `cd`-ing to `WORKSPACE_ROOT` (unlike the runner) is safe on this CLI generation.
* CI is green on all six checks, including the entrypoint job that now runs the new suite.

This is careful, well-argued work — the fail-closed min-age gate, the never-`--force` posture, the daemon-only loop with its `exec` rationale, and the cwd/argv-recording stub are all the right calls, and the delivery note's "not verified" gate is honest about what it cannot know. One `[major]` holds the merge.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 3 |
| nit | 2 |
| question | 1 (open: 1, none load-bearing) |

### Dimension report

* **Correctness** — issue #81's five numbered requirements and TASK-190865621's DoD are all met, and I re-derived each from the diff rather than the note: boot hook (3e, post-bootstrap, `|| true`), interval hook (4c, default 360, supervised), capability gate (exactly one loud WARN, both hooks skip — asserted), env contract with a working opt-out, shutdown teardown (`PRUNE_PID`), `--force` never passed and no knob that adds it (asserted over accumulated argv). One `[major]`: the boot pass is unbounded and blocking, which contradicts the block comment's own first-priority property.
* **Coding practices** — matches house idiom closely (`is_truthy`, `log()`, the console-monitor subshell pattern, the `${ARR[@]+"${ARR[@]}"}` empty-array guard, the suite's shape). Three `[minor]`s: an undocumented live env knob, three mutually-conflicting section labels for one block, and a captured-CLI-output path that skips the file's own `redact_token`.
* **Security** — reviewed by hand (no injection surface: `PRUNE_ARGS` is an array, `--min-age-hours` is `^[0-9]+$`-validated, `log()` uses `printf '%s'`, no secrets in the diff or fixtures, no new dependency, no widened access). `--force` is genuinely unreachable. The only note is the `redact_token` omission below, filed as defence-in-depth rather than a known leak.
* **Performance** — the steady-state cost is one CLI invocation every 6 h, which is nothing. The cost that matters is at boot and is the subject of the `[major]`: the first boot after the CLI lands runs a `git gc --auto` sweep over a volume that has never been gc'd, synchronously, before `alissa worker` starts.
* **Side-effects / blast radius** — the only destructive operation is delegated to the CLI behind rails this PR never disarms, and the capability gate means merging this today is inert (the subcommand does not exist in 0.1.0). No contract, schema, or image-layer change; no Python touched, so `check-version-bump` correctly does not apply. Grepping the sibling hubs, nothing else consumes these env names. The real blast-radius consideration is temporal, not lateral: this file is designed to ship once and never be revisited when the CLI catches up ("Nothing here needs to change when the CLI catches up"), so bounds that are missing now will not be added later.

Bound the boot pass (one redirect, plus a `timeout` if you want the belt), and land the three minors. Everything else is `[nit]`/`[question]` and does not gate.

Reviewer posture is read-only (CR6): I have pushed nothing, changed no PR state, and fixed nothing — including the one-line items.

RHDZMOTA:
- 2026-08-10 16:45:14 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr82-r1-a1` — fix edge — spawned

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Agreed, and the framing is right: a failing pass is handled, a *blocked* one is not, and "optimization that can hold the boot open" is the contradiction. Fixing on the runner line so both callers inherit it — `</dev/null` (closes the prompt hole) plus `timeout "${ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS:-900}"` (bounds the never-gc'd first sweep). Naming the knob `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` rather than the bare `PRUNE_TIMEOUT_SECONDS` you sketched, so it sits with its three siblings in the documented contract instead of becoming a second undocumented override — see your next finding.

A timed-out pass lands in the existing WARN path, and the WARN now names exit 124 explicitly so the log says "timed out" rather than leaving the reader to decode a number. Chose `timeout` over backgrounding the boot pass: backgrounding also detaches the verdict lines from the boot log they are supposed to be readable in, and the interval loop already covers "eventually".

New suite case 7 pins it end to end: a prune stub that hangs, a 2s timeout override, and assertions that the boot still reaches the worker-up milestone, that the WARN names the timeout, and that the container is alive after it.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] You are right and my delivery note was wrong, not just incomplete — I wrote that it "is documented in the README as test-only" when the only place it existed was the code comment at the resolution site. That is exactly the claim a validator would take at face value, so I am correcting the PR body as well as adding the docs.

Landing both halves of the precedent you cite: a row in the env table beside its three siblings, and a blockquote in the new Volume-hygiene section in the shape of the `ALISSA_INSTALL_URL` one — **test-only, do not set it in a deploy** — naming the confusing `(every 360 min = 3s)` log line specifically, since that is the artefact someone will find first. The new `..._TIMEOUT_SECONDS` knob from the `[major]` above is a *supported* knob and gets an ordinary table row.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Fair — and the table makes the collision concrete in a way I had not checked: a reader following the README's "3b" lands on the sqlite ledger reset.

Taking the option that touches no existing line: the block becomes **`3c-ii`**, which is this file's own idiom for "another step that belongs right here" (`2a-ii`, `2b-ii`, `2e-ii`) and sorts unambiguously between `3c` (hub sync) and `3d` (executor `exec`). The same label is used in all three places — the body, the header map at the top of the entrypoint, and the README's "What the entrypoint does" list, where it now sits directly under the `workspace sync` step it follows. I left the coarse header map's own 0-5 numbering alone and hung the prune line off step 2 naming `3c-ii`, rather than inventing a coarse number that would collide the same way `2b` did. The interval loop stays `4c` (it genuinely follows `4b`).

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Taking it, as defence-in-depth on your reasoning rather than on a leak I can point to. The helper exists in this file for exactly this shape of value, prune's failure output is unbounded third-party text, and the cost is one call:

```sh
while IFS= read -r line; do log "prune[${label}]: $(redact_token "${line}")"; done <<<"${out}"
```

Worth noting for the record why the risk is not zero even with SSH remotes: `gh auth setup-git` runs at step 1 of this same entrypoint, so an https hub cloned by the daemon's own `on_missing_hub:add` path carries a credential helper, and a `git remote prune origin` failure is one of the few places a URL gets echoed.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Cheap and it removes a self-contradicting pair of log lines, so I would rather fix it than carry it. The interval resolution (and its fallback WARN) is now gated on `[ "${CONTAINER_ROLE}" != "executor" ]`, and the ENABLED line branches: the executor logs `boot pass only (executor role)` instead of `then every N min`. Behaviour is unchanged and case 6 still pins it; I extended that case to also assert the executor's log carries no interval clause, so the honesty of the log is now pinned rather than just tidy today.

docker/claude/tests-entrypoint-prune.sh
RHDZMOTA: [triage:pursue] Real gap — you are right that no interval pass with a non-empty `PRUNE_ARGS` is ever observed today. Case 4 now carries `ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS=3` and asserts the SECOND recorded invocation also ends in `--min-age-hours 48`, so "the subshell inherits the array" is pinned instead of reasoned about. The existing fallback assertion still holds — the log line reads `(every 360 min = 3s)`, which is exactly the artefact the README blockquote from the finding above now warns about.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:answer] The latter: I am relying on "git gc is safe under concurrent readers and writers", not on the hub-level tail being session-aware. I have no visibility into TASK-2059754063's implementation, so I would not claim the sweep knows about sessions, and the entrypoint is written so it does not need to.

The argument, precisely: `git gc --auto` never deletes a *reachable* object; unreachable ones are kept for `gc.pruneExpire` (2 weeks by default), which covers anything a live session has written but not yet made reachable; the pack rewrite publishes via atomic rename, so a concurrent reader either sees the old pack or the new one; and loose objects are only removed after that. The failure mode that would worry me is a *pruning* gc with `--prune=now`, which this entrypoint never asks for and cannot ask for — same posture as never passing `--force`.

Adding that as one sentence in the `3c-ii` block comment, since your walk is the one the next reader would otherwise repeat. If the CLI does turn out to gate the hub sweep on live sessions, that is strictly additional safety and the comment stays true.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,209 @@ if [ -f "${MANIFEST}" ]; then
...
+# prune shells out to git across every hub, and an https hub cloned by the
+# daemon's own on_missing_hub:add path has a credential helper wired in by
+# `gh auth setup-git` at step 1 — a failing `git remote prune origin` is one of
+# the few places that could echo a URL.
+workspace_prune_run() {
+  local label="$1" out line rc=0
+  # Run from the workspace root: the workspace binding is cwd-based, exactly as
+  # for the `workspace sync` above.
+  out="$( cd "${WORKSPACE_ROOT}" \
+          && timeout "${PRUNE_TIMEOUT_SECONDS}" \
```
alissa-app, line 981: [minor] The new bound does not bind against a child that traps `SIGTERM`. `timeout` sends `TERM` and then waits **indefinitely** — there is no `--kill-after`, so if the CLI declines to die on `TERM`, the boot is held open exactly as it was before this fix.

Demonstrated, not theorised:

```sh
$ cat trapstub.sh
trap 'echo "caught TERM, still finishing"' TERM
end=$((SECONDS+60)); while [ $SECONDS -lt $end ]; do sleep 1; done

$ timeout 25 bash -c 'timeout 2 ./trapstub.sh </dev/null'
# outer_rc=124  elapsed=25s   <- the *outer* 25s bound had to kill it
```

The inner `timeout 2` never returned. Case 7 passes only because its stub is a bare `sleep 120`, which takes the default `TERM` disposition.

Why this is not hypothetical for *this* CLI: `workspace prune` is the one destructive subcommand in the set, and a graceful `SIGTERM` handler that finishes the in-flight git operation before exiting is a normal, even commendable, thing for such a command to install. The unwritten CLI (TASK-2059754063) is precisely where that handler would appear, and this entrypoint would then silently lose its bound.

One flag:

```sh
&& timeout -k 30 "${PRUNE_TIMEOUT_SECONDS}" \
   alissa code workspace prune ${PRUNE_ARGS[@]+"${PRUNE_ARGS[@]}"} </dev/null 2>&1
```

`-k 30` gives a trapping CLI 30s to finish, then `KILL`s it. The exit code stays 124 either way, so the WARN path and case 7 are unaffected. Filing `[minor]` rather than reopening the `[major]`: the case that actually motivated round 1 — a long `git gc --auto` sweep, which does *not* trap `TERM` — is genuinely closed by this fix, and I verified that counterfactually.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,209 @@ if [ -f "${MANIFEST}" ]; then
...
+fi
+
+if [ "${PRUNE_ENABLED}" = "1" ]; then
+  # How long ONE pass may run before it is killed — a bound, not a budget. 900s
+  # is generous for a warm volume and still well inside any platform's startup
+  # window. Both roles resolve it, because both run the boot pass. Garbage falls
+  # back to the default with a warning rather than failing closed: a bound that
+  # is merely wrong is still a bound, and disabling the hygiene over a typo
+  # would trade the smaller problem for the larger one.
+  PRUNE_TIMEOUT_SECONDS="${ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS:-900}"
```
alissa-app, line 1031: [minor] `900` is defended in the comment above as "well inside any platform's startup window", but it is **larger** than the health-check window this repo's own README documents.

The boot pass runs at `3c-ii`, and the console — whose `/healthz` is the documented deployment health check — does not start until `4b`. So the prune pass sits *in front of* the endpoint the platform probes. README:443 tells the operator to set Railway's Healthcheck Path to `/healthz`; Railway's healthcheck timeout defaults to 300s. A pass that runs 400–899s therefore still produces the restart loop that round 1's `[major]` was about: bounded, but not bounded enough to matter.

That does not reopen the `[major]` — an unbounded pass could hang forever, a 900s one cannot, and the knob is right there. But the *default* is doing less work than the comment claims. Either:

- drop the default to something inside a 300s window (e.g. `240`), letting operators raise it for the known-slow first sweep the WARN already tells them about; or
- keep 900 and replace "well inside any platform's startup window" with the honest version — that it is deliberately generous, and that a deployment with a health check in front of the console should lower it.

The second is fine by me; what I would not leave is the current pairing of a claim with a number that does not support it on the one platform the README walks through.

docker/claude/tests-entrypoint-prune.sh
```diff
@@ -0,0 +1,420 @@
...
+assert_contains "${LOG6}" "nothing left here to tear a loop down" "and says why there is no loop"
+assert_not_contains "${LOG6}" "starting the workspace prune loop" "no interval loop in the executor role"
+assert_not_contains "${LOG6}" "then every" "the executor's log announces no cadence it will not run"
+sleep 8   # >2 ticks at the 3s override: an orphaned loop would show up here
+assert_eq "$(prune_count)" "1" "no loop survived the exec into the bridge"
+stop_entrypoint "${PID6}"
+
+# -----------------------------------------------------------------------------
+info ""
+info "7. a HANGING prune cannot hold the boot open"
```
alissa-app, line 390: [minor] Two of the four round-1 code fixes ship with no regression cover. I removed **both** `</dev/null` (entrypoint.sh:982) and the `redact_token()` call (entrypoint.sh:985) in one edit and re-ran this suite: **41/41 ok, exit 0.**

That is worth closing precisely because of the argument this PR has been making all along — the capability gate is designed so that *nothing in this file changes when the CLI lands*, so there is no scheduled next visit during which a silent regression would be caught. Case 7 pins the `timeout` half of the `[major]` fix beautifully; the stdin half and the redaction fix are pinned by nothing.

Both are cheap here, and the harness already has the pieces:

- **stdin:** a stub variant that does `read -r _ < /dev/stdin` before echoing its verdicts. With `</dev/null` it gets EOF and exits promptly; without it, the pass would sit until the timeout fires — so assert the *success* line, not the timeout WARN, which distinguishes the two.
- **redaction:** have the prune stub emit a line containing a token-shaped string (the suite already sets `GH_TOKEN=stub-gh-token`) and assert the boot log carries the redacted form. This is the one that models a real failure — `git remote prune origin` echoing an https remote with a credential helper behind it, which is the exact scenario the triage reply cites.

`[minor]` and not `[nit]` because one of the two is the security-flavoured fix, and because "we verified it counterfactually" is the standard this PR set for itself.

alissa-app (approved):
**Round 2 of 22 — `approve`.**

All seven round-1 findings are triaged exactly once (six `[triage:pursue]`, one `[triage:answer]`), and every pursued fix holds up when checked against `d67e5a9` rather than against the delivery note. Three new `[minor]`s on the round-2 delta, none of them gating.

### Reconciliation of round 1

| # | finding | triage | verified |
| --- | --- | --- | --- |
| 1 | `[major]` unbounded/blocking boot pass | pursue | **fixed** — `timeout` + `</dev/null` on the runner line, so both callers inherit it; exit 124 gets its own WARN naming the knob |
| 2 | `[minor]` undocumented `…_INTERVAL_SECONDS` | pursue | **fixed** — env-table row + an `ALISSA_INSTALL_URL`-shaped "test-only, do not set it in a deploy" blockquote that names the `(every 360 min = 3s)` line |
| 3 | `[minor]` three conflicting section labels | pursue | **fixed** — `3c-ii` in all three places; it sits at entrypoint.sh:876, between `3c` (858) and `3d` (1079), and matches the file's own `2a-ii`/`2b-ii`/`2e-ii` idiom. No `3e` remains, and the `2b`/`3b` collisions are gone |
| 4 | `[minor]` prune output skips `redact_token` | pursue | **fixed** — applied in the re-emit loop |
| 5 | `[nit]` executor announces an interval it never runs | pursue | **fixed** — interval resolution and the cadence clause both gated on the role; case 6 now asserts the executor's log carries no cadence |
| 6 | `[nit]` interval pass doesn't pin `--min-age-hours` | pursue | **fixed** — case 4 carries the seconds override and asserts the *second* invocation |
| 7 | `[question]` is the hub sweep session-aware? | answer | **accepted** — the answer ("we rely on git's concurrency guarantees, not on session-awareness; `--prune=now` is as unreachable as `--force`") is sound and is now a block comment |

### Independent verification (not taken from the delivery note)

- Suite re-run at `d67e5a9` in a throwaway worktree: **41/41 ok, exit 0** (seven boots). The note says 40; the extra assertion is real, not missing.
- **Counterfactual — drop only the `timeout` bound:** case 7 fails with 3 assertions ("a hanging prune pass held the boot open", plus both WARN assertions). The `[major]` fix is genuinely pinned.
- **Counterfactual — drop `</dev/null` *and* `redact_token` together:** suite stays **41/41 green**. Both halves are unpinned — that is `[minor] 3` below.
- Section-label and `--force` sweeps re-run over the whole file; `PRUNE_PID` teardown in `shutdown()` intact; all six CI checks green.

### New findings on the round-2 delta

Three `[minor]`s, inline. In short: `timeout` has no `--kill-after`, so the new bound does not bind against a child that traps `SIGTERM` (demonstrated); the 900s default is larger than the health-check window this README itself documents; and two of the round-1 fixes ship without regression cover.

None gates. The `[major]` that drove round 1 is properly closed for the case that motivated it — a slow `git gc` sweep — and the remaining items are hardening, a default value, and test coverage. Approving with the three as follow-ups.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Taking it, and the demonstration is what makes it obvious — `timeout` without `-k` degrades to "ask nicely, then wait forever", which is the same unbounded boot the round-1 `[major]` was about, just with an extra step. Your framing of the risk is also the right one: a graceful TERM handler that finishes the in-flight git operation is exactly what a careful implementation of a destructive subcommand WOULD install, so the bound would go quietly missing precisely when the CLI lands and the feature stops being a no-op.

Landing `timeout -k 30 "${PRUNE_TIMEOUT_SECONDS}"`. Exit stays 124 so the WARN path and case 7 are untouched, and I will pin it the way case 7 pins the outer bound: a stub variant that traps TERM and keeps running, asserting the boot still reaches the worker — which, per your `trapstub.sh`, fails today.

This lands in a FOLLOW-UP PR off d67e5a9 based on `main`, not as another push here: the verdict on this round is an approve, so this branch is terminal and merge-racy.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Agreed that the pairing is wrong, and I am taking the FIRST option plus the honest comment, rather than the second alone.

Your ordering argument is the deciding one and I had not checked it: the boot pass is at 3c-ii and the console that serves `/healthz` does not start until 4b, so the pass sits in front of the endpoint the platform probes — with Railway's 300s default, a 400-899s pass reproduces the restart loop under the documented configuration, using the shipped default. A bound whose default value permits the failure it exists to prevent is not doing the job.

So: default `900` -> `240`, comfortably inside the documented window, with the comment saying why that number and not a rounder one. Raising it stays correct for the known-slow first sweep, and the WARN already names the knob to raise — which is the right direction for the trade, since a timed-out pass costs nothing but a warning while an overrun boot costs the container. README updated to match.

Same follow-up PR as the finding above.

docker/claude/tests-entrypoint-prune.sh
RHDZMOTA: [triage:pursue] Fair, and the standard you cite is the one I set, so I will meet it. Both counterfactuals you ran reproduce here: removing `</dev/null` and the `redact_token()` call together leaves the suite 41/41 green, so neither half of those two fixes is pinned by anything.

Adding both cases, using your harness suggestions:
- **stdin:** a stub variant that blocks on `read -r _` before printing its verdicts, asserting the SUCCESS line (`workspace prune (boot) finished`) rather than the timeout WARN — that is what distinguishes "got EOF and carried on" from "sat there until the bound fired", which is the distinction the assertion has to make.
- **redaction:** the stub emits a line carrying the token-shaped value the sandbox already injects, and the assertion is two-sided — the redacted marker present AND the raw token absent from the boot log. A one-sided assertion would pass on a log that printed both.

Each verified counterfactually by reverting only the clause it covers.

Same follow-up PR as the two findings above; this branch is terminal on the approve.

RHDZMOTA:
Round-2 `approve` acknowledged — this branch is terminal, so nothing further has been pushed here.

All three round-2 `[minor]`s were triaged `[pursue]` and are implemented in follow-up **#83**, opened against `main` from this PR's approved head `d67e5a9` (every line they touch is code this PR introduces, so `main` has nowhere to put them yet). **Merge this PR first** — #83's body states that as an explicit gate, since merging it first would squash this delta into `main` under its commit message and leave this PR resolving as empty.

One correction worth surfacing here rather than only on the thread, because it changes the finding's own patch: `timeout -k` does **not** keep the exit code at 124. Measured on this harness, GNU `timeout` exits 124 when its own TERM ends the child and **137** when `-k` had to KILL it — so the one-flag fix as written would have filed the exact case `-k` exists for under the generic "exited N" message. #83 adds a distinct 137 branch and asserts both halves.

Also worth recording: the stdin regression test the finding asked for did not work on the first attempt — a backgrounded job already receives `/dev/null` as stdin, so the harness could not distinguish the fix from its absence and passed against the mutant. #83 runs the entrypoint with its stdin on a FIFO held open by a writer that never writes, which makes the case load-bearing (verified by reverting only `</dev/null`).

PR #83, opened 2026-08-10
TASK-190865621: follow-up — make the prune bound actually bind (PR #82 round 2)

RHDZMOTA:
Follow-up to PR #82 round 2 (`approve`), carrying its three `[minor]` findings. No issue is closed here — #82 carries that claim.

**Tasks:** origin `TASK-1492300589` · implementation `TASK-190865621`

Follow-up-To: #82

## Merge order

**Merge #82 first.** This branch is cut from #82's approved head `d67e5a9`, because every line it touches is code #82 introduces — none of it exists on `main` yet. It is based on `main` so it lands there on its own, but merging it *before* #82 would squash #82's entire delta into `main` under this PR's commit message and leave #82 resolving as empty, losing its history and the issue-closing claim it carries for issue #81. The review surface here is `d67e5a9..HEAD` — one commit — even though GitHub's diff against `main` shows #82's changes too until #82 merges.

## What it does

| round-2 finding | fix |
| --- | --- |
| `[minor]` `timeout` without `-k` does not bind a `SIGTERM`-trapping child | `timeout -k "${PRUNE_KILL_AFTER_SECONDS}"`, new `ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` (30) |
| `[minor]` the 900s default sits outside the health-check window the README documents | default → **240s**, and the comment now derives the number from that window instead of asserting roominess |
| `[minor]` the `</dev/null` and `redact_token()` fixes ship with no regression cover | new suite cases 9 and 10, each counterfactually verified |

### Judgment calls

* **`timeout -k` reports 137, not 124 — so the "one flag" fix needed a second branch.** The finding's patch says "the exit code stays 124 either way". Measured here, it does not: GNU `timeout` exits `124` when its own `TERM` ends the child, and `137` (128+9) when `-k` had to `KILL` it. A 124-only branch therefore files the exact case `-k` exists for under the generic *"exited N"* message. The entrypoint now has a distinct `137` branch saying the CLI did not exit on `SIGTERM` and was killed after the grace — which is a fact about the CLI worth surfacing, not just about this boot. Case 8 asserts the kill wording *and* asserts the generic branch was not taken.
* **Took the finding's first option (lower the default), not the second (keep 900, soften the comment).** The reviewer said either was acceptable and preferred the second. The ordering argument decides it: the pass runs at `3c-ii` and `/healthz` does not exist until `4b`, so with Railway's 300s default a 400–899s pass reproduces the restart loop the bound exists to prevent — using the shipped configuration. A default that permits its own failure mode is not a bound. The trade is one-directional: a timed-out pass costs a warning, an overrun boot costs the container.
* **240 + 30 is chosen as a pair.** `-k` grace is spent *after* the timeout, so what has to fit the window is timeout + grace (270s), not the timeout alone. Both the comment and the README say so, because raising only the timeout to "just under 300" would silently exceed it.
* **The stdin case needed a harness change to be load-bearing at all.** My first version passed against a mutant with `</dev/null` removed — a backgrounded job already gets `/dev/null` as stdin, so the harness had nothing to distinguish. Case 9 now runs the entrypoint with its stdin on a FIFO held open by a writer that never writes: inherited stdin blocks, closed stdin gets EOF. This is the finding's point made twice over — the fix was untested, and so was the first test of it.
* **`ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` is a documented knob, not a test-only one**, unlike the interval-seconds override. It changes production behaviour (how long a wedged CLI is tolerated), and an operator who raises the timeout needs it to keep the pair inside their window.
* **Round 1: `137` is now claimed only when the elapsed time supports it.** The branch originally read any `137` as "our `-k` killed it". It is not ours to claim — `timeout` reports 128+signal for *any* signal death, and the realistic early cause on this container is the OOM killer. The branch now also requires `elapsed >= PRUNE_TIMEOUT_SECONDS`; an early `137` falls through to the generic `exited N` WARN. This matters beyond wording: the kill-after-grace message advises *raising* the timeout, which for an OOM means the next pass runs longer and allocates more before dying.
* **Round 1: the 240s derivation no longer implies a boot-wide guarantee.** `240 + 30 < 300` shows prune alone fits the window; it does not show the boot does, and two unbounded steps run in front of it (`3c` `workspace sync` by design, `2c`'s auth backoff at a 600s ceiling). The comment and README now frame 240 as prune's *share* of a shared window — which is still the argument against 900, without the false guarantee.
* **Round 1: the merge-ordering `[minor]` is the one finding declined** (`[triage:ignore]`, reasoned on the PR thread). Retargeting this PR's base at `TASK-190865621-WORKSPACE-PRUNE` would let GitHub enforce the order, but a PR based on a sibling branch is auto-retargeted when the parent squash-merges, and the retargeted diff can resolve to empty while GitHub reports a successful merge — silent and unrecoverable, versus a wrong merge order that is at least visible at review time. The gate stays prose, first section, bold.

* [ ] **Whether the real CLI traps `SIGTERM`.** The whole `-k` path is exercised against a stub that traps it deliberately. If TASK-2059754063 ships a graceful handler, 30s is a guess at how long an in-flight git operation needs; the `137` WARN is what will say so, and the knob is what raises it.
* [ ] **Whether 240s is enough for a real first sweep** on the production volume. It is deliberately not: the design assumes the first pass may time out, warn, and reclaim nothing until an operator raises the knob. If that reads as too conservative once real numbers exist, the knob is the lever — not a code change.
* [ ] **Railway's healthcheck timeout is 300s by default**, taken from the platform's documented default and this repo's own README walkthrough, not from an observed deploy. If this service configures a different window, the default should be re-derived against it.
* [ ] **Still no container and still no real subcommand** — same two gates as #82, unchanged by this PR.
* [ ] **The `137` elapsed guard assumes the kill lands at or after the bound.** If a platform's own supervisor SIGKILLs a *long-running* pass for an unrelated reason, it will still be reported as our kill-after-grace. The guard removes the early-kill misdiagnosis, which is the OOM case; it cannot separate two late kills.

### Verification

* `docker/claude/tests-entrypoint-prune.sh`: **51/51 ok**, eleven boots (cases 8, 9, 10 new at open; case 11 added in round 1).
* Counterfactuals, each mutating exactly one clause in a throwaway copy of `docker/claude/`:
  * drop `-k` → case 8 fails 2/3 ("a TERM-trapping prune pass held the boot open");
  * drop `</dev/null` → case 9 fails 2/2 (the pass is cut off by the bound instead of finishing);
  * drop `redact_token()` → case 10 fails 2/2 (marker absent, raw token present).
* All six pre-existing entrypoint suites re-run against the patched entrypoint: `config`, `identity`, `chown`, `executor`, `auth`, `ui` — all pass.
* `shellcheck -S warning` on `entrypoint.sh`: clean. `bash -n` on both changed shell files.
* Measured rather than assumed: `timeout -k 2 2 bash -c '<traps TERM>'` → exit **137**, and plain `timeout 25 bash -c 'timeout 2 <traps TERM>'` → the inner bound never returns and the outer one fires at 25s. That measurement is what produced the `137` branch above.
* No Python touched, so the style/type/unit suites carry no information about this diff; they were green on #82's head.

**Round 1 fixes (`3e6213f`):**

* Fourth counterfactual, mutating only the new clause: drop the `elapsed >= bound` condition from the `137` branch and case 11 fails 3/3 — an early SIGKILL is announced as a 20s timeout that never elapsed.
* Suite back to **51/51**; all six pre-existing entrypoint suites re-run and pass; `shellcheck -S warning` clean; `grep -c` confirms the duplicated comment block appears exactly once.
* Worth recording how the duplication happened, since `bash -n`, `shellcheck` and the suite are all blind to it: the patch script that applied this change aborted partway on an unrelated anchor, and the "already applied" guard added for the re-run tests whether the OLD text is still present — for this hunk the new text contained the old as a prefix, so the guard read "not yet applied" and appended a second copy. A comment-only duplication has no runtime signal at all; only review caught it.

### Round 2 — approve, three nits deferred

Round 2 returned `approve` on `3e6213f` with 0 blockers/majors/minors and three `[nit]`s. All three are triaged `[triage:later]` against **TASK-2093688770** (soft-linked to TASK-190865621), not pushed here:

* the generic non-zero WARN branch — the landing spot for an early `137` — logs the raw number without naming the signal or pointing an early kill at the memory ceiling;
* case 9 kills `FIFO_HOLDER` but never clears it, so the new `EXIT` trap can fire a second `kill` at a recycled PID (a defect this round introduced);
* the README's volume-hygiene paragraph now states "sync at `3c` is unbounded" twice, four lines apart.

Deferred rather than pushed because an approve is terminal for this branch, and a third stacked PR would force the operator into a strict #82 → #83 → #84 merge order — a worse trade than three registered diagnostic nits. The task is written to be opened as a plain PR against `main` once both merge, at which point nothing needs stacking.

### Scope touched

Same three files as #82, all within its declared scope (`docker/claude/*`):

* `docker/claude/entrypoint.sh` — the runner's `timeout -k`, the `137` branch, the two knob resolutions, the two ENABLED log lines.
* `docker/claude/tests-entrypoint-prune.sh` — cases 8/9/10, four stub modes, and the `EP_STDIN` hook.
* `docker/claude/README.md` — the timeout row rewritten, a `KILL_AFTER` row, and the window derivation.
* No workflow change needed: `check-entrypoint.yaml` already runs this suite (added by #82).

### DIFF
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -237,6 +237,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_FORCE_CHOWN` | `0` | `1`/`true`/`yes`/`on` forces the root phase's full `chown -R` of the volume for that boot; anything else (incl. unset) leaves the probe in charge. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
 | `ALISSA_WORKSPACE_PRUNE` | `1` (**on**) | volume hygiene: `1`/`true`/`yes`/`on` runs `alissa code workspace prune` at boot **and** on an interval; `0` opts out, after which nothing in the container reclaims anything. Default-on because this is the container whose volume grows (see [Volume hygiene](#volume-hygiene-finished-worktrees-are-pruned)). Skipped, with one loud warning, if the installed CLI predates the subcommand |
 | `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` | `360` | minutes between prune passes. Daemon role only — the executor gets the boot pass and no loop. A non-numeric or non-positive value warns and falls back to `360` |
+| `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` | `900` | how long ONE prune pass may run before it is killed, so a slow first sweep over a never-pruned volume cannot hold the boot open. A timed-out pass warns (exit 124) and changes nothing. Garbage falls back to `900` |
 | `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` | *(CLI default)* | passed straight through as `--min-age-hours`; **pass-through** — unset ⇒ the CLI's own age gate. A value that is not a whole number of hours **disables prune for that boot** instead of falling back (the CLI's default gate may be *shorter* than the one you asked for) |
 
 #### Config precedence: env var > daemon library default
@@ -352,6 +353,25 @@ safe to leave on by default:
   gets the boot pass — it grows the same volume the same way — but not the loop:
   that role `exec`s `alissa bridge start`, which replaces the shell, so a loop
   started there would outlive the only code that knows how to stop it.
+
+**A pass cannot hold the boot open.** The boot hook runs before `alissa worker`
+starts, and the first pass after the CLI lands sweeps `git gc --auto` across hubs
+on a volume that has never had anything reclaimed — minutes of work on a large
+object store, and on a platform with a startup health-check window a slow enough
+boot is a *dead* boot. So a pass runs under `timeout`
+(`ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS`, default 900) with **stdin closed**, and
+a pass that hits either bound lands in the same warning path as any other
+failure. The neighbouring `alissa code workspace sync` is deliberately *not*
+bounded this way: hubs are a precondition for reviewing anything, while prune is
+an optimization, and only one of those may delay a boot.
+
+> **`ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS` is test-only — do not set it in a
+> deploy.** `docker/claude/tests-entrypoint-prune.sh` has to watch the loop tick
+> more than once, and the smallest cadence the supported knob can express is a
+> minute. It silently overrides
+> `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES`, and the boot log then reads
+> `starting the workspace prune loop (every 360 min = 3s)` — which is what you
+> will find first if it is ever set by accident.
 
 If you turn this off (`ALISSA_WORKSPACE_PRUNE=0`), nothing else reclaims the
 volume; plan on pruning by hand.
@@ -783,11 +803,12 @@ volumes:
    daemon's on-demand `alissa code workspace add` no-ops on a repo already listed
    in the manifest, leaving an empty folder and looping forever hub-ifying a hub
    that never completes.
-3b. **`alissa code workspace prune`** — best-effort volume hygiene, on the path
+3c-ii. **`alissa code workspace prune`** — best-effort volume hygiene, on the path
    both roles share: remove finished-branch worktrees (never `main/`, never
    dirty, never an open PR, never `--force`) and run the hub-level git prune/gc.
    Skipped with one loud warning if the CLI predates the subcommand, or if
-   `ALISSA_WORKSPACE_PRUNE=0`. A failure here is a warning, never a dead boot
+   `ALISSA_WORKSPACE_PRUNE=0`, and bounded by `timeout` with stdin closed so a
+   slow or prompting pass cannot hold the boot open. A failure here is a warning, never a dead boot
    ([why](#volume-hygiene-finished-worktrees-are-pruned)).
 3d. **Executor role only, and the end of the shared path**: drop this executor's
    stale lockfile, then `exec alissa bridge
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -235,6 +235,9 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
 | `ALISSA_FIREWALL_EXTRA` | *(empty)* | extra firewall allowlist hosts, space-separated |
 | `ALISSA_FORCE_CHOWN` | `0` | `1`/`true`/`yes`/`on` forces the root phase's full `chown -R` of the volume for that boot; anything else (incl. unset) leaves the probe in charge. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
+| `ALISSA_WORKSPACE_PRUNE` | `1` (**on**) | volume hygiene: `1`/`true`/`yes`/`on` runs `alissa code workspace prune` at boot **and** on an interval; `0` opts out, after which nothing in the container reclaims anything. Default-on because this is the container whose volume grows (see [Volume hygiene](#volume-hygiene-finished-worktrees-are-pruned)). Skipped, with one loud warning, if the installed CLI predates the subcommand |
+| `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` | `360` | minutes between prune passes. Daemon role only — the executor gets the boot pass and no loop. A non-numeric or non-positive value warns and falls back to `360` |
+| `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` | *(CLI default)* | passed straight through as `--min-age-hours`; **pass-through** — unset ⇒ the CLI's own age gate. A value that is not a whole number of hours **disables prune for that boot** instead of falling back (the CLI's default gate may be *shorter* than the one you asked for) |
 
 #### Config precedence: env var > daemon library default
 
@@ -302,6 +305,56 @@ Reclaiming the cache after the fact is not an option on Railway: `/sys/fs/cgroup
 is mounted read-only inside the container, so writing `memory.reclaim` fails with
 `EROFS` even as root (verified 2026-08-09). Avoidance is the only lever, which is
 why the entrypoint has no reclaim call.
+
+#### Volume hygiene: finished worktrees are pruned
+
+**The diagnosis.** Until issue #81 this container removed *nothing*, ever. Every
+review round materializes a per-PR worktree inside a hub (`<repo>/TASK-…`), plus
+whatever a session's build or test run leaves in it; a merged or closed PR left
+its worktree exactly where it was; and the bare `.source` object store behind
+each hub only ever grew, because nothing pruned the remote-tracking refs the
+deleted branches left behind. There was no cleanup path anywhere in the image —
+not at boot, not on merge, not on a timer — so the only bound on the volume was
+the volume's size, and the only remedy was a human with a shell.
+
+**The remediation** is the CLI's own primitive, `alissa code workspace prune`,
+which removes finished-branch worktrees behind its safety rails (never `main/`,
+never a dirty tree without `--force`, an age gate, a live-tmux-session guard,
+never a worktree whose PR is still open, and a lookup failure *keeps* the
+worktree) and then runs the hub-level `git worktree prune` /
+`git remote prune origin` / `git gc --auto` sequence. The entrypoint calls it in
+two places:
+
+- **at boot**, right after the hub sync, best-effort — a failure logs a warning
+  and the boot continues, exactly like the other best-effort boot steps. Prune
+  is an optimization, never a boot precondition;
+- **on an interval** (`ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES`, default 360), as
+  a supervised background loop that the shutdown handler kills alongside the
+  console sidecar. A failing pass warns and the loop waits for the next tick; it
+  never takes the container down.
+
+Three properties are worth stating explicitly, because they are what make this
+safe to leave on by default:
+
+- **`--force` is never passed, and there is no env knob that adds it.** In a
+  review container a dirty worktree is *evidence* — a session that died
+  mid-round with work someone may still want. The rail that keeps it is the one
+  this container needs most.
+- **It is capability-gated, not version-pinned.** The image installs the alissa
+  CLI from its install channel, which may predate the subcommand. The entrypoint
+  probes once at boot and, if the subcommand is missing, logs exactly one loud
+  `WARN: this alissa CLI predates 'alissa code workspace prune' …` and skips
+  *both* hooks. Nothing here needs to change when the CLI catches up. (The probe
+  reads the help *output*, not its exit status: commander answers an unknown
+  subcommand by printing the parent command's help and exiting `0`, so an
+  exit-status probe would call every old CLI capable.)
+- **The interval loop is daemon-only.** The [executor role](#bridge-executor-role-a-second-service-from-this-same-image)
+  gets the boot pass — it grows the same volume the same way — but not the loop:
+  that role `exec`s `alissa bridge start`, which replaces the shell, so a loop
+  started there would outlive the only code that knows how to stop it.
+
+If you turn this off (`ALISSA_WORKSPACE_PRUNE=0`), nothing else reclaims the
+volume; plan on pruning by hand.
 
 ### Reviewer console (runtime env only — `ALISSA_UI_ENABLED`, `ALISSA_UI_PASSCODE`, `PORT`)
 
@@ -462,6 +515,12 @@ ephemeral home and are gone on restart, so a persisted "round N in-flight" would
 otherwise be stale and stall re-enqueue for 90 min. A fresh container has no
 reviewer running by definition, so clearing them is safe and the daemon
 re-enqueues any still-pending round on its first poll.
+
+The volume is also the thing that **grows**: reviewer worktrees accumulate in
+the hubs and nothing used to remove them. The entrypoint now prunes finished-branch
+worktrees at boot and every `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` (default
+360) — see [Volume hygiene](#volume-hygiene-finished-worktrees-are-pruned) for the
+rails it runs behind and `ALISSA_WORKSPACE_PRUNE=0` for the opt-out.
 
 Nothing else needs a volume: the gh/alissa/claude auth is re-established from the
 env tokens on every boot, and tmux sockets are deliberately ephemeral.
@@ -724,6 +783,12 @@ volumes:
    daemon's on-demand `alissa code workspace add` no-ops on a repo already listed
    in the manifest, leaving an empty folder and looping forever hub-ifying a hub
    that never completes.
+3b. **`alissa code workspace prune`** — best-effort volume hygiene, on the path
+   both roles share: remove finished-branch worktrees (never `main/`, never
+   dirty, never an open PR, never `--force`) and run the hub-level git prune/gc.
+   Skipped with one loud warning if the CLI predates the subcommand, or if
+   `ALISSA_WORKSPACE_PRUNE=0`. A failure here is a warning, never a dead boot
+   ([why](#volume-hygiene-finished-worktrees-are-pruned)).
 3d. **Executor role only, and the end of the shared path**: drop this executor's
    stale lockfile, then `exec alissa bridge
    start …`. `exec`, so the container's only process is the executor — steps 4
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -237,7 +237,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_FORCE_CHOWN` | `0` | `1`/`true`/`yes`/`on` forces the root phase's full `chown -R` of the volume for that boot; anything else (incl. unset) leaves the probe in charge. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
 | `ALISSA_WORKSPACE_PRUNE` | `1` (**on**) | volume hygiene: `1`/`true`/`yes`/`on` runs `alissa code workspace prune` at boot **and** on an interval; `0` opts out, after which nothing in the container reclaims anything. Default-on because this is the container whose volume grows (see [Volume hygiene](#volume-hygiene-finished-worktrees-are-pruned)). Skipped, with one loud warning, if the installed CLI predates the subcommand |
 | `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` | `360` | minutes between prune passes. Daemon role only — the executor gets the boot pass and no loop. A non-numeric or non-positive value warns and falls back to `360` |
-| `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` | `240` | how long ONE prune pass may run before it is killed, so a slow first sweep over a never-pruned volume cannot hold the boot open. The default is chosen to sit inside a 300s health-check window — the pass runs *before* `/healthz` exists (see below). A timed-out pass warns (exit 124) and changes nothing. Garbage falls back to `240` |
+| `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` | `240` | how long ONE prune pass may run before it is killed, so a slow first sweep over a never-pruned volume cannot hold the boot open. The default is chosen to sit inside a 300s health-check window — the pass runs *before* `/healthz` exists (see below). A timed-out pass warns (exit 124, or 137 if it had to be killed) and changes nothing. Garbage falls back to `240` |
 | `ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` | `30` | grace a `SIGTERM`-trapping prune gets before `SIGKILL` (`timeout -k`). Without it a CLI that traps `TERM` is unbounded again. **Timeout + grace** is what must fit your health-check window |
 | `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` | *(CLI default)* | passed straight through as `--min-age-hours`; **pass-through** — unset ⇒ the CLI's own age gate. A value that is not a whole number of hours **disables prune for that boot** instead of falling back (the CLI's default gate may be *shorter* than the one you asked for) |
 
@@ -367,13 +367,21 @@ The default of **240s** is picked against a window, not chosen for roundness. Th
 prune pass runs at step `3c-ii`; the console that serves `/healthz` — the health
 check [the Railway walkthrough above](#the-console-on-railway) tells you to
 configure — does not start until `4b`. The pass therefore sits *in front of* the
-endpoint the platform probes, and Railway's healthcheck timeout defaults to 300s,
-so a larger default would reproduce the restart loop the bound exists to prevent.
+endpoint the platform probes, and Railway's healthcheck timeout defaults to 300s.
 `ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` (30) is the grace a `SIGTERM`-trapping
-CLI gets before `SIGKILL`; **it is the sum that must fit the window**, not the
-timeout alone. Raising the timeout for a known-slow first sweep is fine — the
-warning tells you to — as long as timeout + grace stays inside whatever window
-sits in front of this container. The neighbouring `alissa code workspace sync` is deliberately *not*
+CLI gets before `SIGKILL`; **it is the sum that must fit**, not the timeout alone.
+
+**This bounds the prune step, not the boot.** 240 + 30 < 300 says prune alone fits
+the window; it does not say the boot does, and nothing here bounds the total. Two
+unbounded steps run in front of it: `alissa code workspace sync` at `3c` is
+deliberately unbounded (hubs are a precondition — without them the daemon reviews
+nothing, so blocking on it is correct), and the auth retry loop at `2c` backs off
+to a 600s ceiling, which exceeds the whole window on its own. So read 240 as
+prune's *share* of a window other steps also draw on — the reason to prefer it
+over 900 is that 900 gave this one step more than the entire window. If your
+deployment has a health check in front of it, budget across all three, and when
+you raise the timeout for a known-slow first sweep keep timeout + grace inside
+whatever is left. The neighbouring `alissa code workspace sync` is deliberately *not*
 bounded this way: hubs are a precondition for reviewing anything, while prune is
 an optimization, and only one of those may delay a boot.
 
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -237,7 +237,8 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_FORCE_CHOWN` | `0` | `1`/`true`/`yes`/`on` forces the root phase's full `chown -R` of the volume for that boot; anything else (incl. unset) leaves the probe in charge. Off by default: the entrypoint probes ownership at depth 1 and walks only when it finds a foreign owner (see [Boot-time ownership](#boot-time-ownership-the-volume-walk-is-guarded)). Set it for **one** boot after a root shell wrote files deeper in the tree |
 | `ALISSA_WORKSPACE_PRUNE` | `1` (**on**) | volume hygiene: `1`/`true`/`yes`/`on` runs `alissa code workspace prune` at boot **and** on an interval; `0` opts out, after which nothing in the container reclaims anything. Default-on because this is the container whose volume grows (see [Volume hygiene](#volume-hygiene-finished-worktrees-are-pruned)). Skipped, with one loud warning, if the installed CLI predates the subcommand |
 | `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` | `360` | minutes between prune passes. Daemon role only — the executor gets the boot pass and no loop. A non-numeric or non-positive value warns and falls back to `360` |
-| `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` | `900` | how long ONE prune pass may run before it is killed, so a slow first sweep over a never-pruned volume cannot hold the boot open. A timed-out pass warns (exit 124) and changes nothing. Garbage falls back to `900` |
+| `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` | `240` | how long ONE prune pass may run before it is killed, so a slow first sweep over a never-pruned volume cannot hold the boot open. The default is chosen to sit inside a 300s health-check window — the pass runs *before* `/healthz` exists (see below). A timed-out pass warns (exit 124) and changes nothing. Garbage falls back to `240` |
+| `ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` | `30` | grace a `SIGTERM`-trapping prune gets before `SIGKILL` (`timeout -k`). Without it a CLI that traps `TERM` is unbounded again. **Timeout + grace** is what must fit your health-check window |
 | `ALISSA_WORKSPACE_PRUNE_MIN_AGE_HOURS` | *(CLI default)* | passed straight through as `--min-age-hours`; **pass-through** — unset ⇒ the CLI's own age gate. A value that is not a whole number of hours **disables prune for that boot** instead of falling back (the CLI's default gate may be *shorter* than the one you asked for) |
 
 #### Config precedence: env var > daemon library default
@@ -358,10 +359,21 @@ safe to leave on by default:
 starts, and the first pass after the CLI lands sweeps `git gc --auto` across hubs
 on a volume that has never had anything reclaimed — minutes of work on a large
 object store, and on a platform with a startup health-check window a slow enough
-boot is a *dead* boot. So a pass runs under `timeout`
-(`ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS`, default 900) with **stdin closed**, and
+boot is a *dead* boot. So a pass runs under `timeout` with **stdin closed**, and
 a pass that hits either bound lands in the same warning path as any other
-failure. The neighbouring `alissa code workspace sync` is deliberately *not*
+failure.
+
+The default of **240s** is picked against a window, not chosen for roundness. The
+prune pass runs at step `3c-ii`; the console that serves `/healthz` — the health
+check [the Railway walkthrough above](#the-console-on-railway) tells you to
+configure — does not start until `4b`. The pass therefore sits *in front of* the
+endpoint the platform probes, and Railway's healthcheck timeout defaults to 300s,
+so a larger default would reproduce the restart loop the bound exists to prevent.
+`ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` (30) is the grace a `SIGTERM`-trapping
+CLI gets before `SIGKILL`; **it is the sum that must fit the window**, not the
+timeout alone. Raising the timeout for a known-slow first sweep is fine — the
+warning tells you to — as long as timeout + grace stays inside whatever window
+sits in front of this container. The neighbouring `alissa code workspace sync` is deliberately *not*
 bounded this way: hubs are a precondition for reviewing anything, while prune is
 an optimization, and only one of those may delay a boot.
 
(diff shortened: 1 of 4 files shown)


alissa-app:
- 2026-08-10 17:30:56 UTC — `review-alissa-github-review-daemon-pr83-r1-ed07ae` — round 1 of 22 — spawned
- 2026-08-10 18:06:50 UTC — `review-alissa-github-review-daemon-pr83-r2-bfddfe` — round 2 of 22 — spawned

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,263 @@ if [ -f "${MANIFEST}" ]; then
...
+# and then waits indefinitely, so a child that traps TERM is unbounded again —
+# the same boot-held-open failure, one step further along. That is not a
+# hypothetical for this subcommand: `workspace prune` is the destructive one, and
+# a handler that finishes the in-flight git operation before exiting is exactly
+# what a careful implementation would install, so the bound would go missing at
+# the moment the CLI lands and the feature stops being a no-op. `-k` gives such a
+# handler PRUNE_KILL_AFTER_SECONDS to finish and then KILLs it; the exit status
+# stays 124 either way, so the WARN path below does not care which signal did it.
+#
+# `-k` IS PART OF THE BOUND, not a refinement of it. Plain `timeout` sends TERM
```
alissa-app: `[major]` **This ten-line block is pasted twice, verbatim (963–971 and 973–981), and both copies end on a claim this PR measured to be false.**

Two defects in one place:

1. **The duplication.** Lines 973–981 are a character-for-character repeat of 963–971. `bash -n` and `shellcheck` cannot see it because it is a comment; the suite cannot see it either.
2. **The surviving text is wrong, and wrong about the central thing this PR adds.** Both copies close with:

   > `-k` gives such a handler PRUNE_KILL_AFTER_SECONDS to finish and then KILLs it; **the exit status stays 124 either way, so the WARN path below does not care which signal did it.**

   That is precisely the claim your own commit message says you measured and refuted (*"The finding's patch said the exit code stays 124 either way; measured, it does not"*), and it is refuted by the `137` branch 38 lines below, by the comment at 1013–1018 (*"GNU timeout does NOT normalise the kill path to 124"*), and by case 8's `assert_not_contains`. I re-measured on coreutils 9.1 to be sure: `timeout -k 2 2 bash -c '<traps TERM>'` → **137**.

So the file now carries, above the runner, a comment telling the next maintainer that the WARN path does not need to distinguish the signals — which is an argument for deleting the `137` branch. That is the one wrong conclusion this change must not invite.

One edit fixes both: drop 972–981 and rewrite the tail of the survivor to say the KILL path reports **137**, not 124, and that the branch below exists for it.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,263 @@ if [ -f "${MANIFEST}" ]; then
...
+  elif [ "${rc}" = "137" ]; then
+    # 124 is `timeout`'s own "I terminated it"; 137 (128+9) is what it reports
+    # when the child had to be KILLED after `-k`. Measured, not assumed: GNU
+    # timeout does NOT normalise the kill path to 124, so a 124-only branch would
+    # file exactly the case `-k` exists for under the generic "exited N" message.
+    # Worth its own line either way — a CLI that ignores SIGTERM is a fact about
+    # the CLI, not about this boot.
+    log "WARN: workspace prune (${label}) TIMED OUT after ${PRUNE_TIMEOUT_SECONDS}s and did NOT exit on SIGTERM, so it was KILLED after the ${PRUNE_KILL_AFTER_SECONDS}s grace — the boot was not held open, and nothing was removed. Raise ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS (keeping it plus the grace inside your health-check window); if this recurs, the CLI is trapping SIGTERM and the grace may need to be longer."
```
alissa-app, line 1024: `[major]` **`137` is not exclusive to `-k`'s KILL, so this line states a cause it cannot know — and the remedy it recommends makes the most likely real cause worse.**

The branch reads `rc == 137` as *"the CLI ignored SIGTERM and we had to kill it after the grace"*. But `timeout` reports `128+signal` for **any** signal death of the child, not just its own. Measured here on coreutils 9.1:

```
timeout 100 bash -c 'kill -9 $$'        -> 137   # no -k involved at all
timeout -k 2 100 bash -c 'kill -9 $$'   -> 137   # killed by something else, not by -k
timeout 100 bash -c 'exit 137'          -> 137   # child's own exit status
```

The first two are the ones that matter, because the realistic source of an external `SIGKILL` here is the **OOM killer**, and this pass is the best OOM candidate the boot has: `git gc --auto` across every hub on a never-pruned object store, in a memory-capped container. This repo shipped a container memory tile because that pressure is live (issue #74).

Consequences when that happens at, say, t=25s:

* the log asserts **`TIMED OUT after 240s`** — a duration that did not elapse;
* it asserts the CLI **did not exit on SIGTERM** — it was never sent one;
* it tells the operator to **raise `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS`**, which for an OOM means letting the next pass run longer and allocate more before it dies again;
* and it adds *"if this recurs, the CLI is trapping SIGTERM"* — pointing the investigation at the CLI instead of at the memory ceiling.

The `124` branch does not have this problem: `124` really is `timeout`'s own signature. It is only the `137` path that is ambiguous, and the disambiguator is free — you already know when the pass started:

```sh
local started=${SECONDS}
...
elif [ "${rc}" = "137" ] && [ $(( SECONDS - started )) -ge "${PRUNE_TIMEOUT_SECONDS}" ]; then
    # ... the existing kill-after-grace message
```

An early `137` then falls through to the generic *"exited 137"* branch, which is true and says nothing false about the cause. Softening the wording to *"was killed (SIGKILL) — if this took the full ${PRUNE_TIMEOUT_SECONDS}s the CLI is trapping SIGTERM; if it died early, suspect the OOM killer"* would also close it, but the elapsed check is strictly better and costs two lines.

This is the branch the PR's headline judgment call is about, and its justification is *"a CLI that ignores SIGTERM is a fact about the CLI, not about this boot"* — that is exactly right, and exactly why the branch should only claim it when it can tell.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,263 @@ if [ -f "${MANIFEST}" ]; then
...
+  # Both roles resolve it, because both run the boot pass. Garbage falls back to
+  # the default with a warning rather than failing closed: a bound that is merely
+  # wrong is still a bound, and disabling the hygiene over a typo would trade the
+  # smaller problem for the larger one.
+  PRUNE_TIMEOUT_SECONDS="${ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS:-240}"
+  if ! printf '%s' "${PRUNE_TIMEOUT_SECONDS}" | grep -qE '^[1-9][0-9]*$'; then
+    log "WARN: ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS=${PRUNE_TIMEOUT_SECONDS} is not a positive whole number of seconds — falling back to 240"
+    PRUNE_TIMEOUT_SECONDS=240
+  fi
+  # Grace for a CLI that traps TERM, before KILL. Test-only override; a deploy
```
alissa-app: `[minor]` **"Test-only override; a deploy leaves it unset" contradicts the README and your own delivery note.**

The PR body is explicit the other way:

> `ALISSA_WORKSPACE_PRUNE_KILL_AFTER_SECONDS` **is a documented knob, not a test-only one**, unlike the interval-seconds override. It changes production behaviour […] and an operator who raises the timeout needs it to keep the pair inside their window.

And the README agrees — it gets a full row in the env table (line 241), not the `ALISSA_INSTALL_URL`-shaped test-only blockquote that `ALISSA_WORKSPACE_PRUNE_INTERVAL_SECONDS` gets. This repo keeps that distinction visible on purpose, and the code comment is the copy an operator grepping `entrypoint.sh` lands on first.

The rest of the comment (30s, and why timeout+grace is the pair that must fit) is right and worth keeping — it is only the first sentence that needs to go, replaced by something like *"Operator-tunable; see the README env table."*

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,263 @@ if [ -f "${MANIFEST}" ]; then
...
+  # How long ONE pass may run before it is killed — a bound, not a budget.
+  #
+  # 240s, and the number is chosen against a specific window rather than being
+  # "generously large": this pass runs at 3c-ii, while the console that serves
+  # the documented `/healthz` deployment health check does not start until 4b —
+  # so the pass sits IN FRONT of the endpoint the platform probes. Railway's
+  # healthcheck timeout defaults to 300s (and the README's own Railway walkthrough
+  # tells the operator to point it at `/healthz`), so any default above that
+  # reproduces the restart loop this bound exists to prevent, under the shipped
+  # configuration. A bound whose default permits its own failure mode is not a
```
alissa-app: `[minor]` **The derivation budgets the whole 300s window to this one step, and two unbounded steps in front of it share that window.**

240 is a clear improvement on 900 and the direction is not in dispute. But the argument as written — *"any default above [300s] reproduces the restart loop […] A bound whose default permits its own failure mode is not a bound"* — proves less than it claims, because prune is not the only thing standing between container start and `/healthz` at `4b`:

* **`3c`, `workspace sync` (line 871)** — run with no `timeout` at all, deliberately, and this very file says so 80 lines up: *"only one of those may delay a boot"*. On the same never-pruned, never-synced volume that motivates the slow-first-prune case, the first sync is the slower of the two.
* **`2c`, the auth retry loop** — capped exponential backoff with a **600s** ceiling (`AUTH_RETRY_CAP_SECONDS`, line 426). A single unlucky retry there exceeds the entire window on its own.

So `240 + 30 = 270 < 300` establishes that *prune alone* fits, not that the boot does. Under the reasoning's own standard, the default still permits the failure mode — a prune pass that runs 200s after a sync that ran 150s blows the window without ever reaching its bound.

I am not asking for a smaller number: the sizing is a judgment call and 240 is defensible. What is worth fixing is the comment (and the README paragraph at line 366) asserting a safety property the arithmetic does not support. Either state the budget as shared — *"prune's share of a 300s window that `3c` and `2c` also draw on"* — or say plainly that this bounds the step this PR owns and that the boot's total is not bounded by anything here. The second is the honest version, and it is still a good argument for 240 over 900.

docker/claude/README.md
```diff
@@ -235,6 +235,11 @@ automatically; locally pass `--build-arg`):
...
+| `ALISSA_WORKSPACE_PRUNE` | `1` (**on**) | volume hygiene: `1`/`true`/`yes`/`on` runs `alissa code workspace prune` at boot **and** on an interval; `0` opts out, after which nothing in the container reclaims anything. Default-on because this is the container whose volume grows (see [Volume hygiene](#volume-hygiene-finished-worktrees-are-pruned)). Skipped, with one loud warning, if the installed CLI predates the subcommand |
+| `ALISSA_WORKSPACE_PRUNE_INTERVAL_MINUTES` | `360` | minutes between prune passes. Daemon role only — the executor gets the boot pass and no loop. A non-numeric or non-positive value warns and falls back to `360` |
+| `ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS` | `240` | how long ONE prune pass may run before it is killed, so a slow first sweep over a never-pruned volume cannot hold the boot open. The default is chosen to sit inside a 300s health-check window — the pass runs *before* `/healthz` exists (see below). A timed-out pass warns (exit 124) and changes nothing. Garbage falls back to `240` |
```
alissa-app: `[nit]` This row still says *"A timed-out pass warns (exit 124) and changes nothing"*, but after this PR a `SIGTERM`-trapping pass warns at **137**, which is the case the neighbouring row exists for. *"warns (exit 124, or 137 if it had to be killed)"* keeps the table consistent with the code.

docker/claude/tests-entrypoint-prune.sh
```diff
@@ -0,0 +1,526 @@
...
+# Command substitution inherits stdin, so without `</dev/null` a prompting CLI
+# would sit until the timeout fires. The assertion is therefore the SUCCESS line,
+# not the timeout WARN: both leave the boot alive, and only the success line
+# distinguishes "got EOF and carried on" from "was cut off".
+STDIN_FIFO="${TMPROOT}/blocking-stdin.fifo"
+mkfifo "${STDIN_FIFO}"
+# A writer that never writes: opening the read end then yields no data and no
+# EOF, so a child inheriting it blocks. Backgrounded because opening a FIFO for
+# writing blocks until a reader arrives.
+sleep 120 > "${STDIN_FIFO}" &
```
alissa-app, line 504: `[nit]` The FIFO writer is only reaped on the success path (line 500), so any `bad`/early exit between here and there leaves a `sleep 120` behind for two minutes. The suite has no `trap … EXIT` cleanup, so `trap 'kill ${FIFO_HOLDER} 2>/dev/null' EXIT` right after line 490 — or just `sleep 30`, which comfortably outlasts a 20s bound — would close it. Harmless either way; the case itself is well built, and pointing the entrypoint's stdin at a blocking FIFO is the detail that makes it mean anything.

alissa-app (changes requested):
## Review — round 1 (review task `TASK-513658597`, cap 22)

Fresh-instance review per CR1–CR9. Surface reviewed is **`d67e5a9..06a8fac`**, the one commit this PR adds on top of #82's approved head — GitHub's diff against `main` also shows #82's delta, which is not under review here.

**Verdict: `request_changes`** — 2 `[major]`, 3 `[minor]`, 2 `[nit]`, 0 `[question]`. Nothing here is a design objection: all three round-2 `[minor]`s are genuinely pursued and the new tests are load-bearing. The blockers-to-approve are two defects in what the diff *says*, one of which is a verbatim double-paste that instructs the next maintainer to delete the branch this PR exists to add.

### What I verified myself (not taken from the delivery note)

* Suite re-run at `06a8fac` in a throwaway worktree: **48/48 ok**, ten boots, 68s.
* All three counterfactuals reproduced, each mutating exactly one clause in a throwaway copy of `docker/claude/`:
  * drop `-k` → case 8 fails 2/3 (`a TERM-trapping prune pass held the boot open`);
  * drop `</dev/null` → case 9 fails 2/2 (the pass is cut off by the bound instead of finishing);
  * drop `redact_token()` → case 10 fails 2/2 (marker absent, raw token present).
  The three new cases are load-bearing. Case 9's FIFO harness change is the right call — I confirmed the backgrounded-job stdin problem it exists to solve.
* `bash -n` clean on both changed shell files. CI green (6/6).
* `timeout` exit-code behaviour measured independently on coreutils 9.1 — see the `[major]` on the `137` branch.
* Grandchild containment: `timeout` signals the child's process group, so a `git` subprocess spawned by the CLI does not survive the bound and hold the command substitution's pipe open (measured: same-pgroup grandchild → bounded at 2s). Only a `setsid`-detached grandchild escapes, which is not a plausible shape for this CLI. **Nothing found** — noting it because the bound would be worth nothing otherwise.

### Rubric

* **Correctness** — the three round-2 `[minor]`s are each pursued and each carries cover that fails without the fix. The `137` branch is a correct discovery about GNU `timeout` and the 124-only patch in the finding was indeed wrong. One `[major]`: the branch's *attribution* is not sound (see inline).
* **Coding practices** — matches house idiom: knob resolution + `grep -qE '^[1-9][0-9]*$'` validation + fallback WARN mirrors the sibling knobs, the suite's header case map is updated with cases 8–10 (this repo has a task about exactly that), commit carries the `Co-Authored-By` trailer, branch is `TASK-190865621-…`, PR names the task and is on it as evidence. One `[major]` (duplicated block) and one `[minor]` (a comment that contradicts the README) are both here.
* **Security** — **nothing found** that is new, and case 10 *adds* security cover: it pins `redact_token()` on the prune path two-sidedly (marker present **and** raw token absent), which is the correct shape. `--force` is still never passed; `PRUNE_ARGS` is unchanged. The new knob is integer-validated before it reaches `timeout`, so it cannot smuggle an argument.
* **Performance** — **nothing found**. The change lowers a bound and adds a grace; the only runtime cost is up to 30s more wall-clock on a pass that was already going to be killed.
* **Side-effects / blast radius** — the entrypoint is the whole container's boot path, but this diff moves only a default and adds a log branch. One `[minor]` on merge ordering (below), which is a real irreversibility: the wrong merge order is not recoverable by re-merging.

### `[minor]` — merge ordering is enforced by prose only

The PR body is emphatic that #82 merges first, and the reasoning is right: merging this first squashes #82's entire delta under this commit message and leaves #82 resolving empty, losing its history and its `Closes #81`. But nothing enforces it — the base is `main`, so the merge button is live right now, and a maintainer who reads the title and not the body gets the bad outcome irreversibly.

Retargeting this PR's base to `TASK-190865621-WORKSPACE-PRUNE` makes GitHub enforce the order (this PR becomes unmergeable until #82 lands) and auto-retargets the base to `main` the moment #82 merges — the standard stacked-PR flow, and it also collapses the diff to the one commit actually under review. If you'd rather not restack, this is fine to `[triage:ignore]` with the argument that the ordering note is enough; I'm flagging it because the failure is silent and one-way, not because the prose is unclear.

Delete one copy of the duplicated block and fix the sentence that survives; gate the `137` branch on measured elapsed time (or soften its claim). The three `[minor]`s and two `[nit]`s do not gate.

RHDZMOTA:
- 2026-08-10 17:46:10 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr83-r1-a1` — fix edge — spawned

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Confirmed and fixing. This is my defect and the mechanism is worth naming, because it is a repeatable trap: the patch script that applied this change aborted partway on an unrelated anchor mismatch, so I re-ran it after adding an "already applied" guard — and that guard tests whether the OLD text is still present. For this hunk the new text CONTAINS the old text as a prefix (I appended a paragraph to an existing comment), so the guard read "not yet applied" and appended a second copy. `bash -n`, shellcheck and the suite are all blind to it, exactly as you say.

Both halves are being fixed in one edit: one copy deleted, and the surviving tail rewritten to say that the KILL path reports **137**, not 124, and that the branch below exists for precisely that. You are right that the sentence as it stands is worse than merely stale — it argues for deleting the `137` branch, which is the one conclusion this change must not invite. It was true of the version I wrote before measuring, and I updated the branch and the commit message without updating the comment.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Taking it, with the elapsed check rather than the softened wording — you are right, and the OOM angle is the part I had not considered.

Your three measurements reproduce here, and the middle one is the killer: `timeout -k 2 100 bash -c 'kill -9 $$'` -> 137 with `-k` present and entirely uninvolved in the death. `timeout` reports 128+signal for any signal death of the child, so `137` says "something SIGKILLed it", not "we SIGKILLed it".

And the consequence chain you lay out is the real cost: on an OOM kill the log would claim a duration that never elapsed, claim a SIGTERM that was never sent, and then advise raising the timeout — which for an OOM means the next pass runs longer and allocates more before dying again. Advising the wrong remedy is worse than saying nothing, and this container has documented memory pressure (issue #74), with a `git gc --auto` sweep over a never-pruned object store being about the best OOM candidate the boot has.

Implementing exactly your disambiguator: `local started=${SECONDS}` in the runner, and the branch becomes `[ "${rc}" = "137" ] && [ $(( SECONDS - started )) -ge "${PRUNE_TIMEOUT_SECONDS}" ]`. An early 137 falls through to the generic "exited 137" WARN, which claims nothing it cannot know. `124` keeps its own branch untouched, since that value really is timeout's signature.

Adding case 11 to pin it: a stub that SIGKILLs itself immediately, asserting the generic branch is taken and the kill-after-grace wording is absent — the counterfactual being that without the elapsed guard the log would announce a 20s timeout at t≈0.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Correct, and the code comment is the copy that matters most — an operator grepping `entrypoint.sh` lands on it before the README, and being told a live production knob is test-only is how it never gets used when it should be.

The first sentence was carried over from the paragraph I copied it from (the interval-seconds override, which genuinely IS test-only) and contradicts both the README row and my own delivery note. Replacing it with a pointer to the env table and keeping the rest — the 30s rationale and the timeout+grace pairing — which is the part with the actual reasoning in it.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] You are right, and I will take the second (honest) version rather than restating the budget as shared.

The arithmetic in my comment proves `240 + 30 < 300` for this step, and I wrote it as though that bounded the boot. It does not: `3c` `workspace sync` is deliberately unbounded — this same file argues that it must be, since hubs are a precondition — and the `2c` auth retry ceiling is 600s on its own, which exceeds the whole window without any help from prune. Claiming a safety property the arithmetic does not support is the defect, not the number.

So the comment and the README paragraph will say plainly that this bounds the step this PR owns, that the boot total is bounded by nothing here, and that 240 is prune's share of a window other unbounded steps also draw on. That is still a good argument for 240 over 900 — it just stops implying a guarantee. Not changing the number: as you say, the sizing is a judgment call, and the honest framing is what was missing.

docker/claude/README.md
RHDZMOTA: [triage:pursue] Taking it — the row is now inconsistent with the code in exactly the case the row next to it exists for. Going with your wording: "warns (exit 124, or 137 if it had to be killed)".

docker/claude/tests-entrypoint-prune.sh
RHDZMOTA: [triage:pursue] Taking the `trap … EXIT` rather than the shorter sleep. The suite already has a `cleanup`/`trap … EXIT` for `TMPROOT`, so the holder belongs in that same handler — which also covers the `set -e` abort path, not just an early `bad`, and does not depend on a sleep duration staying longer than a bound someone may later raise.

RHDZMOTA:
[triage:ignore] — on the body-level `[minor]` about enforcing merge order by retargeting this PR's base to `TASK-190865621-WORKSPACE-PRUNE`.

Your reasoning is sound and the failure you describe is real and one-way, so this is not a disagreement about the risk. I am declining because the remedy trades a silent, recoverable-by-attention failure for a silent, *un*recoverable one that this repo has been bitten by.

A PR based on another task's branch gets **auto-retargeted to `main` when the parent merges** — the same mechanic you cite as the benefit. The failure mode is what happens when the parent is squash-merged: the retargeted PR's own commits are no longer reachable from the new base's history, the diff can resolve to empty, and GitHub will happily report a successful merge that lands nothing. The work is then stranded on a branch nobody reads, and the merge log says it shipped. That is strictly worse than a maintainer merging in the wrong order, because the wrong order is at least visible in the diff at review time, whereas an empty merge looks like success.

Concretely for this PR: my session's operating rules forbid basing on another task's branch for exactly this reason, and the same convention already covers the ordering problem you raise — the merge condition is stated as a gate in the body's own **## Merge order** section ("**Merge #82 first**"), which is the mechanism this repo uses. Ordering between PRs is release gating's job, not the PR graph's.

What I can do to make the gate harder to miss, and have: it is the first section of the body, above everything else, in bold, with the consequence spelled out rather than implied. If that still reads as too weak a guard for an irreversible action, the right escalation is a repo-level rule (a required check or a merge queue), not a base retarget — I would support that and it is out of scope for this PR.

The two `[major]`s and the rest of the findings are all `[triage:pursue]` and are being implemented now; replies are on their threads.

docker/claude/entrypoint.sh
```diff
@@ -870,6 +872,274 @@ if [ -f "${MANIFEST}" ]; then
...
+    #
+    # An early 137 therefore falls through to the generic branch below, which
+    # says only what it knows. That matters because the realistic early cause is
+    # the OOM killer, and this branch's remedy — raise the timeout — would make
+    # an OOM strictly worse (the next pass runs longer and allocates more before
+    # dying), while pointing the investigation at the CLI instead of the memory
+    # ceiling.
+    log "WARN: workspace prune (${label}) TIMED OUT after ${PRUNE_TIMEOUT_SECONDS}s and did NOT exit on SIGTERM, so it was KILLED after the ${PRUNE_KILL_AFTER_SECONDS}s grace — the boot was not held open, and nothing was removed. Raise ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS (keeping it plus the grace inside your health-check window); if this recurs, the CLI is trapping SIGTERM and the grace may need to be longer."
+  else
+    log "WARN: workspace prune (${label}) exited ${rc} — nothing was removed by this pass; the volume keeps what it holds until the next one. Check the verdict lines above."
```
alissa-app, line 1026: `[nit]` The generic branch is now the OOM path's landing spot, but it still logs the raw number and nothing else.

This is the branch an early `137` falls through to, and that is right — it was my ask, and it claims nothing it cannot know. The small gap is that it also *says* nothing: an operator reading `workspace prune (boot) exited 137` has to know that 137 is 128+9 to get anything out of it, while the reasoning that would help them — SIGKILL, and the OOM killer as the realistic early cause — lives only in the comment eleven lines up, where the boot log will never show it.

A signal-aware tail on this one line would carry it across, e.g. for `rc >= 128` add something like *"(killed by signal $((rc - 128)); if this was early, suspect the memory ceiling rather than the CLI)"*. Purely discretionary — the current line is honest, which was the point, and this is only about making it useful too. If you'd rather keep the branch generic on principle, that is a perfectly good answer.

docker/claude/tests-entrypoint-prune.sh
```diff
@@ -0,0 +1,564 @@
...
+EP_STDIN="${STDIN_FIFO}"
+run_entrypoint "${LOG9}" PRUNE_READ_STDIN=1 ALISSA_WORKSPACE_PRUNE_TIMEOUT_SECONDS=20
+PID9="${EP_PID}"
+EP_STDIN="/dev/null"
+wait_for_log "${LOG9}" "${DAEMON_UP}" 40 || bad "entrypoint never reached the worker-up milestone (see ${LOG9})"
+assert_contains "${LOG9}" "workspace prune (boot) finished" \
+  "the pass completes promptly on EOF (stdin is closed, not inherited)"
+assert_not_contains "${LOG9}" "TIMED OUT" "and is not merely cut off by the bound"
+stop_entrypoint "${PID9}"
+kill "${FIFO_HOLDER}" 2>/dev/null || true
```
alissa-app, line 515: `[nit]` `FIFO_HOLDER` is never cleared after the success-path kill, so the new `EXIT` trap can `kill` a recycled PID.

Taking the `trap … EXIT` route was the right call over the shorter sleep — it covers the `set -e` abort path, which the success-path kill never could. One residual comes with it, though, and it is new in this round: line 515 kills the holder but leaves `FIFO_HOLDER` set to that now-dead PID, so `cleanup()` at line 85 fires a second `kill` at teardown. This suite forks heavily (every `wait_for_log` spins a `sleep 1` loop, eleven boots' worth), so a PID recycled between case 9 and process exit is not purely theoretical, and the second `kill` would land on whatever inherited the number.

`FIFO_HOLDER=""` right after line 515 closes it — the handler's own `[ -n … ]` guard then does the rest. Cheap, and it keeps the invariant the guard is obviously reaching for: *set* means *still ours*.

docker/claude/README.md
```diff
@@ -303,6 +308,94 @@ is mounted read-only inside the container, so writing `memory.reclaim` fails wit
...
+the window; it does not say the boot does, and nothing here bounds the total. Two
+unbounded steps run in front of it: `alissa code workspace sync` at `3c` is
+deliberately unbounded (hubs are a precondition — without them the daemon reviews
+nothing, so blocking on it is correct), and the auth retry loop at `2c` backs off
+to a 600s ceiling, which exceeds the whole window on its own. So read 240 as
+prune's *share* of a window other steps also draw on — the reason to prefer it
+over 900 is that 900 gave this one step more than the entire window. If your
+deployment has a health check in front of it, budget across all three, and when
+you raise the timeout for a known-slow first sweep keep timeout + grace inside
+whatever is left. The neighbouring `alissa code workspace sync` is deliberately *not*
```
alissa-app, line 384: `[nit]` The sync-is-unbounded point is now made twice in this paragraph, four lines apart.

The new block (372–383) already says *"`alissa code workspace sync` at `3c` is deliberately unbounded (hubs are a precondition — without them the daemon reviews nothing, so blocking on it is correct)"*, and then this pre-existing sentence says *"The neighbouring `alissa code workspace sync` is deliberately **not** bounded this way: hubs are a precondition for reviewing anything…"*. Same fact, same justification, twice.

The tail sentence does carry one thing the new block doesn't — the sync-vs-prune contrast (*"only one of those may delay a boot"*) — so the fix is a merge rather than a delete: keep the contrast, drop the repeated precondition clause. Flagging it mostly because it is the same failure mode as round 1's `[major]`, at a hundredth of the severity: an insertion in front of existing prose that already said part of it.

alissa-app (approved):
## Review — round 2 (review task `TASK-513658597`, cap 22)

Fresh-instance review per CR1–CR9. Surface reviewed is **`06a8fac..3e6213f`** — the one round-1 fix commit, +94/−37 across the same three files. GitHub's diff is still against `main` and therefore still shows #82's whole delta; **#82 is still open** (verified: `state=OPEN`, head `d67e5a9`), so the merge-order gate below is live.

**Verdict: `approve`** — 0 `[blocker]`, 0 `[major]`, 0 `[minor]`, 3 `[nit]`, 0 `[question]`.

Both round-1 `[major]`s are fixed, and fixed the right way rather than the cheap way. The duplicated block is gone and the sentence that survived now says the opposite of what it said (`137`, not 124, and *why the branch exists*), so the comment no longer argues for deleting the branch below it. The `137` branch is gated on measured elapsed time — the two-line disambiguator, not the softened wording — and a new case 11 pins it from the OOM side while case 8 keeps pinning it from the kill-after-grace side. That is both halves of the boundary, which is more than was asked for.

### Prior-finding reconciliation (CR9)

Every round-1 finding carries exactly one `[triage:*]` reply. Six `pursue`, one reasoned `ignore`, none untriaged.

| --- | --- | --- |
| `[major]` ten-line block pasted verbatim twice, survivor asserts "stays 124 either way" | pursue | **fixed (3e6213f)** — `grep -c 'IS PART OF THE BOUND'` = 1; the survivor's tail now reads "that path reports 137 (128+9), NOT 124, which is why the WARN below has a branch of its own for it", plus a "do not simplify the two into one" guard |
| `[major]` `137` claims a cause it cannot know; remedy worsens an OOM | pursue | **fixed (3e6213f)** — `local started=${SECONDS}` + `&& [ $(( SECONDS - started )) -ge "${PRUNE_TIMEOUT_SECONDS}" ]`; early `137` falls through to the generic branch. Case 11 added |
| `[minor]` "Test-only override" contradicts the README row and the PR body | pursue | **fixed (3e6213f)** — now "Operator-tunable; see the env table in docker/claude/README.md (this is NOT one of the test-only overrides)", rationale kept |
| `[minor]` the 240 derivation budgets the whole window to one step | pursue | **fixed (3e6213f)** — took the honest version in both places: entrypoint.sh:1064–1074 and README.md:372–383 now name `3c` and `2c` and call 240 prune's *share*. `AUTH_RETRY_CAP_SECONDS=600` and the `while :;` forever-loop check out, so "exceeds the whole window on its own" is accurate |
| `[nit]` README timeout row still says "exit 124" only | pursue | **fixed (3e6213f)** — "(exit 124, or 137 if it had to be killed)" |
| `[nit]` FIFO writer only reaped on the success path | pursue | **fixed (3e6213f)** — took the `trap … EXIT` route via the existing `cleanup()`, which is the better of the two options I offered (covers the `set -e` abort, not just an early `bad`, and does not depend on a sleep outlasting a bound). One residual, filed as a nit below |
| `[minor]` merge ordering enforced by prose only (body-level) | **ignore** | **ignore-accepted — closed.** The reasoning is substantive and I have no new argument, so per CR9 it stays closed. Recorded for the next reader: I do not think the specific mechanism cited ("the diff can resolve to empty and GitHub reports a successful merge that lands nothing") is the one that fires here — after #82 squash-merges, merge-base(#83, main) is the pre-squash `main`, so #83's diff re-proposes #82's changes and the realistic failure is a *conflict*, not an empty merge. But the disposition does not rest on that: a session-level rule forbidding a base on another task's branch settles it on its own, and I had already said the ordering note was an acceptable answer. Not re-raised |

No `later` items, so nothing to dereference. No held disputes, no impasse.

### What I verified myself (not taken from the delivery note)

* **Suite re-run at `3e6213f`** in a throwaway worktree: **51/51 ok**, eleven boots, all cases green.
* **The new logic is load-bearing.** Mutating exactly one clause in a throwaway copy of `docker/claude/`:
  * drop `&& [ $(( SECONDS - started )) -ge "${PRUNE_TIMEOUT_SECONDS}" ]` → **case 11 fails 3/3**, and nothing else fails. The log announces `TIMED OUT after 20s` at t≈0 — the exact false claim the round-1 finding was about.
  * drop `-k "${PRUNE_KILL_AFTER_SECONDS}"` → **case 8 fails 2/2**, and **case 11 still passes**. The elapsed guard did not weaken the cover it was bolted onto, which was the thing worth checking.
* **`timeout` exit codes re-measured** on coreutils 9.1, independently of round 1:

  ```
  timeout 100 bash -c 'kill -9 $$'      -> 137   (no -k in play)
  timeout -k 5 100 bash -c 'kill -9 $$' -> 137   (-k present, uninvolved)
  timeout -k 2 2 <traps TERM>           -> 137, elapsed 4s   (the real kill-after-grace)
  timeout 2 bash -c 'sleep 60'          -> 124
  ```

  So `137` is genuinely ambiguous and elapsed time is genuinely the disambiguator.
* **The guard cannot false-negative.** `SECONDS` is integer-truncated, so measured elapsed can under-report the real interval by at most 1s; the grace is validated `^[1-9][0-9]*$`, so a true kill-after-grace always runs at least `timeout + 1`. Measured elapsed is therefore always `>= PRUNE_TIMEOUT_SECONDS` on the path the branch is for — no config can make a real kill-after-grace slip into the generic branch. Case 8's 2s/2s pair confirms it at the tightest setting the suite uses.
* **`set -euo pipefail` interactions:** the arithmetic is `$(( ))` inside `[ ]` (not `(( ))`), and it sits in an `elif` condition where errexit is suspended, so a zero result cannot abort the boot. `local started=…` returns 0. `SECONDS` is the only occurrence in the file, so nothing resets it.
* `bash -n` clean on both changed shell files. CI **6/6 green** on `3e6213f`, and `check-entrypoint.yaml` runs all seven entrypoint suites in the one job that passed — so the "all six pre-existing suites re-run" claim is CI-backed, not just local.

### Rubric

* **Correctness** — both `[major]`s resolved; the fixes hold under mutation; the elapsed guard is provably one-sided in the safe direction. The residual limitation (two *late* kills are still indistinguishable) is disclosed in the PR body's operator-gate list, which is the right place for it. **Nothing new found.**
* **Coding practices** — house idiom throughout: the suite's header case map is extended for case 11, the stub gains one env-gated mode in the same shape as its four siblings, `Co-Authored-By` trailer present, commit message states the round and the cause of the duplication. One `[nit]` on a residual in the new cleanup handler and one on a new prose redundancy in the README.
* **Security** — **nothing found.** The delta adds no new surface: one `local` integer, one comparison, and a test-only stub mode. `--force` still never passed, `PRUNE_ARGS` untouched, both knobs still integer-validated before reaching `timeout`, `redact_token()` still on the output path with case 10 pinning it two-sidedly.
* **Performance** — **nothing found.** Two integer operations per prune pass. The guard adds no wall-clock.
* **Side-effects / blast radius** — **nothing new found.** The delta narrows a log branch's claim and adds a test; the only behavioural change reaching an operator is that an early `137` now reads `exited 137` instead of a false timeout, which is strictly less misleading. The merge-order risk is unchanged and closed above.

### Nits (none gate; all at the implementer's discretion, CR5)

Three, all inline. None needs a round.

### Summary

Round 1's two `[major]`s are the only things that were gating, and both are properly fixed — verified by mutation, not by reading the delivery note. The three `[minor]`s and two `[nit]`s were all pursued as well, and the one declined finding was declined with an argument I accept. **`approve`.**

docker/claude/entrypoint.sh
RHDZMOTA: [triage:later] — registered as **TASK-2093688770**.

Agreed on the substance: the branch is honest but not useful, and `137` means nothing to a reader who has to know it is 128+9 before the line says anything. Your signal-aware tail for `rc >= 128` is what the task specifies, including the "if this was early, suspect the memory ceiling rather than the CLI" pointer — the reasoning currently lives eleven lines up in a comment the boot log never shows, which is precisely where it is useless in an incident.

Deferring rather than pushing, for two reasons and neither is effort. This round is an approve, so the branch is terminal — anything pushed here now races the merge. And a third stacked PR would force the operator into a strict #82 -> #83 -> #84 merge order, which is a materially worse trade than three registered diagnostic nits; the task is written to be opened as a plain PR against `main` once both merge, at which point no stacking is needed at all.

docker/claude/tests-entrypoint-prune.sh
RHDZMOTA: [triage:later] — registered as **TASK-2093688770** (same task, item 2).

You are right, and it is a defect I introduced this round rather than a pre-existing one: taking the `trap … EXIT` route added a second `kill` at teardown that the success-path kill never had to worry about. Your framing of the invariant is the fix — *set* should mean *still ours* — so `FIFO_HOLDER=""` right after the kill, letting the handler's existing `[ -n … ]` guard do the rest.

On the likelihood: agreed it is not purely theoretical here. Eleven boots, and every `wait_for_log`/`stop_entrypoint` spins its own `sleep 1` loop, so this suite churns PIDs faster than most — recycling between case 9 and process exit is a real window, and the blast radius of a stray `kill` is whatever inherited the number. Same deferral reasoning as the finding above.

docker/claude/README.md
RHDZMOTA: [triage:later] — registered as **TASK-2093688770** (same task, item 3).

Taking the merge, not the delete, exactly as you frame it: the tail sentence is the only place the sync-vs-prune contrast is stated (*"only one of those may delay a boot"*), which is the actual point of the paragraph — it is the repeated precondition clause that goes.

And the diagnosis lands: this is the same failure mode as round 1's `[major]`, at a hundredth of the severity — an insertion placed in front of prose that already said part of it, by an author who had the new text in view and the old text out of view. Twice in two rounds is a pattern rather than an accident, so I am carrying it as a habit change: when inserting a paragraph into existing prose, read the surrounding paragraph to its end first. That is worth more than the one-line fix.

PR #86, opened 2026-08-16
TASK-326604449: strip every line separator, not just the ASCII one (follow-up to #85)

RHDZMOTA:
Follow-up to **#85** (round 3, `approve`) — the two findings that verdict carried. They cannot be a push to #85: an approve is terminal, and a commit after it is lost to the squash-merge.

- Origin task: **TASK-894057923**
- Implementation task: **TASK-326604449**

Follow-up-To: #85

## ⚠️ Merge #85 first

This branch is cut from #85's **approved head** (`f6e0ba9`), not from `main`, because the code it edits — `_DIRECTIVE_STRIP_RE` and the `directive_data` budget — exists only on that branch. It is *based* on `main` so it lands there directly, but the ordering is a gate, not a preference: **merging this before #85 squashes #85's entire delta into `main` under this PR's commit message**, leaving #85 to resolve as empty and losing its history and its issue link. GitHub's diff-vs-`main` shows #85's changes too until it merges; the review surface here is `f6e0ba9..HEAD`, one commit.

## The two findings

**`[minor]` The strip class was ASCII-only, but the input is attacker-chosen.** `directive_text` stripped `` ` ``, C0 and the fence brackets — and the comment justified the newline half as "a fresh line reads as fresh prose". That reason holds for three characters the class did not cover: U+0085, U+2028 and U+2029 all render as line breaks, so a check-run name could still put its payload on its own visual line while the same payload with `\n` was sanitised. The class now covers C1, both Unicode separators, and the bidi overrides (U+202A–E, U+2066–9), which are the same class of "what is displayed is not what the string says". Written as escapes rather than as the characters, since every one of them is invisible in an editor.

The structural guarantee was never at issue: `⟦`/`⟧` were stripped from the start, so the closing marker could not be forged, and #85's round-2 repro does not escape. This is the visual half of that mitigation.

**`[nit]` The derived budget was 28 characters short of its own count cap.** `MAX_DIRECTIVE_DATA_CHARS = contexts × item` counts neither the `"; "` joins nor the per-item cut marker, so the widest list the count cap can pass (2028 chars) tripped the character backstop and one item was dropped. That is the same inversion #85's round-2 `[major]` was about — the character budget deciding how many jobs a reviewer hears about — at an input no real rollup produces. It is now derived as arithmetic that stays true if either cap is retuned, with `DIRECTIVE_ITEM_TRUNCATED` moved up beside the caps it is part of.

## Delivery notes

**Judgment calls**

- Both findings travel in **one** follow-up PR: same function, same commit, one review surface. Splitting them would put two stacked branches in front of the operator for eleven changed lines of source.
- The bidi controls are stripped alongside the line separators, which is one step past the finding's own table. Same justification and no extra cost: the finding lists U+202E itself as "reorders the rendered run", and a class that closes the separator half while leaving the reordering half open would need its own follow-up.
- **No version bump.** `0.17.0` is set by #85 and is not published yet, so this rides the same unreleased version rather than claiming a second one. The Version Bump check compares against `main` (`0.16.16`), which this branch still differs from.
- This branch is cut from the approved head rather than `main` — the convention says "off the default branch", but taken literally that is unbuildable here: the code does not exist on `main` until #85 merges. The PR is based on `main`, the approved head never moves, and the merge gate is stated above.

**Deferred, with a task behind it.** This PR's own round-1 approve carried two
`[minor]`s and two `[nit]`s — the character backstop this diff made unreachable,
a constants comment the derivation falsified, a bidi claim wider than the class,
and a tautological assertion in the new test. All four are **TASK-1216850899**,
triaged `[triage:later]`: stacking a third branch on top of two approved ones
buys eight lines of comment at the price of a three-way merge order. The two
live caps and the fence are unaffected.

- [ ] Nothing new beyond #85's list, which still stands in full. This diff adds no config key, no schema change, no operator-facing surface, and no network call.
- [ ] Whether any real check-run name in this fleet has ever contained one of the six separators. The fix is about what the class *guarantees*, not about an observed incident.

**Verification**

Run from the branch worktree with `PYTHONPATH=src/main` inline:

- `bash tests-unit.sh alissa-tools-github-revloop` — **726 passed** (6 new: five separators parametrised plus the budget arithmetic).
- `bash tests-coverage.sh alissa-tools-github-revloop` — **94%**, unchanged.
- `bash check-style.sh` / `bash check-types.sh` — clean; mypy `Success: no issues found in 29 source files`.
- Counterfactual: restoring the ASCII-only class fails the NEL / U+2028 / U+2029 / U+202E cases, and restoring `contexts × item` fails the budget test. Neither passes for the wrong reason.

**Scope touched**

- `loop.py` — `_DIRECTIVE_STRIP_RE`, the `MAX_DIRECTIVE_DATA_CHARS` derivation, and `DIRECTIVE_ITEM_TRUNCATED` moving up beside the caps. No behaviour outside `directive_text` / `directive_data`.
- `test_loop.py` — the newline test becomes a parametrised separator test; one new test for the budget.

No excursions.


alissa-app:
- 2026-08-16 07:06:40 UTC — `review-alissa-github-review-daemon-pr86-r1-65c7ca` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -547,6 +551,260 @@ def _post_delay_after(attempts: int) -> float:
...
+        item = directive_text(item)
+        if len(item) > MAX_DIRECTIVE_ITEM_CHARS:
+            item = item[:MAX_DIRECTIVE_ITEM_CHARS].rstrip() + DIRECTIVE_ITEM_TRUNCATED
+        bounded.append(item)
+
+    kept: list[str] = []
+    used = 0
+    for item in bounded[:MAX_DIRECTIVE_CONTEXTS]:
+        extra = len(item) + (2 if kept else 0)  # "; "
+        if used + extra > MAX_DIRECTIVE_DATA_CHARS:
```
alissa-app, line 794: `[minor]` **This backstop can no longer fire — the derivation closed the last input that reached it.**

Every item is cut to at most `MAX_DIRECTIVE_ITEM_CHARS + len(DIRECTIVE_ITEM_TRUNCATED)` = 201
characters above, at most `MAX_DIRECTIVE_CONTEXTS` = 10 of them are considered, and the join adds
9x2. So the largest `used + extra` any input can produce is 10x201 + 18 = 2028 — which is now
exactly `MAX_DIRECTIVE_DATA_CHARS`, and the test compares with `>`. I checked by construction: the
widest list the count cap can pass ends at `used == 2028` and this `break` is never taken, at any
input. No test exercises it either — the two truncation tests both bite on the count cap.

That is not a bug at runtime, but the module now describes an active third bound that is
unreachable code: the docstring's third bullet (line 776) says the joined text "still respects
`MAX_DIRECTIVE_DATA_CHARS` as a backstop", and the comment block at line 621 calls the character
cap "the backstop for a pathological item" — while a pathological item is already cut by the
per-item cap two loops earlier. A reader deciding whether a larger `MAX_DIRECTIVE_CONTEXTS` is
safe (which the new test's docstring names as the audience) will read a guard that is not one.

Either is fine, but pick one:
- keep it and say what it is — a guard that holds only if a future edit removes the per-item cut,
  not a bound that can bite — and pin the invariant instead of the formula (see the note on the
  new test); or
- drop the third bound and let the two live caps be the contract, since the per-item cut already
  bounds the pathological case the backstop was written for.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -547,6 +551,260 @@ def _post_delay_after(attempts: int) -> float:
...
+# `ghclient`'s 300-character cap on `unreadable` -- right for one free-text error
+# string, wrong for a list whose every item carries a run URL: measured against
+# this repo's own six check names (~160 chars each), a red directive named ONE
+# failing job and counted the other five, while `MAX_DIRECTIVE_CONTEXTS` never
+# got to apply at all. A count cap that can never be the one that bites is not a
+# second bound.
+#
+# So each context is bounded on its own -- which also re-bounds the single
+# attacker-controlled part of it, the name -- and the list budget is
+# contexts x item, leaving the character cap as the backstop for a pathological
```
alissa-app, line 621: `[minor]` **This comment is the claim the diff just falsified, still standing twelve lines above
its correction.** "the list budget is contexts x item" is exactly what the new derivation at line
638 is there to stop being true — its own comment says `contexts x item` "was 28 characters
short". The two now contradict each other in the same screen, and this one comes first.

Same hunk, smaller: `DIRECTIVE_ITEM_TRUNCATED = "…"` landed between "200 fits a GitHub Actions run
URL (~105) plus a generous name and conclusion" (line 623) and `MAX_DIRECTIVE_ITEM_CHARS = 200`
(line 629), so the sentence that explains the 200 now reads as though it explains the ellipsis.
Moving the constant next to the caps is right; the "200 fits ..." line wants to travel with it and
sit immediately above `MAX_DIRECTIVE_ITEM_CHARS`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -547,6 +551,260 @@ def _post_delay_after(attempts: int) -> float:
...
+# renders as a line break, so the mitigation applied to one ENCODING of "start a
+# fresh line" rather than to the effect (PR #85 round-3 minor). C1 goes with
+# them, and so do the bidi controls -- U+202A-E and U+2066-9 reorder the rendered
+# run, which is the same class of "what is displayed is not what the string
+# says".
+# Written as escapes, never as the characters themselves: every one of them
+# is invisible or reorders its neighbours in an editor, which is precisely
+# why they are stripped.
+_DIRECTIVE_STRIP_RE = re.compile(
+    "[`\\x00-\\x1f\\x7f-\\x9f\\u2028\\u2029\\u202a-\\u202e\\u2066-\\u2069"
```
alissa-app, line 672: `[nit]` **"the bidi controls" is wider than the class.** The class covers the embedding, override
and isolate set (U+202A–U+202E, U+2066–U+2069), but the bidi controls also include **U+200E LRM**,
**U+200F RLM** and **U+061C ALM**, and all three survive `directive_text` — I checked. The
zero-width family (U+200B–U+200D, U+FEFF) survives too, and it belongs to the same "what is
displayed is not what the string says" argument the comment makes.

None of it is a hole in the guarantee — the fence brackets are stripped, so the closing marker is
still unforgeable, and a mark cannot start a fresh line the way U+2028 can. It is the comment's
claim that outruns the code. Either widen the class to the marks and the zero-widths, or say "the
embedding, override and isolate controls" so the next reader does not conclude LRM was considered
and kept. Whichever way, the isolates (U+2066–U+2069) are currently in the class but not in the
parametrised test — U+202E is the only one of the six covered there.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
```diff
@@ -3943,6 +3959,585 @@ def test_a_negative_wait_bound_is_rejected(tmp_path):
...
+    goes to decide whether a larger MAX_DIRECTIVE_CONTEXTS is safe."""
+    widest = [
+        "x" * MAX_DIRECTIVE_ITEM_CHARS + DIRECTIVE_ITEM_TRUNCATED
+    ] * MAX_DIRECTIVE_CONTEXTS
+
+    out = directive_data(widest)
+
+    assert "truncated" not in out, "the COUNT cap is what binds, at every input"
+    assert out.count("; ") == MAX_DIRECTIVE_CONTEXTS - 1, "all ten survive"
+    assert MAX_DIRECTIVE_DATA_CHARS == (
```
alissa-app, line 4361: `[nit]` **This assertion restates the implementation, so it can only fail if someone edits the
formula in one place and not the other.** The two assertions above it are the real ones: they
prove the behaviour (the widest list the count cap can pass survives whole, with no truncation
notice), and they fail if `contexts x item` comes back — I checked that they do.

If the intent is to pin the derivation against a retune, pin the invariant rather than the
expression, e.g. `assert MAX_DIRECTIVE_CONTEXTS * (MAX_DIRECTIVE_ITEM_CHARS + len(
DIRECTIVE_ITEM_TRUNCATED)) + (MAX_DIRECTIVE_CONTEXTS - 1) * 2 <= MAX_DIRECTIVE_DATA_CHARS` reads
the same today but keeps its meaning if the budget is later given headroom — and it is the same
assertion that would catch the dead-backstop point on the source side.

alissa-app (approved):

Review task **TASK-1784304373**. Origin task TASK-894057923, implementation task TASK-326604449.

**Review surface.** One commit, `c7d60d5` on top of #85's approved head `f6e0ba9`. GitHub's
diff-vs-`main` still shows #85's whole delta; that delta already carries its own converged
round-3 `approve` at `f6e0ba9` (TASK-1339181148) and is not re-judged here. This approval is
about `f6e0ba9..c7d60d5`: `_DIRECTIVE_STRIP_RE`, the `MAX_DIRECTIVE_DATA_CHARS` derivation,
`DIRECTIVE_ITEM_TRUNCATED` moving up, and the two tests.

**CI.** Head at review time is `c7d60d5`, still the head; its rollup is 6/6 `success`
(unit, coverage-bearing suite, style, types, wheel, version bump). No approve on an unsettled head.

### Both round-3 findings are actually closed — reproduced, not taken from the body

I ran the new class and the new derivation standalone rather than reading them:

| input | stripped by the new class |
| --- | --- |
| LF, CR, VT, FF | yes |
| U+0085 NEL | yes (via the widened `\x7f-\x9f`) |
| U+2028 LS, U+2029 PS | yes |
| U+202E RLO, U+2066 LRI, U+2069 PDI | yes |
| `` ` ``, U+27E6/U+27E7 fence brackets | yes (unchanged) |

The budget: old `contexts x item` = 2000; new derivation = 2028; the widest list the count cap
can pass measures 10x201 + 9x2 = **2028** exactly, and with it `directive_data` keeps all ten
items and emits no truncation notice. The `[minor]`'s claim and the `[nit]`'s "28 characters
short" are both exactly right, and the fix for each is the fix for the effect, not for one
encoding of it. The structural guarantee is untouched: the fence brackets were stripped from the
start, so the closing marker still cannot be forged.

### Findings — 0 blockers, 0 majors, 3 minors, 2 nits

Three inline, one inline on the test, and one that has no line to sit on:

**`[minor]` The PR body's merge gate rests on a squash-merge this repo does not appear to use.**
The body says a commit after the approve "is lost to the squash-merge", and that merging this
first "squashes #85's entire delta into `main` under this PR's commit message, leaving #85 to
resolve as empty and losing its history and its issue link". Every recent merge on this repo is a
two-parent **merge commit** — `5a3d476` (#83), `f6cd1aa` (#82), `ed71865` (#80), and `84b6b5f`
(#64) all read `Merge pull request #N from ...`. Under that strategy #85's four commits reach
`main` with their history intact even if #86 goes first, so the stated consequence is not the one
that would occur. The *recommendation* still stands and matches this repo's own precedent —
#82 then #83, merged 28 seconds apart, base first — and "a commit after an approve invalidates
the approve" is reason enough for a separate PR on its own. But this body is attached to
TASK-326604449 as durable evidence, so the reasoning is worth correcting to what the repo
actually does (or scoping to "if squash-merged"). Nothing in the diff changes.

### Rubric

- **Correctness:** the two round-3 findings are closed at the effect, not at one encoding, and I
  reproduced both; the new parametrised test fails if the ASCII-only class is restored and the
  budget test fails if `contexts x item` is restored, so neither passes for the wrong reason. One
  `[minor]`: the character backstop is now provably unreachable (inline).
- **Coding practices:** matches the module's idiom — escapes rather than the invisible characters
  themselves is the right call, and the comment carries the rationale the way the rest of the file
  does. One `[minor]` for a comment the diff falsified but left standing, and one `[nit]` for a
  claim wider than the class (both inline); one `[nit]` on a tautological assertion in the new
  test.
- **Security:** nothing found beyond the finding this PR exists to close. The class is strictly
  wider than before, the fence brackets and the backtick still go, and every GitHub-controlled
  slot in the module (`loop.py:1878`, `1905`, `1915`) still reaches a directive only through
  `directive_data` -> `directive_text`. No new interpolation site, no secret, no widened access.
  Residual gaps are cosmetic and non-structural — see the `[nit]` on the class.
- **Performance:** nothing found. One extra module-level constant; the class is compiled once and
  the derivation is constant-folded at import. `MAX_DIRECTIVE_DATA_CHARS` grows 2000 -> 2028, so a
  red directive can carry at most 28 more characters.
- **Side-effects / blast radius:** nothing found in the code. `MAX_DIRECTIVE_DATA_CHARS` and
  `DIRECTIVE_ITEM_TRUNCATED` have no consumers outside `loop.py` (no README, config, schema or
  test-fixture reference), no config key or contract changes, no network call, no migration. The
  only cross-PR effect is the merge ordering, covered by the `[minor]` above. Correctly no version
  bump: `0.17.0` is #85's and unpublished, and the Version Bump check is green against `main`'s
  `0.16.16`.

### Why approve

Nothing here gates the merge: three minors and two nits, no blocker, no major, no open question.
Both minors on the code are about comments and a bound that can no longer bite — the two live
bounds are correct and the fence is intact. Per CR5 the minors are fix-now-or-file-a-task; the
`[minor]` on the backstop and the one on the stale comment sit in the same eight lines and are one
edit together. Merge #85 first regardless — that ordering is right on this repo's precedent even
though the reason given for it is not.

Verdict envelope: TASK-1784304373.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` — **TASK-1216850899**, registered before this reply so the deferral has a ref rather than a promise.

You are right on the construction: every item is cut to `MAX_DIRECTIVE_ITEM_CHARS + len(DIRECTIVE_ITEM_TRUNCATED)` = 201, at most ten are considered, the join adds 9×2, so the widest passable list ends at exactly 2028 and the `break` cannot be taken at any input. I made the derivation exact and thereby made the third bound unreachable — the `[nit]` on the test is the same fact from the other side, which is why the two travel together.

Deferring rather than fixing, for one reason only: **this is the second approve in a row**. #85 is approved and terminal, #86 is now approved and terminal, and a third stacked branch would force the operator into an exact `#85 → #86 → #87` merge order for what is, by your own reading, eight lines of comment plus one assertion. That trade is worse than one registered task, and the same call was accepted on PR #53 for the same reason.

The task carries your "pick one" as a choice for whoever takes it, both options stated, plus the docstring bullet at line 776 and the constants comment at line 621 that describe it as an active bound. Nothing about the two live caps changes, and the fence is unaffected.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` — **TASK-1216850899**, same task as the backstop finding: the two sit in the same eight lines and are one edit together, exactly as your review body says.

Both halves are accurate. "the list budget is contexts x item" is precisely what the new derivation exists to stop being true, and having the falsified claim twelve lines *above* its correction is the worst arrangement of the two. And moving `DIRECTIVE_ITEM_TRUNCATED` up left "200 fits a GitHub Actions run URL (~105) plus a generous name and conclusion" explaining the ellipsis instead of the 200 — I moved the constant and did not re-read the paragraph it landed in.

Deferred for the merge-ordering reason on the thread above, not because it is small: a comment that contradicts its code within one screen is the same hazard round 1 of #85 raised about `_waiting.pop`, and I would not defer it if this branch could still take a commit.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` — **TASK-1216850899**.

Fair, and the correction is mine to make: the class covers the embedding, override and isolate controls, and I wrote "the bidi controls", which reads as though LRM, RLM and ALM were considered and kept. They were not considered at all. The zero-width family is the same argument.

The task states both routes — narrow the sentence, or widen the class to the marks and the zero-widths — and carries your last point too, which is the sharpest of the three: the isolates are *in* the class but not in the parametrised test, so U+202E is the only one of the six the test actually exercises. Whichever route is taken, the parametrisation gets an isolate.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py
RHDZMOTA: `[triage:later]` — **TASK-1216850899**.

Correct, and it is the assertion I was least happy with when I wrote it: restating the formula can only fail when someone edits one copy and not the other, which is a typo detector, not a test. Your two neighbours are the real ones — and you verified they fail when `contexts x item` comes back, which is the property I actually wanted.

The task carries your inequality form, and notes that it is also what pins the backstop finding on the source side: `contexts × (item + marker) + (contexts − 1) × 2 <= MAX_DIRECTIVE_DATA_CHARS` reads the same today, keeps its meaning if the budget is later given headroom, and states the invariant the module is really claiming.

PR #88, opened 2026-08-16
revloop: bound the Alissa task-list read — negative cache for unmapped PRs + probed narrowing

### LINKED ISSUE #87: revloop: stop full-corpus Alissa list reads — narrowed list call + negative cache for unmapped PRs
## Context

The revloop daemon is a major contributor to `apiActorOps.apiListMyTasks`, the #1 Database I/O offender on the Alissa prod deployment (1.67 GB over Aug 12–16). Its task-list call is the widest possible query, fired on a cache-miss path that can become permanent.

## Current behavior (what to fix)

- `list_tasks` shells out to `alissa task list --json` with no flags (src/main/alissa/tools/github/revloop/alissa.py:228) → `GET /v1/tasks` with NO query string → the actor's ENTIRE non-terminal corpus, sponsor-union scoped (~460+ rows, ~490 KB wire; the code's own comment at alissa.py:217-227 admits this is "the expensive call").
- The daemon keeps only `taskNumber, title, status` per row (`_task_from_row`, alissa.py:170-190) to run one title regex (`is_review_task_for`, alissa.py:149-168).
- The list fires on a review-task cache miss in `_review_task` (loop.py:1174-1263, fetch via `_pass_task_list` at loop.py:1157-1172). A PR with NO review task yet (or a validated/retitled one) is a PERMANENT miss — `find_review_task` returns None and nothing is recorded (loop.py:1241-1246) — so every poll (60s) refetches the whole corpus: 1,440 full-corpus reads/day from one unmapped PR. The read-only-ledger degradation at loop.py:1252-1256 has the same effect.

## Changes

1. **Negative cache for unmapped PRs.** When `find_review_task` comes up empty, record "no review task for PR <owner>/<repo>#<n> as of pass P" in the SQLite ledger (alongside `state.record_review_task`, loop.py:1248) and skip the corpus fetch for that PR for a configurable number of polls (new config knob, e.g. `review_task_miss_ttl_polls`, default ~10, floor 1). A recorded mapping or TTL expiry re-arms the lookup. The read-only-ledger path must fail open to today's behavior (fetch every pass) — never silently skip reviews.
2. **Narrow the list call.** Probe the installed CLI's supported flags at boot (pattern to copy: devloop's `--self` probe, alissa-github-develop-daemon webui/sources.py:255-274) and pass the narrowest supported combination:
   - a status filter covering every status a live review task can hold (verify against the statuses the daemon itself observes/writes before hardcoding),
   - `--self` IF review tasks are owned by the revloop actor — VERIFY ownership first (who creates review tasks: this daemon's own sessions or the operator?). If they are sponsor-owned, do not pass `--self`; the status filter still applies.
   - If the CLI supports a digest/lean view flag (shipping separately in the studio repo as `--view digest` on `task list`), prefer it — the daemon only needs taskNumber/title/status. Adopt via probe, never hard-require it.
3. Keep the per-pass memo (`_pass_task_list`) as-is.

## Out of scope

- Server/API changes (studio repo owns those).
- orcloop/devloop.
- Changing the title-regex mapping mechanism itself.

## Tests

- Negative cache: unmapped PR triggers exactly one corpus fetch, then zero for TTL polls, then re-arms; mapping recorded mid-TTL short-circuits; read-only ledger falls back to per-pass fetch.
- Flag probe: each supported-flag combination produces the expected CLI argv; absent flags degrade gracefully to the current call.
- Existing loop tests stay green.

## Definition of Done

revloop no longer refetches the full task corpus every poll for PRs without a mapped review task (ledger-backed negative cache with configurable TTL, failing open on read-only ledger), and its list call passes the narrowest CLI-supported query (status filter; --self and/or digest view only when probed supported and semantically correct). Config documented in the example config + README. Tests per house bar; honest results table in the PR.

## Acceptance criteria

- [ ] Negative cache implemented in the SQLite ledger with configurable TTL in polls: unmapped PR costs one corpus fetch per TTL window instead of one per pass; re-arms on TTL expiry or recorded mapping; read-only ledger fails open; covered by tests
- [ ] List call narrowed via boot-time CLI flag probe (status filter; --self / digest view only when supported AND ownership semantics verified), degrading gracefully to the current unfiltered call; covered by tests
- [ ] Config knobs documented (example config + README) and the expensive-call comment block updated to describe the new bounded behavior

Alissa-Task: TASK-805662381

RHDZMOTA:
Closes #87

**Tasks** — origin: `TASK-805662381` · implementation: `TASK-115878658` (downstream, soft dependency)

## What this does

`alissa task list` is the widest query the revloop daemon issues — the actor's whole live non-terminal corpus, sponsor-union scoped — and it is the only way to *find* a review task, because the mapping is by title. The PR → review-task cache (issue #66) bounded it for every PR whose review task **exists**; a PR with **none** — a third-party PR, or one whose task was validated or retitled — missed that cache on every pass and re-fetched the whole corpus for the same answer. At a 60s poll that is 1,440 whole-corpus reads a day, per unmapped PR, indefinitely.

Two bounds:

**1. A negative cache in the SQLite ledger** (`review_task_misses`). A search that *ran and found nothing* is recorded with a countdown of `review_task_miss_ttl_polls` polls (default 10, floor 1). Each later pass of that PR spends one and skips the corpus fetch; the row self-cleans at zero and the search re-arms. Recording a mapping clears it, so the two tables can never both speak for one PR. Only a search that **ran** may record one — a search that raised propagates and the pass turns it into that PR's `skipped` decision, exactly as before.

Fail-open is the contract, and one detail carries it: a suppression is granted **only when the countdown's decrement actually persisted**. A volume that flips read-only *after* a row was written can still read that row, so a decoupled implementation would answer "no review task" on every pass forever. Coupling it means such a ledger degrades to one fetch per pass — precisely today's behaviour.

**2. The list call itself, narrowed from a boot-time probe** of `alissa task list --help`. Probed off the help **output**, never the exit status (this CLI is commander-based and answers an unknown subcommand by printing the parent help and exiting 0), and anchored to option lines so a flag merely named in another option's prose is not read as an offer — the shipped 0.1.0 help contains exactly that trap. Three narrowings, all probe-gated:

| narrowing | on 0.1.0 | gate |
| --- | --- | --- |
| `--status <open statuses>` | **absent** → not sent | derived from the daemon's own `OPEN_STATUSES`, so it cannot change which task resolves |
| `--view digest` | **absent** → not sent | adopted wherever it exists, no knob — the daemon keeps only `taskNumber`/`title`/`status` |
| `--self` | advertised, but **off** | opt-in via `task_list_self_scope` |

So on today's CLI the daemon makes byte-for-byte the call it always made. A narrowed call that fails, or that answers with an *empty* corpus, is retried plain and disables the narrowing for the process — both are how a CLI advertising a flag its API does not serve would present, and either would otherwise read as "this actor has no review tasks", which is a skipped review rather than a slower one.

## Delivery notes

**Judgment calls**

- **`--self` is implemented but OFF by default**, against the issue's "pass the narrowest supported combination". The issue conditions it on verifying ownership, and the verification came out *against* it: on the live fleet corpus (2026-08-16, 932 non-terminal rows via `alissa task list`) `--self` removes 36 rows — 4% of the payload — and **3 of the 371 `Review PR …` tasks are among the rows it removes**, owned by an actor other than the agent actor whose sessions write the other 368. Review tasks are predominantly but not exclusively actor-owned, and a review task the list cannot see is a round the daemon cannot count (it changes `on_missing_review_task` behaviour and the round count). 4% of the wire does not buy that risk, so the flag ships as a documented per-deployment opt-in (`task_list_self_scope`) rather than as the default.
- **TTL counted per-PR, not as a global pass number.** The issue words it as "no review task … as of pass P". There is no durable pass counter in this daemon, and adding one would make a PR nobody is requesting a review on burn its window while idle. A per-PR countdown decremented on each pass that consults it is the same contract in the unit the knob is written in ("skip for N polls"), is self-cleaning, and needs no global state.
- **`review_task_miss_ttl_polls < 1` is refused at config load rather than clamped.** The issue says "floor 1". A silent clamp of `0` would accept an operator saying "turn the cache off" and then not do it; the daemon has no off value (0 would write rows nothing reads), so it says so at load. Consistent with `round_cap`/`max_concurrent_sessions`.
- **The negative answer is consulted only when there is no cached mapping.** A cached ref that is *disproved* must still search on the strength of that fresh disproof, and one that could not be *read* must fall back to the search (an unreadable task is not a wrong one). Both were already the module's rules; the new branch is an `elif` so it cannot compete with either.
- **`--pr OWNER/REPO#N` re-arms the PR before deciding.** That mode exists to tell "the search did not find it" apart from "the decision was no"; a suppressed pass would answer it without looking. Done in `__main__` rather than by plumbing a bypass through `evaluate`, so the poll path keeps exactly one way in.
- **`alissa-pr-review` opts out of the status filter** (`list_tasks(narrow_status=False)`). It deliberately matches review tasks whose status `is_open` rejects, so applying the daemon's open-status filter there would re-impose the filter that function exists to avoid. Every other narrowing still applies to it.
- **A narrowed call answering an empty corpus is cross-checked against the plain one.** Not asked for, but the issue's own "never silently skip reviews" requires it: `find_review_task` reads an empty corpus as "this PR has no review task". The cost when the corpus is *genuinely* empty is one extra empty list per pass; the cost without it is every review this actor owns.
- **Container pass-through added** (Dockerfile `ARG`/`ENV`, `revloop-config.sh`, docker README, `tests-entrypoint-config.sh`) though the issue named only "example config + README". A config key with no pass-through cannot be set on the deployed image; the round-1 reviewer on PR #85 raised exactly that as a `[major]`. `ALISSA_TASK_LIST_SELF_SCOPE` is this renderer's first **boolean** pass-through: it accepts `1/0`, `true/false`, `yes/no`, `on/off` and **refuses** anything else, because a typo silently rendering `false` would be indistinguishable from the default it was trying to change.

- [ ] **The narrowed call has never been sent**, because no CLI generation offers `--status` or `--view digest` yet. Everything about them is unit-tested against a synthesised future help text; the first real deployment against a CLI that grows either flag is the first live exercise. The runtime disproof (fail → retry plain; empty → cross-check) is the designed landing for a wrong guess about their syntax, but it too has only been tested against fakes.
- [ ] **`--self` ownership on the revloop actor specifically.** The 932-row corpus was read with *this* session's token (the develop-daemon lane). The evidence that it is the same shared agent actor is strong — 368 of 371 review tasks are owned by the actor `--self` keeps, and reviewer sessions run under the daemon's own `alissa` credentials — but the revloop daemon's token was not read directly. Anyone turning `task_list_self_scope` on should first check `alissa task list --json --self` from inside that container and confirm its open `Review PR …` rows are all present.
- [ ] **No container was built or run.** There is no docker in this environment. The renderer and the boot path were exercised via `docker/claude/tests-entrypoint-config.sh` and `tests-entrypoint-ui.sh` (real entrypoint, stubbed CLIs); the `ARG`/`ENV` additions themselves are unexecuted text until an image is built.
- [ ] **`REVLOOP_VERSION` in the Dockerfile is untouched** and still pins `0.16.14` — it already lagged the published `0.17.0` before this change. The two new env vars need `>= 0.18.0` (documented as such in the docker README), so they are inert on the currently pinned image. The re-pin is a separate release-gated step, as in this repo's prior lanes.
- [ ] **The negative cache trusts the corpus the narrowed call returns, and a PARTIALLY-served filter would make an unmapped-PR answer sticky.** Round-1 `[minor]` on this PR. The two runtime disproofs in `list_tasks` catch a narrowed call that *fails* and one that answers *empty*; neither catches one that answers **non-empty but incomplete** — a `--status` the API accepts and serves partially, say. `find_review_task` then legitimately finds nothing, a miss is recorded, and the re-search that would correct it is suppressed for the window, where before this change the wrong answer self-corrected on the next pass. **Unreachable today** — no shipping CLI offers `--status` or `--view digest`, so the narrowed call has never been sent. Whoever enables the first narrowed CLI generation must prove it serves the full filter before the negative cache is allowed to trust its misses; that proving gate is registered as **TASK-1573578582** (*prove a narrowed `alissa task list` generation before the negative cache may trust its misses*) rather than bolted on here. Raised again as a `[minor]` on the follow-up PR #89 round 1, which is where the task ref comes from.
- [ ] **The I/O reduction itself is unmeasured here.** The claim is arithmetic (one fetch per 10 polls instead of one per poll for an unmapped PR); confirming it means watching `apiActorOps.apiListMyTasks` on the Alissa deployment after this ships.

**Verification**

| what | command | result |
| --- | --- | --- |
| unit | `PYTHONPATH=src/main bash tests-unit.sh alissa-tools-github-revloop` | **770 passed** (was 726; +44) |
| coverage | `PYTHONPATH=src/main bash tests-coverage.sh …` | **94%** total; `alissa.py` 88%, `state.py` 96%, `loop.py` 97%, `config.py` 97% |
| style | `bash check-style.sh …` | clean (`pycodestyle`, 120 cols) |
| types | `bash check-types.sh …` | `Success: no issues found in 30 source files` |
| entrypoint renderer | `bash docker/claude/tests-entrypoint-config.sh` | **ALL PASS** (incl. the 11 new boolean/TTL assertions) |
| entrypoint boot | `bash docker/claude/tests-entrypoint-ui.sh` | all reviewer-console checks passed |
| live CLI probe | `Alissa().probe_task_list()` against the installed `alissa` 0.1.0 | `TaskListFlags(status=False, self_scope=True, digest=False)` → argv `alissa task list --json`, unchanged |
| live list | `Alissa().list_tasks()` | 933 tasks parsed off the real API |
| CI | all six required workflows on `9273722` | **green** (entrypoint 2m45s, style 13s, types 20s, unit 49s, version-bump 5s, wheel 10s) |

Two counterfactuals were run rather than assumed: disabling the suppression branch fails `test_an_unmapped_pr_costs_one_corpus_fetch_per_ttl_window` and `…_window_re_arms`; decoupling the suppression from the decrement's success fails `test_a_read_only_ledger_never_suppresses_a_search`. No failures, skips or xfails anywhere in the run.

**Scope touched**

Within the issue's declared scope (revloop only; no server/API, no orcloop/devloop, no change to the title-regex mapping mechanism):

- `alissa.py` — the probe, `TaskListFlags`, `task_list_argv`, `list_tasks` (+ `narrow_status`), the rewritten expensive-call comment block.
- `state.py` — the `review_task_misses` table and its five methods; `_replace_review_task` now also clears the miss row.
- `loop.py` — the `elif` in `_review_task`, the miss recording, the boot probe/warning in `preflight`, the `Alissa(...)` construction, two comment blocks.
- `config.py`, `__main__.py` — the two knobs, their validation, CLI flags, and the `--pr` re-arm.
- `prreview.py` — one call site opts out of the status filter (see judgment calls).
- tests — new `test_task_list.py`; additions to `test_loop.py` and `test_state.py`; three fakes updated for the new signatures; one existing test (`test_a_pass_never_inherits_the_previous_pass_corpus`) re-armed between passes so it still tests the *memo*.

Excursions, called out explicitly:

- **Container plumbing** (`docker/claude/Dockerfile`, `revloop-config.sh`, `README.md`, `tests-entrypoint-config.sh`) — beyond "example config + README", justified above.
- **Version bump** `0.17.0` → `0.18.0` (`src/main/.../version`) — required by `check-version-bump.yaml` for any packaged change.


alissa-app:
- 2026-08-16 20:22:37 UTC — `review-alissa-github-review-daemon-pr88-r1-542ffa` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1548,6 +1570,12 @@ def _review_task(self, pr: PullRequest) -> "ResolvedTask":
                     "an unreadable task is not a wrong one)",
                     pr.slug, cached,
                 )
+        elif self.state.consume_review_task_miss(pr.full_name, pr.number):
```
alissa-app, line 1573: [minor] The negative cache suppresses this PR even when **this pass has already fetched the corpus**.

`_pass_task_list()`'s memo is populated by the first unmapped PR of the pass that re-arms. Any *other* unmapped PR evaluated after it, whose window is still open, reaches this `elif`, burns a poll, and returns "no review task" — while a fresh corpus that could answer the question definitively sits in `self._pass_tasks`, already paid for.

Failure scenario: two unmapped PRs, `#1` (window: 3 polls left) and `#2` (window expired). Pass P evaluates `#2` first, which fetches the corpus. `#1` is then suppressed and answers from the ledger. If `#1`'s review task was created since its last search, it is invisible for up to `review_task_miss_ttl_polls` more polls — ten minutes of added round-1 latency at the default cadence — even though the daemon was holding the evidence for it, for free, in the same pass. The read saving in that case is exactly zero, so the latency is bought with nothing.

The bound is what the DoD asks for, so this is not a spec violation — it is value left on the table. Gating the branch on `self._pass_tasks is None` (i.e. only trust the negative answer while the corpus has *not* been fetched this pass) keeps the whole saving and gives the free answer back. Order-dependent within a pass, so it is a partial win, but a free one.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1557,6 +1585,16 @@ def _review_task(self, pr: PullRequest) -> "ResolvedTask":
...
             if cached is not None:
                 self.state.forget_review_task(pr.full_name, pr.number)
+            # ...and it is the ONLY thing this daemon may record a negative
+            # answer from. Recorded here rather than at the call sites because
+            # this is the one place that knows the search RAN: a search that
+            # raised never reaches this line at all (it propagates out of
+            # `_pass_task_list` and the pass turns it into one PR's SKIPPED),
+            # so a transient CLI failure can never buy itself a window of
+            # silence.
+            self.state.record_review_task_miss(
```
alissa-app, line 1595: [minor] A narrowed call that returns a **non-empty but incomplete** corpus escapes both runtime disproofs — and now also buys a TTL window of silence.

The two disproofs in `list_tasks` cover a narrowed call that *fails* and one that answers *empty*. Neither covers the middle case: a `--status` (or `--view digest`) the API accepts and serves *partially* — rows come back, the review task's row does not. `find_review_task` then legitimately returns `None`, and this line records a negative miss, so the daemon stops re-searching for `review_task_miss_ttl_polls` polls.

Failure scenario: a future CLI grows `--status` but treats the argument as a single status rather than a comma list, matching only `committed`. The corpus comes back non-empty (every `committed` task), a review task sitting in `pending_validation` is absent, the PR is recorded as unmapped, and the re-search that would have corrected it is suppressed for the window. Before this PR that PR would have been re-searched every pass — still wrong, but self-correcting the moment the flag was fixed; now the wrong answer is also sticky.

Not reachable today: no shipping CLI offers either flag, and the PR is explicit that the narrowed call has never been sent — which is why this is `[minor]` and not blocking. Worth either a defensive note next to the operator's-gate checklist, or making the first narrowed generation prove itself before the negative cache is allowed to trust its misses.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -213,19 +279,156 @@ class TaskDetail:
...
+                raise
+            log.warning(
+                "`%s` failed — retrying the plain task list and dropping the "
+                "narrowing for this process", " ".join(argv),
+            )
+            self._task_list_narrowing_disabled = True
+            data = run_json(plain, timeout=90) or []
+
+        tasks = self._tasks_from(data)
+        if tasks or argv == plain:
```
alissa-app, line 411: [nit] When the narrowed call **failed**, this guard lets the empty-corpus cross-check fire a *second* plain call.

The `except CommandError` branch above already retried `plain` and set `_task_list_narrowing_disabled`. If that retry legitimately returns an empty corpus, `tasks` is empty and `argv != plain`, so control falls through to the cross-check and calls `plain` again — three subprocess calls where two happened — and logs ``"`<narrowed argv>` returned no tasks"`` about a call that in fact *errored* and never answered.

Rare (needs a failing flag *and* a genuinely empty corpus) and harmless beyond the extra call and the misleading log, but it is an extra whole-corpus fetch inside the change whose purpose is to remove them. `if tasks or argv == plain or self._task_list_narrowing_disabled:` covers it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -16,6 +16,72 @@
...
+    """Whether `flag` appears as an OPTION in a CLI help listing.
+
+    Anchored to the start of a help line (allowing a short alias in front, as in
+    `-h, --help`) so a flag merely NAMED in some other option's prose -- "Pair
+    with --include-shared" is in this very help text -- is not read as an offer
+    of that flag. The trailing guard rejects a longer flag that merely starts
+    with this one (`--self-only` is not `--self`).
+    """
+    return re.search(
+        rf"(?m)^\s*(?:-\w,\s+)?{re.escape(flag)}(?![\w-])", helptext
```
alissa-app, line 81: [nit] The anchoring does not fully deliver what the docstring claims.

`^\s*(?:-\w,\s+)?--flag(?![\w-])` rejects a flag named *mid-line* in prose, but not one that a wrapped description line happens to *start* with. Commander wraps descriptions at roughly 80 columns onto continuation lines indented exactly like option lines — the shipped 0.1.0 help already wraps `--self`'s description onto its own line — so a help text such as

```
  --include-shared    Also list tasks shared with you; combine
                      --status with it to filter
```

reads as an offer of `--status`.

The consequence is bounded: the flag is sent, `alissa` exits non-zero, and `list_tasks`' retry-plain disproof catches it at the cost of one extra call — so this is a hardening gap, not a defect. `test_a_flag_named_only_in_prose_is_not_an_offer` places the prose flag at end-of-line, which is the case the regex does handle; a wrapped-line variant would document the real boundary.

README.md
```diff
@@ -606,6 +610,69 @@ The gate bounds what the *daemon* starts. Hand-spawned sessions can still push
 the live count past `reap_session_cap`, and that is exactly when the alarm should
 fire.
 
+### Bounding the task-list read
+
+`alissa task list` is the widest query this daemon issues — the calling actor's
+entire live task corpus, sponsor-union scoped, several hundred rows — and it is
+the *only* way to FIND a review task, because the mapping is by title. Between
+2026-08-12 and 08-16 this loop was the largest single contributor to the Alissa
+deployment's top database-I/O endpoint. Three mechanisms bound it, innermost
```
alissa-app, line 619: [nit] "Three mechanisms" over a four-row table.

The table below lists the mapping, the per-pass memo, the negative cache and the narrowed call — four. Either the count or the table is off by one; the table looks right, so the prose is the half to fix.

alissa-app (approved):

Review task: **TASK-561001883** · origin task: **TASK-805662381** · implementation task: **TASK-115878658**
Reviewed head: `9273722` · CI rollup at that sha re-read at submit time: **6/6 concluded, all `success`**.

Reviewed per the `alissa-code-review` CR4 rubric. Five **minor/nit** findings, no blockers and no majors, so this approves. The two `[minor]`s are worth a follow-up but neither breaks a contract this PR claims.

### What this does, in my own words

Three additions to bound `alissa task list`, the daemon's widest read: a per-PR **negative cache** (`review_task_misses`) that suppresses the corpus fetch for `review_task_miss_ttl_polls` polls after a *completed* search finds nothing; a **boot-time `--help` probe** that assembles the narrowest `task list` argv the installed CLI actually advertises; and two config knobs with CLI, example-config and container pass-through.

### Correctness

Judged against TASK-805662381's DoD and its single validation criterion, all four clauses hold and are pinned by tests that fail when the behaviour is removed (the PR ran the counterfactuals; I verified the assertions themselves):

- *One corpus fetch per TTL window* — `test_an_unmapped_pr_costs_one_corpus_fetch_per_ttl_window` runs a window and a half, so it pins the re-arm as well as the suppression.
- *Mapping recorded mid-TTL short-circuits* — enforced in SQL, in the same transaction as the mapping insert (`_replace_review_task`), not by a second call that could be skipped. Good choice.
- *Read-only ledger falls back to per-pass fetch* — this is the one that could have gone wrong, and it did not: `consume_review_task_miss` returns `True` **only when the decrement persisted**, so a volume that flips read-only *after* a row exists degrades to one fetch per pass instead of answering "no review task" forever. The coupling is the load-bearing detail and it is tested directly.
- *Probed flags produce the expected argv; absent flags degrade* — `test_task_list.py` probes against the **real, verbatim** 0.1.0 help text (I diffed it against the installed CLI's actual output — byte-identical, including the `--include-shared`-in-prose trap), and asserts an old CLI yields the exact call the daemon always made.

Two structural things I checked rather than took on trust:

- The suppression is an `elif` on `cached is not None`, so a cached-and-matching ref, a *disproved* ref, and an *unreadable* ref all reach it never / re-search / re-search respectively. That ordering is right, and `test_a_disproved_mapping_still_searches_despite_an_outstanding_miss` locks it.
- A miss is recorded on the one line that knows the search *ran*. A raising search propagates out of `_pass_task_list` before that line, so a CLI hiccup cannot buy itself a window of silence. Confirmed by reading the call graph, not just the test.
- `ResolvedTask(task=None)` from a suppressed pass is the same value a searching pass produces, and `evaluate`'s `on_missing_review_task` handling is unchanged, so the suppression cannot invent a decision. There is no review-task *creation* path downstream, so no duplicate-task hazard.

Findings: `[minor]` on the pass-memo interaction, `[minor]` on the residual partial-corpus gap, `[nit]` on a redundant third list call.

### Coding practices

Matches the repo's idiom closely — the why-not-what comment density, the constant-derived-from-the-predicate pattern (`TASK_LIST_STATUS_FILTER` from `OPEN_STATUSES`, with a test asserting they stay equal), refuse-don't-clamp validation consistent with `round_cap`/`max_concurrent_sessions`, and best-effort telemetry writes routed through `_write_telemetry` like every other optimization write in `State`. No dead code, no debug leftovers, no untracked TODOs, no new dependencies. `prreview.py`'s bare `Alissa()` is correct — that CLI loads no daemon config, and the wider corpus is the safe direction there. One `[nit]` on the README's "Three mechanisms" over a four-row table.

### Security

Nothing found. No secrets in the diff, config samples or fixtures. All new SQL is parameterised. The probe regex escapes its needle and only ever sees module constants. `task_list_argv` builds an argv **list** — no shell. The new `jq` boolean coercion in `revloop-config.sh` fails closed via `error()` on anything outside `1/0|true/false|yes/no|on/off` rather than rendering a silent `false`, and its error text does not echo the value; the entrypoint test pins both the accepted spellings and the refusal. No auth/permission surface is touched, and no new log line carries task content beyond refs already logged.

### Performance

This is the dimension the PR exists for and the arithmetic holds: an unmapped PR goes from one whole-corpus read per poll to one per `review_task_miss_ttl_polls` polls, and the ledger read that replaces it is a single indexed primary-key lookup plus one small write. `--self`'s measured 4% saving being declined because 3/371 review tasks would vanish from the list is the right call — a review task the list cannot see is a round the daemon cannot count, and that is not a 4%-of-the-wire trade. The one place the bound leaves value on the table is the pass-memo interaction (`[minor]` below). The genuinely-empty-corpus cross-check costs one extra *empty* list per pass forever, which is correctly priced in the PR body.

### Side-effects / blast radius

- **Schema**: `CREATE TABLE IF NOT EXISTS` with an in-place migration test; an older revloop opening a newer ledger simply ignores the extra table, so rollback is clean.
- **Signatures**: `Alissa.__init__` and `list_tasks` gained keyword-only params with defaults — source-compatible for any caller, and the three in-repo fakes were updated.
- **Config**: `config.build` rejects unknown keys, so the two new keys hard-fail an older library at load. That is why the docker README gates them at `REVLOOP_VERSION >= 0.18.0` and the entrypoint cross-check unsets them — the same treatment `ALISSA_CHECKS_SPAWN_WAIT_SECONDS` already gets. Worth restating for whoever deploys: the image still pins **0.16.14**, so setting either new build-arg on today's image renders a key the pinned library will refuse. That is pre-existing pin lag disclosed in the PR body, not a regression from this diff, and the re-pin is correctly a separate release-gated step.
- **Irreversible operations**: none. No external side-effects, no destructive migration.
- **Failure containment**: every new degradation lands on "what the daemon did before this existed" — probe failure, narrowed-call failure, empty narrowed answer, unwritable ledger, unreadable ledger, hand-edited exhausted row. I looked for a path where a failure makes the daemon *quieter* rather than slower and did not find one.

### Shortest path from here

Nothing is required. If you take the two `[minor]`s, the pass-memo one is the higher-value fix: it removes avoidable round-1 latency at zero I/O cost.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]`

Agreed, and the failure scenario is exactly right: when the pass memo is already
populated, the negative answer buys a window of latency for **zero** saved reads.
The corpus is sitting in `self._pass_tasks`, already paid for, and answering from
it is a list comprehension.

Taking the suggested gate — `self._pass_tasks is None and
self.state.consume_review_task_miss(...)` — so the negative answer is only
trusted while the corpus has *not* been fetched this pass. It is order-dependent
within a pass, as you say, but the direction is strictly better: never a read
added, sometimes a whole window of latency removed. A pass that then searches the
free corpus and still finds nothing records a fresh miss, which is the same
disproof the re-armed path already writes.

Not pushed here — this round approved `9273722` and an approved head is
merge-racy, so a commit on this branch could be lost behind the squash. It lands
in a follow-up PR off `main`, opened and handed off in this same session; I will
post its number on this PR once it is cut.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:pursue]`

Correct, and the sharp part of the finding is the second half: before this PR a
partially-served filter was wrong but *self-correcting* on the next pass, and the
negative cache makes the wrong answer sticky for a window. That is a real
regression in the failure mode even though the good path is unchanged.

Taking your first option, and adding to it:

* a defensive bullet in this PR's **Not verified — operator's gate** checklist,
  saying plainly that the negative cache trusts the corpus the narrowed call
  returns and that a partially-served filter would make an unmapped-PR answer
  sticky for the window. A body edit does not move `headRefOid`, so it is safe on
  this approved branch;
* a comment at the recording site itself, in the same follow-up PR as the finding
  above — the hazard belongs next to the line that writes the row, not only in a
  PR body the next reader may never open.

Not taking the stronger "make the first narrowed generation prove itself before
the cache may trust its misses" option in this session: it is a real design
(a per-generation proving gate on top of the two existing disproofs) and it wants
its own issue rather than being bolted onto a nit-fix follow-up. Nothing is
reachable today — no shipping CLI offers either flag — so the ordering costs
nothing, and whoever ships the first `--status` generation reads the gate bullet
before they turn it on.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]`

Confirmed by reading it back: after the `except CommandError` branch retries
`plain`, `argv` is still the narrowed argv, so an empty result from that retry
falls straight into the cross-check and calls `plain` a second time — three
subprocess calls, and a warning that names a call which errored rather than
answered. Inside a change whose whole purpose is removing whole-corpus fetches,
that is worth the one-token fix.

Taking your guard verbatim (`if tasks or argv == plain or
self._task_list_narrowing_disabled:`), with a test that pins the three-call path
down to two. Lands in the follow-up PR off `main` described on the thread above.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]`

You are right that the docstring over-claims. The regex anchors to line start,
and a commander continuation line is indented like an option line, so a wrapped
description whose second line *begins* with `--status` reads as an offer — and
0.1.0's own help already wraps `--self`'s description, so the shape is not
hypothetical.

Tightening the anchor to the option COLUMN rather than merely to line start:
commander puts options at two spaces and wraps descriptions to the description
column (22 in the shipped help), so a small bounded indent separates the two
without depending on the exact width. Plus the wrapped-line variant of
`test_a_flag_named_only_in_prose_is_not_an_offer` you asked for, which is the
part that documents the real boundary.

Noting for the record that the consequence stays bounded either way — a
false-positive flag is sent, `alissa` exits non-zero, and `list_tasks`' retry
disproof catches it at the cost of one call. Follow-up PR, as above.

README.md
RHDZMOTA: `[triage:pursue]`

Off by one, and the table is the half that is right — the per-pass memo is a
mechanism in its own row and the prose forgot to count it. Fixing the prose.
Follow-up PR off `main`, with the code nits above.

RHDZMOTA:
Round-1 triage is complete: all five findings `[triage:pursue]`, replies on their threads above.

The four with a diff are in **#89**, cut from this PR's approved head `9273722` and based on `main` — nothing was pushed here after the approve, so this head is exactly the commit the round approved. **#88 must merge before #89**: merging the follow-up first would squash this PR's whole delta into `main` under its commit message and leave this one with an empty diff.

The fifth (the `[minor]` on a partially-served narrowed filter making a miss sticky) needed no diff on this branch — it is now a bullet in the **Not verified — operator's gate** checklist in this PR's body, plus a comment at the recording site in #89. The stronger remedy the reviewer offered, a proving gate for the first narrowed CLI generation, is deliberately left to its own issue; it is unreachable until a CLI ships `--status` or `--view digest`.

PR #89, opened 2026-08-16
revloop follow-up (PR #88 r1): free-corpus gate for the negative cache, and three hardening nits

RHDZMOTA:
Follow-up to **PR #88, review round 1** (`alissa-app`, approve, head `9273722`) — all five findings, every one triaged `[pursue]`.

Follow-up-To: #88

**Tasks** — origin: `TASK-805662381` · implementation: `TASK-115878658` (same implementation task as #88; this carries that PR's round-1 findings, not new scope)

## ⚠️ Merge #88 first

This branch is cut from **#88's approved head `9273722`**, not from `main`, and it is based on `main`. Every line it touches is code #88 *introduces*, so the follow-up is unbuildable off `main` while #88 is open. Two consequences for whoever merges:

- **`#88` must merge before this PR.** Merging this one first would squash #88's entire delta into `main` under *this* PR's commit message, leaving #88 with an empty diff — losing its history and the issue-closing claim it carries for issue #87.
- GitHub's diff-vs-`main` below shows #88's changes too until it merges. The review surface of *this* PR is the single commit `9273722..0208dfb`.

No third level is stacked on this one. This PR carries the blocking label **`blocked:merge-order`** as the mechanical half of that gate (round-1 `[minor]`): **remove it once #88 has merged.**

## Round-1 outcome

**Approved** on `0208dfb`, carrying 4 `[minor]` and 2 `[nit]` — all real, all answered on their threads. One (`merge order`) was `[triage:pursue]` and is done above via the label. The other five are `[triage:later]` against registered tasks, because an approve makes this branch terminal and their remedies need a branch off `main` that cannot exist until #88 and #89 merge:

- **TASK-1573578582** — prove a narrowed `alissa task list` generation before the negative cache may trust its misses (the `KNOWN GAP` deferral now has something to open, assign and close; the comment cites it as part of the task below).
- **TASK-1006125108** — the four documentation/clause cleanups: the README's three pre-gate latency sites (the knob's unit is now *polls in which no corpus was fetched*, a better contract than the one documented), the call-local `retried_plain` flag replacing the process-level one in `list_tasks`' guard, the second precondition in `_review_task`'s docstring, and `MAX_OPTION_INDENT`'s silent false-negative direction plus the boundary test the constant currently lacks.

Both are soft-downstream of `TASK-115878658`. Nothing from this round is deferred in prose alone.

## The five findings

| # | finding | remedy |
| --- | --- | --- |
| `[minor]` | the negative cache spent a window even when the pass memo already held the corpus | gate the branch on `self._pass_tasks is None` |
| `[minor]` | a narrowed call answering **non-empty but incomplete** escapes both runtime disproofs, and the miss then becomes sticky | defensive comment at the recording site + a bullet in #88's operator's-gate checklist |
| `[nit]` | a *failed* narrowed call could pay a third list call | guard also honours `_task_list_narrowing_disabled` |
| `[nit]` | `_advertises` anchored to line start, which a wrapped description line satisfies | anchor to the option **column** |
| `[nit]` | README said "Three mechanisms" over a four-row table | "Four" |

The first `[minor]` is the one with runtime value. When some earlier PR of a pass re-arms and pays for the corpus, answering a *later* in-window PR from the ledger saves exactly nothing — the search is a list comprehension over rows already in memory — while still spending the window, so a review task created since that PR's last search stays invisible for up to `review_task_miss_ttl_polls` more polls with the evidence for it sitting right there. Gating on the memo is order-dependent within a pass and therefore a partial win, but never adds a read.

## Delivery notes

**Judgment calls**

- **Cut from `9273722`, not from `main`.** The convention says "new branch off the default branch"; taken literally it is unbuildable here, because none of the code being fixed exists on `main` yet. The property the convention protects — the approved head never moves — holds: nothing was pushed to `TASK-115878658-LIST-BOUND` after the approve. Same shape as this repo's PR #48/#53.
- **The second `[minor]` is answered with documentation, not a mechanism.** The reviewer offered two options; the stronger one ("make the first narrowed generation prove itself before the negative cache may trust its misses") is a real design — a per-generation proving gate layered on the two existing disproofs — and bolting it onto a nit-fix follow-up would be the wrong place to get it right. Nothing can reach the gap today: no shipping CLI offers `--status` or `--view digest`, so the narrowed call has never been sent. It is recorded both at the recording site in `loop.py` and as an operator's-gate bullet on #88, and it wants its own issue before any deployment turns a `--status` generation on. That deferral is stated on the review thread rather than left implied.
- **No further version bump.** #88 already moves `0.17.0 → 0.18.0` and that release is unpublished, so these changes ride it. `check-version-bump` compares against PyPI, and `0.18.0` is not there yet.
- **`MAX_OPTION_INDENT = 3`, not the exact description column.** Commander's description column moves with the longest option name in the listing (22 in the shipped help, wider elsewhere), so pinning it would be brittle. Options sit at two spaces in every generation of this CLI; a small bounded indent separates the two columns without depending on either width. A test pins that a listing indenting its option column by three is still matched.

- [ ] **The merge order is not enforced by anything but this text.** GitHub will happily merge this PR first, and doing so strands #88 (see the banner above). Someone has to read it.
- [ ] **The second `[minor]`'s underlying hazard is documented, not closed.** A future CLI that serves `--status` partially would still produce a sticky wrong answer for one window. Unreachable today; it needs an issue before any deployment enables a narrowed generation.
- [ ] Everything still open from #88's own delivery note — the narrowed call has never been sent live, `--self` ownership was measured with the develop lane's token, no container was built, `REVLOOP_VERSION` still pins `0.16.14`, and the I/O reduction itself is arithmetic rather than a measurement.

**Verification**

| what | command | result |
| --- | --- | --- |
| unit | `PYTHONPATH=src/main bash tests-unit.sh alissa-tools-github-revloop` | **776 passed** (was 770; +6) |
| coverage | `PYTHONPATH=src/main bash tests-coverage.sh …` | **94%** total, unchanged |
| style | `bash check-style.sh …` | clean |
| types | `bash check-types.sh …` | `Success: no issues found in 30 source files` |
| entrypoint renderer | `bash docker/claude/tests-entrypoint-config.sh` | **ALL PASS** |
| CI | all six required workflows on `0208dfb` | **green** (entrypoint 2m51s, style 13s, types 20s, unit 44s, version-bump 8s, wheel 9s) |

Counterfactual, run rather than assumed: reverting all three code fixes at once fails exactly the four new tests (`…_already_holds_the_corpus_does_not_spend_a_window`, `…_free_corpus_is_searched_rather_than_trusted_blind`, `…_wrapped_description_line_is_not_an_offer`, `…_does_not_also_pay_the_empty_cross_check`) and nothing else — 4 failed, 772 passed. No skips or xfails in any run.

**Scope touched**

Strictly the five findings, no new scope:

- `loop.py` — the `_pass_tasks is None` gate on the suppression branch (+ its rationale), and the known-gap comment at `record_review_task_miss`.
- `alissa.py` — `MAX_OPTION_INDENT` and the option-column anchor in `_advertises`; the `_task_list_narrowing_disabled` clause in `list_tasks`' early-return guard.
- `README.md` — one word.
- tests — three in `test_loop.py` (the gate in both directions, plus the first-PR-of-a-pass case so the refinement cannot undo the saving it refines), two in `test_task_list.py` (the wrapped line, the three-call path), one asserting the anchor still tolerates a shifted option column.

No excursions.


alissa-app:
- 2026-08-16 20:59:15 UTC — `review-alissa-github-review-daemon-pr89-r1-a3a317` — round 1 of 22 — spawned

README.md
```diff
@@ -124,6 +126,8 @@ extending it. `--dry-run` / `--no-dry-run` override the config in both direction
...
 | `checks_spawn_wait_seconds` | `900` | **pre-spawn CI gate**: how long an owed round waits for the head's checks to *conclude* before its reviewer is queued at all. The key above gates the verdict the *daemon* posts; this is the only structural gate on the verdict a reviewer *session* posts. Past the bound the round is queued anyway, told it may not approve. `0` disables the *hold* and relies on the directive alone. **It also bounds the reviewer session's own in-round wait**: the same number is written into every directive as how long a session may wait for a running check before submitting, floored at 5 minutes so a `0` here cannot read as "do not wait at all" — see *Never approve a red head* |
+| `review_task_miss_ttl_polls` | `10` | how many polls a PR with **no** review task is taken on trust before the task corpus is searched for one again. Trades **latency** for reads: a review task created mid-window is picked up at the re-arm rather than the next poll. Floor `1` — there is no value that turns it off — see *Bounding the task-list read* |
```
alissa-app, line 129: `[minor]` **The README still describes the negative cache's pre-gate latency contract — in three places, including the one an operator reads to size the knob.**

The gate this PR adds is precisely a change to *when the window is spent* and *when a mid-window review task is picked up*, and none of the three sites moved:

- **this line (129)** — "a review task created mid-window is picked up at the re-arm rather than the next poll". After the gate that is the **worst case**, not the rule: on any pass where the corpus was already fetched, the PR is searched for free and the task is picked up on that very poll.
- **626** (mechanism table) — "the fetch, for a PR that has *no* review task, for `review_task_miss_ttl_polls` polls". It is now for up to that many polls *in which no corpus was fetched*; a poll that fetches neither spends the window nor skips a fetch.
- **634** — "the next `review_task_miss_ttl_polls` polls of that PR skip the fetch and answer from the ledger", same drift.

That last shift is the real one: the knob's unit quietly changed from *polls* to *polls with no corpus fetch*. It is a better contract than the one documented, which is exactly why it is worth documenting — an operator reading line 129 today will under-estimate how fresh the daemon actually is, and the PR's own best argument for itself is nowhere in the docs.

The PR did touch README, so the one-word fix and this sit in the same file. A sentence on each of the three, or one sentence here plus a pointer, closes it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1557,6 +1597,29 @@ def _review_task(self, pr: PullRequest) -> "ResolvedTask":
...
+            # search "ran" is only as good as the corpus it ran over. The two
+            # runtime disproofs in `Alissa.list_tasks` catch a narrowed call that
+            # FAILS and one that answers EMPTY; neither catches one that answers
+            # NON-EMPTY but INCOMPLETE -- a `--status` the API accepts and serves
+            # partially, say. The search then legitimately finds nothing and this
+            # line makes that wrong answer STICKY for a window, where before the
+            # negative cache it was wrong but self-correcting on the next pass.
+            # No CLI generation offers `--status` or `--view digest` yet, so
+            # nothing can reach it today; the first one that does must prove
+            # itself before this row may be trusted. See the operator's-gate
```
alissa-app, line 1618: `[minor]` **The deferral names no issue and no task, against this module's own convention.**

"the first one that does must prove itself before this row may be trusted. See the operator's-gate checklist on PR #88." — the only durable pointer this leaves is a PR body, and the PR description says the proving gate "wants its own issue before any deployment turns a `--status` generation on."

Compare the two other deferrals in this same docstring, forty lines up, which do it the other way:

> relocating the check is **TASK-317167904**
> Settling that is **TASK-1897198077**

That is the pattern that makes a deferral findable from the code. Here, once #88 and #89 merge, someone enabling a `--status` CLI generation reaches this comment, is told to go read a checkbox on a closed PR, and has nothing to open, assign, or close. The hazard is genuinely unreachable today — I am not asking for the proving gate, and leaving it to its own issue is the right call — but the issue it is being left to does not exist yet.

Register it and cite the id here (and in #88's gate bullet). It is a one-line edit once the task exists.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -213,19 +301,163 @@ class TaskDetail:
...
+                raise
+            log.warning(
+                "`%s` failed — retrying the plain task list and dropping the "
+                "narrowing for this process", " ".join(argv),
+            )
+            self._task_list_narrowing_disabled = True
+            data = run_json(plain, timeout=90) or []
+
+        tasks = self._tasks_from(data)
+        if tasks or argv == plain or self._task_list_narrowing_disabled:
```
alissa-app, line 433: `[minor]` **The third clause expresses a call-local fact through a process-level flag, and is only correct because of a coupling three methods away.**

The comment says exactly what is meant — "the `except` branch above having already retried `plain`" — but that is not what the condition tests. `self._task_list_narrowing_disabled` is process-wide and outlives the call. It happens to be equivalent here only because `task_list_argv` early-returns bare `plain` when the flag is set, so a *pre-existing* disabled flag can never reach this line with `argv != plain`; the flag can only be newly-True-and-narrowed-argv inside this call.

That coupling is invisible from here, and it is load-bearing for a safety net. `--self` is already a deployment opt-in rather than a probe guess, so "keep honouring `task_list_self_scope` even after the probed narrowings are disabled" is a plausible future edit to `task_list_argv`. The moment it lands, `argv != plain` on a narrowing-disabled call, this clause fires on every subsequent call, and the empty-corpus cross-check — the disproof that stands between a filter the API does not serve and "this actor has no review tasks" — is silently gone for the life of the process. Failure mode: skipped reviews, no warning, no test failure.

A call-local flag says the intended thing and cannot be broken from a distance:

```python
retried_plain = False
try:
    data = run_json(argv, timeout=90) or []
except CommandError:
    ...
    self._task_list_narrowing_disabled = True
    retried_plain = True
    data = run_json(plain, timeout=90) or []

tasks = self._tasks_from(data)
if tasks or argv == plain or retried_plain:
    return tasks
```

Same behaviour today, same test (`test_a_failed_narrowed_call_does_not_also_pay_the_empty_cross_check`) still pinning it, and the comment then describes the code rather than compensating for it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1507,9 +1509,29 @@ def _review_task(self, pr: PullRequest) -> "ResolvedTask":
...
 
+        ...and a FOURTH outcome that is none of those (issue #87): no cached ref
+        and a recent search that already found nothing. The three above bound
+        the cost of a PR whose review task EXISTS; a PR that has none missed
+        every one of them on every pass and paid the corpus fetch for the same
+        answer forever -- 1,440 whole-corpus reads a day, per unmapped PR, at a
+        60s poll. So a completed search that finds nothing is recorded as such
+        and taken on trust for `review_task_miss_ttl_polls` further polls.
+
+        That negative answer is consulted ONLY when there is no cached ref at
```
alissa-app, line 1520: `[nit]` **The docstring's account of when the negative answer is consulted is now incomplete.**

"That negative answer is consulted **ONLY** when there is no cached ref at all, which is what keeps it from ever competing with the three outcomes above" — as of this PR there is a second precondition, `self._pass_tasks is None`, and the paragraph that enumerates the conditions does not mention it. The full reasoning is in the inline comment fifty lines below, which is the right place for the *why*, but this docstring is where the module states the mechanism's contract, and a reader who stops here now has a slightly wrong model.

An added clause is enough — something like "…and only while this pass has not already fetched the corpus, because an answer the memo can give for free is never worth a window (see the branch below)."

Purely a documentation point in a module whose documentation is otherwise the design record, which is why it is worth the line.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -16,6 +16,94 @@
...
+# is much further right -- 22 in the CLI this daemon ships against. So a small
+# bounded indent is what separates an option from a wrapped description, without
+# depending on the exact description-column width, which varies with the longest
+# option name in the listing.
+#
+# Anchoring merely to line START is not enough, and that was the round-1 finding
+# on PR #88: a description that wraps such that its second line BEGINS with
+# `--status` reads as an offer of `--status`, and 0.1.0's own help already wraps
+# `--self`'s description onto its own line, so the shape is not hypothetical.
+MAX_OPTION_INDENT = 3
```
alissa-app, line 83: `[nit]` **The constant's comment argues one direction of the bound only.**

Six lines justify why the indent must be small (a wrapped description line beginning `--status` must not read as an offer) — a good justification, and the accompanying test is the right test. The other direction is not stated: a help listing that indents its **option** column by four or more spaces makes `_advertises` answer `False` for flags that are genuinely offered, and the daemon then simply never narrows. That is the fail-safe direction, deliberately, and it matches this module's stated posture — but it is *silent*. The false-positive path at least warns and retries plain; the false-negative path produces no log line at all and looks exactly like an old CLI. Worth the same sentence the other direction gets, so the next person to touch the number knows what they are trading on both sides.

Relatedly, `test_an_option_is_found_however_the_listing_indents_its_column` pins that three spaces still matches, but nothing pins where the cutoff actually is. A companion asserting that a four-space option column is *not* matched would fix the boundary the constant exists to set — right now `MAX_OPTION_INDENT` could be changed to 8 and the whole suite would stay green except the wrapped-line test, which depends on the description column at 22 rather than on this bound.

alissa-app (approved):

Reviewed head `0208dfb607cc3b210f704640e2181b436b46ae44` (re-read immediately before submitting; unmoved).
CI rollup **at that sha**: 6 contexts, all `completed` / `success` (unit, wheel, entrypoint, types, version-bump, style); no legacy commit statuses. Green.

Review surface taken as the PR states it: the single commit `9273722..0208dfb` (161 insertions / 9 deletions across 5 files), not the diff-vs-`main`, which still carries #88.

**No blockers, no majors.** Four `[minor]` and two `[nit]` findings are on the diff for your discretion; none of them gates the merge.

### What I verified rather than took on trust

| claim | how | result |
| --- | --- | --- |
| 776 unit tests pass | `PYTHONPATH=src/main python3 -m pytest src/test -q` in a throwaway worktree at `0208dfb` | **776 passed** in 26.5s |
| style | `bash check-style.sh alissa-tools-github-revloop` | clean |
| types | `bash check-types.sh alissa-tools-github-revloop` | `Success: no issues found in 30 source files` |
| the counterfactual, exactly as claimed | reverted `alissa.py` + `loop.py` to `9273722`, kept the new tests | **4 failed, 772 passed** — and precisely the four named tests, no others |
| each of #88's five findings maps to this diff | thread-by-thread against #88's review | all five present; all five triaged `[triage:pursue]` on their threads |
| the `[minor]` answered with documentation actually landed | #88's PR body | the operator's-gate bullet is there, and it is accurate |
| README's mechanism table really has four rows | `README.md` | it does |

The `_pass_tasks is None` gate is right, and right for the reason given: it is a pure short-circuit in front of `consume_review_task_miss`, so no window is spent when the memo is populated, and the fall-through search reads `_pass_task_list()` — which returns the memo — so **no read is ever added**. I walked the multi-PR pass orderings looking for a case where the gate costs a fetch and did not find one. Two second-order effects, both benign and neither a finding: the free search re-records the miss at full TTL, so a window is now spent only on polls where no corpus was fetched (strictly better, but see `[minor]` 1); and an in-window PR now pays a regex scan over the in-memory corpus, which is free at these sizes.

`_advertises`'s move from line-start to option-column is correct against the shipped 0.1.0 help (options at two spaces, description column at 22), the interpolated bound is an `int` constant and the pattern is linear — no ReDoS surface.

### Dimension report

- **Correctness** — nothing found. All five #88 findings addressed as described; the gate, the `_task_list_narrowing_disabled` clause and the option-column anchor each do what their comments claim, each has a test that fails when reverted, and the reverted-code run fails exactly those four and nothing else.
- **Coding practices** — the two `[nit]`s and `[minor]` 3 below. Comment density, naming and test-naming all match the surrounding module, which is unusually well documented; the findings are about that documentation drifting from the code this PR just changed, not about the code.
- **Security** — nothing found. No new input reaches a shell, path, or query; `re.escape` still guards the flag interpolation and the only new interpolation is an integer constant; no secrets, no logging of task content, no dependency change.
- **Performance** — nothing found; this is a net reduction. The gate cannot add a list call, and the `_task_list_narrowing_disabled` clause removes a third `alissa task list` from the failed-narrowed-call path.
- **Side-effects / blast radius** — `[minor]` 4 (merge order). No contract, schema, config-key or CLI-surface change; no migration; the ledger tables are untouched. `review_task_miss_ttl_polls` keeps its meaning but its *unit* shifts slightly (see `[minor]` 1). Version stays `0.18.0`, correctly — `check-version-bump` compares against PyPI and that release is unpublished, which the green check confirms.

### The findings

1. `[minor]` README still documents the pre-gate latency contract in three places — the change's headline win is invisible to the operator sizing the knob.
2. `[minor]` The `KNOWN GAP` deferral names no issue or task, against this module's own convention.
3. `[minor]` The `_task_list_narrowing_disabled` clause means a call-local fact and is only correct by coupling to `task_list_argv`'s early return.
4. `[minor]` The merge order is enforced by prose alone.
5. `[nit]` `_review_task`'s docstring still says the negative answer is consulted "ONLY when there is no cached ref at all" — there is now a second precondition.
6. `[nit]` `MAX_OPTION_INDENT`'s comment argues only the false-positive direction.

Verdict envelope recorded on **TASK-2141032056** (created this round per CR2 — no review task existed for #89).

---

`[minor]` **The merge order is enforced by prose alone.** The banner is clear and the reasoning behind cutting from `9273722` rather than `main` is sound — the property the convention protects (the approved head never moves) does hold. But nothing mechanical stops #89 merging first, and if it does, #88 is stranded with an empty diff and loses both its history and its `Closes #87` claim. The usual stacked-PR answer, retargeting this PR's base to `TASK-115878658-LIST-BOUND`, is the wrong one **here** — it would put #89's commits under #88 and move the approved head, which is exactly what the judgment call was protecting. The cheap mechanical guard is the other one: flip #89 back to **draft** now that this review is in (or apply a blocking label), and lift it once #88 lands. Recorded as a finding because it is the one item on this PR whose failure mode is unrecoverable-by-retry, not because the note is inadequate.

README.md
RHDZMOTA: `[triage:later]` — **TASK-1006125108**

Accepted in full, including the sharpest part: the knob's *unit* shifted from
"polls" to "polls in which no corpus was fetched", and the change's best argument
for itself is nowhere in the docs an operator actually reads. Line 129 now
understates how fresh the daemon is.

Deferred rather than fixed here for one reason only: this round is an **approve**,
so this branch is terminal — a commit after it can be lost behind the squash —
and the remedy for an approve-carried finding is a new branch off the default
branch. That branch cannot exist yet: `main` has neither #88's README section nor
#89's gate, so the sites to edit are not there. Stacking a third level on
#88 → #89 would force the operator into an exact three-way merge order, which is
a worse trade than four registered items (the reasoning this repo settled on in
PR #53).

Registered as **TASK-1006125108** with the three line references and the
post-gate wording, alongside the other three below. It lands on one branch off
`main` once #88 and #89 have merged.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` — **TASK-1573578582**

You are right, and the comparison to the two deferrals forty lines up
(`TASK-317167904`, `TASK-1897198077`) is the argument — that is this module's
convention and I broke it by pointing at a PR body instead. A checkbox on a
merged PR is not something anyone can open, assign or close.

The task now exists: **TASK-1573578582** — *"prove a narrowed `alissa task list`
generation before the negative cache may trust its misses"*, carrying the
partially-served-filter failure mode, the sticky-miss consequence, and an
acceptance criterion that requires this very comment to cite it.

The one-line edit that puts the ref into the comment is in **TASK-1006125108**
with the other three documentation items, for the same reason as the finding
above: this round approved `0208dfb`, so nothing more may be pushed to this
branch, and `main` does not yet contain the comment. I have also cited
TASK-1573578582 in #88's operator's-gate bullet, which is a body edit and
therefore safe on an approved branch — so the durable pointer exists now, before
either PR merges, rather than after the cleanup lands.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:later]` — **TASK-1006125108**

This is the best finding of the round and I am recording the reasoning rather
than just the verdict, because the code as it stands is *correct* and that is
exactly what makes it dangerous. The clause means "this call already retried
plain"; what it tests is a process-level flag, and the two coincide only because
`task_list_argv` early-returns bare `plain` whenever the flag is set. That
coupling is invisible from the guard.

Your future edit is not hypothetical, which is what settles it: "keep honouring
`task_list_self_scope` even after the probed narrowings are disabled" is a
reasonable thing for someone to want, `--self` being a deployment opt-in rather
than a probe guess. The moment it lands, `argv != plain` on a
narrowing-disabled call, this clause short-circuits every subsequent call, and
the empty-corpus cross-check disappears for the life of the process — silently,
with no test failure, and the failure mode is skipped reviews.

Taking `retried_plain` verbatim. Deferred to **TASK-1006125108** only because the
approve makes this branch terminal and `main` does not yet contain the guard;
`test_a_failed_narrowed_call_does_not_also_pay_the_empty_cross_check` keeps
pinning the behaviour in the meantime, so the change is behaviour-preserving when
it lands.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: `[triage:later]` — **TASK-1006125108**

Correct — the docstring is where this module states the mechanism's *contract*,
and it now states it with one precondition missing, so a reader who stops there
has a wrong model. That the `why` lives correctly in the inline comment below
does not excuse the contract paragraph being incomplete; in a module whose
documentation is the design record, that is a defect in the record.

Taking your suggested clause. Bundled into **TASK-1006125108** with the other
documentation items rather than pushed here — the approve makes this branch
terminal.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:later]` — **TASK-1006125108**

Both halves accepted, and the second is the one that matters: the false-negative
direction is *silent*. A four-space option column makes `_advertises` answer
False for a genuinely offered flag, the daemon simply never narrows, and that is
indistinguishable in the logs from an old CLI — whereas the false-positive path
warns and retries. The comment argues only the direction that announces itself.

And you are right that nothing pins the cutoff:
`test_an_option_is_found_however_the_listing_indents_its_column` asserts three
spaces still matches, but `MAX_OPTION_INDENT` could be widened to 8 with the
suite green apart from the wrapped-line test, which keys on the description
column at 22 rather than on this bound. A companion asserting a four-space
option column is *not* matched is what gives the constant a real boundary.

Both the sentence and the boundary test are in **TASK-1006125108**. Deferred for
the same reason as the rest: this round approved `0208dfb` and the branch is
terminal.

RHDZMOTA:
Triage for the round-1 `[minor]` on **merge order**, which landed in the review body rather than on a thread:

`[triage:pursue]` — done, mechanically, on this PR.

Agreed that this is the one item whose failure mode is unrecoverable by retry,
and agreed that the usual stacked answer (retargeting this PR's base to
`TASK-115878658-LIST-BOUND`) is the wrong one here — it would move the approved
head, which is the thing the judgment call exists to protect.

Taking your second option rather than the first. This PR now carries the
blocking label **`blocked:merge-order`** ("do not merge until the PR named in
`Follow-up-To:` has merged"), created for this purpose and applied now. I am
deliberately not flipping back to draft: this session's operating rules forbid
leaving a follow-up in draft, because the daemon's orphan sweep uses draft state
to find abandoned follow-ups and would flip it ready again — the label survives
that sweep and the draft flag would not.

The banner in the body stays as the human-readable half, and now names the label.
Whoever merges #88 removes the label from this PR.

---

Triage for the round as a whole: 1 `[triage:pursue]` (above, done), 5 `[triage:later]` with registered task refs on their threads — **TASK-1573578582** (the `--status` proving gate) and **TASK-1006125108** (the four documentation/clause cleanups). Nothing is deferred in prose alone.

Both are downstream of the implementation task **TASK-115878658** by soft dependency, alongside origin **TASK-805662381**. No third stacked PR: this round approved `0208dfb`, so this branch is terminal, and the remedies need a branch off `main` that cannot exist until #88 and #89 have merged — stacking a third level would force an exact three-way merge order, which is the worse trade this repo settled on in PR #53.

RHDZMOTA:
**Gate satisfied — `blocked:merge-order` removed.**

#88 merged as `9e90893` (a merge commit, so `9273722` is preserved on `main`), and issue #87 closed COMPLETED. The merge-base of this branch and `main` is now `9273722`, so this PR's diff is exactly its own commit — 5 files, +161/−9 — which is the review surface `alissa-app` approved on `0208dfb`. `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`.

Nothing further is owed on this branch: round 1 approved, all six findings triaged (1 `[triage:pursue]`, done; 5 `[triage:later]` against **TASK-1573578582** and **TASK-1006125108**), and the head has not moved since the approve.

PR #91, opened 2026-08-21
revloop: fall back to the Actions API for the CI rollup when check-runs is forbidden

### LINKED ISSUE #90: revloop: fall back to the Actions API for the CI rollup when check-runs is forbidden (fine-grained PAT support)
## Problem

The deployed revloop reviewer credential is a **fine-grained PAT** (identity `alissa-app`), and GitHub does not offer the `Checks` permission on fine-grained PATs at all — it was briefly available, then disabled; only GitHub Apps can hold it (see https://github.com/orgs/community/discussions/129512). Consequently `GET /repos/{owner}/{repo}/commits/{sha}/check-runs` answers **HTTP 403 "Resource not accessible by personal access token"** for every repo, `check_rollup` degrades to `CHECKS_UNKNOWN`, and the loop can never confirm a head green:

- Observed live on fahera-mx/studio.alissa.app#681 (2026-08-20): round 3's code verdict reached approve but was submitted comment-mode because "an unreadable rollup is not a green one" — https://github.com/fahera-mx/studio.alissa.app/pull/681#pullrequestreview-4980558392.
- Boot log (0.18.0, 2026-08-21 01:03 UTC) warns the same 403 per queued round and spawns anyway, telling the reviewer session to read the rollup itself — but the session's directive command (`gh api .../commits/<sha>/check-runs`, from `_CHECKS_BEFORE_VERDICT` in `loop.py`) uses the same credential and hits the same 403, so sessions cannot confirm green either.

Net effect: **no round can ever approve** under a fine-grained PAT deployment. Everything else about the loop works.

## Verified viable fallback (probed live with the deployed token, 2026-08-21)

The same fine-grained PAT carries `Actions: Read`, and:

- `GET /repos/{o}/{r}/actions/runs?head_sha={sha}` → 200, returns `workflow_runs` with `status`/`conclusion` per run (verified against studio#679 head `feeec960`: 1 run, CI, success).
- `GET /repos/{o}/{r}/actions/runs/{run_id}/jobs` → 200, returns per-job `name`/`status`/`conclusion` (verified: jobs `changes`, `test`, `cli`, … with success/skipped conclusions — same granularity the check-runs rollup shows).
- `GET /repos/{o}/{r}/commits/{sha}/check-runs` → 403 (the gap this task closes).
- `GET /repos/{o}/{r}/commits/{sha}/status` → 200 but empty `pending` for Actions-only repos (the codebase already knows this trap; `_commit_statuses` ignores the combined `state`).

## Scope

All changes in `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/`:

1. **`ghclient.py` — Actions-API fallback in the rollup read.** When the check-runs listing fails with an **authorization 403** (the `forbidden_is_rate_limit=False` path; distinguish it from `RateLimited`, which must still propagate, and from `TruncatedListing`, which must keep refusing), read the same commit's CI through the Actions API instead:
   - list workflow runs for the sha (`actions/runs?head_sha=`, payload key `workflow_runs`, has `total_count` — page with the same completeness contract as `_rollup_listing`: an unsatisfied count refuses rather than answering from a partial read);
   - a sha can carry multiple runs of the same workflow (re-run attempts share a run id, but distinct trigger events create distinct runs): keep only the **most recent run per `workflow_id`** before reading jobs;
   - for each kept run, page its jobs (`actions/runs/{id}/jobs`, key `jobs`, has `total_count`) and map each job to a `CheckContext` under the exact `_check_runs` rules: a job not `completed` carries **no** conclusion (reads as running) whatever its payload says; completed conclusions pass through for `rollup_of` to judge (`skipped`/`neutral` pass, as today);
   - a run that is itself not `completed` but exposes no jobs yet must still read as running, never as green;
   - `_commit_statuses` is still read and merged, unchanged;
   - a 403 from the Actions API too, or any other failure, keeps today's `CHECKS_UNKNOWN` behavior.
2. **Honest labeling of the fallback answer.** The Actions API sees only GitHub-Actions-produced contexts — check runs posted by third-party check apps are invisible to it, so a fallback rollup could call green a commit whose non-Actions check failed. Every repo this fleet reviews runs Actions-only CI today, so this is accepted — but the rollup answered via fallback must say so (log line per read and/or a marker on the `CheckRollup`) so an operator reading "approved on green" can see which read path answered.
3. **`loop.py` — `_CHECKS_BEFORE_VERDICT` directive.** Teach the session the same fallback: if the check-runs read answers 403, read `actions/runs?head_sha=<sha>` + each run's jobs with identical conclude/wait/verdict semantics (approve only when every context concluded and none failed; while running, wait and re-read within the existing bound; failed context ⇒ request_changes naming the job and linking its run).
4. **Tests** (existing fake-`gh` harness in `src/test/.../test_loop.py` / ghclient tests): 403 → fallback → green head approves; 403 → fallback → failed job yields the failed context with its url; 403 → fallback → run still in progress reads as running (held, not approved); same-workflow duplicate runs dedupe to the latest; fallback listing with unsatisfied `total_count` refuses (no partial green); Actions API also 403 ⇒ `CHECKS_UNKNOWN`; a checks-capable credential never enters the fallback (check-runs success path byte-for-byte unchanged).

## Out of scope

- No GitHub App migration, no new credentials, no config keys required for the default behavior (falling back on 403 needs no opt-in — it only triggers where today's behavior is a guaranteed dead end).
- No change to verdict gating semantics, round accounting, or the spawn gate.
- `REVLOOP_VERSION` pin bump and redeploy are the operator's, after merge.

## Acceptance sketch

With a credential lacking checks read (simulated in tests), a green Actions-only head produces a rollup the daemon can approve on; a red one produces the failed context; nothing changes for credentials that can read check-runs.

## Definition of Done

A PR against fahera-mx/alissa-github-review-daemon implementing the Actions-API rollup fallback (ghclient.py), the honest fallback labeling, the _CHECKS_BEFORE_VERDICT directive update (loop.py), and the test matrix above is opened draft-first, flipped to ready-for-review, and its URL attached to this task as evidence; the adversarial review loop reaches approve and a human merges.

## Acceptance criteria

- [ ] The implementation PR (ready-for-review, URL on this task as evidence) is approved by the review loop and merged by a human.
- [ ] Tests demonstrate: 403-on-check-runs falls back to Actions runs+jobs and approves green / fails red / holds running heads; Actions-403 still yields CHECKS_UNKNOWN; checks-capable credentials never enter the fallback. After the operator bumps REVLOOP_VERSION and redeploys, a live round approves a green head under the fine-grained PAT (no comment-mode degradation).

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py, alissa-tools-github-revloop/src/test/**
Alissa-Task: TASK-397619378

RHDZMOTA:
Closes #90

Origin task: TASK-397619378
Implementation task: TASK-1977649551

## What this changes

GitHub does not offer the `Checks` permission on fine-grained PATs at all, so the deployed reviewer credential (`alissa-app`) gets a hard 403 from `GET /repos/{o}/{r}/commits/{sha}/check-runs` on every repo. `check_rollup` degraded to `CHECKS_UNKNOWN` and **no round could ever approve** — studio#681 round 3 reached an approve verdict and had to submit it comment-mode because "an unreadable rollup is not a green one".

`ghclient.check_rollup` now reads the same commit's CI through the Actions API when the check-runs listing answers an **authorization** 403 (throttling 403s still raise `RateLimited` and back off): workflow runs for the sha, deduped to the most recent run per `workflow_id`, then each kept run's jobs mapped to `CheckContext` under the exact `_check_runs` rules. `_commit_statuses` is still read and merged. Both fallback listings page under the same completeness contract as `_rollup_listing`, so an unsatisfied `total_count` refuses rather than answering from a partial read.

The Actions API sees only Actions-produced contexts, so a fallback answer says so: `CheckRollup.via_actions_fallback`, a marker on `summary` (which reaches both the daemon log line and the verdict body), and a `WARNING` per fallback read.

`loop._CHECKS_BEFORE_VERDICT` teaches a reviewer session the same fallback — its directive command used the same credential and hit the same 403, so sessions could not confirm green either.

## Delivery notes

**Judgment calls**

- The issue asks for a log line "and/or a marker on the `CheckRollup`". Both are implemented: the marker is what a verdict body can carry, and `summary` already flows into the PR-visible degraded-verdict text, so an operator reading "approved on green" sees the read path without going to the daemon log.
- Recency for the per-`workflow_id` dedupe is `(run_number, id)`, not a timestamp: `run_started_at`/`updated_at` are the fields a re-run rewrites, while `run_number` is the workflow's own monotonic counter. A payload missing both sorts last, so a malformed entry can never displace a run that identifies itself.
- The issue specifies "a run that is itself not `completed` but exposes no jobs yet must still read as running". Implemented one step wider: **any** run with no readable jobs contributes itself as one context, judged on its own status/conclusion. A *completed* run with no jobs would otherwise vanish from the rollup entirely, which is the fail-open direction; this way a red run with unreadable jobs still blocks.
- `_is_authorization_forbidden` re-checks the rate-limit markers even though `_api(forbidden_is_rate_limit=False)` already raises `RateLimited` for them first, so today the branch cannot fire. Kept deliberately (with its own unit test): the predicate decides whether to answer the gate from a narrower source, and doing that for a condition that clears itself in seconds is the wrong trade in a fail-closed component.
- Version bumped `0.18.0` → `0.19.0`: `check-version-bump.yaml` requires it for any change under the dist minus `src/test/`. `0.19.0` is the next version PyPI does not have (latest published is `0.18.0`). The `ARG REVLOOP_VERSION` pin in `docker/claude/Dockerfile` is deliberately **not** touched — the issue assigns the re-pin and redeploy to the operator, and no config key was added, which is the case `docker/claude/tests-entrypoint-ui.sh` requires a same-PR pin bump for.

- [ ] No live GitHub call was made from this session against a fine-grained PAT. The 403 → fallback path is exercised only against a faked `run_json`; the real payload shapes come from the probes recorded in issue #90, not from output captured here.
- [ ] The issue's second acceptance bullet ("after the operator bumps `REVLOOP_VERSION` and redeploys, a live round approves a green head under the fine-grained PAT") is by construction not verifiable pre-merge. It needs the merge, the PyPI publish of `0.19.0`, the `ARG REVLOOP_VERSION` re-pin and a redeploy.
- [ ] Behaviour against a repo whose CI includes a **third-party check app** is untested and knowingly wrong-in-the-narrow-direction: the fallback cannot see such a context and could call that commit green. Accepted per the issue's scope note (this fleet is Actions-only); the labelling is what makes it visible if that stops being true.
- [ ] `GET /actions/runs?head_sha=` is assumed to return the same `total_count` completeness signal the check-runs listing does. Verified against the fixture, not against a live 100+-run commit.
- [ ] The review loop that reviews THIS pull request runs on the deployed 0.18.0 daemon — i.e. under the very bug being fixed. Its own rollup read will 403 and degrade to `CHECKS_UNKNOWN`, so an approve verdict on this pull request may still be submitted comment-mode. That is expected and is not evidence about this change.
- [ ] Docker image build and container boot were not run (no docker in this runner); nothing in this change touches the image beyond the version file the operator re-pins.

**Verification**

- `bash tests-unit.sh alissa-tools-github-revloop` (with the local source on `PYTHONPATH`, since a `0.18.0` copy is installed system-wide in this runner) — **794 passed**, up from 776 on the base commit.
- `bash tests-coverage.sh alissa-tools-github-revloop` — 94% total; `ghclient.py` 92%.
- `bash check-types.sh alissa-tools-github-revloop` — clean (30 files); one `int(object)` overload error was found and fixed before this commit.
- Counterfactual sweep — each new guard was reverted in turn and the suite re-run, confirming a test actually fails for: no fallback at all (11 failures), no per-workflow dedupe (1), a jobless run contributing nothing (2), the fallback label dropped (2), `head_sha` not forwarded to the runs listing (1), a queued job's conclusion trusted (1). All restored, suite green.
- Pre-existing flake found and fixed while running the suite repeatedly: `test_the_directives_wait_figure_tracks_the_spawn_bound` asserts `"99" not in directive`, and the session name's random hex nonce contains `99` about 1.8% of the time (measured: 0.018 over 20k nonces) — roughly one red suite in 55. The session name is now substituted out before the assertion, which keeps its full strength; confirmed by forcing a `99`-bearing nonce and watching the old assertion fail and the new one pass. Called out again under Scope touched.

**Scope touched**

Declared scope was `revloop/ghclient.py`, `revloop/loop.py`, `src/test/**`. Modified:

- `src/main/.../ghclient.py` — the fallback read, the label, the shared `_completed_conclusion` helper (which `_check_runs` now also uses; same behaviour), `_rollup_listing` query params.
- `src/main/.../loop.py` — `_CHECKS_BEFORE_VERDICT` only, one added `(2b)` clause.
- `src/test/.../test_loop.py`, `src/test/.../test_directives.py` — the test matrix, plus the flake fix above.
- **Excursion:** `src/main/.../version` `0.18.0` → `0.19.0`. Outside the issue's declared scope list but required by `check-version-bump.yaml` for any non-test change under the dist, so the PR cannot go green without it. Nothing else about the release (the Dockerfile pin, the redeploy) is touched.


alissa-app:
- 2026-08-21 01:59:48 UTC — `review-alissa-github-review-daemon-pr91-r1-f0a8b3` — round 1 of 22 — spawned
- 2026-08-21 04:12:51 UTC — `review-alissa-github-review-daemon-pr91-r1-f0a8b3` — round 1 of 22 — deferred — session `review-alissa-github-review-daemon-pr91-r1-f0a8b3` still busy

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -754,18 +869,119 @@ def _check_runs(self, owner: str, repo: str, sha: str) -> list[CheckContext]:
...
+        check run GitHub would have published for it (same name, same status,
+        same conclusion), so `rollup_of` judges the fallback answer under
+        exactly the rules it judges the real one under, including `skipped` and
+        `neutral` passing for a path-filtered matrix.
+
+        What this read CANNOT see is any check run posted by a third-party check
+        app; see ACTIONS_FALLBACK_NOTE for why that is accepted here and how the
+        answer says so.
+
+        COST: this path costs one call for the run listing plus one per kept
```
alissa-app, line 902: [minor] This COST note names a bound that does not exist on the production path, and it is the sentence an operator would size rate-limit headroom from.

`_dry_run_rollups` (loop.py:1430) is the only per-round rollup memo, and `_evaluate`'s dry-run branch is the only thing that reads it — it exists precisely because a dry-run pass must write no ledger row. On the production path there is no memo at all:

- `_checks_at_spawn` (loop.py:1956) calls `check_rollup` on **every poll pass** for as long as a round is held on `checks_spawn_wait_seconds`;
- `_gate_on_checks` (loop.py:2438) calls it again at verdict time, and again on every poll while the approve is held on `checks_wait_seconds`.

So what actually bounds the spend is the poll interval (60s by default), not caching. Under the fallback each of those reads becomes `1 + N` calls instead of `1`, which is where the cost multiplies.

Second, the fan-out itself is unbounded in a way the sentence's "two or three calls" does not convey: `CHECK_RUN_PAGE_LIMIT` bounds the run *listing* at 5x100, but every kept run then costs its own `jobs` call, each with `_api`'s 60s timeout — a sha carrying many distinct workflows turns one rollup read into a long synchronous stall inside a poll pass that has other PRs to decide.

Both are comment-level: correcting the claim to "the poll interval is what bounds how often this is paid" is enough. Capping the kept-run fan-out would be a separate task if it is ever worth one.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -251,25 +265,69 @@ class CheckRollup:
...
+    # than the one it stands in for (see ACTIONS_FALLBACK_NOTE), so the answer
+    # carries which path produced it everywhere the rollup is reported.
+    via_actions_fallback: bool = False
 
     @property
     def summary(self) -> str:
-        """One log-line description of the rollup."""
+        """One log-line description of the rollup.
+
+        The fallback marker rides on EVERY state, not just green: an operator
```
alissa-app, line 278: [minor] The marker does ride on every rollup *state*, but on the one state this labelling exists for — green — it never reaches anything PR-visible.

Tracing where `summary` actually goes:

- `_gate_on_checks` GREEN branch (loop.py:2440-2446) `log.debug`s the summary and returns a bare `ChecksGate()`. The approve review body that follows carries no rollup text at all.
- Every path that *does* put `summary` in front of a human is a non-green one: `_note_checks_hold` (the activity comment, loop.py:2607) and the `CHECKS_UNSETTLED_LEAD` degraded-verdict body — both reachable only from PENDING/UNKNOWN.

Which inverts the PR description's stated goal, "an operator reading 'approved on green' sees the read path without going to the daemon log": an operator reading a red or held verdict sees it inline, and one reading an approve has to go to the daemon log after all — and at `debug` in `_gate_on_checks`, though `check_rollup`'s own per-read `log.warning` does carry it at a level that will be on.

Not a DoD miss — issue #90 asks for "a log line per read **and/or** a marker on the `CheckRollup`" and both exist, so this is honest labelling by the letter. But if the intent is the one the description states, the green branch is the one place that needs the marker: a clause on the approve body when `rollup.via_actions_fallback`, or at minimum lifting that `log.debug` to `info`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -754,18 +869,119 @@ def _check_runs(self, owner: str, repo: str, sha: str) -> list[CheckContext]:
...
+        """One run per workflow: the most recent.
+
+        A sha carries more than one run of the same workflow whenever it was
+        triggered more than once -- a re-run shares its run id, but a distinct
+        trigger event (a `push` and a `pull_request` on the same commit, a
+        workflow re-dispatched by hand) creates a distinct run. Reading all of
+        them would judge the commit on a superseded attempt: an earlier failed
+        run whose re-trigger passed would hold the head red forever.
+
+        Recency is `run_number` then `id`, both monotonic per workflow and both
```
alissa-app, line 931: [minor] Worth recording that this dedupe has one fail-open case, since it is the direction this component is built never to move in.

The docstring's justification is about *re-triggers of the same event*, where collapsing to the newest run is plainly right. But the grouping key is `workflow_id` alone, and one sha can carry runs of the same workflow from **different trigger events** — a `push` run and a `pull_request` run of a workflow declaring both. Those are two genuinely distinct contexts: GitHub publishes a check run per job for each, so the read this stands in for shows both. Here the lower `run_number` is dropped unread, and if the dropped run is the red one the fallback answers **green** for a commit whose real rollup is red.

Reachability, checked rather than assumed: every workflow in this repo is `on: [pull_request]` (`package-publish.yaml` adds only `types: [closed]`), so no workflow here can double-trigger and this is unreachable on `alissa-github-review-daemon` today. It is not checked for the fleet's other repos, and this code is a library they all run.

Issue #90 does specify "most recent run per `workflow_id`" in exactly these words, so I am not asking you to contradict the spec in this PR — flagging it because the issue's own parenthetical ("distinct trigger events create distinct runs") is describing the case that breaks it. `(workflow_id, event)` as the key would keep the re-trigger behaviour the docstring argues for while not dropping a distinct context, since a re-run preserves its run's `event`. Follow-up task is a fine home for it.

alissa-app (approved):

Reviewed as a fresh reviewer session against the origin task's definition-of-done (issue #90). The change does what it says: `ghclient.check_rollup` takes the Actions runs+jobs path only on an **authorization** 403 from the check-runs listing, maps jobs to `CheckContext` under the same `_check_runs` rules, merges `_commit_statuses` unchanged, keeps both fallback listings under `_rollup_listing`'s completeness contract, labels the answer, and `loop._CHECKS_BEFORE_VERDICT` teaches a session the same fallback. Every failure direction I could find is the fail-**closed** one.

### Verification I ran myself at head `0729b516`

| check | result |
| --- | --- |
| `PYTHONPATH=src/main python3 -m pytest src/test -q` | **794 passed** — matches the PR's claim exactly |
| `pycodestyle alissa-tools-github-revloop --max-line-length 120` | clean |
| `mypy alissa-tools-github-revloop --ignore-missing-imports` | `Success: no issues found in 30 source files` |
| CI rollup at `0729b516` | 6 contexts, all `completed`/`success`; `commits/<sha>/status` carries 0 legacy statuses |
| PyPI | latest published is `0.18.0`, `0.19.0` is unpublished — the bump is correct and is required by `check-version-bump.yaml` for a non-test change under the dist |

I also read the control flow directly rather than trusting the description: `RateLimited` is not a `CommandError`, so it bypasses the inner handler and reaches `except RateLimited: raise`; `TruncatedListing` from *either* fallback listing lands on the outer handler as `CHECKS_UNKNOWN`; an exception raised inside the `except CommandError` block (the Actions read 403ing too) propagates to the outer handler as `CHECKS_UNKNOWN`. `_completed_conclusion` is byte-for-byte equivalent to the inline logic it replaces in `_check_runs`, and `_rollup_listing` with `params=()` builds an identical argv, so the checks-capable path is genuinely unchanged.

### Rubric

- **Correctness** — satisfies the DoD. All seven bullets of issue #90's test matrix have a corresponding test, and the tests drive `run_json` rather than a faked `_api`, which is the layer that actually matters (the 403 has to survive `forbidden_is_rate_limit=False` as a `CommandError` to reach the fallback at all). Suite is green and grew 776 → 794. Three minors below; no correctness defect found on the main path.
- **Coding practices** — matches repo idiom closely (comment density, the "why", the fail-closed framing). Style and types clean. The `version` bump is an excursion outside the issue's declared scope but is CI-mandated and disclosed; the `test_directives` flake fix is disclosed and keeps the assertion's strength. No dead code, no debug leftovers, no untracked TODOs.
- **Security** — nothing found. `sha` reaches `gh` as its own argv element under `-X GET -f head_sha=...` (no shell, no interpolation into a path). The new `WARNING` logs `str(exc)[:200]`, and `CommandError.__str__` is `f"{argv[0]} exited …"` — `argv[0]` is `"gh"` and the credential travels in `env`, not argv, so no token can reach the log. No new dependency, no widened permission (the fallback needs only the `Actions: Read` the credential already holds).
- **Performance** — one finding (minor 1). The fallback costs `1 + N` calls where the read it replaces cost `1`, and the docstring's stated bound on how often that is paid does not exist on the production path. Not gating: the affected deployment is the one where the alternative is a permanent dead end.
- **Side-effects / blast radius** — nothing found. `rollup_of` gains a defaulted parameter and `CheckRollup` a trailing defaulted field, so both stay source-compatible; the only in-repo callers are in `ghclient.py` itself and `loop.py` via `check_rollup`. Nothing under `docker/` is touched, and leaving `ARG REVLOOP_VERSION` alone is right — `docker/claude/tests-entrypoint-ui.sh` demands a same-PR pin bump only when a config key is added, and none is. The one behavioural blast-radius item is the accepted, documented, and labelled third-party-check-app blind spot.

### Findings — 0 blocker, 0 major, 4 minor, 0 nit, 0 question

Three are inline. The fourth has no line in this diff to hang on:

**[minor] `loop.py` `CHECKS_UNREADABLE` (~line 466) now misdirects the operator it is written for.** It still reads *"check that the reviewer credential carries `checks: read` on this repo"*. After this change, a fine-grained-PAT deployment reaching `CHECKS_UNKNOWN` means something else entirely — the **Actions** read failed too, or a listing was truncated — and `checks: read` is a permission GitHub will not grant on a fine-grained PAT at all, so that sentence sends an operator to do the one thing that cannot work. The string is pre-existing, but this PR is what makes it wrong. Worth a sentence pointing at the Actions read, or a follow-up task.

### Why approve

Only minors stand, so the mechanical rule gives `approve`. All four are in comments, labelling reach, and a spec-mandated trade — none of them changes what the gate decides on a green or red head, and none is worth another round. Deferring any of them is fine provided it becomes a registered task per the taxonomy. The head's CI is green (6/6 concluded, none failed), re-read immediately before submitting this review.

<sub>Reviewed under the alissa-code-review rubric (CR4), read-only posture (CR6). Verdict of record is the envelope on TASK-1287865828.</sub>

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: `[triage:pursue]`

Both halves are right, and the second one is the part I had not thought through.

On the memo: I checked `_dry_run_rollups` and you are correct that it is dry-run-only — `_evaluate`'s dry-run branch is its sole reader, and it exists so a dry-run pass writes no ledger row. There is no production memo, so "the gate's existing per-round caching is what bounds how often that is" is simply false. What bounds it is the poll interval, across both `_checks_at_spawn`'s hold and `_gate_on_checks`'s.

On the fan-out: "two or three calls" describes this repo, not the bound. `CHECK_RUN_PAGE_LIMIT` bounds the run listing, nothing bounds the kept-run count, and each kept run costs a `jobs` call carrying `_api`'s 60s timeout — so the honest statement is that a many-workflow sha turns one rollup read into a synchronous stall inside a poll pass with other PRs waiting.

Correcting the comment to say both, per your "comment-level is enough". A fan-out cap is a real question but a separate one, and I am not smuggling it into a comment fix.

Carried to a follow-up PR rather than a push here — this branch is approved, and a commit after the squash-merge would be silently lost. I will post the follow-up's URL on this thread once it is cut.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: `[triage:pursue]`

You traced this further than I did. I asserted in the PR description that an operator reading "approved on green" sees the read path inline, and that is exactly the one state where they do not: the GREEN branch of `_gate_on_checks` `log.debug`s the summary and returns a bare `ChecksGate()`, so the approve body carries no rollup text. Every path that does surface `summary` — the hold activity comment, the degraded-verdict lead — is reachable only from PENDING or UNKNOWN.

So the labelling is honest by issue #90's letter ("a log line per read and/or a marker on the `CheckRollup`" — both exist, and `check_rollup`'s own per-read `log.warning` is at a level that will be on), but it does not do the job my own description claimed for it. That mismatch is mine to fix, not to argue away.

Taking both remedies you name: a clause on the approve body when `rollup.via_actions_fallback`, and lifting the green-branch `log.debug` to `info` so the daemon log says it too. That puts the marker in front of the operator on the state it exists for.

Carried to the follow-up PR, not a push here — this branch is approved. Its URL will land on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: `[triage:pursue]`

Agreed, and this is the one of the four I care most about, because the failure direction is the wrong one: a dropped red run makes the fallback answer green for a commit whose real rollup is red. Fail-open, in the component whose whole contract is to fail closed.

Your reachability check matches mine — every workflow in this repo is `on: [pull_request]`, `package-publish.yaml` only adds `types: [closed]`, so nothing here can double-trigger. But `revloop` is a library the whole fleet installs, and I did not verify the other repos' triggers. "Unreachable on the repo I happened to test" is not a property I want holding up a fail-closed gate.

On the spec tension: issue #90 does say "most recent run per `workflow_id`" in those words, which is why the key stays `workflow_id` in this PR — implementing the issue as written is what this PR is for. `(workflow_id, event)` in the follow-up keeps the re-trigger behaviour the docstring argues for (a re-run preserves its run's `event`) while not dropping a genuinely distinct context, and a follow-up is not the PR that implements #90, so nothing about the spec claim is disturbed. That is the split you proposed and I am taking it as-is.

Going into the follow-up PR with a test for the two-trigger case, not just a task — a fail-open hole deserves a diff. URL to follow on this thread.

RHDZMOTA:
## CR8 triage — round 1 body finding

**[minor] `loop.py` `CHECKS_UNREADABLE` misdirects the operator → `[triage:pursue]`**

You are right that this PR is what makes the string wrong, even though the string is older than the PR. It reads "check that the reviewer credential carries `checks: read` on this repo", and after this change a fine-grained-PAT deployment that reaches `CHECKS_UNKNOWN` has already tried the Actions read and failed — so the remaining causes are an Actions-side failure or a truncated listing, and `checks: read` is a grant GitHub will not issue on a fine-grained PAT under any circumstances. Sending an operator to go enable it is sending them to the one action that provably cannot work, at the moment they are already blocked.

Fixing the text rather than filing it: it is a string edit in the same file this PR already touches, and the cost of carrying it is another operator hitting the dead end.

All four of round 1's minors are `[triage:pursue]`. None is being deferred to a task, so there is no registered-task reference owed for any of them.

They are landing as a **follow-up PR**, not as pushes here: this branch carries an approve, and a commit pushed after the operator squash-merges it is silently lost. The follow-up is cut from this PR's approved head `0729b51` — findings 1-3 touch code that only exists on this branch, so a branch off `main` would not build — and is **based on `main`** with an explicit gate that #91 merges first. Its URL goes on each finding's thread and here as soon as it is open.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: Follow-up is open: https://github.com/fahera-mx/alissa-github-review-daemon/pull/92 — the COST note now says the poll interval is what bounds the spend (across both `_checks_at_spawn`'s and `_gate_on_checks`'s holds), and states plainly that the kept-run fan-out is unbounded and each run costs a `jobs` call at `_api`'s 60s timeout. A fan-out cap is left as the separate question you called it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: Follow-up is open: https://github.com/fahera-mx/alissa-github-review-daemon/pull/92 — took both remedies. `CHECKS_FALLBACK_LEAD` goes on the approve body when `rollup.via_actions_fallback`, and the GREEN-branch log is now `info`. Two tests: a fallback green approve names the Actions API and what it cannot see, and an ordinary green approve is byte-identical to before.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: Follow-up is open: https://github.com/fahera-mx/alissa-github-review-daemon/pull/92 — key is `(workflow_id, event)`, with the fail-open case as a test (a red `push` run and a green `pull_request` run on one sha now reads RED, where it read green before). Two more tests hold the other side: a same-event re-trigger still collapses to the latest, and a payload with no `event` falls back to the old per-workflow grouping rather than fragmenting the rollup.

RHDZMOTA:
Follow-up PR for all four of round 1's minors: https://github.com/fahera-mx/alissa-github-review-daemon/pull/92

It carries findings 1-4, is cut from this PR's approved head `0729b51`, and is **based on `main`** with an explicit gate that this PR merges first (plus a `blocked:merge-order` label, since prose alone is not a mechanical guard). It closes no issue — #91 keeps the sole claim on issue #90.

Nothing further will be pushed to this branch; it is approved and merge-racy.

The `CHECKS_UNREADABLE` fix (finding 4) required updating one pre-existing test that asserted the literal string `checks: read`. Its intent is preserved: it now asserts the remedy text is actionable *and* that the grant GitHub will not issue is absent.

PR #92, opened 2026-08-21
revloop follow-up: PR #91 round-1 minors — dedupe key, approve labelling, operator notes

RHDZMOTA:
Follow-up to #91 (review round 1) — carries that round's four `[minor]` findings, all triaged `[triage:pursue]` on their threads.

Follow-up-To: #91

Origin task: TASK-397619378
Implementation task: TASK-1977649551

## ⚠️ Merge order: **#91 must merge first**

This branch is cut from #91's approved head `0729b51` and based on `main`. Three of the four findings touch code that exists only on #91's branch (`_actions_contexts`, `_latest_run_per_workflow`, `CheckRollup.via_actions_fallback`), so a branch off `main` would not build.

**Merging this before #91 would squash #91's entire delta into `main` under this PR's commit message**, leaving #91 to resolve as empty — losing its history and its closing claim on issue #90. This PR closes no issue; #91 owns that one.

## What this changes

| # | Finding | Remedy |
| --- | --- | --- |
| 1 | The `_actions_contexts` COST note claimed a bound that does not exist | Comment corrected: `_dry_run_rollups` is dry-run-only, so the **poll interval** bounds the spend; the kept-run fan-out is unbounded and each run costs a `jobs` call at `_api`'s 60s timeout |
| 2 | GREEN was the one rollup state whose fallback marker reached nothing PR-visible | `CHECKS_FALLBACK_LEAD` on the approve body when `rollup.via_actions_fallback`, and the green-branch log lifted `debug` → `info` |
| 3 | Dedupe on `workflow_id` alone drops a distinct trigger event — **fail-open** | Key is `(workflow_id, event)`; a re-run preserves its `event`, so the superseded-attempt collapse is unchanged |
| 4 | `CHECKS_UNREADABLE` sent the operator to grant `checks: read` | GitHub will not issue that on a fine-grained PAT, and reaching UNKNOWN now means the Actions read failed too — the string names the real causes |

Finding 3 is the one with teeth: a workflow declaring `on: [push, pull_request]` produces two runs for one sha and GitHub publishes check runs for both, so the old key dropped the lower `run_number` unread — and when the dropped run was the red one, the fallback answered **green for a commit whose real rollup is red**.

## Delivery notes

**Judgment calls**

- Branch cut from #91's approved head rather than `main`, per the finding set: three of four remedies edit code `main` does not have. The PR is still **based on `main`** so nothing is stranded, and the merge-order gate above is stated as a condition rather than implied.
- Finding 3 changes behaviour issue #90 specifies literally ("most recent run per `workflow_id`"). Deliberately not done in #91 — implementing the issue as written is what that PR is for — and done here, which is not the PR that implements #90. The reviewer proposed exactly this split.
- Finding 3 got a diff rather than a deferral task, against the reviewer's "follow-up task is a fine home for it". A fail-open hole in a fail-closed gate is worth the diff, and the fix is three lines plus tests.
- Finding 2 took both remedies the reviewer offered (body clause **and** the log level), not one: the log line serves the operator reading daemon output, the body clause serves the one reading the PR, and only the second is the merge cue.
- Version `0.19.0` → `0.20.0`. `0.19.0` is what #91 publishes on merge, so this needs the next number; `check-version-bump.yaml` requires a bump for any non-test change under the dist.

- [ ] No live GitHub call under a fine-grained PAT. The two-trigger case (finding 3) is exercised against faked `run_json` payloads, not a real sha carrying a `push` run and a `pull_request` run.
- [ ] The `event` field is assumed present and stable on `workflow_runs` entries, and assumed preserved across a re-run. Not probed live; a test covers the missing-`event` payload so its absence degrades to the old per-workflow grouping rather than fragmenting the rollup.
- [ ] The approve-body lead (finding 2) has not been seen on a real verdict — that needs the merge, the publish, the `REVLOOP_VERSION` re-pin and a redeploy, same as #91.
- [ ] Whether any repo in this fleet actually has a double-triggering workflow is unchecked. This repo has none (every workflow is `on: [pull_request]`; `package-publish.yaml` adds only `types: [closed]`), which is why the old key was unreachable here.

**Verification**

- `PYTHONPATH=src/main bash tests-unit.sh alissa-tools-github-revloop` — **799 passed** (794 on #91's head; 5 new tests).
- `bash check-style.sh` — clean. `bash check-types.sh` — clean, 30 files. `bash tests-coverage.sh` — 94% total.
- Counterfactual sweep: reverting the dedupe key, the approve-body lead, and the `CHECKS_UNREADABLE` text each fails exactly one test. All restored, suite green.
- One pre-existing test had to change: `test_an_unreadable_rollup_is_never_treated_as_green` asserted the literal string `checks: read`, which finding 4 removes. Its intent ("and how to fix it") is preserved — it now asserts the advice is actionable and that the impossible grant is *absent*.

**Scope touched**

- `src/main/.../ghclient.py` — the COST docstring, the dedupe key.
- `src/main/.../loop.py` — `CHECKS_FALLBACK_LEAD` (new), the `_gate_on_checks` GREEN branch, `CHECKS_UNREADABLE`.
- `src/main/.../version` — `0.19.0` → `0.20.0`.
- `src/test/.../test_loop.py` — 5 new tests, 1 pre-existing assertion updated.

No excursion beyond the finding set. Nothing under `docker/` touched; the `ARG REVLOOP_VERSION` re-pin remains the operator's.


alissa-app:
- 2026-08-21 04:34:19 UTC — `review-alissa-github-review-daemon-pr92-r1-5ed647` — round 1 of 22 — spawned
- 2026-08-21 04:35:29 UTC — `review-alissa-github-review-daemon-pr92-r1-6fd959` — round 1 of 22 — spawned
- 2026-08-21 04:38:51 UTC — `review-alissa-github-review-daemon-pr92-r1-4e4a60` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -2419,10 +2462,24 @@ def _gate_on_checks(
...
                 pr.slug, round_, judged[:8], rollup.summary,
             )
+            if rollup.via_actions_fallback:
+                # The one state where the marker had nowhere PR-visible to go.
+                # Every other path that surfaces `summary` (the hold activity
+                # comment, the degraded-verdict lead) is reachable only from
+                # PENDING or UNKNOWN, so "approved on green" -- the claim an
+                # operator acts on -- was the single verdict that did not say
+                # it came from the narrower read.
+                return ChecksGate(
```
alissa-app, line 2479: [minor] This is a fourth `ChecksGate` shape, and the class docstring still says there are three.

`ChecksGate`'s docstring (`loop.py:1350-1360`, unchanged by this PR) enumerates the shapes exhaustively and ends with:

> * neither -- green (or nothing to gate): the post proceeds unchanged.

That is now false for exactly the case this PR adds. This return is green, `event` is unchanged, and yet the post does **not** proceed unchanged — `body = gate.lead + NATIVE_VERDICT_BODY…` prepends the lead. The docstring is the only place a reader learns what the three fields mean together, so the one shape a future reader most needs (green + lead) is the one it denies exists.

Suggest a fourth bullet: `lead` set with `event` unchanged — post as an approve, but say which read produced the green.

Nothing about the runtime behaviour is wrong here; `state=CHECKS_GREEN` keeps `gate_note` empty and `event` at `APPROVE`, which the new test pins.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -407,6 +407,22 @@ def _post_delay_after(attempts: int) -> float:
 
 # Prepended when the gate held the approve for the whole wait bound and the
 # rollup still had not settled, so the verdict lands as a COMMENT.
+# Prepended to an APPROVE whose green came from the Actions-API fallback rather
+# than the check-runs rollup. Not a caveat on the review -- the code verdict is
+# unchanged -- but on the EVIDENCE behind the approve: the fallback cannot see a
+# check run posted by a third-party check app, so "green" here means "every
+# Actions-produced context passed", which is the same thing only on an
+# Actions-only repo.
+CHECKS_FALLBACK_LEAD = (
```
alissa-app, line 416: [minor] The comment above this constant belongs to `CHECKS_UNSETTLED_LEAD`, and inserting here orphaned it.

Lines 408-409 are pre-existing:

```
# Prepended when the gate held the approve for the whole wait bound and the
# rollup still had not settled, so the verdict lands as a COMMENT.
```

They documented `CHECKS_UNSETTLED_LEAD`, which now sits ten lines further down. As the file reads at `af0a52c8`, that block is the first half of the comment stack directly above `CHECKS_FALLBACK_LEAD` — and it is false about it: this lead rides on an **APPROVE**, not a COMMENT, and no wait bound is involved. Every other constant in this block carries exactly one comment block immediately above it, so the misattachment reads as intentional.

Fix is to move the new constant + its comment below `CHECKS_UNSETTLED_LEAD`, or move lines 408-409 back down to sit against their own constant.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -466,10 +482,18 @@ def _post_delay_after(attempts: int) -> float:
...
+# permission on fine-grained PATs at all, and a rollup that reaches UNKNOWN on
+# such a deployment has ALREADY tried the Actions read and failed -- so that
+# advice pointed at the one action that provably cannot work, at the moment the
+# operator is already blocked. The remaining causes are the ones named here.
 CHECKS_UNREADABLE = (
     "The rollup could not be read: `{why}`. An unreadable rollup is not a green "
-    "one — check that the reviewer credential carries `checks: read` on this "
-    "repo."
+    "one. On a credential that cannot read check runs the loop already fell "
+    "back to the Actions API, so reaching this point means that read failed "
```
alissa-app, line 494: [minor] "reaching this point means that read failed too" is not exhaustive — the same string is reached from failures that never touched the Actions API.

Removing the `checks: read` advice is right. But the sentence that replaced it makes a claim `check_rollup` does not support. Reading `ghclient.check_rollup`, `CHECKS_UNKNOWN` is reached from any of:

- `_check_runs` raising a **non-403** `CommandError` — re-raised past the fallback branch (`if not _is_authorization_forbidden(exc): raise`) into the outer `except Exception`. No Actions read is attempted.
- `TruncatedListing` from the **check-runs** listing — same, no fallback.
- `_commit_statuses` failing — and note this call runs *after* a **successful** fallback, so even on the fine-grained-PAT deployment the premise holds ("a credential that cannot read check runs") while the conclusion does not: the Actions read succeeded and the legacy-status read is what failed.

In each case the operator is told to check `Actions: Read` or the page bound, and neither is the cause. `{why}` carries the real error, so this is a wrong signpost rather than a dead end — but it is the same class of overclaim that finding 4 set out to remove, one step narrower.

Something like "…so on such a deployment reaching this point usually means the Actions read failed too — check `Actions: Read`, whether a listing was truncated (more contexts than the page bound), or whatever `{why}` names" would keep the correction without asserting the closed set.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -754,18 +869,147 @@ def _check_runs(self, owner: str, repo: str, sha: str) -> list[CheckContext]:
...
+
+        DISTINCT EVENTS DO NOT. A workflow declaring `on: [push, pull_request]`
+        produces a `push` run and a `pull_request` run for the same sha, and
+        GitHub publishes a check run per job for BOTH -- so the read this
+        stands in for reports both. Keying on `workflow_id` alone dropped the
+        lower `run_number` of the pair unread, and when the dropped one was the
+        red one the fallback answered GREEN for a commit whose real rollup is
+        red. That is the fail-open direction this whole component exists to
+        avoid, so the key carries `event` as well. A re-run preserves its run's
+        `event`, so the superseded-attempt collapse above is unaffected.
```
alissa-app, line 957: [minor] Adding `event` to the key silently drops the hand-re-dispatch case the previous docstring named as a must-collapse, and nothing states or tests the change.

The fix itself is right — the fail-open hole is real and this closes it. What went unremarked is the other half. At `0729b516` this docstring read:

> a re-run shares its run id, but a distinct trigger event (a `push` and a `pull_request` on the same commit, **a workflow re-dispatched by hand**) creates a distinct run. Reading all of them would judge the commit on a superseded attempt: an earlier failed run whose re-trigger passed would hold the head red forever.

That second case is now the behaviour. Concretely, on a workflow declaring `on: [pull_request, workflow_dispatch]`: the `pull_request` run fails, the operator re-dispatches by hand, the `workflow_dispatch` run passes — the keys are `(900, "pull_request")` and `(900, "workflow_dispatch")`, both kept, and the failed run holds the head **red until a new push**, which is exactly what the old paragraph said must not happen.

I think it is the right trade — it is the fail-closed direction, and it is *parity with the read this stands in for*, since the check runs of the failed `pull_request` run are still on the commit and the real rollup is red too. GitHub's "Re-run all jobs" keeps the run id and the `event`, so the common recovery path still collapses correctly. But:

1. This paragraph asserts "the superseded-attempt collapse above is unaffected", which is true for re-runs and not for re-dispatches — the sentence directly above it still mentions re-dispatch, so a reader is left with the wrong conclusion.
2. No test pins it. `test_a_retrigger_of_the_same_event_still_collapses_to_the_latest` covers same-event only; a cross-event re-dispatch case (assert RED, `total == 2`) would fix the behaviour in place and document the trade at the same time.

Worth one sentence saying the trade is deliberate and why parity makes it correct.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -754,18 +869,147 @@ def _check_runs(self, owner: str, repo: str, sha: str) -> list[CheckContext]:
...
+        exactly the rules it judges the real one under, including `skipped` and
+        `neutral` passing for a path-filtered matrix.
+
+        What this read CANNOT see is any check run posted by a third-party check
+        app; see ACTIONS_FALLBACK_NOTE for why that is accepted here and how the
+        answer says so.
+
+        COST, stated exactly, because it is the sentence an operator would size
+        rate-limit headroom from: this path costs one call for the run listing
+        plus one per KEPT workflow, where the read it replaces cost one.
```
alissa-app, line 904: [minor] "one per KEPT workflow" is the unit this same PR changed — the note is now off by the factor it was rewritten to be exact about.

This paragraph opens by promising to be exact "because it is the sentence an operator would size rate-limit headroom from", and `_actions_contexts` fans out one `jobs` call per element of `_latest_run_per_workflow(runs)`. After the key change in this PR that is one call per **(workflow, event)** pair, not per workflow — so on the very repo shape the fix is for (`on: [push, pull_request]`), the fan-out is up to **2× the workflow count** and this sentence says 1×.

Line 914 has the same issue: "the number of kept workflows is whatever the sha carries".

The `(one or two workflows per commit) it is two or three calls` sizing for this fleet stays right — every workflow here is `on: [pull_request]`, per the PR body — but the general statement an operator would extrapolate from does not. Suggest "one per KEPT RUN — one per (workflow, trigger event) pair, so a workflow that triggers on both `push` and `pull_request` costs two".

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -754,18 +869,147 @@ def _check_runs(self, owner: str, repo: str, sha: str) -> list[CheckContext]:
...
+            "workflow runs",
+            params=(f"head_sha={sha}",),
+        )
+        out: list[CheckContext] = []
+        for run_ in self._latest_run_per_workflow(runs):
+            out.extend(self._run_contexts(owner, repo, run_))
+        return out
+
+    @staticmethod
+    def _latest_run_per_workflow(runs: list[dict]) -> list[dict]:
```
alissa-app, line 937: [nit] The name is now narrower than what the function does — its own docstring's first line has to correct it ("One run per workflow AND trigger event"). `_latest_run_per_workflow_event`, or `_latest_run_per_trigger`, would let the summary line just restate the name. Implementer's discretion; the docstring does carry the truth either way.

alissa-app (approved):
## Review — round 1 (fresh reviewer, alissa-code-review CR1–CR9)

**Verdict: approve.** 0 blockers, 0 majors, 5 minors, 1 nit. Verdict of record: **TASK-30628631** (this PR's CR2 review task; created this round — none existed. Round cap for this loop is **22**, the daemon's own `ALISSA_ROUND_CAP`, not the skill default of 3).

### What I reviewed

The PR's diff against `main` carries #91's delta as well, because the branch is cut from #91's approved head `0729b516`. #91 was reviewed and approved on its own review task (TASK-1287865828), so this review's rubric pass is on the **round-2 delta `0729b516..af0a52c8`** — 203 insertions across `ghclient.py`, `loop.py`, the `version` file and `test_loop.py` — and reads #91's code only as the base the remedies land in. Each of the four `[minor]` findings from #91 round 1 is answered here, and each answer is the right one:

1. **COST note** — the corrected paragraph is *verified true*: `loop._dry_run_rollups` is the only rollup memo and is read at exactly one site (`loop.py:2315`, the dry-run branch), so nothing caches the production read; `_api`'s default timeout really is 60s; `CHECK_RUN_PAGE_LIMIT` really bounds only the run listing.
2. **Fallback labelling on GREEN** — correctly diagnosed: `summary` reaches PR-visible text only from the PENDING/UNKNOWN paths, so GREEN was the one approve that never said which read answered. `ChecksGate(lead=…, state=CHECKS_GREEN)` keeps `event` at `APPROVE` and leaves `gate_note` empty, so the verdict is not downgraded — only labelled.
3. **Dedupe key** — the fail-open hole is real and the fix is right. Taking the diff rather than a deferral task was the correct call: a fail-open hole in a fail-closed component is worth three lines.
4. **`CHECKS_UNREADABLE`** — `checks: read` is genuinely unobtainable on a fine-grained PAT, so removing it is right (see the one caveat below).

### Verification I ran myself

- `PYTHONPATH=src/main python3 -m pytest src/test -q` at `af0a52c8` → **799 passed**, matching the PR's claim exactly.
- **Counterfactual on finding 3**, reverting only the key to `("workflow", workflow_id)`: **1 failed, 798 passed** — `test_two_trigger_events_of_one_workflow_are_both_kept`, and nothing else. The new test genuinely pins the fix, and the `event` addition costs no existing behaviour.
- CI rollup at `af0a52c8` re-read immediately before submitting, through the **full check-runs endpoint** (this reviewer credential can read checks; no Actions fallback involved): **6/6 contexts completed, all `success`**, 0 legacy commit statuses. Head sha unchanged from the one reviewed.

### Rubric

- **Correctness** — the dedupe fix is sound and the gate change cannot downgrade an approve. Two documentation-vs-behaviour gaps found (`ChecksGate`'s docstring, the dropped re-dispatch case) — both `[minor]`, neither a defect in what the code *does*.
- **Coding practices** — matches the file's idiom exactly: reason-first comment blocks, `--` em-dashes in docstrings, the "why, and what it would cost to get it wrong" register. Branch `TASK-1977649551-FU-ROLLUP-NOTES`, PR names both tasks, co-author trailer present. Two comment-hygiene minors and one naming nit.
- **Security** — nothing found. No new input handling; `event` reaches only a dict key via `str(...)`. The approve-body lead discloses that the deployment's credential lacks `Checks` — a GitHub platform fact, not a secret, and no scope information beyond it.
- **Performance** — nothing found beyond the fan-out the PR itself documents. Worth noting the note now undercounts (see the COST comment below): the keyed unit changed in this same PR.
- **Side-effects / blast radius** — the merge-order gate the PR states is real and correctly stated; `#91` must land first. `CHECKS_UNREADABLE`'s literal text changed, which one existing test asserted on — updated in-PR, and I confirmed the replacement still asserts intent rather than prose. `0.19.0 → 0.20.0` is the correct next number given #91 publishes `0.19.0`; Version Bump Check is green. The `ARG REVLOOP_VERSION` re-pin is correctly left to the operator, same as #91. This daemon reviews this repo, so the change is self-affecting on deploy — the fail-closed direction of every change here is the right posture for that.

### Findings

All five minors are documentation/precision items in a component whose comments are its specification; none of them gates the merge. They can be fixed in place before merge or carried as one follow-up task, implementer's call — but per CR5 a deferred `[minor]` must become a registered task.

Merge order stands as the PR states it: **#91 first**, then this.

PR #98, opened 2026-08-28
TASK-2038401416: drop the non-canonical `todo` from the task-list status filter (issue #97)

### LINKED ISSUE #97: revloop: drop the non-canonical `todo` from the task-list status filter (it rejects the narrowed call outright)
## Context

`fahera-mx/studio.alissa.app` PR #830 adds `--status <csv>` and `--view <digest|full>` to `alissa task list` (CLI 0.2.0) so this daemon's `Alissa.probe_task_list` finds them and narrows its 30s poll — the whole point being to stop `apiActorOps.apiListMyTasks` shipping the operator's entire non-terminal corpus (621 rows / ~734 KB per billed execution, 378 MB read on 2026-08-27).

That PR's round-1 review found the narrowed call this daemon would assemble is **rejected by the new CLI**, so the adoption is a no-op — and worse than shipping `--view` alone.

At `origin/main`, `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py`:

```python
OPEN_STATUSES = {"committed", "in_progress", "pending_validation", "todo"}   # :17
TASK_LIST_STATUS_FILTER = ",".join(sorted(OPEN_STATUSES))                    # :35
```

`todo` is **not** one of Alissa's seven canonical task statuses (`draft`, `committed`, `in_progress`, `blocked`, `pending_validation`, `validated`, `cancelled` — `TaskStatusSchema`, `packages/client/src/schemas/common.ts`). CLI 0.2.0 validates `--status` client-side against exactly those seven and exits 1 on anything else, before any request.

So once the fleet's CLI advertises the flags:

```
argv  -> alissa task list --json \
           --status committed,in_progress,pending_validation,todo --view digest
stderr-> Unknown --status value: todo. Valid: draft, committed, ...
exit  -> 1
```

`list_tasks` catches the `CommandError`, sets `self._task_list_narrowing_disabled = True` and retries `["alissa","task","list","--json"]` plain **for the rest of the process**. That flag also gates `--view` and `--self`, so the digest projection is lost with the status filter. Net effect: zero read saving, one extra subprocess per daemon process.

The CLI side is not the defect and is not up for change — the issue it implements names the seven canonical statuses, and passing `todo` through would only move the rejection to the API.

## Scope

`alissa-tools-github-revloop` — the only tools package in this repo, and the only definition of `OPEN_STATUSES` at `origin/main`.

> **Correction (2026-08-28).** This section originally named an `alissa-tools-github-reviewloop` "twin" holding the same line at `alissa.py:16`, and the `Autonomous-Scope` trailer below named its path. There is no such twin. `674ca6d` (2026-07-23, *"rename reviewloop -> revloop repo-wide; first revloop release (0.12.0)"*) **renamed** that path to today's `revloop`; git reports it as a rename, and `alissa.py:16` at `674ca6d^` is byte-identical to today's `revloop/alissa.py:17`. It is this same file's pre-rename ancestor, not a sibling — which is why the line number looked like a second live copy. The rename collapsed the copies five weeks before this issue was filed, so there is no drift to guard against and nothing to reconcile. Raised as a `[major]` by round 2 of the review of fahera-mx/studio.alissa.app#830, which is where the wrong claim originated.

Either:

1. drop `todo` from `OPEN_STATUSES` (it looks dead — Alissa has no such status, so `is_open` can never match it), or
2. keep `OPEN_STATUSES` as-is and intersect `TASK_LIST_STATUS_FILTER` with the canonical set.

The comment at `alissa.py:29-34` argues the filter is deliberately derived from `OPEN_STATUSES` "so any status added to `is_open` is added here by construction". Whichever option is taken, that invariant needs to stay true or be restated — option 2 breaks it silently unless the intersection is itself pinned.

## Out of scope

- The studio CLI (`fahera-mx/studio.alissa.app`): PR #830 ships the flags as specified and keeps rejecting `todo`.
- Any other narrowing behaviour: `--self` opt-in, the negative cache and the probe (issue #87) all stay as they are.

## Definition of Done

The narrowed `alissa task list` argv this daemon builds against a CLI 0.2.0 help is accepted by that CLI (no `CommandError`, `_task_list_narrowing_disabled` stays `False`), `TASK_LIST_STATUS_FILTER` carries only canonical Alissa statuses, and a test pins the filter against the canonical set rather than only against `OPEN_STATUSES` — so a non-canonical value cannot re-enter through `is_open`.

## Acceptance criteria

- [ ] `TASK_LIST_STATUS_FILTER` contains only canonical Alissa statuses; pinned by a test that fails if a non-canonical value re-enters via `OPEN_STATUSES`.
- [ ] `task_list_argv` against a help text advertising `--status`/`--view` produces an argv CLI 0.2.0 accepts; covered by a test with the real flag values, not a stub filter.
- [ ] `OPEN_STATUSES` still has exactly one definition in this repo, and the change lands in it — no second copy is introduced alongside it. *(Corrected 2026-08-28; see the Scope note. This criterion originally asked for a `reviewloop` copy to be reconciled or excused. That path was renamed away by `674ca6d`, so the original was unsatisfiable as written.)*
- [ ] Manual, operator: after the base-image CLI refresh, a live poll issues the narrowed request instead of falling back to the plain whole-corpus call.

Alissa-Task: TASK-1593278191

Raised by the round-1 review of fahera-mx/studio.alissa.app#830 (blocker on `cli/src/commands/task.ts`).

RHDZMOTA:
Closes #97

**Origin task:** TASK-1593278191
**Implementation task:** TASK-2038401416 (downstream of the origin, soft-linked)

## What was wrong

`TASK_LIST_STATUS_FILTER` is derived from `OPEN_STATUSES`, and `OPEN_STATUSES`
carried `todo` — a status Alissa has never had. Because `is_open` could never
match it, it read as inert. It was not: CLI 0.2.0 (studio PR #830) validates
`--status` client-side against the seven canonical statuses of `TaskStatusSchema`
and exits 1 on anything else *before* issuing a request. So the narrowed argv
this daemon assembles would have been refused outright, `list_tasks` would have
caught the `CommandError`, set `_task_list_narrowing_disabled` and retried the
plain whole-corpus call for the rest of the process — dropping `--view digest`
and `--self` along with the status filter. Zero read saving, one extra
subprocess per daemon process.

## What this branch does

- `CANONICAL_TASK_STATUSES` — the seven canonical statuses, mirrored the way the
  studio CLI mirrors them into `TASK_STATUSES`; the vocabulary the filter is
  allowed to speak.
- `todo` is out of `OPEN_STATUSES`, which is now entirely canonical (the issue's
  **option 1**).
- `_status_filter()` keeps the derivation whole — the filter is still
  `OPEN_STATUSES` itself, so the "any status added to `is_open` is added here by
  construction" invariant survives — and answers `""` (send no `--status` at
  all), loudly, for any set carrying a non-canonical value.

  Deliberately **not** the issue's option 2. An intersection with the canonical
  set makes the filter a strict *subset* of the open set the moment the two
  disagree, so a review task in the dropped status never crosses the wire and
  its absence reads as "this PR has no review task" — a skipped review. Wide but
  complete is the direction every other failure on this path degrades in.
- The test fixture for a CLI advertising both flags was invented. It is now the
  **real** 0.2.0 help: rendered by the commander bundled in the installed
  `alissa`, with PR #830's own `.option()` calls spliced into `task list`, and
  captured byte-for-byte. `cli_0_2_0_task_list()` transcribes #830's
  `parseStatusFilter` / `parseListView`, so the suite asserts the assembled argv
  is one the CLI **accepts**, with the pre-0.22.0 argv as the counterfactual
  (refused with `Unknown --status value: todo`).
- Version 0.21.0 → 0.22.0; README's narrowing section restated.

## Acceptance criteria

| criterion | where |
| --- | --- |
| filter holds only canonical statuses, pinned against the canonical set | `test_the_open_set_and_its_filter_hold_only_canonical_alissa_statuses`, `test_a_non_canonical_status_added_to_the_open_set_disables_the_filter` |
| `task_list_argv` against a `--status`/`--view` help produces an argv 0.2.0 accepts, real flag values | `test_the_narrowed_call_is_one_cli_0_2_0_accepts` (+ `test_the_narrowed_call_this_daemon_used_to_build_is_rejected`) |
| `reviewloop` twin reconciled or left alone with a reason | left alone — it no longer exists in this repo (see below) |
| operator: a live poll issues the narrowed request | operator's gate, see the delivery note |

The `alissa-tools-github-reviewloop` twin the issue names is **gone from this
repo**: the rename (PR #33) removed the directory and the dist is frozen on PyPI
at 0.11.0, receiving no further releases. Its `OPEN_STATUSES` still holds `todo`
there and is harmless — that copy has no status filter and never sends
`--status`. There is nothing to keep from diverging.

### Judgment calls

- **Took the issue's option 1 (drop `todo`) and explicitly rejected option 2
  (intersect).** An intersection can only ever make the wire filter *narrower*
  than `is_open`, which silently skips reviews; a non-canonical value now
  disables the status filter instead, which is merely wide. Reasoned in the
  comment above `TASK_LIST_STATUS_FLAG`.
- **Added a runtime guard, not only the mandated test.** The DoD asks that a
  non-canonical value "cannot re-enter through `is_open`"; a test enforces that
  in CI, `_status_filter` enforces it in a running daemon. It is ~10 lines and
  its failure mode is the one the daemon already handles.
- **Replaced the invented `HELP_FULL` fixture with the real 0.2.0 help** rather
  than adding a second fixture beside it. Every argv assertion in the file now
  runs against output the fleet will actually emit; an invented column layout is
  how a text-matching probe stays green against a shape it never sees.
- **Also fixed one unrelated stale fixture**: `test_prreview.py` used
  `status="todo"` on a task whose title is what makes the assertion (still
  passes either way). Left in the tree it would have been the last live use of a
  status that does not exist.
- **The `--digest`/`--view` divergence is left alone.** Today's shipped 0.1.0
  advertises a boolean `--digest` — the same projection under a different
  spelling — and the probe looks for `--view`, so the digest view is not adopted
  today even where it exists. That is new narrowing behaviour, which the issue
  puts out of scope; noted in the README instead of changed here.
- **No Dockerfile re-pin.** `ARG REVLOOP_VERSION` is 0.16.14 and this ships
  0.22.0, but `docker/**` is outside the issue's declared `Autonomous-Scope` and
  the pin is its own lane.

- [ ] **A live poll of the real daemon issuing the narrowed call.** No CLI 0.2.0
  exists on this runner (the installed one is 0.1.0, and 0.2.0 is unmerged in
  the studio repo), so the daemon's own `probe_task_list` cannot see `--status`
  here. This is the issue's own manual criterion.
- [ ] **The published CLI 0.2.0.** Everything about 0.2.0 in this branch —
  the help text, the validation — comes from PR #830's diff as of 2026-08-28,
  transcribed and rendered locally. If that PR's flags or messages change before
  it merges, `HELP_0_2_0` and `cli_0_2_0_task_list` are what needs re-checking.
- [ ] **The base-image refresh.** The fix reaches the fleet only once the image
  installs revloop ≥ 0.22.0 (`ARG REVLOOP_VERSION`, currently 0.16.14) *and*
  ships a CLI that advertises `--status`. Neither is in this diff.
- [ ] **`bash tests-coverage.sh` reports "No data to report" on this runner.**
  Pre-existing local-environment artifact, not a regression: `tests-unit.sh`
  runs `coverage run` with no `PYTHONPATH`, so the installed copy is measured as
  third-party and only `src/test` files land in the data, which the script's own
  `--omit=test_*` then strips. CI installs the package editable, where it works.
  Coverage was measured directly instead (below).

### Verification

- `bash tests-unit.sh alissa-tools-github-revloop` — **820 passed** (25 in
  `test_task_list.py`, 5 of them new).
- `bash check-style.sh alissa-tools-github-revloop` — clean (pycodestyle, 120).
- `bash check-types.sh alissa-tools-github-revloop` — **Success: no issues found
  in 30 source files** (mypy).
- `bash tests-coverage.sh alissa-tools-github-revloop` — "No data to report", see
  the gate above. Run directly with the worktree source on `PYTHONPATH`:
  **94% total**, `alissa.py` 88%, both new branches of `_status_filter` covered.
- **Live end-to-end, against the real API** — the installed `alissa` bundle with
  #830's option definitions *and* its validation injected, driven with the argv
  this branch builds:

  | argv | result |
  | --- | --- |
  | `--status committed,in_progress,pending_validation --view digest` | exit 0, 1285 rows, 349,586 bytes, digest row shape, only the three open statuses |
  | `--status committed,in_progress,pending_validation,todo --view digest` (pre-0.22.0) | exit 1, `Unknown --status value: todo`, no request issued |
  | no narrowing (today's call) | exit 0, 1326 rows, 1,630,642 bytes |

  79% of the payload, on this actor's corpus. That is the saving issue #87 built
  the probe for, and issue #97 is what stood in front of it.
- **Counterfactual**: reverting only the filter value (`monkeypatch` to the
  0.21.0 string) makes `test_the_narrowed_call_this_daemon_used_to_build_is_rejected`
  the failing side of the same harness — the acceptance test cannot pass
  vacuously.

### Scope touched

Declared scope: `alissa-tools-github-revloop/**`, `alissa-tools-github-reviewloop/**`.

- `alissa-tools-github-revloop/src/main/.../alissa.py` — the canonical set, the
  open set, `_status_filter`, one guard in `task_list_argv`.
- `alissa-tools-github-revloop/src/main/.../version` — 0.21.0 → 0.22.0.
- `alissa-tools-github-revloop/src/test/.../test_task_list.py` — real 0.2.0 help
  fixture, the CLI acceptance harness, 5 new tests.
- `alissa-tools-github-revloop/src/test/.../test_prreview.py` — one fixture
  status, `todo` → `committed`.
- `alissa-tools-github-reviewloop/**` — nothing: the directory does not exist at
  `main`.
- **Excursion:** `README.md` at the repo root, outside the declared scope. Its
  "Bounding the task-list read" section documents this exact filter and would
  otherwise have been left stale; two bullets, no other section touched.


alissa-app:
- 2026-08-28 03:26:57 UTC — `review-alissa-github-review-daemon-pr98-r1-5ef974` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -26,13 +54,48 @@
...
+            "not narrowing `alissa task list` by status: %s not canonical "
+            "Alissa status(es) -- the CLI rejects the whole call over one "
+            "unknown value, so this daemon lists the corpus unnarrowed",
+            ", ".join(unknown),
+        )
+        return ""
+    return ",".join(sorted(statuses))
+
+
+TASK_LIST_STATUS_FILTER = _status_filter(OPEN_STATUSES)
```
alissa-app, line 98: [minor] The guard's one diagnostic is emitted at **import** time, which is the one moment the daemon's logging is not configured yet.

`__main__.py` imports `.loop` (line 22), which pulls this module in and evaluates `_status_filter(OPEN_STATUSES)` here; `logging.basicConfig(...)` does not run until line 241. So the `log.warning` in `_status_filter` fires with no handler on the root logger and falls through to `logging.lastResort` — bare message on stderr, no timestamp, no logger name, and outside whatever stream/handler the deployment configured. `_status_filter`'s own docstring makes loudness the point ("Loud rather than silent ... this is the last line rather than the first"), and this is precisely where it is quietest.

It is not a correctness problem — the message still reaches stderr, and CI's canonical-set test catches the condition long before a daemon runs — so it is fine to leave for a follow-up. If you do want it: computing the filter lazily inside `task_list_argv` (memoized) would put the warning in the configured stream, or the daemon could re-assert the condition once at start-up next to its other boot-time logging.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_task_list.py
```diff
@@ -240,18 +365,112 @@ def test_the_status_filter_is_the_daemons_own_open_set(cli):
...
+    filter entirely is wide but complete.
+    """
+    with caplog.at_level(logging.WARNING):
+        assert _status_filter(OPEN_STATUSES | {"todo"}) == ""
+
+    assert "todo" in caplog.text, "and it says which value cost the narrowing"
+    assert _status_filter(OPEN_STATUSES) == TASK_LIST_STATUS_FILTER
+
+
+def test_the_narrowed_call_is_one_cli_0_2_0_accepts(cli):
```
alissa-app, line 415: [minor] The acceptance harness never sees the argv a `task_list_self_scope=True` deployment builds.

`cli_0_2_0_task_list` is only ever fed argvs from a default `Alissa()`, whose `_task_list_self_scope` is False — so `--self` is in the fixture help and in `TaskListFlags`, but never in an argv that is put through the CLI's parse. The one test that does turn the knob on (`test_narrow_status_false_drops_only_the_status_filter`) asserts on the argv only. The result is that the full three-flag call — `--status ... --self --view digest` — is the one shape a real deployment can emit that no test asserts CLI 0.2.0 accepts, and this file's stated purpose (per its module docstring) is exactly "the argv is one the CLI ACCEPTS".

The risk is genuinely low: `--self` is a plain boolean in both the real option table and the transcription. But closing it is one line — `assert cli_0_2_0_task_list(argv) == {...}` added to the `self_scope=True` case — and it makes the coverage claim true rather than nearly true.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -13,8 +13,36 @@
...
+# The seven canonical Alissa task statuses -- `TaskStatusSchema` in the studio
+# repo (`packages/client/src/schemas/common.ts`). Mirrored here for the reason
+# the studio CLI mirrors it into `TASK_STATUSES` rather than importing it: what
+# this module needs is the SET of names, not the schema that validates them.
+#
+# It is the vocabulary the `--status` filter below is allowed to speak. CLI
+# 0.2.0 validates `--status` against exactly these seven client-side and exits 1
+# on anything else BEFORE issuing any request, so a single non-canonical value
+# does not narrow the call slightly less -- it costs the whole narrowed call.
+CANONICAL_TASK_STATUSES = frozenset(
```
alissa-app, line 25: [nit] The comment explains why this set is mirrored rather than imported, but not what happens when it goes stale, which is the reader's next question.

If Alissa ever gains an eighth status and someone legitimately adds it to `OPEN_STATUSES` without touching this frozenset, `_status_filter` reads it as non-canonical and silently drops the narrowing for the whole fleet — a performance regression that no test fails on (the canonical-set test would still pass: `OPEN_STATUSES <= CANONICAL` is what breaks, so it *would* fail — but only because the new status is absent here, which is a confusing signal pointing at the wrong constant). The failure direction is the safe one, which is why this is a nit and not a finding about behaviour. One sentence naming that — "a status Alissa adds must land here first" — would aim a future editor at the right line.

README.md
```diff
@@ -685,14 +685,24 @@ What that trades is latency, not correctness:
...
-becomes a skipped review, not a slower one. Today's CLI (0.1.0) offers only
-`--self`, so the call the daemon makes is the same one it always made:
+becomes a skipped review, not a slower one. Of the three narrowings below,
+today's CLI (0.1.0) offers only `--self`, so the call the daemon makes is the
+same one it always made:
 
 - **status filter** — sent as exactly the daemon's own open-status set, so it
   cannot change which task resolves; adopted automatically when the CLI grows a
-  `--status`;
+  `--status`. That set holds only *canonical* Alissa statuses (`draft`,
```
alissa-app, line 694: [nit] `That set holds only *canonical* Alissa statuses (draft, committed, in_progress, blocked, pending_validation, validated, cancelled)` reads on a first pass as if the open set holds all seven, when the parenthetical is enumerating the canonical vocabulary rather than the set's members. `holds only canonical Alissa statuses — i.e. names drawn from draft, committed, ...` (or moving the list into the following clause about what the CLI validates against) removes the ambiguity.

alissa-app (approved):

Review task: **TASK-1916071167** (created for this PR per CR2, downstream of origin TASK-1593278191). Reviewed at head `9b25e9e`.

The fix is right, the diagnosis behind it is right, and — unusually — most of the PR's own claims are independently checkable from this runner, so I checked them rather than taking them:

- **`todo` really is dead.** Listed this actor's live corpus (1,328 non-terminal rows): statuses present are `pending_validation` (897), `committed` (356), `draft` (40), `in_progress` (35), `blocked` (1). No `todo`. Dropping it from `OPEN_STATUSES` cannot narrow `is_open` against anything Alissa actually issues, so the removal has no behavioural blast radius — the concern that would have made this a blocker.
- **`HELP_0_2_0` is the real help.** Ran `alissa task list --help` against the installed CLI (0.1.0) here: the `--self` / `--include-shared` / `--digest` / `--project` / `--bow` rows and their exact wrap columns are byte-identical to the fixture. The claim that this is the shipped help with #830's two `.option()` rows spliced in holds up, and it is a materially better fixture than the invented `HELP_FULL` it replaces — the probe is a column-anchored text matcher, so column widths are load-bearing.
- **`--digest` in 0.1.0** — confirmed present in that same help, so the README's note about the `--digest`/`--view` spelling divergence is accurate and leaving it out of scope is the right call.
- **The guard behaves as documented.** Imported the module at this head: `TASK_LIST_STATUS_FILTER == "committed,in_progress,pending_validation"`, `_status_filter(OPEN_STATUSES | {"todo"}) == ""` with the warning naming `todo`, `_status_filter(set()) == ""`, `Task(status="todo").is_open is False`.
- **The acceptance tests are not vacuous.** Reverting only the filter breaks `test_the_open_set_and_its_filter_hold_only_canonical_alissa_statuses` (`OPEN_STATUSES <= CANONICAL_TASK_STATUSES` fails) *and* `test_the_narrowed_call_is_one_cli_0_2_0_accepts` (the argv collapses to no `--status`), independently of the counterfactual test that exists for exactly this purpose.
- **The twin is genuinely gone** — no `alissa-tools-github-reviewloop/` at this head, so "left alone with a reason" satisfies that DoD line.

The design call — drop the filter entirely on a non-canonical value rather than intersecting with the canonical set — is the right one and for the right reason: an intersection degrades *narrower* than `is_open`, and a review task that never crosses the wire is read as "this PR has no review task", i.e. a skipped review. Degrading wide-but-complete matches how every other failure on this path already degrades (`list_tasks`' failed-narrowing retry, its empty-answer cross-check). The `if ... and TASK_LIST_STATUS_FILTER` guard in `task_list_argv` correctly stops `--status ""` — which would have been the rejected call with an extra step.

### CI

Rollup re-read at `9b25e9e` immediately before submitting (full check-runs API, not the Actions fallback): **7/7 contexts completed, all `success`** — Python Style, Python Types, Unit-Test, Wheel Package, Version Bump, Image contract, Entrypoint config renderer. No legacy commit statuses on the sha. Note that `pytest`/`coverage` are not installed on this reviewer runner, so the suite was not re-run locally; CI's green Unit-Test Check at this sha is the evidence, plus the direct module-level checks above.

### Findings

2 `[minor]`, 2 `[nit]`, 0 blockers, 0 majors, 0 questions — all inline. None gate the merge; neither minor needs to be fixed in this loop, though both are cheap.

One further nit that has no diff line to hang on: `prreview.py`'s `_review_task_ref` docstring still says "the daemon's server-side status filter is exactly its `is_open` set". After this change that is true *or* the filter is absent entirely. The conclusion it draws (`narrow_status=False` here) is unaffected either way, so this is wording only.

### Dimensions

- **Correctness** — satisfies the origin DoD and criteria c1/c2; c3 is the operator's live-poll gate and is correctly left open. Edge cases behave: empty set and non-canonical set both answer `""`, and `task_list_argv` drops the flag rather than sending an empty one. Nothing found.
- **Coding practices** — matches this module's idiom exactly (dense why-comments, ASCII `--`, an explanation of the *rejected* alternative next to the chosen one). Style and mypy green in CI. Branch, PR title, task refs and co-author trailer all per convention. The `test_prreview.py` fixture touch is a one-word drive-by but is disclosed and is the removal of the last live `todo` in the tree.
- **Security** — nothing found. The filter is a constant joined from a hardcoded literal set, passed as a discrete argv element to list-form `run`/`run_json` (no shell); no new inputs, secrets, dependencies, or logged data beyond a status name.
- **Performance** — this *is* the performance fix, and it is the whole saving rather than a fraction of it: without it the first narrowed call fails and `_task_list_narrowing_disabled` pins the process to the whole-corpus list, losing `--view digest` and `--self` too. Nothing found.
- **Side-effects / blast radius** — `OPEN_STATUSES` reaches `Task.is_open` → `is_review_task_for` → `find_review_task`; the removed member is unreachable on the live corpus (measured above), so no consumer changes. `TASK_LIST_STATUS_FILTER` has exactly one production consumer and it is guarded. Nothing found. For context, not a finding: the fix reaches the fleet only once `ARG REVLOOP_VERSION` (0.16.14, already five minors behind `main`) is re-pinned to ≥ 0.22.0 *and* the image ships a CLI advertising `--status` — both correctly out of this PR's scope, and both already named in the delivery note.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: [triage:pursue]

Right, and the docstring makes it worse by claiming loudness at the one call site that cannot deliver it: `__main__.py` imports `.loop` → this module at line 22 and does not `basicConfig` until 241, so the only warning the guard emits lands on `logging.lastResort`.

Taking the lazy-and-configured option, shaped to your second suggestion rather than the memoized-property one, because it also makes the message *conditional on the loss*:

- `_status_filter` becomes pure (no logging), and the non-canonical members are captured next to it as `NON_CANONICAL_OPEN_STATUSES` so the diagnostic can still name them;
- `task_list_argv` warns once per client, and only when `flags.status` is true — i.e. only when a CLI that *would* have served the filter is being handed a call without one. At import, nothing yet knows whether the narrowing was even available, so the import-time message was both unconfigured and premature.

That keeps `TASK_LIST_STATUS_FILTER` a module constant (the tests monkeypatch it, and `_status_filter` stays directly callable), and puts the message in the deployment's configured stream.

Cannot go on this branch — the verdict is an approve, so its head is terminal. It goes in a follow-up PR carrying `Follow-up-To: #98`, cut from this approved head `9b25e9e` because `_status_filter` does not exist on `main`, based on `main`, gated on merging this PR first. URL in a top-level comment once it is cut.

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_task_list.py
RHDZMOTA: [triage:pursue]

Fair — the file's docstring promises "the argv is one the CLI ACCEPTS", and the one shape a real deployment can emit that no test puts through the parse is the three-flag call. `--self` being a plain boolean on both sides is why this is a gap in the claim rather than a bug, but the claim is the point of the file.

`test_narrow_status_false_drops_only_the_status_filter` is the wrong home for it (it exists to show the *absence* of `--status`), so the follow-up adds the assertion where the knob is on and every narrowing applies — `--status ... --self --view digest` through `cli_0_2_0_task_list`, expecting `{"scope": "self", "view": "digest", "status": [...]}`, plus the `narrow_status=False` variant so the opt-out shape is covered too.

Same follow-up PR as the guard finding above; not this branch, which is terminal on the approve.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: [triage:pursue]

Agreed, and your parenthetical is the part worth writing down: the canonical-set test *does* fail if an eighth status is added to `OPEN_STATUSES` alone, but it fails pointing at `OPEN_STATUSES`, when the edit that is actually missing is here. A one-sentence pointer ("a status Alissa adds lands here first") aims the next editor at the right constant.

Adding it to the `CANONICAL_TASK_STATUSES` comment in the follow-up, along with the failure direction — narrowing silently dropped fleet-wide, never a wrong task resolved — since that is what makes it a nit rather than a blocker and a future reader deserves to know which way it breaks.

README.md
RHDZMOTA: [triage:pursue]

You are right that it reads as an enumeration of the set's members on first pass — the parenthetical sits immediately after "That set holds only", so the seven names attach to the wrong noun.

Taking the second of your two options: the list moves into the clause about what the CLI validates against, where it is unambiguously the vocabulary rather than the contents. The bullet then reads "holds only canonical Alissa statuses, because the CLI validates `--status` against exactly the seven (`draft`, ...) and refuses the whole call over one unknown value".

Same follow-up PR.

RHDZMOTA:
[triage:pursue] — the fifth finding, the one with no diff line to hang on.

`prreview._review_task_ref`'s docstring says "the daemon's server-side status filter is exactly its `is_open` set". After this change that is true *or* the filter is absent entirely, and the `narrow_status=False` conclusion it draws holds either way — wording only, as you said. It is picked up in the same follow-up as the other four, since it is one sentence in a file that branch is not otherwise touching.

Round-1 triage, for the record: **4 pursued inline** (the import-time diagnostic, the `--self` argv gap in the acceptance harness, the stale-canonical-set pointer, the README parenthetical) **+ 1 pursued here**. Nothing ignored, nothing deferred, so no deferral tasks to register.

None of them lands on this branch: the verdict is an approve and its head `9b25e9e` is terminal, so a push here would race the merge. They go to a follow-up PR carrying `Follow-up-To: #98`.

**On that follow-up's base, stated up front because it deviates from the usual "cut off `main`":** every one of the five findings is in code or prose this PR *introduces* — `_status_filter`, `CANONICAL_TASK_STATUSES`, `cli_0_2_0_task_list`, the two README bullets — none of which exist on `main`. A branch cut from `main` could only re-introduce them, i.e. duplicate this diff. So it is cut from this approved head, **based on `main`**, and merging this PR is an explicit gate on it — the same shape settled on PR #48/#53 and #89 in this repo, including the `blocked:merge-order` label rather than a re-draft. The approved head does not move.

PR #99, opened 2026-08-28
TASK-2038401416: PR #98 round-1 findings — diagnostic placement, harness gap, two wordings

RHDZMOTA:
Follow-up to **PR #98, review round 1** (verdict: approve, 2 `[minor]` + 3 `[nit]`, none blocking). Carries all five, triaged `[pursue]` on their threads before this branch existed.

Follow-up-To: #98

**Origin task:** TASK-1593278191
**Implementation task:** TASK-2038401416
No issue reference: #98 is the PR that answers issue #97, and this one answers #98's review.

## ⚠️ Merge order: merge #98 first

**This PR must not be merged before #98.** Squash-merging it first would fold #98's entire delta into `main` under this PR's message, leaving #98 to resolve as empty — losing its history and its issue-closing claim. The `blocked:merge-order` label is on this PR as the mechanical guard; drop it once #98 is in.

**On the base**, stated plainly because it deviates from the usual "cut a follow-up off `main`": every one of the five findings is in code or prose that #98 *introduces* — `_status_filter`, `CANONICAL_TASK_STATUSES`, `cli_0_2_0_task_list`, two README bullets. A branch cut from `main` could only re-introduce them, i.e. carry #98's diff a second time. So this is cut from #98's **approved head `9b25e9e`** and based on `main`; #98's head does not move, and the review surface here is `9b25e9e..HEAD` — one commit. GitHub's diff-against-`main` will show #98's changes too until it merges. Same shape as PRs #48, #53 and #89 in this repo.

## The five findings

1. **`[minor]` the diagnostic was emitted at import time.** `__main__` imports `.loop` → `alissa` at line 22 and calls `logging.basicConfig` at line 241, so the guard's one `log.warning` fell through to `logging.lastResort`: bare on stderr, no timestamp or logger name, outside whatever handler the deployment configured — at the exact point the docstring claimed loudness.

   `_status_filter` is now pure; the non-canonical members are kept beside it as `NON_CANONICAL_OPEN_STATUSES`; `Alissa._warn_status_filter_dropped` emits the message once per client from `task_list_argv`. That site is also the only one where the message is *true*: at import nothing yet knows whether the CLI serves `--status`, and on a CLI that does not, a dropped filter costs nothing. It now fires exactly when the flag was on offer and could not be used.

2. **`[minor]` the acceptance harness never saw a `--self` argv.** `cli_0_2_0_task_list` was only ever fed a default `Alissa()`, so the full three-flag call a `task_list_self_scope` deployment emits was asserted on but never put through the CLI's parse. Now it is, in both the narrowed and the `narrow_status=False` form.

3. **`[nit]` the mirrored canonical set says why, not what-if-stale.** A status added to `OPEN_STATUSES` alone drops the narrowing fleet-wide, and the canonical test that fails points at the constant that is *right*. Named in the comment, with the failure direction.

4. **`[nit]` README parenthetical** read as if the open set held all seven statuses. The list moves into the clause about what the CLI validates against.

5. **`[nit]`, raised in the review body with no diff line to hang on** — `prreview._review_task_ref`'s docstring said the server-side filter "is exactly its `is_open` set"; since 0.22.0 it is that *or* absent. Wording only; the `narrow_status=False` conclusion holds either way.

Version 0.22.0 → 0.23.0, so each merge is its own release.

### Judgment calls

- **Cut from #98's approved head, not from `main`** — reasoned above. The rule this bends exists to stop the approved head moving and to stop a PR being *based* on a task branch; neither happens here. The alternative was deferring all five to a registered task, which for four one-to-five-line edits would cost more than it protects.
- **Warn from `task_list_argv`, not from a memoized property.** The review offered both. The call site keeps `TASK_LIST_STATUS_FILTER` a plain module constant (the tests monkeypatch it, and `_status_filter` stays directly callable), and makes the diagnostic conditional on `flags.status` — i.e. on the narrowing actually being lost.
- **Once per `Alissa` client, not once per process.** The daemon builds one client; a module-level "already warned" flag would go stale across tests and hide the message from a second deployment in one process.
- **Bumped to 0.23.0** rather than reusing #98's unpublished 0.22.0, so the two merges are two releases. PyPI's latest is 0.21.0.

- [ ] **Merge order.** Nothing in CI enforces "#98 first"; the label is a marker, not a branch-protection rule.
- [ ] **The live-poll criterion** from issue #97 is still open and still the operator's, unchanged by this PR: it needs a base image carrying revloop ≥ 0.23.0 *and* a CLI advertising `--status`.
- [ ] **CLI 0.2.0 remains unpublished** (studio PR #830 open as of 2026-08-28); the acceptance harness and `HELP_0_2_0` are transcribed from its diff.
- [ ] `bash tests-coverage.sh` still reports "No data to report" on this runner — the same pre-existing install-shadowing artifact described on #98, not a regression.

### Verification

- `bash tests-unit.sh alissa-tools-github-revloop` — **823 passed** (3 new).
- `bash check-style.sh` — clean. `bash check-types.sh` — **no issues in 30 source files**.
- **The finding itself, directly**: importing the module writes nothing to stderr; with a non-canonical member spliced into `OPEN_STATUSES` and logging configured the way `__main__` configures it, the warning arrives fully formatted — `2026-08-28 03:40:02,993 WARNING alissa.tools.github.revloop.alissa: ... (non-canonical status(es) in OPEN_STATUSES: todo) ...` — and is emitted **once** across two `task_list_argv` calls. Both are what the old placement could not do.
- Pinned by tests as well as observed: one test asserts the warning is emitted once per client from the argv call, another that a CLI without `--status` produces no such warning.

### Scope touched

- `alissa-tools-github-revloop/src/main/.../alissa.py` — `_status_filter` made pure, `NON_CANONICAL_OPEN_STATUSES`, `_warn_status_filter_dropped`, the canonical-set comment.
- `alissa-tools-github-revloop/src/main/.../prreview.py` — one docstring sentence (finding 5).
- `alissa-tools-github-revloop/src/main/.../version` — 0.22.0 → 0.23.0.
- `alissa-tools-github-revloop/src/test/.../test_task_list.py` — 3 new tests, one existing test loses its `caplog` arm to the new pair.
- `README.md` — one bullet reworded (finding 4). Same documented excursion as #98: the root README is outside the issue's declared scope but documents this filter.


RHDZMOTA:
- 2026-08-28 04:00:27 UTC — `orphan-fahera-mx-alissa-github-review-daemon-pr99` — orphan sweep — handed off — marked ready and requested review from alissa-app; the session that opened this draft died before handing it off (head 57d66e1c public 20 min), so the daemon completed its protocol (no session spawned, once per PR/head)

alissa-app:
- 2026-08-28 04:08:54 UTC — `review-alissa-github-review-daemon-pr99-r1-a1e31e` — round 1 of 22 — spawned
- 2026-08-28 04:15:57 UTC — `review-alissa-github-review-daemon-pr99-r1-ddbea2` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_task_list.py
```diff
@@ -240,18 +365,170 @@ def test_the_status_filter_is_the_daemons_own_open_set(cli):
...
+
+    The answer is deliberately NOT to intersect the filter with the canonical
+    set: that would keep narrowing on a filter narrower than `is_open`, so a
+    review task in the dropped status would never cross the wire and the daemon
+    would read its absence as "this PR has no review task". Dropping the status
+    filter entirely is wide but complete.
+    """
+    assert _status_filter(OPEN_STATUSES | {"todo"}) == ""
+    assert _status_filter(set()) == "", "and an empty set is not a filter either"
+    assert _status_filter(OPEN_STATUSES) == TASK_LIST_STATUS_FILTER
```
alissa-app, line 410: [minor] The property this PR exists to establish — that `_status_filter` is **silent** — is now pinned by nothing.

This test's `caplog` arm was removed here (correctly: it asserted the old behaviour), and the two new tests pin where the diagnostic *is* emitted (`task_list_argv`) rather than where it must *not* be. Neither touches `_status_filter`. So if a `log.warning` is reintroduced into `_status_filter` tomorrow, the whole suite still passes: `test_the_dropped_filter_is_reported_once_from_the_call_that_loses_it` never calls `_status_filter`, and its `count("todo") == 1` is measured on a different code path.

That is the exact regression the commit message says was verified by hand ("importing the module writes nothing to stderr") — hand-verification is what a test is for. Import-time silence is genuinely awkward to assert under pytest, but the *mechanism* is trivially testable right here, one line below:

```python
with caplog.at_level(logging.WARNING):
    assert _status_filter(OPEN_STATUSES | {"todo"}) == ""
assert caplog.text == "", "pure: the diagnostic belongs to _warn_status_filter_dropped"
```

Same `caplog` fixture the file already uses, and it fails the moment purity is lost.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -365,6 +437,38 @@ def probe_task_list(self) -> "TaskListFlags":
...
+        self._status_filter_warned = True
+        reason = (
+            "non-canonical status(es) in OPEN_STATUSES: "
+            + ", ".join(NON_CANONICAL_OPEN_STATUSES)
+            if NON_CANONICAL_OPEN_STATUSES
+            else "OPEN_STATUSES is empty"
+        )
+        log.warning(
+            "`alissa task list` offers --status and this daemon is not using "
+            "it (%s) -- the CLI rejects the whole call over one unknown value, "
```
alissa-app, line 467: [nit] The trailing clause is hardcoded to the non-canonical case, so it contradicts the other branch of `reason`. When `OPEN_STATUSES` is empty the emitted line reads:

> `alissa task list` offers --status and this daemon is not using it (**OPEN_STATUSES is empty**) -- the CLI rejects the whole call over one unknown value, so the corpus is listed unnarrowed

There is no unknown value in that branch — nothing was rejected, the set simply had nothing to send. (Reproduced by pointing `NON_CANONICAL_OPEN_STATUSES` at `()` with an empty filter.)

It is unreachable in a healthy tree and only misleads a reader already debugging a code defect, hence a nit. If you touch it, folding the *consequence* into `reason` alongside the *cause* would make both branches honest.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -365,6 +437,38 @@ def probe_task_list(self) -> "TaskListFlags":
...
+        filter costs nothing and there is nothing to warn about. Here the
+        message is only emitted when the flag was on offer and could not be
+        used, which is the only case where the defect actually costs reads.
+        """
+        if self._status_filter_warned:
+            return
+        self._status_filter_warned = True
+        reason = (
+            "non-canonical status(es) in OPEN_STATUSES: "
+            + ", ".join(NON_CANONICAL_OPEN_STATUSES)
```
alissa-app, line 461: [nit] Correct, but it leans on operator precedence that is easy to misread across a line break: the conditional binds looser than `+`, so this is `("...: " + ", ".join(...)) if NON_CANONICAL_OPEN_STATUSES else "..."` and not a `+` of a conditional. I checked it renders both branches as intended, so this is purely about the next reader.

An `if`/`else` statement, or parenthesising the concatenation, would say the same thing without the pause. The rest of this module is unusually explicit for exactly this reason.

alissa-app (approved):

Review task: **TASK-213710745** · Origin: TASK-1593278191 · Implementation: TASK-2038401416

**Findings:** 0 blocker · 0 major · 2 minor · 2 nit · 0 question. Nothing here gates the merge.

### Review surface

Judged the delta `9b25e9e..57d66e1` — one commit. Worth stating because the picture changed while this was under review: **#98 merged** (`c33d22d`, a merge commit, so `9b25e9e` is on `main` verbatim), which makes `git merge-base origin/main 57d66e1` exactly `9b25e9e`. The PR body's merge-order hazard therefore never materialised — the squash-folding scenario it guards against is moot, and GitHub's files view showing 428 additions is only the stale `base.sha` (`9755d08`) on the API, not a real duplicate delta. The true delta against today's `main` is 126/24 across 5 files.

The base excursion (cut from #98's approved head rather than `main`) is reasoned in the body and holds: every one of the five findings lives in code #98 introduces, #98's head never moved, and the branch is *based* on `main`. Same shape as #48/#53/#89.

### Verified independently, not taken from the description

- **Finding 1 actually holds at the mechanism level.** Importing `alissa.tools.github.revloop.alissa` with stderr captured writes `''` — the import-time `lastResort` leak is gone. Driving `_warn_status_filter_dropped` with `logging.basicConfig` configured the way `__main__` configures it produces a fully formatted line (`2026-08-28 04:19:22,046 WARNING alissa.tools.github.revloop.alissa: ...`), naming `todo`, **once** across two `task_list_argv` calls and not at all under `narrow_status=False`. Both branches of `reason` render correctly — the `+` binds tighter than the conditional, as intended.
- **The healthy tree is unaffected**: `TASK_LIST_STATUS_FILTER == 'committed,in_progress,pending_validation'`, `NON_CANONICAL_OPEN_STATUSES == ()`, so the new warning path is dead code until someone reintroduces the defect — which is the right shape for a guard.
- **CI at this exact sha**: all 7 check-runs `completed/success` (unit, style, types, wheel, version bump, image contract, entrypoint). Read the full `check-runs` rollup for `57d66e1`, not "the PR's checks"; no legacy commit statuses exist on this commit. HEAD re-confirmed as `57d66e1` immediately before submitting. `pytest` is unavailable on this runner, so the suite result is CI's, not a local re-run.

### Rubric

- **Correctness** — the five findings are each addressed at the mechanism, not just narratively. One gap in *test* coverage of the fix itself, filed as `[minor]` below; the code is right.
- **Coding practices** — matches the file's idiom (dense reasoned comments, `--` dashes, constants above their consumers). Branch `TASK-2038401416-STATUS-FOLLOWUP`, PR titles the task, co-author trailer present. Two `[nit]`s below. No dead code or untracked TODOs.
- **Security** — nothing found. The new log line interpolates only a module-level constant derived from a hardcoded set; no user or CLI input reaches it, no secrets, no injection surface.
- **Performance** — nothing found. The per-poll cost added to `task_list_argv` is one boolean check; `NON_CANONICAL_OPEN_STATUSES` is computed once at import. Making `_status_filter` pure removed work from import, it did not add any.
- **Side-effects / blast radius** — no contract change. `_status_filter`, `NON_CANONICAL_OPEN_STATUSES` and `_warn_status_filter_dropped` have no consumer outside this module (grepped `src/main`); the `alissa-tools-github-reviewloop` twin still does not exist at `main`, so there is nothing to reconcile. `0.22.0 → 0.23.0` is correct for two merges as two releases. Unchanged and correctly declared as the operator's gate: `docker/**`'s `ARG REVLOOP_VERSION` is `0.16.14`, so origin criterion **c3** (a live poll issuing the narrowed request) still needs both a base-image re-pin to ≥ 0.23.0 and a published CLI 0.2.0 — neither is in this diff, and neither is this PR's job.

### One stale artifact to clear, not a code finding

The **`blocked:merge-order` label is still on this PR**, and by the body's own rule ("drop it once #98 is in") it is now false — #98 merged as `c33d22d`. It is the only mechanical guard named, so leaving it on either stalls the merge or teaches readers the label means nothing. I am read-only and cannot drop it; flagging it as the last step before merge.

### Summary

This is a clean, well-evidenced follow-up. All five round-1 findings from #98 are addressed at the level the findings were about — finding 1 in particular moves the diagnostic to the one call site where it can be both *formatted* and *true*, and that reasoning is correct rather than merely plausible. The two `[minor]`s are a missing regression test for the fix's own property and a stale label; neither blocks. Approving; the minors are the implementer's to fix now or carry as a registered task per CR8.

PR #101, opened 2026-08-28
TASK-658833970: revloop — opt-in BOW-scoped task listing (`--bow`) for the 30s poll (issue #100)

### LINKED ISSUE #100: Revloop: BOW-scoped task listing — opt-in `--bow` narrowing for the 30s poll
## Context

`list_tasks` is this daemon's widest query — the operator's entire non-terminal corpus (measured 2026-08-27: 621 rows, ~734 KB) on a 30-second poll (`ALISSA_POLL_INTERVAL=30` in prod), and `apiListMyTasks` is the Alissa deployment's #1 Database I/O offender (378 MB on 2026-08-27). The daemon only needs the `Review PR …` tasks it maps PRs against. The Alissa CLI already advertises `--bow <bowId>` on `task list` (structural narrowing, studio issue #754): candidates come from the BOW's junction rows — which carry denormalized status, so non-matching rows never cost a task-document read — instead of the operator's involvement index. Equally important for cost: a BOW-scoped query's cache is invalidated only by writes inside that BOW, so the operator's unrelated task churn stops re-billing every poll.

Upstream (soft dependency): studio TASK-703031741 gives `create_downstream_task` a create-time `bodyOfWorkId`, which is what lets review tasks be born into a dedicated review BOW. Until sessions actually create review tasks there, a configured BOW would list empty — which is why this narrowing must be **opt-in per deployment**, exactly like `task_list_self_scope`.

## Scope

In `alissa-tools-github-revloop` (module `alissa.py`, plus `config.py` / `__main__.py`):

1. **Config key `task_list_bow_id`** (string, default unset) — env `ALISSA_REVIEW_TASK_BOW`, CLI flag `--task-list-bow <id>`, following the exact precedence and plumbing conventions of `task_list_self_scope`. Unset ⇒ today's call shape, unconditionally.
2. **Probe-gated adoption:** extend `TaskListFlags` with a `bow` field discovered by the existing `_advertises` probe (`TASK_LIST_BOW_FLAG = "--bow"`). The flag is sent only when (a) the config id is set AND (b) the probe advertises `--bow` — an issue's claim about a flag is not evidence; same doctrine as the existing flags. Non-zero `alissa` exit retries plain, same as today.
3. **Composition:** `--bow` composes with the existing `--status` / `--self` / `--view` narrowing when those are advertised — it swaps the candidate set; the others still narrow within it. Every `list_tasks` call site in the daemon (the per-pass poll in `loop.py`, and `prreview.py`'s discovery listing) goes through the same narrowing.
4. **Docs:** config reference documents the semantics: the id must be the Convex `_id` of the BOW review tasks are created into (NOT a repo's `autodev:` feed BOW, and note the known trap that a BOW's `_id` is not its `mirrorInstanceId`); a review task NOT in the configured BOW is invisible to the daemon — a missed review round, which is why the default is off.

## Out of scope

- The CLI flag itself (already shipped), studio/server changes, orcloop.
- Creating the review BOW or updating the review protocol to attach tasks to it (operator steps).
- Changing the default behavior of any deployment that does not set the new config.

## Definition of done

With `task_list_bow_id` set and the CLI advertising `--bow`, every task listing the daemon issues carries `--bow <id>` composed with the other advertised narrowing flags; with the config unset, or the flag not advertised, the call shape is byte-identical to today's; config precedence (file < CLI < env) and the probe gate are test-pinned; the config docs carry the review-BOW semantics and the `_id` trap.

## Definition of Done

Revloop gains an opt-in `task_list_bow_id` config (env `ALISSA_REVIEW_TASK_BOW`, flag `--task-list-bow`, same precedence conventions as `task_list_self_scope`); when set AND the installed CLI's help advertises `--bow` (probe-gated via the existing `_advertises` machinery), every `list_tasks` call site sends `--bow <id>` composed with the other advertised narrowing flags; unset or unadvertised leaves the call shape byte-identical to today; failure retries plain as today; docs state the id must be the review BOW's Convex `_id` (never an autodev feed BOW, `_id` ≠ mirrorInstanceId) and that unattached review tasks become invisible when narrowed.

## Acceptance criteria

- [ ] With config unset, or --bow not advertised in the probed help, the daemon's task-list invocations are byte-identical to today's (test-pinned); with both satisfied, --bow <id> is present on every list_tasks call site including prreview.py.
- [ ] --bow composes with --status/--self/--view when those are advertised — pinned by a test asserting the full argv of a maximally-narrowed call.
- [ ] task_list_bow_id follows the established precedence (file < CLI flag < env ALISSA_REVIEW_TASK_BOW) with tests mirroring the task_list_self_scope suite.
- [ ] Config docs name the review-BOW contract: Convex _id required (not mirrorInstanceId), never an autodev feed BOW, and the visibility consequence for unattached review tasks.

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/*.py, alissa-tools-github-revloop/src/test/**, README.md
Alissa-Task: TASK-135578267

RHDZMOTA:
Closes #100

Alissa tasks — origin: `TASK-135578267` · implementation: `TASK-658833970` (soft-dependent on the origin).

## What this does

`list_tasks` is the widest query this daemon issues — the operator's whole non-terminal corpus on a 30-second poll — and `--bow` is the only narrowing available to it that changes the **candidate set** rather than filtering one. Candidates come from a body of work's junction rows, which carry a denormalized status, so a non-matching row never costs a task-document read; and a BOW-scoped query's cache is invalidated only by writes *inside* that BOW, so the operator's unrelated task churn stops re-billing every poll.

Opt-in per deployment, because unlike every other narrowing on this path it trades **visibility** rather than wire bytes: a review task outside the configured BOW is invisible to the daemon, and that is a missed round rather than a slower one.

- **`task_list_bow_id`** — config key, `--task-list-bow <id>` flag, env `ALISSA_REVIEW_TASK_BOW`. Precedence file < CLI < env.
- **Probe-gated** — `TaskListFlags.bow`, discovered by the existing `_advertises` machinery. Both gates required: a configured id **and** the installed CLI advertising `--bow`.
- **Ordered, not composed** — `--bow` goes first (it picks the set). Round 1 measured what it does to the others: it **replaces** the actor's corpus and **ignores `--self` and `--include-terminal`**; only the digest projection still applies. The flags are still sent — a filter the server ignores widens the answer and `is_open` runs client-side, so it costs bytes, not reviews. Both call sites go through `task_list_argv`: the per-pass poll in `loop.py` and `prreview.py`'s discovery listing. A non-zero `alissa` exit still retries plain.
- **Byte-identical when off** — unset, or the flag unadvertised, and the argv is exactly what it was before this key existed.
- **Preflight** warns when an id is configured and the installed CLI cannot serve it — read off the *client*, not the config, because the env layer outranks both the file and the flag.

### Round 1 (`request_changes`, head `3b793a7`) — what changed

Six findings, all triaged on their threads before this push, per CR8.

| # | severity | triage | disposition |
| --- | --- | --- | --- |
| F1 | `blocker` | `[triage:pursue]` | `version` bumped `0.23.0` → `0.24.0` (PyPI's latest is `0.23.0`, so the merge would have published nothing silently) |
| F2 | `major` | `[triage:pursue]` | an empty BOW answer now drops **only** the BOW, not every narrowing; warning rewritten; README documents the restart |
| F3 | `minor` | `[triage:pursue]` | README now names all three `on_missing_review_task` modes — the default spawns the round *untethered*, it does not miss it |
| F4 | `minor` | `[triage:later]` | `revloop.config.example.json` is outside the declared scope → **TASK-1671410669** (soft-dependent on the implementation task) |
| F5 | `question` | `[triage:answer]` | probed the live CLI; answer below, and it corrected claims in both the README and the code comments |
| F6 | `nit` | `[triage:pursue]` | a non-string `task_list_bow_id` is refused at load instead of coerced into `--bow 0` |

**F5's answer moved something the operator gate previously listed as unverifiable.** The installed CLI *does* have `--bow`, so it was testable against real bodies of work after all (measured 2026-08-28, plain answer 1335 rows):

- **swap, not intersection** — across five BOWs, no answer was a subset of the plain one (24/35, 4/4, 5/5, 10/10, 1/2 rows outside it), and every outside row was non-terminal. A BOW-scoped list can surface tasks the actor's involvement index does not contain;
- **`--self` is ignored under `--bow`** — `--bow X --self` returns exactly `--bow X`, and that set is not a subset of the `--self` answer;
- **`--include-terminal` is ignored** — a BOW holding 52 tasks, 17 `validated`, answers 35 rows either way;
- **`--digest` still applies** (a projection, not a filter). **`--status` could not be probed** — this CLI has none.

No code behaviour follows from that: the ignored flags widen the answer rather than narrowing it wrongly, so they are still sent. What changed is the claims — the README gained a *What `--bow` actually does to the other flags* subsection and the `TASK_LIST_BOW_FLAG` comment was rewritten, both dated as a snapshot rather than a contract.

### Round 2 (`approve`, head `7cc6b8b`) — where its findings went

Round 2 approved and carried four documentation findings: F7/F8 (`minor`) the blanket "the narrowing is dropped" claim left stale in the README and in `list_tasks`' docstring by round 1's F2 split, F9 (`nit`) `task_list_argv`'s "today's CLI, which offers none of them" — disproved by the `HELP_LIVE_2026_08_28` fixture this PR itself adds — and F10 (`nit`) a note that the `except CommandError` path is deliberately blunter than the empty path.

All four are triaged `[triage:pursue]` on their threads and land in **follow-up PR #104**, not here: an approve is terminal, so a commit pushed onto this branch after the squash-merge would be silently lost. #104 is cut from this branch's approved head (three of the four correct prose about behaviour that exists only here), is based on `main`, carries `Follow-up-To: #101` and an explicit **merge #101 first** gate, and makes no closing claim — issue #100 stays attached to this PR alone.

### Judgment calls

- **The env layer is implemented in Python, not in the container's config renderer.** Every other `ALISSA_*` knob is rendered into the config *file* by `docker/claude/revloop-config.sh`, which would put env at the *file* layer — the opposite of the issue's stated `file < CLI < env`. `docker/**` is also outside the declared scope. Reading `ALISSA_REVIEW_TASK_BOW` in `config.env_task_list_bow_id` satisfies the stated precedence *and* is the only thing that reaches `alissa-pr-review`, which runs in a task worktree with neither the daemon's config file nor its argv. It is the only key with a fourth layer, and `config.py`'s module docstring says why.
- **`--bow` is placed first in the argv**, before `--status`/`--self`/`--view`. The issue says it "swaps the candidate set; the others still narrow within it", so reading the argv left to right reads them in that order. Round 1 measured that the second half of that sentence is not true of this CLI (see above); the ordering is kept because it still reads correctly, and the flags are kept because sending one the server ignores is free of correctness risk.
- **Empty means unset in every layer.** An exported-but-empty variable is how a container spells "not configured", and `--bow ""` is a call the CLI takes and answers with nobody's tasks. An empty *variable* does not blank an id the file or the flag already set.
- **`--project` is advertised by the live CLI and deliberately not probed for.** The daemon has no project to scope to; probing for a flag it would never send grows an answer nothing consumes.
- **`revloop.config.example.json` and the container env pass-through were not touched.** Both are outside `Autonomous-Scope`, and neither is needed for this key: the README config table is the config reference the issue asks for, and the variable reaches the daemon by plain inheritance rather than by being rendered into a file. The example-config gap became round-1's F4 and now carries **TASK-1671410669**.
- **Dropping the BOW on an empty answer is a one-way door for the process.** A wrong id and a correct-but-empty BOW are indistinguishable under a swap, so the first empty BOW answer stops `--bow` for the life of the process and a populated BOW needs one daemon restart to take effect. The alternative — holding the BOW — costs two list calls per pass for ever on a typo. Documented in the README and in the warning itself.
- **Two existing tests changed meaning rather than being added to.** `test_todays_cli_advertises_only_self` was renamed to `test_the_2026_08_16_cli_advertised_only_self` (its fixture is a dated capture and is no longer "today"), and `test_cli_0_2_0_advertises_everything_the_daemon_asks_for` now expects `bow=True` — the 0.2.0 fixture already contained `--bow`, so the probe finding it is the correct new answer, not a regression.

- [ ] **The id itself.** No *review* BOW exists yet on this deployment. `--bow` itself has now been exercised against five real bodies of work (F5), so the flag works; what is unverified is that any particular id is the right one for review tasks. Upstream studio `TASK-703031741` (create-time `bodyOfWorkId` on `create_downstream_task`) has to land, and the review protocol has to start attaching review tasks, before the key can be set anywhere without listing an empty corpus.
- [ ] **The empty-corpus safety net against a *wrong* id.** A mistyped id, a `mirrorInstanceId`, or an `autodev:` feed BOW all present as an empty answer, which the pre-existing runtime disproof retries plain and drops the narrowing for. That path is unit-tested with a fake CLI; it has not been exercised against a real BOW that genuinely holds nothing.
- [ ] **Container pass-through.** `docker/claude/entrypoint.sh` re-execs via `gosu` (which preserves the environment) and invokes `alissa-revloop` directly, so `ALISSA_REVIEW_TASK_BOW` set on the container should reach the daemon unmodified. That is read off the script, not booted — there is no docker in this runner.
- [ ] **The measured saving.** The 378 MB/day and 621-row/~734 KB figures come from the issue text; nothing here re-measured them, and this PR ships no telemetry that would.
- [x] **The CLI's server side — resolved in round 1.** Argv acceptance is still pinned through the transcription of CLI 0.2.0's option parse, but the *semantics* were measured directly against the installed CLI and five real bodies of work (F5 above). What remains unmeasured there is `--status` under `--bow`, because this CLI has no `--status`.
- [ ] **`--status` under `--bow`.** Untestable here. On the `--include-terminal` evidence it is likely ignored, which would cost bytes only. If a future CLI serves it *and* serves it wrongly, the failure would be a narrower-than-intended answer — worth a re-probe when `--status` ships.
- [ ] **The container version pin is unrelated to this bump and already behind.** `docker/claude/Dockerfile` pins `ARG REVLOOP_VERSION=0.16.14` while PyPI is at `0.23.0`, so publishing `0.24.0` does not put this change in the image. Pre-existing drift, outside this issue's scope, and stated here only so the validator does not read the version bump as shipping it.

### Verification

- `bash tests-unit.sh alissa-tools-github-revloop` — **844 passed** at head `7cc6b8b` (the head round 2 approved) (838 on `main`; 841 at round 1, plus three for the round-1 fixes). Two pre-existing assertions were re-aimed as described above, and `test_a_disproved_narrowing_drops_the_bow_with_everything_else` was rewritten in round 1 because it pinned the behaviour F2 rejected. Note: the runner imports the *installed* distribution, so the working tree has to be `pip install`ed first or collection fails on the new symbol.
- `bash check-style.sh alissa-tools-github-revloop` — clean (one E305 found and fixed).
- `bash check-types.sh alissa-tools-github-revloop` — `Success: no issues found in 30 source files`.
- Coverage on the touched modules (round 1): `alissa.py` 89%, `config.py` 98%, `loop.py` 97%, `prreview.py` 83%.
- **Counterfactual pass — every production change was reverted one at a time and the suite re-run.** Probe blind to `--bow` → 6 failures; argv never sends it → 4; env layer removed from `Config.build` → 1; `prreview` builds a bare client → 2; `loop` drops the id → 1; preflight warning removed → 1; `overrides_from` drops the key → 1; argparse `dest` typo → 14. The last three initially failed to fail, which is why `test_the_polls_own_client_carries_the_configured_bow` and `test_the_bow_id_walks_all_three_layers_through_the_real_entry_point` exist.
- **The round-1 fixes were counterfactualled the same way.** Restoring the old blanket disable in the empty branch → 2 failures; un-gating the argv on `_task_list_bow_disabled` → 1; coercing a non-string id instead of refusing it → 1. F1 has no unit test by nature — the `Version Bump Check` workflow is the test, and the value came from the `version` file plus PyPI, not from a PR title.
- **The fixtures and the semantics are live output, not invented.** `HELP_LIVE_2026_08_28` is `alissa task list --help` captured verbatim from the installed CLI today. It advertises `--self`, `--digest`, `--project` and `--bow` while still reporting version `0.1.0` — the same string it reported on 2026-08-16 when it offered none of the last three. Round 1's swap/`--self`/`--include-terminal` findings come from running that CLI against five real bodies of work, not from reading its help.

### Scope touched

Declared scope was `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/*.py`, `alissa-tools-github-revloop/src/test/**`, `README.md`. Everything changed is inside it:

- `alissa.py` — `TASK_LIST_BOW_FLAG`, `TaskListFlags.bow`, the ctor knob, the `task_list_bow_id` property, one argv branch; round 1 added `_task_list_bow_disabled` and split the empty-answer branch.
- `config.py` — `TASK_LIST_BOW_ENV`, `env_task_list_bow_id`, the `task_list_bow_id` field and `CONFIG_KEYS` entry, the `environ` layer on `Config.build`.
- `__main__.py` — `--task-list-bow` and its `overrides_from` entry.
- `loop.py` — pass the id into the client; one preflight warning.
- `prreview.py` — build its client with the env-resolved id (this file previously ignored *every* narrowing knob, including `task_list_self_scope`; only the BOW is wired here, since widening that to `--self` is not this issue's).
- tests — `test_task_list.py` (fixture + probe + argv + precedence), `test_loop.py` (client construction, preflight, CLI wiring, one fake attribute), `test_prreview.py` (the env reaches the discovery listing).
- `README.md` — config-table row, a fourth narrowing bullet, and a *Naming the review BOW* subsection carrying the `_id` contract, the `autodev:` feed trap, the `mirrorInstanceId` trap and the visibility consequence. The narrowing section's "of the three narrowings, today's CLI offers only `--self`" intro was also corrected against the live 2026-08-28 probe.

Round 1 added one file: `…/revloop/version` (`0.23.0` → `0.24.0`), which is inside the declared `…/revloop/*.py` area only by adjacency — it is the version file the publish workflow reads, and the `Version Bump Check` job requires it. Called out explicitly rather than folded in.

No other excursions outside the declared scope.


alissa-app:
- 2026-08-28 13:29:58 UTC — `review-alissa-github-review-daemon-pr101-r1-af872f` — round 1 of 22 — spawned
- 2026-08-28 13:52:33 UTC — `review-alissa-github-review-daemon-pr101-r2-970bae` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -122,6 +122,25 @@ def _status_filter(statuses: "set[str] | frozenset[str]") -> str:
...
+# invalidated only by writes INSIDE that BOW, so the operator's unrelated task
+# churn stops re-billing this daemon's 30-second poll.
+#
+# Opt-in per deployment for the same reason `--self` is, only sharper: a review
+# task that is not attached to the configured BOW is INVISIBLE to the daemon --
+# a missed review round, not a slower one -- and review tasks are only born into
+# a BOW once the sessions that create them say so. A configured BOW that nothing
+# writes into lists empty, which `list_tasks` treats as a disproved narrowing
+# and retries plain; that is a safety net, not a licence to guess the id.
+TASK_LIST_BOW_FLAG = "--bow"
```
alissa-app, line 165: [blocker] **CI is red at this head.** `Version Bump Check for alissa-tools-github-revloop` — **failure** — https://github.com/fahera-mx/alissa-github-review-daemon/actions/runs/33175574112/job/98863190707

This is not an unrelated flake: it is squarely about this diff. The PR changes published files under `alissa-tools-github-revloop/` (`alissa.py`, `config.py`, `__main__.py`, `loop.py`, `prreview.py`) and leaves `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version` at `0.23.0` — the same value `main` carries. The job's own message spells out the consequence: on merge, `Python Package Publish` uploads with `twine upload --skip-existing`, so re-publishing 0.23.0 is skipped *silently*, the workflow reports SUCCESS, and `main` ends up ahead of PyPI with nothing released and everything pinned to 0.23.0 still running the old code — the 2026-07-28 `alissa-github-develop-daemon` #62/#64 incident.

Fix: bump the `version` file to the next unpublished version. (`version.py` beside it is a derived reader — editing that alone changes nothing.)

The other six contexts at `3b793a75` are green: Unit-Test, Python Types, Python Style, Wheel Package, Image contract, and Entrypoint config renderer. This one context is the whole of the red.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -479,6 +522,14 @@ def task_list_argv(self, *, narrow_status: bool = True) -> list[str]:
         if self._task_list_narrowing_disabled:
             return argv
         flags = self.probe_task_list()
+        # The BOW goes FIRST because it is the only one of these that chooses a
+        # candidate set rather than filtering one: `--status`/`--self`/`--view`
+        # then narrow WITHIN it, and reading the argv left to right is reading
+        # them in that order. Composition is the point -- they are independent
+        # server-side, so scoping to a BOW does not make the status filter or
+        # the digest projection redundant.
+        if flags.bow and self._task_list_bow_id:
```
alissa-app: [major] **The empty-corpus disproof cannot tell a *wrong* BOW id from a *correct but currently empty* one — and its false positive silently disables every narrowing for the life of the process.**

`list_tasks` treats "narrowed call answered empty, plain call answered non-empty" as proof the filter is broken: it sets `_task_list_narrowing_disabled = True`, which `task_list_argv` reads three lines above this one as an unconditional early return of `PLAIN`. That premise held for `--status` / `--self` / `--view`, which all filter the *same* corpus — an empty answer from them really is anomalous. `--bow` breaks it, because it swaps the candidate set: "the review BOW holds no open tasks while this actor holds 621 other ones" is the **expected** answer, not a broken filter.

And that state is not hypothetical — it is the documented rollout state, in this PR's own README: *"Nothing creates review tasks into a BOW until the operator's review protocol says to ... so a deployment that sets this key before that is true lists an empty corpus."* So the first poll pass after an operator sets the key turns the narrowing off — the BOW **and** `--status`/`--self`/`--view` with it — and it stays off until the process restarts, since the flag is instance state and `ReviewWatcher` builds one `Alissa` per process. The daemon quietly reverts to exactly the whole-corpus poll issue #100 exists to remove.

The operator's only signal is a warning that says the opposite of what happened: *"the plain task list returned %d task(s) — the narrowed call is dropping rows, so this process stops narrowing."* Dropping rows is what `--bow` is **for**.

`test_a_disproved_narrowing_drops_the_bow_with_everything_else` pins this as desirable, so it is a deliberate choice rather than an oversight — but that test only exercises the wrong-id reading of an empty answer, never the right-id one. Either scope the disproof so a BOW-empty answer does not condemn the other three flags (or does not condemn the BOW permanently), or — if the current behaviour is genuinely what you want — say so in the README beside the empty-corpus paragraph, so an operator knows that a transiently empty review BOW costs them the optimization until the next restart.

README.md
```diff
@@ -709,12 +716,53 @@ same one it always made:
...
 once unnarrowed and the narrowing is dropped for the rest of the process — both
 are how a CLI that advertises a flag its API does not serve would present, and
 either would otherwise read as "this actor has no review tasks".
 
+#### Naming the review BOW
+
+`task_list_bow_id` is opt-in per deployment, and the reason is sharper than the
+one behind `--self`: every other narrowing on this page trades **wire bytes**,
+while this one trades **visibility**. A review task that is not attached to the
+configured BOW is invisible to the daemon — not a slower round, a **missed**
```
alissa-app: [minor] **"a *missed* one" understates what the default actually does.**

`on_missing_review_task` defaults to `ON_MISSING_SPAWN` (`spawn_anyway`), and in `loop.py`'s `_refused_before_start` only the `skip` mode turns an unresolved review task into `NO_REVIEW_TASK_REASON` (`ON_MISSING_CREATE` is `warn_and_spawn`, which also spawns). So on a default deployment a review task hidden by the BOW scope does not produce a missed round — it produces a round that is **spawned anyway, untethered from the task**: the reviewer session has no task to hydrate and nowhere to record its verdict envelope.

Worth saying the reassuring half too, since it is not obvious: round counting itself survives, because `completed` is derived from the PR's own review records, not from the task — so the cap still holds.

A sentence naming the three `on_missing_review_task` modes here would put the real consequence on the page an operator is already reading when they set the key.

README.md
```diff
@@ -129,6 +131,7 @@ extending it. `--dry-run` / `--no-dry-run` override the config in both direction
...
 | `review_task_miss_ttl_polls` | `10` | how many polls a PR with **no** review task is taken on trust before the task corpus is searched for one again. Trades **latency** for reads: a review task created mid-window is picked up at the re-arm rather than the next poll. Floor `1` — there is no value that turns it off — see *Bounding the task-list read* |
 | `task_list_self_scope` | `false` | narrow `alissa task list` to this actor's own rows (`--self`). **Off by default on evidence**: a small minority of review tasks on the live fleet are owned by another actor, and a review task the list cannot see is a round the daemon cannot count — see *Bounding the task-list read* |
+| `task_list_bow_id` | `null` | scope `alissa task list` to one body of work (`--bow`), so candidates come from that BOW's junction rows instead of the operator's whole involvement index. The **only key the environment can set** (`ALISSA_REVIEW_TASK_BOW`, which wins over both the file and `--task-list-bow`). Off by default: a review task **outside** the configured BOW is invisible to the daemon — see *Bounding the task-list read* for the id's contract and the two ways to get it wrong |
```
alissa-app: [minor] **`revloop.config.example.json` is now the one config surface missing this key.**

That file carries an entry *and* a `_`-prefixed doc string for every other key, `task_list_self_scope` included — its `_task_list_self_scope` text is where the sibling narrowing's semantics actually live for an operator working from the example. The PR notes the file sits outside the declared `Autonomous-Scope`, which is a fair call and I am not asking you to widen scope for it. But the gap is real: someone copying the example gets a config that cannot express the new key, and nothing in the file hints it exists.

A registered follow-up task is a fine disposition here; silently dropping it is not.

README.md
```diff
@@ -709,12 +716,53 @@ same one it always made:
   by the agent actor whose sessions write the rest. A review task the list cannot
   see is a round the daemon cannot count. Turn it on only where every review task
   is created by this daemon's own reviewer sessions.
+- **`--bow <id>`** — off unless `task_list_bow_id` names a body of work. The only
```
alissa-app, line 719: [question] **Does `--bow` intersect with the actor's involvement index, or replace it?**

This bullet — and the `TASK_LIST_BOW_FLAG` comment in `alissa.py` — says candidates "come from that BOW's junction rows" *instead of* the operator's involvement index. Read literally that is a **swap**, not an intersection, and it has two consequences the docs do not draw out:

1. `--bow` could *widen* visibility in one direction — a review task inside the BOW but owned by **another actor** would become visible, which is precisely the case that keeps `task_list_self_scope` off by default on this fleet.
2. A BOW-scoped answer would then not be a subset of the plain answer at all — which is the other half of why the disproof I flagged in `alissa.py` is comparing two sets that are not comparable.

If instead the server intersects (owned ∩ BOW), the junction-row cost argument still holds, but the widening does not and point 1 is moot.

Your own operator gate is straight about this being unverified — *"It asserts the argv is one the CLI *accepts*, not what the API returns for it."* That is the right disclosure and I am not asking for a live call against a BOW that does not exist yet. I am asking which of the two readings the README should be **claiming**, because it currently asserts the stronger one and the answer changes what an operator should expect.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -419,6 +467,19 @@ def build(
...
+        # file on its volume and no argv in its unit carries, so it has to be
+        # able to say something the other two do not.
+        env_bow = env_task_list_bow_id(environ)
+        if env_bow is not None:
+            raw["task_list_bow_id"] = env_bow
+        bow_id = raw.get("task_list_bow_id")
+        # Normalised the same way the env layer is, so `""` from a config file
+        # or a flag means "unset" too rather than `--bow ""` -- a call the CLI
+        # takes and answers with nobody's tasks.
+        bow_id = (str(bow_id).strip() or None) if bow_id is not None else None
```
alissa-app: [nit] `str(bow_id)` coerces a non-string config value rather than rejecting it, so `{"task_list_bow_id": 0}` becomes `--bow 0` and `{"task_list_bow_id": false}` becomes `--bow False` — well-formed argv for an id that can never resolve, landing the deployment in the empty-answer path above rather than in the `unknown config key` error that a *misspelled* key gets. This does match the file's idiom (`bool(...)` / `int(...)` coercion elsewhere in `build`), so it is genuinely your call; I mention it only because the failure mode is silent rather than loud.

alissa-app (changes requested):

Reviewer: fresh instance per CR3. Alissa review task **TASK-91380496** (created this round — no review task existed for this PR). Origin `TASK-135578267`, implementation `TASK-658833970`.

**Verdict: `request_changes`** — one blocker, one major, two minors, one open question, one nit.

### CI gate

Head `3b793a75dff29213c895a12d5842c6e35d96b23c`, rollup read from `commits/<sha>/check-runs` (the full read, not the Actions-API fallback). All seven contexts have concluded: **six green, one failed.** `Version Bump Check for alissa-tools-github-revloop` — https://github.com/fahera-mx/alissa-github-review-daemon/actions/runs/33175574112/job/98863190707 — is red, and it is red *about this diff*: published files changed with the `version` file left at `0.23.0`. That alone forbids an approve this round, and it is F1 below.

### Findings

| # | severity | where | one-liner |
| --- | --- | --- | --- |
| F1 | `blocker` | `version` (absent from the diff) | Version Bump Check failed — a merge would publish nothing, silently |
| F2 | `major` | `alissa.py:531` | An empty *correct* BOW reads as a broken filter and permanently drops **all** narrowing for the process |
| F3 | `minor` | `README.md:738` | "a missed round" understates the `spawn_anyway` default: the round is spawned untethered from the task |
| F4 | `minor` | `README.md:134` | `revloop.config.example.json` is the one config surface missing the new key |
| F5 | `question` | `README.md:719` | Does `--bow` swap the candidate set or intersect it? The docs assert the stronger reading, unverified |
| F6 | `nit` | `config.py:481` | `str(bow_id)` coerces `0` / `false` into a well-formed unresolvable id |

### Dimension report

- **Correctness** — F2 (the disproof's premise does not survive a candidate-set narrowing), F5 (unverified server-side semantics the docs assert). The gate logic itself is right: both gates required, argv byte-identical when off, and `test_an_unset_bow_leaves_todays_call_byte_identical` pins that across all four help fixtures. The precedence chain file < CLI < env is correct and pinned end-to-end through the real `resolve_config`, not just through `Config.build`. I verified the `HELP_LIVE_2026_08_28` fixture against the live CLI in this container: it is verbatim, `--bow` and `--project` are really advertised, and `alissa --version` really still says `0.1.0` — the "version is not a capability signal" claim holds.
- **Coding practices** — F1 (the version-bump convention this repo enforces in CI), F6. Otherwise this matches the repo's idiom closely: comment density, dated help fixtures, counterfactual test pass, and the two renamed/retargeted assertions are correctly re-aimed rather than weakened. The `Config.build` `environ` parameter is additive and every existing caller (`__main__.py:247`, `webui/__main__.py:77`) still resolves.
- **Security** — nothing found. No new input reaches a shell, path, or query; the BOW id is a config value interpolated into an `argv` list (never a shell string) and Convex BOW ids are not secrets, so the id appearing in the narrowing warning is not a leak. No dependency, authz, or deserialization surface is touched.
- **Performance** — this is a performance change and its shape is sound (junction rows over the involvement index, BOW-scoped cache invalidation). F2 is the performance finding: the mechanism disables itself on the exact state the rollout starts in, and the fallback costs *two* list calls on that pass rather than one.
- **Side-effects / blast radius** — small and contained. Both `list_tasks` call sites are covered (`loop.py` poll, `prreview.py` discovery); off by default, so no deployment changes behaviour without an operator setting the key — `ALISSA_REVIEW_TASK_BOW` is confirmed unset on this fleet. `env_task_list_bow_id` means the `webui` process now also reads the variable through `Config.build`, which is harmless (the webui renders and never writes config back). The `gosu` re-exec in `docker/claude/entrypoint.sh` does preserve the environment, so the container pass-through claim in the delivery notes reads correctly off the script. F3 is the blast-radius doc gap.

Bump the version file (F1) and answer F2 — either narrow the disproof so a legitimately empty BOW does not condemn the other three flags for the process, or document that it does. F5 wants an answer on the thread, not necessarily a code change. F3/F4/F6 are triage-and-move-on.

Per CR8, every blocker/major/minor/question above needs exactly one reasoned `[triage:*]` reply before the next round; nits are optional. A green re-run of the same code can approve in the next round.

<sub>Read-only reviewer — no commits, no state changes. Verdict of record is the envelope on TASK-91380496.</sub>

RHDZMOTA:
- 2026-08-28 13:38:56 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr101-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]`

Confirmed and mine. The `version` file is `0.23.0` on both this branch and `main`, and PyPI's latest release for `alissa-tools-github-revloop` is `0.23.0` — so a merge would hit `--skip-existing` and publish nothing while reporting success, exactly as the job says.

Bumping to **`0.24.0`** — read from the version file plus PyPI, not from merged PR titles (a stale-base merge has clobbered that file before, so the published set is the authority). Only the `version` file changes; `version.py` beside it is a derived reader.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]`

This is right, and the live CLI makes it worse than you had evidence for. I probed `alissa task list --bow <id>` against five real BOWs in this container:

| BOW | rows | rows **not** in the plain answer |
| --- | --- | --- |
| `k178h9xh…` | 35 | 24 |
| `k17a8154…` | 4 | 4 |
| `k17b25bd…` | 5 | 5 |
| `k17astce…` | 10 | 10 |
| `k17bk1jx…` | 2 | 1 |

Not one BOW answer is a subset of the plain one (1335 rows), and the outside rows are all non-terminal — so it is a genuine candidate-set **swap**, not a terminal-inclusion artifact. Your second point in the F5 thread is therefore measured fact: the cross-check is comparing two sets that are not comparable, and "plain returned more" is not evidence of anything about `--bow`.

**The fix.** The retry itself stays — it is what guarantees the pass still sees review tasks, and that is correctness. What changes is the blast radius of the disproof: when the narrowed argv carried `--bow`, an empty answer now drops **the BOW alone** for the process and leaves `--status` / `--self` / `--view` intact. Those three filter the same corpus, so an empty answer really is anomalous for them; `--bow` is the one flag for which "empty" is a legitimate answer, so it is the only one the evidence condemns. The warning is rewritten to name both causes (wrong id vs. a BOW nothing has been created into yet) instead of asserting the filter is broken.

Dropping the BOW for the process rather than keeping it is deliberate, and it is the safe direction: with a swap there is no way to tell a mistyped id from an empty-but-correct one, and holding the BOW would mean paying two list calls every pass forever on a typo. The cost is that a BOW that fills up later needs one daemon restart, which the README now says out loud beside the empty-corpus paragraph — your second option, taken as well as the first rather than instead of it.

`test_a_disproved_narrowing_drops_the_bow_with_everything_else` was pinning the old behaviour and is rewritten; a second test covers the right-id-empty-BOW reading you point out it never exercised.

README.md
RHDZMOTA: `[triage:pursue]`

Checked and you are right. `loop.py:1960` reads `if task is None and self.config.on_missing_review_task == ON_MISSING_SKIP`, so only `skip` produces `NO_REVIEW_TASK_REASON`; `spawn_anyway` (the default) and `warn_and_spawn` both spawn. "A missed round" is the `skip` deployment's consequence, and I wrote it as if it were everyone's.

Rewriting the paragraph to say what the default actually does — a round spawned untethered from its task, with no task to hydrate and nowhere to put the verdict envelope — and naming all three modes, since the operator setting this key is already on that page. Taking the reassuring half too: `completed` is derived from the PR's own review records rather than from the task, so the CR9 cap still holds either way.

README.md
RHDZMOTA: `[triage:later]` — registered as **TASK-1671410669** (soft-dependent on the implementation task TASK-658833970).

Agreed on both halves: the gap is real, and it is outside issue #100's declared `Autonomous-Scope` (`…/revloop/*.py`, `src/test/**`, `README.md`). The task carries the full remedy — the `task_list_bow_id` entry plus the `_task_list_bow_id` doc string beside it, matching how `_task_list_self_scope` documents the sibling narrowing, and covering the `_id` contract, the two traps, the `ALISSA_REVIEW_TASK_BOW` override and the visibility consequence — so it is deferred with a ref behind it rather than dropped in prose.

README.md
RHDZMOTA: `[triage:answer]` — **swap, measured.** And the measurement found something worse than the question asks about, which turns part of this into a code/doc fix as well.

The installed CLI has `--bow`, so this was testable after all. Against five real BOWs (plain answer: 1335 rows):

| BOW | rows | rows **not** in the plain answer |
| --- | --- | --- |
| `k178h9xh…` | 35 | 24 |
| `k17a8154…` | 4 | 4 |
| `k17b25bd…` | 5 | 5 |
| `k17astce…` | 10 | 10 |
| `k17bk1jx…` | 2 | 1 |

No BOW answer is a subset of the plain one, and every outside row is non-terminal (`committed` / `in_progress` / `pending_validation` / `blocked` / `draft`), so it is not the terminal filter creating the difference. **It replaces the involvement index.** Your point 1 holds: a BOW-scoped list can surface tasks the actor's own index does not contain. Your point 2 holds too, and it is why F2's cross-check is comparing incomparable sets.

Two further things the same probe turned up, both of which make my own composition claim wrong:

1. **`--self` is ignored under `--bow`.** `--bow X --self` returns the same 35 rows as `--bow X` alone, and that set is not a subset of the 1299-row `--self` answer. So a deployment running `task_list_self_scope` *and* a BOW gets the BOW's whole membership, not the actor's slice of it.
2. **`--include-terminal` is ignored too.** The BOW holds 52 tasks of which 17 are `validated`; `--bow X` returns 35 and `--bow X --include-terminal` still returns 35. So the default non-terminal filter is applied server-side, but the flag that would lift it is not.

`--digest` does still apply (the rows come back in digest shape), which is expected — a projection, not a filter.

I cannot test `--status`, since the installed CLI has none. Given `--include-terminal` is ignored I would not assume `--status` composes either.

**What changes.** No code behaviour: sending flags the server ignores under `--bow` costs bytes, never correctness — the answer is *wider* than advertised, `is_open` still filters client-side, and the argv is one the CLI accepts. What changes is the claims. The README's "the BOW picks the set, they narrow within it" and the matching `TASK_LIST_BOW_FLAG` comment are replaced by what was measured, including the `--self` widening, with the probe dated so the next reader treats it as a snapshot rather than a contract — this CLI has now grown flags twice without moving off version `0.1.0`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: `[triage:pursue]`

Taking it. You are right that it matches the file's coercion idiom, but the idiom is load-bearing in the other cases — `int(...)` on `poll_interval` turns a JSON string into the number the validation below it needs. Here there is nothing downstream to validate the coerced value, so `0` and `false` become well-formed argv for an id that can never resolve, and the deployment lands in the empty-answer path instead of getting told.

`task_list_bow_id` now refuses a non-string value with a message naming the type it got, so a JSON `0` fails at load like a misspelled key does rather than silently at the first poll. `null` still means unset, and a string is still stripped.

README.md
```diff
@@ -709,12 +716,102 @@ same one it always made:
...
+  one of the four that changes the **candidate set** rather than filtering it, and
+  the only one worth more than bytes: candidates come from that BOW's junction
+  rows, which carry a denormalized status, so a non-matching row never costs a
+  task-document read — and a BOW-scoped query's cache is invalidated only by
+  writes *inside* that BOW, so the operator's unrelated task churn stops
+  re-billing every 30-second poll. It does **not** compose the way the other
+  three do — see below.
 
 A narrowed call that fails, or that answers with an *empty* corpus, is retried
```
alissa-app, line 728: [minor] **This paragraph now contradicts the behaviour this round introduced.**

> A narrowed call that fails, or that answers with an *empty* corpus, is retried once unnarrowed and **the narrowing is dropped for the rest of the process**

That is exactly the claim the F2 fix stopped being true: an empty answer from a BOW-scoped call now drops `--bow` **alone**, and `--status` / `--self` / `--view` survive. The corrected statement does exist — 60-odd lines further down under *Naming the review BOW* ("The other three narrowings are **kept**") — but a reader working through the narrowing bullets hits the blanket version first, and it is the version that sits directly under the new `--bow` bullet.

One clause is enough: name the BOW carve-out here and point at the later subsection, rather than leaving the two paragraphs asserting opposite things about the same code.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -502,8 +584,9 @@ def list_tasks(self, *, narrow_status: bool = True) -> list[Task]:
         filtering. Newer CLIs offer more, so the call is now assembled from a
         boot-time probe of the installed CLI's help (`task_list_argv`): a status
         filter covering exactly the statuses a live review task can hold, a lean
-        `--view digest`, and `--self` when the deployment says its review tasks
-        are actor-owned. None of it is required; an absent flag is simply not
+        `--view digest`, `--self` when the deployment says its review tasks are
+        actor-owned, and `--bow <id>` when it has named the body of work they
```
alissa-app, line 588: [minor] **Same stale claim, in the docstring of the function whose behaviour actually changed.**

You correctly extended this docstring's flag list to mention `--bow <id>` (line 588), but its closing paragraph twenty lines down is untouched:

> A narrowed call that FAILS, or that answers with an empty corpus, is retried once unnarrowed and turns **the narrowing** off for the rest of the process.

For the empty-corpus half that is no longer what the code below it does — `list_tasks` now splits on `TASK_LIST_BOW_FLAG in argv` and turns off only the BOW in one of the two branches. This is the docstring a reader consults *before* reading the branch, so it is the one place the correction matters most, and it is the only one of the three claim sites (README bullet, `TASK_LIST_BOW_FLAG`, here) that the round-1 sweep missed.

The `_task_list_bow_disabled` attribute comment and the in-branch comment are both excellent; this is just the summary above them not having caught up.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -479,6 +552,15 @@ def task_list_argv(self, *, narrow_status: bool = True) -> list[str]:
...
             return argv
         flags = self.probe_task_list()
+        # The BOW goes FIRST because it is the only one of these that chooses a
+        # candidate set rather than filtering one, and reading the argv left to
+        # right is reading them in that order. The others are still sent: the
+        # projection (`--view digest`) genuinely applies within a BOW, and the
+        # ones the server ignores there cost bytes rather than reviews -- see
+        # TASK_LIST_BOW_FLAG for which is which, and for why that list is a
+        # dated snapshot rather than a contract.
+        if flags.bow and self._task_list_bow_id and not self._task_list_bow_disabled:
```
alissa-app, line 562: [nit] The docstring fourteen lines up (`task_list_argv`, line 548) still says:

> Every addition is probe-gated, so an older CLI -- **today's, which offers none of them** -- produces exactly the call the daemon has always made.

This PR's own `HELP_LIVE_2026_08_28` fixture disproves the parenthetical: today's installed CLI advertises `--self`, `--digest`, `--project` **and** `--bow`, which is precisely what made F5 answerable by measurement rather than by reading the option parse. You corrected the identical claim in the README's narrowing intro this round; this copy is pre-existing text that the same sweep did not reach.

The load-bearing half of the sentence — *every addition is probe-gated* — is true and is what the reader needs, so this is genuinely discretionary: dropping the dash-clause, or dating it the way the `TASK_LIST_BOW_FLAG` block now dates its claims, would do it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
```diff
@@ -558,7 +641,34 @@ def list_tasks(self, *, narrow_status: bool = True) -> list[Task]:
             "empty corpus from a filter this API does not serve", " ".join(argv),
         )
         tasks = self._tasks_from(run_json(plain, timeout=90) or [])
-        if tasks:
+        if not tasks:
+            return tasks
+        if TASK_LIST_BOW_FLAG in argv:
```
alissa-app, line 646: [nit] **The one branch F2's reasoning was not carried across.** Pre-existing, safe, and mentioned only so the asymmetry is on the record rather than discovered later.

This branch is now careful that "the plain list found rows" is not evidence against `--bow`. The `except CommandError` path at lines 615-623 makes no such distinction: a narrowed call that *errors* sets `_task_list_narrowing_disabled = True` unconditionally, so if the failing argv carried `--bow` the operator loses `--status` / `--self` / `--view` for the life of the process too — and any transient CLI failure (timeout, a 5xx from the API) does the same, having nothing to do with the flags at all.

I am **not** asking for a change, for two reasons. A hard failure genuinely is ambiguous evidence — unlike an empty answer, it does not say which flag the server choked on — and the degradation is in the wide-but-complete direction this whole path is designed to fail in, so it costs the optimization and never a review. It is also pre-existing (PR #88), not introduced here.

Worth a sentence in the comment above line 615 saying the error path is deliberately blunter than the empty path and why, so the next reader does not read the asymmetry as an oversight in the F2 fix.

alissa-app (approved):
## Review round 2 of cap 22 — `approve`

Reviewer: fresh instance per CR3. Alissa review task **TASK-91380496** — reused per CR2/CR7 (it already existed from round 1; the directive's "no review task yet" was stale, and its description already carries the effective cap **22**, so no correction was needed). Origin `TASK-135578267`, implementation `TASK-658833970`.

**Verdict: `approve`** — no blockers, no majors, no open questions. Two minors and two nits, all documentation, all listed below and none of them merge-blocking.

### CI gate

Head re-read immediately before submitting: still `7cc6b8b06d2a5034f689d53fce29bb7d9978171e`, unmoved since I started. Rollup read from `commits/<sha>/check-runs` — **the full read, not the Actions-API fallback**. All **seven** contexts concluded, **all `success`**: Entrypoint config renderer, Unit-Test, Wheel Package, Python Style, Image contract, Version Bump, Python Types. The legacy `commits/<sha>/status` rollup carries **zero** contexts, so it contributes nothing to fail. `Version Bump Check` — round 1's blocker — is now green.

### Prior-finding reconciliation (CR9 step 2)

All six round-1 findings carry exactly one reasoned `[triage:*]` reply. Nothing untriaged, so the round does not bounce.

| # | severity | triage | disposition |
| --- | --- | --- | --- |
| F1 | `blocker` | `pursue` | **fixed** (`7cc6b8b`) — `version` `0.23.0` → `0.24.0`, and `Version Bump Check` is green at this head. Only the `version` file changed; the derived `version.py` reader was correctly left alone |
| F2 | `major` | `pursue` | **fixed** (`7cc6b8b`) — verified in the code, not just in the reply: a new `_task_list_bow_disabled` flag is set in the `TASK_LIST_BOW_FLAG in argv` branch of `list_tasks` and gates `--bow` alone in `task_list_argv:562`, while `_task_list_narrowing_disabled` is now reached only by the `else`. The warning names both causes and the restart remedy. Self-correcting, too: with the BOW dropped, a *next* pass that still answers empty falls into the `else` and disables the rest, so a genuinely broken `--status`/`--self` is still caught one pass later at no correctness cost |
| F3 | `minor` | `pursue` | **fixed** (`7cc6b8b`) — README now names all three `on_missing_review_task` modes and says the default spawns the round untethered; the reassuring half (round counting survives, because `completed` derives from the PR's review records) is there too |
| F4 | `minor` | `later` | **deferred** — **TASK-1671410669** exists, status `committed` (open), and its definition-of-done is the full remedy I asked for, not a placeholder. Dangling-reference check passes |
| F5 | `question` | `answer` | **answered, and better than asked** — the probe against five real BOWs settles it as a **swap**, and turned up two things the PR had claimed wrongly (`--self` and `--include-terminal` are both ignored under `--bow`). The README gained a dated *What `--bow` actually does to the other flags* subsection and the `TASK_LIST_BOW_FLAG` comment was rewritten to match. The "no code behaviour follows" argument is right: an ignored filter widens the answer, `is_open` still runs client-side. Closed |
| F6 | `nit` | `pursue` | **fixed** (`7cc6b8b`) — `task_list_bow_id` now raises `ValueError` naming the type it got, matching `build`'s own idiom for `round_cap`/`on_missing_review_task` rather than the numeric-coercion idiom. `None` still means unset and the `(bow_id or "").strip() or None` normalisation is still correct on the str-only path |

No `ignore` dispositions, so nothing to re-judge and no impasse.

### New findings this round (delta `3b793a7..7cc6b8b`)

| # | severity | where | one-liner |
| --- | --- | --- | --- |
| F7 | `minor` | `README.md:728` | The blanket "the narrowing is dropped for the rest of the process" now contradicts the behaviour this round introduced |
| F8 | `minor` | `alissa.py:606` | The same stale blanket claim in `list_tasks`' own docstring — the docstring of the function whose behaviour changed |
| F9 | `nit` | `alissa.py:548` | "today's [CLI], which offers none of them" is disproved by the `HELP_LIVE_2026_08_28` fixture this PR adds |
| F10 | `nit` | `alissa.py:615` | The `except CommandError` path still condemns all narrowing when the failing argv carried `--bow` — the one branch F2's reasoning was not applied to |

All four are documentation or discretionary. None of them changes what the code does, and none is worth holding the merge for; F7/F8 are the two I would actually fix, and a follow-up task is a fine home for them.

### Dimension report

- **Correctness** — nothing found in behaviour. The F2 fix is right where it matters: the two disproof signals are now separated by the kind of evidence they carry, the BOW is gated in exactly one place, and the pass still returns the complete plain corpus on the disproving pass, so no review is ever missed as a result. I checked the branch the fix does *not* cover (F10, hard CLI failure) and it degrades in the safe direction — wider, never narrower. `Config.build`'s new `ValueError` fires before the strip, `env_task_list_bow_id` normalises `""`/whitespace to `None` on every layer, and the env-after-CLI ordering really does implement `file < CLI < env`. The new key cannot regress any existing deployment: it did not exist before this PR, so no config file in the field can carry the non-string value that now raises.
- **Coding practices** — F7/F8/F9. The code itself matches the repo's idiom closely: the ValueError is the file's own validation idiom rather than its coercion idiom, and the reason is written down; the new comments are dated snapshots rather than assertions; the rewritten tests replace a test that pinned the rejected behaviour rather than being bolted on beside it, and `test_an_empty_answer_without_a_bow_still_disables_everything` pins the un-changed half so the split cannot silently collapse. The findings above are all the *prose* lagging the code, which is a narrow and consistent class.
- **Security** — nothing found. Unchanged from round 1: no new input reaches a shell, a path, or a query; the BOW id is interpolated into an argv **list**, never a shell string. The round-2 delta adds no new surface — the new `ValueError` message includes only `type(bow_id).__name__`, not the value, and the new warning logs a row count and no id.
- **Performance** — nothing found; this round improves it. The false positive that disabled the whole optimization on the rollout's own starting state is gone, which was the point of the change. Cost accounting is unchanged: one extra list on the disproving pass, one call per pass thereafter, byte-identical argv when the key is unset.
- **Side-effects / blast radius** — nothing found. Small and contained, off by default, both `list_tasks` call sites covered. The only new failure mode is a load-time `ValueError` on a non-string `task_list_bow_id`, which is loud, boot-time, and reachable only by a config that opts into a key that did not exist before this PR. `docker/**` and `revloop.config.example.json` are correctly still untouched, and the version bump to `0.24.0` is the only publishing consequence.

### Why approve

The reconciliation is clean — every round-1 finding is either fixed in the diff or deferred behind a real open task — and the new delta carries no blocker, no major and no open question. Per CR9 that is an approve even with minors outstanding; F7–F10 are listed here so they can be picked up as a follow-up rather than lost.

<sub>Read-only reviewer — no commits, no fixes, no state changes. Verdict of record is the envelope on TASK-91380496.</sub>

README.md
RHDZMOTA: `[triage:pursue]`

Right, and it is the worse of the two placements: that paragraph sits directly under the new `--bow` bullet, so a reader meets the blanket claim before the carve-out sixty lines later. I made the correction where I was writing new prose and left the paragraph I was writing *around* untouched.

Going into a follow-up PR off `main`, not onto this branch — the round-2 verdict is an approve and this head is merge-racy, so nothing more gets pushed here. The follow-up carries a clause naming the BOW carve-out at that paragraph and pointing at *Naming the review BOW*, so the two stop asserting opposite things.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]`

Agreed, and this is the site that matters most of the three — it is the summary a reader consults before reaching the branch it describes. I extended the flag list in that docstring and did not re-read its closing paragraph, which is exactly the sweep miss you name.

Same follow-up PR as F7. The closing paragraph will split the two halves it currently runs together: the FAILS half is unchanged, and the empty-corpus half says the BOW is dropped alone.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]`

Taking it rather than leaving it, because this PR is what disproved it: `HELP_LIVE_2026_08_28` is in the diff, and it shows `--self`, `--digest`, `--project` and `--bow` on the CLI the parenthetical calls "today's, which offers none of them". Shipping the fixture and the claim it refutes in one commit is worse than the pre-existing staleness was.

Same follow-up PR as F7/F8. Dating it rather than dropping it, to match how the `TASK_LIST_BOW_FLAG` block now dates its claims — the load-bearing half (*every addition is probe-gated*) stays exactly as it is.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py
RHDZMOTA: `[triage:pursue]` — for the comment you actually asked for, not for a behaviour change.

To be explicit that I read the reasoning and agree with it: the `except CommandError` path stays blunt. A hard failure does not say which flag the server choked on, so there is nothing to attribute it to — unlike an empty answer, where `--bow` is the one flag for which "empty" is a legitimate result. And it degrades wide-but-complete, so it costs the optimization and never a review. It is also pre-existing (PR #88) rather than something the F2 split introduced.

What goes in the follow-up is the sentence: a note above that branch saying the error path is deliberately blunter than the empty path and why, so the asymmetry reads as a decision rather than as a corner the F2 fix missed. That is the whole of it — no code moves.

RHDZMOTA:
**Round 2 findings F7–F10 → follow-up PR #104** (draft at the time of writing; being driven to review in the same session).

All four were triaged `[triage:pursue]` on their threads above. None of them could land here: round 2 returned an **approve**, so this head is terminal and merge-racy — a commit pushed after the squash-merge would be silently orphaned.

#104 is cut from this PR's approved head `7cc6b8b` rather than from `main`, because F7/F8/F10 correct prose describing the empty-vs-error split that only exists on this branch. Its **base is `main`**, and it carries an explicit **merge #101 first** gate plus `Follow-up-To: #101`; it makes no closing claim of its own, so issue #100 stays attached to this PR alone.

It also bumps `version` `0.24.0` → `0.25.0`, since it changes published files and will sit on a base that already carries `0.24.0` once this merges.

PR #106, opened 2026-08-29
TASK-437845115: revloop product-stability guard — hold the loop once the shipped-product diff has been empty for N rounds (issue #105)

### LINKED ISSUE #105: revloop: product-stability guard — hold the loop once the shipped-product diff has been empty for N rounds (mechanical CR9 converged-by-stability)
## Why

The adversarial review loop converges on the product in rounds 1–2 and then keeps spending rounds on non-shipped surfaces. Measured on `fahera-mx/studio.alissa.app` merged PRs, 2026-08-28/29 (executable non-test lines moved per fix round):

- `#847` — 16 rounds. Rounds 1–5 moved shipped code (130/15/17/3/10 lines); **rounds 6–16 moved zero shipped lines** — only `tests/**` — and ran 22:29 → 06:16 (≈7 h, ≈22 agent sessions). The round-7/8/9 majors were the wording of one MCP description and then the test that bans that wording.
- `#866` (open at authoring time) — round-1 fix moved 6 shipped lines; rounds 2, 3, 4 moved only comments and tests, each re-litigating one justification paragraph.
- `#834` — 13 rounds, product byte-identical since round 2 (recorded on TASK-1059759424). `#808` — capped out at 23, merged by operator waiver.

Nothing in the sample that was not approved by round 5 ever needed a shipped-code change after round 5.

The skill-side fix — `alissa-code-review` amendment **TASK-1059759424** in `fahera-mx/skills.alissa.app` (severity ceiling for non-shipped surfaces, a "converged-by-stability" CR9 trigger, non-normative prose rule) — asks the *reviewer* to compute the shipped-product delta and stop. That is a judgment rule in a fresh-instance session, and the fresh reviewer's judgment is precisely the thing that fails in the tail (it has no memory that the sentence was litigated three times). This task adds the **mechanical** guard in the daemon that already owns the round cap: the daemon computes the delta itself, tells the reviewer, and refuses to spawn further plain `request_changes` rounds when the product is provably unchanged. The two are complementary and independent — this task must not depend on the skill PR merging, and must not edit the skill.

### 1. Config

- New config key `stability_rounds: int` (default **3**; `0` disables the guard and leaves every decision bit-identical to today). Add it to `CONFIG_KEYS` in `config.py`, to the `Config` dataclass beside `round_cap`, to `Config.from_raw` validation (`>= 0`, `ValueError` otherwise, same style as `round_cap`), to the `--stability-rounds` CLI flag in `__main__.py`, and to the `revloop.config.example.json` with a one-line `_stability_rounds` doc string in the file's existing style.
- New config key `stability_nonshipped_globs: list[str]` — path globs that count as **non-shipped**. Default: `tests/**`, `test/**`, `**/*.test.*`, `**/*.spec.*`, `**/*.md`, `docs/**`, `**/__snapshots__/**`, `**/_generated/**`. Match with `fnmatch`-style semantics where `**` crosses directory separators (`pathlib.PurePosixPath.match` does NOT do this — pick an implementation and pin it with a test that includes a nested `src/a/b/x.test.ts`).
- Env mapping in `docker/claude/entrypoint.sh`: `ALISSA_STABILITY_ROUNDS` → `stability_rounds`, following exactly how `ALISSA_ROUND_CAP` reaches `round_cap` there. Document both keys in the README's config table and the docker README's env table.

### 2. Measurement — `ghclient.py`

- New method `GitHub.compare_files(owner, repo, base_sha, head_sha) -> list[str]` over `GET /repos/{owner}/{repo}/compare/{base}...{head}` returning the changed filenames (paginate `files` per the API's 300-file page; a `renamed` file contributes both `previous_filename` and `filename`). Follow the module's existing error-handling and 403/404 conventions (this credential is the same PAT that cannot read check-runs — the compare endpoint works with `contents: read`; if it 403s, the guard must be **inert for that PR with one WARN log line**, never a hold).
- Shipped delta for a PR at the moment the daemon would spawn round `k+1`: take the reviewer identity's own `request_changes` reviews (`my_reviews`, which carry `commit_id`), ordered by `submitted_at`; let `base` = `commit_id` of the review submitted `stability_rounds` rounds ago and `head` = the PR's current `head_sha`. `shipped_delta = [f for f in compare_files(base, head) if not any(fnmatch(f, g) for g in stability_nonshipped_globs)]`. The PR is **product-stable** iff there are at least `stability_rounds` completed `request_changes` rounds, every review in the window has a non-empty `commit_id` (older records lack one → not stable, fail toward rounds), and `shipped_delta` is empty.
- Comment-only hunks inside shipped files are OUT of scope: a shipped file that changed at all counts as product movement. The guard therefore fails toward another round, never toward a false hold.

### 3. Action — `loop.py`, at the round-`k+1` spawn decision (the same place `round_cap` is consulted)

Two-stage, both required:

1. **Grace round with notice.** When product-stable is first detected, round `k+1` still spawns, but its directive carries a new injected block `{stability}` (same mechanism as `{checks}` / `{credential}`; empty string when not stable), text to the effect of:

   > PRODUCT-STABILITY NOTICE — the shipped-product diff between `<base8>` (round k−N head) and `<head8>` (this head) is EMPTY for N consecutive request_changes rounds; only non-shipped paths moved: `<up to 10 paths, then "+M more">`. Per alissa-code-review CR9 (converged-by-stability), a plain request_changes is out of policy on this round: either APPROVE (open non-shipped findings become registered follow-up tasks, cite them in the verdict) or, if you hold that an unwaivable defect remains, your request_changes body MUST name the shipped file:line that is wrong at this head. A request_changes that names none is treated by the daemon as a hold.

   Apply the same GitHub-text caps/escaping the module already applies to directive-bound text (the `⟦⟧`-fenced, count- and length-capped list pattern used for check names) — path names are repo-controlled text reaching a directive.

2. **Hold.** If round `k+1` closes with `request_changes` and the PR is still product-stable at the next spawn decision, the daemon does **not** spawn round `k+2`. It records the hold in `state.db` keyed by head (`stability_kind(head_sha, rounds)` beside `capout_kind`), posts **one** operator page per head on the PR (mirroring the cap-out page: what was measured, both shas, the non-shipped paths that moved, and the existing `alissa-review: re-enter +N` grammar as the override), and appends a line to its own activity comment. A push that moves `head_sha` re-evaluates from scratch (a shipped-file change clears the hold naturally; a tests-only push re-holds on the new head with a fresh once-only page). An honoured operator ack lifts the hold for exactly the granted rounds, each spawned WITH the notice.

- `approve` at any point converges exactly as today. The guard never touches approve/withdrawn/drift/checks paths.
- The console (`webui`) shows the PR as `stability-held` with the two shas where it today shows cap-out; keep the page-render test's HTML assertions honest.
- Log every stability evaluation at INFO with the two shas and the delta size, in the existing log style, so an operator reading the log can see why a round did or did not spawn.

### 4. Tests (`src/test/.../test_revloop/`) — fail-on-revert, in the existing harness style

- `test_loop.py`: (a) `stability_rounds=3`, a shipped file moves in the window → not stable, round spawns without notice; (b) only `tests/**` moves for 3 rounds → round 4 directive contains the notice with both shas; (c) round 4 `request_changes` + delta still empty → no round-5 spawn, one page, second poll pages nothing; (d) `alissa-review: re-enter +1` from an operator → round 5 spawns with the notice; (e) `approve` after a hold → converged as today; (f) a review in the window with empty `commit_id` → guard inert; (g) `stability_rounds=0` → every decision identical to the pre-change fixture; (h) compare 403 → inert + one WARN; (i) tests-only push after a hold → new head, fresh single page.
- `test_directives.py`: the notice is present exactly when injected, absent otherwise, and the path list is capped/escaped.
- A `ghclient` test for `compare_files` with a two-page fixture including a rename.
- A glob-matcher test proving `**` crosses directories.
- `tests-unit.sh`, `check-types.sh`, `check-style.sh` all green; results table in the PR body.

### 5. Release

- Bump the package version per the repo's release convention (README release section / `version` file) — the deploy to Railway is the operator's step and is out of scope.

## Out of scope

- Any change to `fahera-mx/skills.alissa.app` or the `alissa-code-review` skill text (TASK-1059759424 owns that).
- Lowering `round_cap` / `ALISSA_ROUND_CAP` (operator env lever, not code).
- Comment-only hunk detection inside shipped files.
- devloop / orcloop.

## Definition of Done

A PR against fahera-mx/alissa-github-review-daemon implementing the product-stability guard per the contract (config keys + env/CLI mapping, GitHub compare measurement with non-shipped globs, grace-round notice injection, once-per-head hold + operator page lifted by the existing re-entry ack, console state, fail-on-revert tests, version bump) is open, flipped draft → ready-for-review, has passed the adversarial review loop to an approve verdict (or a recorded operator decision), and the PR URL is attached to this task as evidence.

## Acceptance criteria

- [ ] `stability_rounds` (default 3, 0 = inert) and `stability_nonshipped_globs` exist in config.py / CONFIG_KEYS / --stability-rounds / example json, and ALISSA_STABILITY_ROUNDS reaches the daemon through docker/claude/entrypoint.sh exactly as ALISSA_ROUND_CAP does; both documented in the READMEs.
- [ ] GitHub.compare_files exists (paginated, renames counted both ways); product-stable is computed from the reviewer's own request_changes reviews' commit_id over the window, with empty commit_id or a compare 403 making the guard inert (never a hold).
- [ ] When product-stable is first detected the next round still spawns and its directive carries the PRODUCT-STABILITY NOTICE naming both shas and the capped/escaped non-shipped path list; the block is absent otherwise.
- [ ] A request_changes on the grace round with the delta still empty results in no further spawn, exactly one operator page per head (re-entry grammar quoted), an activity-comment line, console `stability-held`; a head move re-evaluates, an honoured ack lifts the hold for the granted rounds with the notice.
- [ ] The nine test_loop cases, the directive tests, the compare_files fixture test and the ** glob test are present and green; reverting the guard fails them; stability_rounds=0 fixture is bit-identical to pre-change behaviour.
- [ ] The implementing PR's review loop converged: approve verdict on the downstream review task, or a recorded operator decision.

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/**, alissa-tools-github-revloop/src/test/**, alissa-tools-github-revloop/version*, revloop.config.example.json, docker/claude/entrypoint.sh, docker/claude/README.md, README.md
Alissa-Task: TASK-1005298589

RHDZMOTA:
Closes #105

**Origin task:** TASK-1005298589
**Implementation task:** TASK-437845115

## What this does

The review loop converges on the *product* in rounds 1–2 and then spends rounds on surfaces that never ship. studio #847 ran 16 rounds; rounds 6–16 moved nothing but `tests/**` — ~7 hours and ~22 agent sessions litigating the wording of one description and then the test that bans that wording.

The skill-side amendment (TASK-1059759424, a different repo) asks the *reviewer* to notice and stop. This is the mechanical half, in the daemon that already owns the round cap: it measures the shipped-product delta itself, tells the reviewer, and refuses to spawn further plain `request_changes` rounds when the product is provably unchanged. The two are independent — nothing here depends on that skill PR merging, and no skill text is touched.

**Two stages, both required.**

1. **Grace round with notice.** When the shipped diff between the head judged `stability_rounds` `request_changes` rounds ago and the current head is empty, round *k+1* still spawns — carrying a `{stability}` PRODUCT-STABILITY NOTICE naming both shas and the (fenced, count- and length-capped) non-shipped paths that moved, and telling the reviewer to approve or name the shipped `file:line` that is still wrong.
2. **Hold.** If that grace round comes back `request_changes` with the product still unmoved, no further round is queued: one operator page per head (quoting the existing `alissa-review: re-enter +N` grammar), one activity-comment line, a `stability_notices` row in `state.db`, and `stability-held` in the console inbox. A push that moves a shipped file clears the hold with no ack at all; a tests-only push re-holds on the new head with a fresh single page; an honoured ack lifts it for exactly the rounds granted, each queued *with* the notice.

Every unknown fails toward another round, never toward a hold: too few rounds, a review record with no `commit_id`, a compare the credential cannot read, a listing too long to prove absence from. `stability_rounds: 0` disables the guard whole — no comparison call, no directive block, no hold.

### Judgment calls

- **The env mapping lives in `docker/claude/revloop-config.sh`, not `entrypoint.sh`.** The issue says "env mapping in `docker/claude/entrypoint.sh`, following exactly how `ALISSA_ROUND_CAP` reaches `round_cap` there" — but `ALISSA_ROUND_CAP` does not reach the config in `entrypoint.sh`; it reaches it through the `render_revloop_config` jq renderer in `revloop-config.sh`, which the entrypoint sources. Following the *mechanism* the issue names rather than the *file* it names: pass-through `--arg stab` beside `--arg cap`, plus the `Dockerfile` ARG/ENV pair every other pass-through knob has, and the entrypoint's own comment updated. Excursion listed under *Scope touched*.
- **The Decision `Action` is reused, not extended.** A hold reports `ESCALATED` on the poll that pages and `CAPPED` afterwards, exactly as a cap-out does. A new enum member would need a new `poll_snapshots` column, a snapshot migration and console plumbing for a state the operator already reads as "the loop stopped and paged"; the console distinction the issue asks for is carried where an operator actually looks — the inbox kind `stability-held`, with both shas.
- **The gate sits below the local refusals, not literally beside the `round_cap` check.** Both are in `evaluate`. Placing it after `_refused_before_start` and the in-flight/stale checks means a hub-less PR never buys a comparison and is never paged about a round it was never going to run, and an in-flight round is not re-measured every poll. A held round gives up its FIFO seat exactly as a refused one does.
- **A held PR is answered from the ledger once paged.** The hold is re-decided every poll; paying a compare call a minute for a PR that is by definition not moving is the one cost this guard could plausibly add to a steady state. The short-circuit reads grants first, so an ack still lifts the hold on the very next poll.
- **The grace round is per *episode*, not per PR.** The hold fires only when exactly one `request_changes` round has closed since the notice. If the product moves and stability re-establishes later, that is a new convergence the reviewer was never warned about, so it earns a fresh grace round. Pinned by a test.
- **Re-entry grants are one pool, but only forward.** An `alissa-review: re-enter +N` ack lifts whichever gate is holding the loop — cap-out or stability. Round 1 found the missing half: `State.granted_rounds` is a lifetime `SUM` the cap consumes only implicitly, so an ack given earlier *for a cap-out* was still in the total and the stability guard would spend it a second time — disarming the guard for exactly as many rounds as that ack bought, on exactly the capped-out PR shape the issue cites. `stability_notices.grants_seen` is now **seeded** with the PR's ack total when a fresh episode's notice is written, so the predicate means "only an ack posted after this guard last spoke can lift this hold". Seeded rather than derived from `completed - round_cap`: that arithmetic double-counts once a lifted round pushes `completed` past the cap, re-arming the guard a round early.
- **`{stability}` is in the round-1 template too.** Round 1 cannot be product-stable in practice (the window needs `stability_rounds` prior `request_changes` reviews). It is in both templates anyway because the round number comes from the verdict-envelope count while the window comes from GitHub reviews, and a divergence between them would otherwise drop the notice silently.
- **The path list gets its own untrusted-data lead-in.** A filename is repo-controlled text exactly as a check-run name is, but the existing `UNTRUSTED_LEAD` says the data came from "this PR's workflow file and check runs" — wrong provenance for a diff. Same `⟦data⟧` fence, same caps, accurate sentence.
- **`glob_matches` is segment-by-segment, not bare `fnmatch`.** The issue asked for `**` to cross separators and pinned the nested case. Bare `fnmatch` would pass that pin but breaks the defaults in two other directions: `docs/*` would swallow `docs/a/b/c`, and `**/*.md` would *not* match a top-level `README.md`. Matching per segment gives both.
- **`compare_files` makes ONE request and refuses a listing at the API's file cap.** The guard reads *absence*, so a short list would be read as "nothing else moved"; it follows `pull_request_commits`, which refuses for the same reason. Round 1 corrected *how* it refuses: `files` on the compare endpoint is a hard-capped array, not a paginated one, so a listing that comes back at `COMPARE_FILE_PAGE` cannot be distinguished from a truncated one by paging further — 300 files means "cannot prove absence". `per_page` is asked for `1`, since it sizes the commits array that no caller here reads.

- [ ] **The container.** No docker in this runner: the `Dockerfile` ARG/ENV pair and the entrypoint's comment are not built. `docker/claude/tests-entrypoint-config.sh` *was* run locally and passes, which covers the renderer (key omitted when unset **and when empty**, emitted as a JSON number when set, and `ALISSA_STABILITY_ROUNDS=0` emitted rather than treated as unset).
- [ ] **The pinned `REVLOOP_VERSION` skew.** `docker/claude/README.md` records `Needs REVLOOP_VERSION >= 0.26.0`; the image still pins 0.16.14, so setting `ALISSA_STABILITY_ROUNDS` before that pin moves would render a config key an older library rejects as unknown and the daemon would not start. Re-pinning the image is a separate operator step, as it has been for every key since 0.16.3.
- [x] ~~**The live GitHub compare endpoint's paging.**~~ **Settled in round 1, and it was wrong.** The reviewer measured the endpoint against `api.github.com`; I reproduced it from this container before triaging. `per_page` sizes the *commits* array, `files` is hard-capped at 300, and page 2 carries no files at all — so the original paging loop returned 300 paths as a complete diff for any larger comparison, and since `files` comes back in **path order** the survivors are systematically the non-shipped early paths. That is a false hold, the one direction the guard is built never to take. Fixed; the fixture now carries the measured response shape.
- [ ] **Whether the deployed PAT can read `/compare` at all.** Still open, and a fixture cannot settle it: the inert-on-403 path is tested, the credential is not. The verification above ran under this container's `gh` token against public repos, which does not prove the reviewer identity's fine-grained PAT has `contents: read` on the watched private repos.
- [x] ~~**The 300-file constant.**~~ Confirmed correct — it is the cap, not a page size. What was wrong was the assumption that paging past it was possible.
- [ ] **The console rendering.** `stability-held` is asserted in the payload and the pill's CSS rule is present; nothing renders the page in a browser here.
- [ ] **Behaviour on a real 16-round PR.** The measurement that motivates the default (`stability_rounds: 3`) comes from the issue's sample, not from replaying those PRs through this code.
- [ ] **PyPI release.** Version bumped to `0.26.0` in the version file; publishing and the Railway deploy are the operator's steps.

### Verification

Run in this worktree, on the final tree:

| check | command | result |
| --- | --- | --- |
| unit tests | `bash tests-unit.sh alissa-tools-github-revloop` | **894 passed** (863 before this branch) |
| types | `bash check-types.sh alissa-tools-github-revloop` | **Success: no issues found in 30 source files** |
| style | `bash check-style.sh alissa-tools-github-revloop` | **clean** (exit 0) |
| entrypoint config | `bash docker/claude/tests-entrypoint-config.sh` | **ALL PASS** |

PR CI on the round-1 head was green on all seven checks (the eighth, `[code]smith`, skips).

Fail-on-revert was checked by mutation, not assumed. Ten mutations were applied to the shipped code and the suite re-run each time; nine were caught (guard switched off entirely → 12 failures; window base read as the current head → 5; bare `fnmatch` for the globs → 5; rename counted only under its destination → 1; hold branch disabled → 7; notice dropped from the directive → 2; the caller's once-only short-circuit removed → 1; the grace-round condition loosened from `==` to `>=` → 1). **One mutation was not caught**, and it found a real defect rather than a test gap: `_hold_on_stability` re-checked the ping ledger, which the caller already does *before* the comparison, so the inner check was unreachable. It is removed in the second commit and the caller's check is now pinned by a test.

**Round 1 fixes** were mutated the same way, on a committed tree this time: restoring the old paging stop condition fails `test_compare_files_refuses_a_listing_at_the_api_cap`; carrying `grants_seen` forward instead of seeding it fails both grant-pool tests; dropping the `cap` recompute fails the ack-bought-directive test.

Known reservations, stated rather than hidden: the ack tests and every other multi-round fixture now run with a `FakeGitHub.compare_files` that answers "a shipped file moved" by default, which is what keeps them byte-identical — the guard's *inertness* on a moving PR is therefore exercised broadly, while its active paths are exercised by the fixtures that set `changed_files` explicitly. The `stability_rounds=0` case has its own test asserting the pre-change decision **and** that no comparison call is made.

### Scope touched

Inside the issue's declared `Autonomous-Scope`:

- `alissa-tools-github-revloop/src/main/…/revloop/`: `config.py` (both keys, validation, `glob_matches`/`is_nonshipped`), `ghclient.py` (`compare_files`, two page constants), `loop.py` (the gate, the notice, the hold, the ping kind, the `{stability}` slot), `state.py` (`stability_notices` table + two accessors), `__main__.py` (`--stability-rounds`), `webui/sources.py` + `webui/page.py` (the `stability-held` inbox kind), `version` → `0.26.0`
- `alissa-tools-github-revloop/src/test/**`: `test_loop.py`, `test_directives.py`, `test_state.py`, `test_webui_sources.py`
- `revloop.config.example.json`, `README.md`, `docker/claude/README.md`, `docker/claude/entrypoint.sh` (comment only)

**Excursions**, all in service of the env mapping the issue asks for, none of them optional for it to work:

- `docker/claude/revloop-config.sh` — where `ALISSA_ROUND_CAP` actually becomes `round_cap`; the mapping is inert without it
- `docker/claude/Dockerfile` — the `ARG`/`ENV` pair every pass-through knob needs to be settable at build time
- `docker/claude/tests-entrypoint-config.sh` — three assertions for the new key, and adding it to the pinned-library cross-check's `env -u` list so a version skew cannot fail that check for the wrong reason

Not touched, deliberately: `fahera-mx/skills.alissa.app` and the `alissa-code-review` skill text (TASK-1059759424 owns those), `round_cap` / `ALISSA_ROUND_CAP`, comment-only hunk detection inside shipped files, devloop and orcloop.


alissa-app:
- 2026-08-29 23:50:11 UTC — `review-alissa-github-review-daemon-pr106-r1-8c0f56` — round 1 of 22 — spawned
- 2026-08-30 00:36:35 UTC — `review-alissa-github-review-daemon-pr106-r2-acce36` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
```diff
@@ -1196,6 +1207,69 @@ def issue_comments(self, owner: str, repo: str, number: int) -> list[IssueCommen
...
+                or {}
+            )
+            files = payload.get("files") or []
+            for entry in files:
+                for key in ("filename", "previous_filename"):
+                    name = str(entry.get(key) or "")
+                    if name and name not in seen:
+                        seen.add(name)
+                        out.append(name)
+            if len(files) < COMPARE_FILE_PAGE:
```
alissa-app: [blocker] **This stop condition can never detect truncation on the real compare endpoint, so `compare_files` silently returns a partial listing — and the caller reads absence from it.**

The docstring's contract is right: *"ABSENCE is the load-bearing answer, so the listing is paged to exhaustion and a listing that runs past `COMPARE_PAGE_LIMIT` raises `TruncatedListing` rather than returning a short list."* The endpoint does not behave the way that contract assumes.

I checked it live against `api.github.com` on `python/cpython` `v3.11.0...v3.12.0` (a comparison with thousands of changed files):

```
per_page=300 page=1  ->  {"commits": 300, "files": 300}
per_page=300 page=2  ->  {"commits": 300, "files": 0}     <-- files, not more files
per_page=100 page=1  ->  {"commits": 100, "files": 300}   <-- per_page does NOT bound files
per_page=500 page=1  ->  {"commits": 500, "files": 300}
```

Two facts fall out, and both break this loop:

1. `per_page` paginates **commits**, not `files`. The `files` array is hard-capped at 300 by the API.
2. `files` is served **only on page 1**. Page 2 returns `files: []` — not because the listing ended, but because that array is not paginated at all.

So for any comparison with more than 300 changed files, this loop runs: page 1 returns 300 (`300 < 300` is false, continue), page 2 returns 0 (`0 < 300` is true, **return**) — and `compare_files` hands back exactly 300 paths as if that were the complete diff. `TruncatedListing` is unreachable in production; `COMPARE_PAGE_LIMIT` and the warning above it are dead code on this endpoint.

The consequence is the one direction `_gate_spawn_on_stability`'s docstring says the guard must never take. GitHub returns `files` in **path order**, so the 300 that survive are the alphabetically first ones — on the cpython comparison the list does not even get past `Doc/library/enum.rst`. Non-shipped globs like `docs/**` and `**/*.md` sort early; `src/**`, `lib/**`, `Modules/**` sort late. Truncation therefore does not sample the diff randomly, it systematically drops the shipped files and keeps the non-shipped ones.

Reproduced against this branch's own code (`GitHub.compare_files` with `_api` stubbed to the response shape measured above — 300 `docs/**/*.md` entries on page 1, `[]` on page 2, real diff also containing `src/app.py`):

```
compare_files -> 300 paths, no TruncatedListing (real diff had 301)
shipped delta computed by the guard: []
shipped delta in TRUTH:              ['src/app.py']
```

`moved` is empty, so with `graced` set the daemon holds and pages the operator for a PR whose shipped code demonstrably moved.

This is adjacent to the PR body's fourth "Not verified" bullet, but the diagnosis there is not quite the defect: `COMPARE_FILE_PAGE = 300` is the *correct* constant. The bug is that a 300-entry page can never be distinguished from a truncated one by continuing to page, because the next page is always empty.

Suggested fix, in the fail-toward-rounds direction the rest of the guard takes — a single request, and exactly `COMPARE_FILE_PAGE` files means "cannot prove absence":

```python
payload = self._api(..., f"per_page={COMPARE_FILE_PAGE}", forbidden_is_rate_limit=False) or {}
files = payload.get("files") or []
if len(files) >= COMPARE_FILE_PAGE:
    raise TruncatedListing(
        f"{owner}/{repo} comparison {base_sha[:8]}...{head_sha[:8]} reported "
        f"{len(files)} files, the API's per-comparison cap; absence cannot be proven from it"
    )
```

The caller already turns `TruncatedListing` into an inert gate plus one WARN, so this needs no change at the call site. Note that `test_compare_files_refuses_to_answer_from_a_truncated_listing` currently passes only because its fixture serves 300 files on *every* page, which the real API never does — that fixture is worth reshaping to the measured response so it pins the real contract.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1974,6 +2157,220 @@ def _refused_before_start(
...
+        granted = self.state.granted_rounds(pr.full_name, pr.number)
+        if graced:
+            # The only place an ack for a stability hold can be DISCOVERED --
+            # `completed` is nowhere near the round cap, so the cap's own scan
+            # never runs. Same ledger, same once-per-comment accounting: an
+            # operator saying "run N more rounds" lifts whichever gate is
+            # holding the loop.
+            granted = self._collect_acks(pr, granted)
+            self._announce_grants(pr)
+            if lifts >= granted and self.state.pinged(
```
alissa-app: [minor] **`granted` here is the PR's lifetime grant total, so re-entry grants acked for a cap-out silently pre-authorise stability grace rounds.**

`State.granted_rounds` is `SUM(rounds)` over every ack ever recorded on the PR and is never decremented — the cap consumes grants implicitly, by `completed` counting up past `round_cap + granted`. The stability guard instead counts its own consumption in `lifts`, and gates on `lifts >= granted`.

Those two accountings do not compose. Concretely:

1. A PR moves shipped code through round 10 and caps out.
2. An operator acks `alissa-review: re-enter +5`. `granted` becomes 5; rounds 11–15 run.
3. By round 13 the PR is tests-only, and stability is detected. The grace round is queued (`lifts` 0 -> 1).
4. Round 14 comes back `request_changes`, product still unmoved. `graced` is true, but `lifts (1) >= granted (5)` is false — **no hold**. The guard keeps queuing "grace" rounds until `lifts` reaches 5.

So the guard is disarmed for five rounds on precisely the PR shape issue #105 cites as motivating it (`#808`, capped at 23). The PR body's "Re-entry grants are one pool" reasoning covers an ack given *while a stability hold is standing* — that case is right and well-tested. What is untraced is an ack given earlier, for a different gate, that the stability guard then spends a second time.

This fails in the safe direction (extra rounds, never a false hold), so it is a minor rather than a blocker. If "one pool" is the intent, the pool likely wants to be grants *not already consumed by rounds past the base cap* — e.g. comparing `lifts` against `max(0, granted - max(0, completed - round_cap))` — or `lifts` could be seeded from the grants outstanding when the first notice is written.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1843,6 +2008,18 @@ def evaluate(self, owner: str, repo: str, number: int) -> Decision:
...
             return refused
 
+        # THE PRODUCT-STABILITY GUARD (issue #105). Between the local refusals
+        # and the slot gate, which is where its two costs land best: it is paid
+        # for only by a round that is genuinely owed and could genuinely start
+        # (so a hub-less PR never buys a comparison, and never gets paged about
+        # a round it was never going to run), and a HELD round gives its FIFO
+        # place up exactly as a refused one does -- a seat a held round keeps is
+        # one the oldest genuine waiter does not get.
+        stability = self._gate_spawn_on_stability(pr, my_reviews, round_)
```
alissa-app, line 2037: [minor] **`cap` was computed above this line, so a grace round bought by a freshly-discovered ack is told a cap one lower than the effective one.**

`evaluate` computes `granted` and `cap = self.config.round_cap + granted` earlier (around L1969–1971), and only re-collects acks there when the loop is at or past the cap. A stability hold fires well below the cap, so an operator ack posted to lift *this* hold is discovered for the first time inside `_gate_spawn_on_stability`, where `self._collect_acks(pr, granted)` records the new grant. That local `granted` never propagates back, so the `cap` handed to `_spawn` on this same pass — and interpolated into the directive's `Round {round} of cap {cap}` — is stale by the granted amount.

It self-corrects on the next poll, and it gates nothing (the cap check already passed). But the number is not merely cosmetic in this system: the reviewer session copies it into the `- **Round:** k of cap M` line of its verdict envelope, which is the durable audit record of how much runway the loop had. A round that exists *because* an operator granted it is exactly the round most likely to be read back later.

Cheapest fix is to have the gate return the collected total alongside the notice and recompute `cap` before `_spawn`.

alissa-app (changes requested):

Reviewed at head `a95f4339`. CI at that sha: **8 contexts, all concluded, none failed** (7 success, `[code]smith` skipped; 0 legacy status contexts) — read via the check-runs rollup, which this credential could read directly.

This is a carefully built change and the reasoning in the comments is unusually good — the fail-toward-rounds invariant is stated everywhere it matters, the notice/hold split is well argued, the untrusted-path fencing is properly done and properly tested, and the mutation-testing pass in the Delivery note is real work that found a real defect. The blocker below is not a design objection; it is one factual assumption about the GitHub compare endpoint that turns out to be wrong, and it inverts precisely the invariant the guard is built on.

### Findings

| severity | count |
| --- | --- |
| blocker | 1 |
| major | 0 |
| minor | 2 |
| nit | 1 |
| question | 0 |

**[blocker]** `compare_files` cannot detect truncation on the real endpoint, so the guard can compute an empty shipped delta on a PR whose shipped code *did* move — a false hold. Inline on `ghclient.py`. Live-verified against `api.github.com`, details in the comment.

**[minor]** Prior re-entry grants pre-authorise stability grace rounds, disarming the guard on exactly the long PRs it targets. Inline on `loop.py`.

**[minor]** The `cap` a grace round is spawned with is computed before the stability gate collects acks, so a round bought by a fresh ack is told a cap one lower than the effective one. Inline on `loop.py`.

**[nit]** The stability gate sits above `_gate_spawn`, so a round that is merely waiting for a slot re-buys a compare call every poll. The placement rationale in the comment argues the *hold* should give up its FIFO seat (correct), but does not address the waiting round's repeated comparisons. Moving the gate below `_gate_spawn` would keep the seat behaviour for holds only if the hold is re-checked there too — probably not worth restructuring for, but worth a sentence in the comment saying the cost was accepted.

### Dimension report

- **Correctness:** one blocker (`compare_files` truncation, below) and two minors. Everything else checks out against issue #105's contract: the window base is `rc[-rounds:][0].commit_id` off `my_reviews`, which is genuinely sorted by `submitted_at`; renames are counted under both names; `stability_rounds=0` is a whole-gate skip with a test asserting no compare call; empty `commit_id`, 403 and `TruncatedListing` all return an inert `StabilityGate()`; `RateLimited` is correctly re-raised rather than swallowed. The grace-round `==` condition and the per-episode reset are right and pinned. All nine issue-mandated `test_loop` cases (a–i) are present, plus the directive, `compare_files` and `**`-glob tests.
- **Coding practices:** matches the module's idiom closely — the comment density, the "why not the obvious thing" docstrings, the ping-kind/parse pair beside `capout_kind`, the `CREATE TABLE IF NOT EXISTS` migration-free table with its own test. `glob_matches`'s segment-by-segment matcher is the right call and the docstring justifies it against both wrong alternatives. Commits carry the co-author trailers; branch and PR title name the task; the excursion into `revloop-config.sh`/`Dockerfile` is declared in the PR body and is genuinely required for the env mapping to work. Nothing found beyond the nit.
- **Security:** nothing found. Paths reaching the directive go through `directive_data` (fenced, count- and length-capped, `directive_text`-stripped) with a correctly-reworded provenance lead-in, and the hostile-filename test covers `U+2028`/`U+2029`, backticks and a forged `DATA_CLOSE`. The hold comment's paths go through `directive_text` too. No new dependencies, no secrets, no widened access; the two shas interpolated into the compare URL are GitHub-supplied commit ids.
- **Performance:** acceptable. The held-PR short-circuit answers from the ledger before paying for a comparison, which is the cost that mattered; the compare is otherwise bought only when a round is genuinely owed. The `**`-recursion is bounded by operator-written patterns, not repo-controlled text. One residual cost noted in the nit. Note also that `per_page=300` on the compare endpoint only enlarges the *commits* payload — it does not affect `files` — so each call currently drags 300 commits it never reads.
- **Side-effects / blast radius:** nothing found. `stability_notices` is additive with a migration-free create; `Action` is reused rather than extended, so `poll_snapshots` and the console need no migration; the inbox kind is a new value rendered as a CSS class that already has a rule; `stability_rounds=0` is a complete off switch with a bit-identical-decision test. The `REVLOOP_VERSION >= 0.26.0` skew is correctly documented in both the Dockerfile ARG comment and the docker README, and the entrypoint cross-check adds the new key to its `env -u` list so a skew fails for the right reason.

### Waived majors

None.

Fix the blocker: treat a first page of exactly `COMPARE_FILE_PAGE` files as unprovable rather than as a complete listing, so the guard goes inert (another round) instead of potentially holding. The two minors are small and can be pursued or reasoned away on the thread; the nit is discretionary.

---
*Reviewer: fresh instance per CR3. Verdict of record: the CR6 envelope on TASK-882968340 (round 1 of cap 22).*

RHDZMOTA:
- 2026-08-30 00:01:45 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr106-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: **[pursue]** — confirmed, and confirmed against the live endpoint rather than taken on the write-up. Reproducing your measurement from this container:

```
per_page=300 page=1 -> {"commits":300,"files":300,"total_commits":4591}
per_page=300 page=2 -> {"commits":300,"files":0,  "total_commits":4591}
per_page=100 page=1 -> {"commits":100,"files":300,"total_commits":4591}
per_page=1   page=1 -> {"commits":1,  "files":300,"first":".azure-pipelines/ci.yml","last":"Doc/library/enum.rst"}
```

All three of your claims hold: `per_page` bounds `commits` only, `files` is hard-capped at 300, and it is served on page 1 alone — so my `len(files) < COMPARE_FILE_PAGE` stop condition reads the empty page 2 as "the listing ended" and hands back 300 paths as a complete diff. `TruncatedListing` is unreachable and `COMPARE_PAGE_LIMIT` is dead. The `per_page=1` row also settles the direction of the bias: `files` comes back in path order, so what survives truncation is systematically the alphabetically-early paths, which is where `docs/**` and `**/*.md` live.

You are also right that my fourth "Not verified" bullet mis-diagnosed this. I wrote it as "the 300 constant may be wrong, and if it is, the guard is biased toward a hold" — the constant is correct and the bias is real, but the mechanism is the unpaginated array, which no re-reading of the docs would have given me. This is the bullet a live call should have retired before I wrote it down; I will say so in the amended note rather than quietly deleting it.

Taking your fix as written, with two additions:

- one request, `per_page=1` rather than 300 — since `per_page` only sizes the commits array and `files` is unaffected, the current call drags 300 commits it never reads on every evaluation. This also removes the loop, `COMPARE_PAGE_LIMIT`, and the now-unreachable warning.
- `test_compare_files_refuses_to_answer_from_a_truncated_listing` is reshaped to the measured response — 300 files on page 1, `files: []` on page 2 — so it fails against the code as it stands on this branch today. I will note in the reply that I checked it does.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: **[pursue]** — the walkthrough is right, and my "one pool" paragraph in the delivery note covers the case that works while missing this one. `granted_rounds` is a lifetime `SUM(rounds)` that the cap consumes only implicitly, so a grant acked for a cap-out is still sitting in the total when the stability guard reads it, and the guard then spends it a second time — disarmed for exactly as many rounds as the earlier ack bought, on exactly the PR shape (#808, capped at 23) the issue cites.

Of your two suggestions I am taking the second, because the first has a counter I could not make disappear: `max(0, granted - max(0, completed - round_cap))` double-counts in the other direction once a stability-lifted round pushes `completed` past `round_cap` — that round is then charged to the cap as well as to `lifts`, and the guard re-arms one round early. Seeding at the notice has no such interaction.

So: when a FRESH episode's notice is written (`not graced`), the column is seeded with the PR's current grant total rather than carried forward; a lift still writes `+1`. The predicate `>= granted` then means what it should — *only an ack posted after the guard last spoke can lift this hold*. In your scenario the round-13 notice seeds 5, round 14 comes back `request_changes`, `5 >= 5`, and it holds.

The column name goes with it: `lifts` no longer means "lifts taken", so it becomes `grants_seen` — "the ack total this guard has already accounted for and will not lift on again". Free to rename, since the table is new in this PR and unreleased.

Two tests: your scenario end to end (a prior cap-out grant must not buy a grace round), and a re-check that an ack posted *while the hold stands* still lifts it for exactly the rounds granted — the case that already worked and must keep working.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: **[pursue]** — correct, and the reason it matters is the one you name rather than the display: the reviewer session copies `cap` into the `- **Round:** k of cap M` line of its CR6 envelope, so the round most likely to be read back later — the one an operator paid an ack for — is the one carrying the wrong number.

Taking your suggested fix. `StabilityGate` grows a `granted` field, set only on the path that actually collected acks (`graced`), left `None` everywhere else so no other decision changes; `evaluate` recomputes `cap = self.config.round_cap + stability.granted` before `_spawn` when it is set. A test asserts the directive of an ack-bought grace round names the raised cap on the same pass, not one poll later.

RHDZMOTA:
## Round 1 triage — 4 findings, 4 `[pursue]`, 0 rejected

Replies are on each inline thread; this covers the nit, which has no thread, and records the triage in one place.

| # | severity | finding | triage |
| --- | --- | --- | --- |
| 1 | blocker | `compare_files` cannot detect truncation on the real endpoint | **[pursue]** — verified live from this container; your three measurements all reproduce |
| 2 | minor | prior cap-out grants pre-authorise stability grace rounds | **[pursue]** — seeding at the notice, not the arithmetic alternative (reasoning on the thread) |
| 3 | minor | `cap` is stale for an ack-bought grace round | **[pursue]** — gate returns the collected total, `evaluate` recomputes before `_spawn` |
| 4 | nit | a slot-gated round re-buys a compare call every poll | **[pursue]**, as a comment |

**[nit] — the repeated comparison while a round waits for a slot.** You read the placement rationale correctly: it argues why a *hold* gives up its FIFO seat and says nothing about the round that is merely waiting. I am taking it as a comment rather than a restructure, and the reason is worth stating because it is not obvious from the code — moving the gate below `_gate_spawn` does not simply relocate the cost, it changes what a full container *means* for this guard. A slot-gated round returns before the gate runs, so a PR that is product-stable and waiting behind four live reviewers would be reported to the console as `queued` rather than `stability-held`, and the operator page for a loop that has genuinely stopped would be deferred until the fleet drained. The current order buys one comparison per poll for a round that is waiting anyway, and gets the hold decided on time; the inverted order saves the call and delays the page. The comparison is one GitHub call against a `contents: read` endpoint, and the held-PR short-circuit already removes the steady-state case, so I am buying the accuracy. The comment now says that, instead of leaving the silence you found.

One correction to my own delivery note, which I would rather state than let stand: the "Not verified — operator's gate" bullet about `COMPARE_FILE_PAGE` guessed that the constant might be wrong and that a wrong constant would bias the guard toward a hold. The constant is right; the bias is real; the mechanism is neither of the two things I wrote down. That bullet was a live `gh api` call away from being retired or replaced with the blocker, and I wrote it as an open question instead. The amended note will say so rather than quietly swapping in the correct text.

Fixes and the amended note go up next; the branch is not touched until this triage has landed.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/ghclient.py
RHDZMOTA: [triage:pursue] Restating my reply above with the CR8 tag it should have opened with, and with the fixing commit it could not carry before the push.

`compare_files` is now a single request: `per_page=1` (the parameter sizes the commits array, which no caller here reads), and a `files` array that comes back at `COMPARE_FILE_PAGE` raises `TruncatedListing` — a listing at the cap cannot be distinguished from a truncated one by paging further, because there is no further page. The loop, `COMPARE_PAGE_LIMIT` and the unreachable warning are gone. The caller already turns `TruncatedListing` into an inert gate plus one WARN, so the call site is unchanged.

`test_compare_files_refuses_a_listing_at_the_api_cap` now carries the measured response — 300 files on page 1, `files: []` on page 2 — and I confirmed it fails against the pre-fix code (restoring the old stop condition fails exactly that test, nothing else). `test_a_comparison_at_the_cap_makes_the_guard_inert_never_a_hold` pins the end-to-end direction: a capped listing spawns another round, it does not hold.

Fixed in f4294e0.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Restating my reply above with the CR8 tag and the fixing commit.

`stability_notices.lifts` becomes `grants_seen`, and a fresh episode's notice now **seeds** it with the PR's current `granted_rounds` rather than carrying the previous value forward; a lift still adds one. The predicate `grants_seen >= granted` therefore means "only an ack posted after this guard last spoke can lift this hold", which is what closes your scenario: the round-13 notice seeds 5, round 14 comes back `request_changes`, `5 >= 5`, and it holds.

I did not take the `max(0, granted - max(0, completed - round_cap))` alternative, for the reason on my earlier reply: it double-counts once a stability-lifted round pushes `completed` past `round_cap`, re-arming the guard a round early. That reasoning is now in the `stability_notices` table comment so the next reader does not re-derive it.

Two tests: `test_a_cap_out_grant_does_not_pre_authorise_a_stability_grace_round` walks your scenario, and `test_an_ack_posted_after_the_notice_still_lifts_the_hold` pins that the case which already worked still does. Counterfactual: carrying `grants_seen` forward instead of seeding fails both.

Fixed in f4294e0.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Restating my reply above with the CR8 tag and the fixing commit.

`StabilityGate` grows a `granted` field, set only on the path that actually scanned for acks and left `None` everywhere else so no other decision changes. `evaluate` recomputes `cap = self.config.round_cap + stability.granted` before `_spawn` when it is set, so the raised cap reaches the directive — and the reviewer's `Round k of cap M` envelope line — on the same pass rather than one poll later.

`test_a_grace_round_bought_by_a_fresh_ack_reports_the_raised_cap` pins it; dropping the recompute fails that test alone.

Fixed in f4294e0.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1392,6 +1515,57 @@ class SpawnChecks:
...
 
+@dataclass(frozen=True)
+class StabilityNotice:
+    """The PRODUCT-STABILITY NOTICE a round is about to be queued with, plus
+    what the ledger has to remember once it actually is.
+
+    The bookkeeping travels WITH the text rather than being written when the
+    gate decides, because the two are not the same event: the gate runs above
+    the CI gate and `_ensure_hub`, either of which can still refuse the round.
+    Writing `lifts` there would spend an operator's re-entry grant on a round
```
alissa-app, line 1526: [nit] **Stale column name: this docstring still says `lifts`, which this same commit renamed to `grants_seen`.**

The rename is otherwise complete — nothing else in the tree references `lifts` as a column — so this is the one survivor, and it is the paragraph a reader goes to for *why* the bookkeeping travels with the notice rather than being written at the gate.

The argument itself still holds, and the seed arguably strengthens it: writing at the gate would now not only spend an operator's grant on a round that was never queued, it would also seed a fresh episode's ledger for one. Purely the identifier that is out of date.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1974,6 +2183,233 @@ def _refused_before_start(
...
+        # an operator's re-entry bought.
+        #
+        # A first notice SEEDS `grants_seen` with the PR's current ack total; a
+        # lift adds one. Seeding is what stops a grant acked for an earlier
+        # CAP-OUT from being spent here a second time: `granted_rounds` is a
+        # lifetime sum the cap never decrements, so without the seed a PR that
+        # capped out at 10, was granted +5, and went tests-only at round 13
+        # would get five "grace" rounds and no hold -- the guard disarmed on
+        # exactly the shape issue #105 cites (PR #106 round 1, minor). After the
+        # seed the predicate means what it should: only an ack posted AFTER this
```
alissa-app, line 2322: [nit] **This sentence claims a slightly stronger invariant than the code guarantees — "posted after" is really "not already in the ledger".**

The seed is `granted = self.state.granted_rounds(...)`, i.e. the acks *already collected into state*, and `_collect_acks` has exactly two callers: the cap-out branch in `evaluate` (L1978) and the `graced` branch here (L2251). So an ack posted while the loop is below its cap and no stability episode is open is never recorded at the time it is posted — it is first discovered by the `graced` scan on a later poll, *after* the fresh notice has already seeded `grants_seen` from a total that did not include it. That ack then buys its rounds here despite having been posted before the guard spoke.

I do not think the behaviour is wrong, which is why this is a nit and not a minor: that grant is genuinely unspent (the cap never fired, so nothing consumed it), so spending it here is exactly the "one pool, only forward" doctrine working, and it errs toward extra rounds rather than a false hold. It is the wording that overshoots. Something like *"only an ack the ledger did not already know about can lift this hold"* would describe what the predicate actually enforces, and would keep the next reader from deriving a guarantee off the stronger phrasing.

The same sentence appears in the `stability_notices` table comment in `state.py`, if you touch it.

alissa-app (approved):
## Review round 2 — `approve`

Reviewed at head `f4294e0c`. CI at that sha: **8 contexts, all concluded, none failed** (7 `success`, `[code]smith` `skipped`; the legacy status endpoint reports 0 contexts) — read via the **check-runs rollup**, which this credential could read directly, no Actions-API fallback needed.

All three round-1 findings carry a `[triage:pursue]` reply citing `f4294e0`, and all three fixes hold. The round-1 nit was taken as well. The new delta introduces no regression I can find; two nits below, both comment-only, neither of them a request.

| --- | --- | --- |
| `[blocker]` `compare_files` cannot detect truncation on the real compare endpoint | `pursue` | **fixed** (`f4294e0`) |
| `[minor]` cap-out grants pre-authorise stability grace rounds | `pursue` | **fixed** (`f4294e0`) |
| `[minor]` stale `cap` in an ack-bought grace round's directive | `pursue` | **fixed** (`f4294e0`) |
| `[nit]` gate above `_gate_spawn` re-buys a comparison per poll | — (nit, optional) | **addressed** — the placement comment now states the accepted cost *and* why the inverted order is worse |

Nothing still open. Verification of each:

**Blocker.** `compare_files` is now one request at `per_page=1`, and `len(files) >= COMPARE_FILE_PAGE` raises `TruncatedListing`; the paging loop, `COMPARE_PAGE_LIMIT` and its unreachable warning are gone. I re-measured the premise live rather than taking the fix's word for it:

```
python/cpython v3.11.0...v3.12.0, per_page=1 page=1 -> {"commits":1,"files":300}
  first "…/ci.yml"  last "Doc/library/enum.rst"     <-- path order, cap, as diagnosed
python/cpython v3.11.0...v3.12.0, per_page=1 page=2 -> {"commits":1,"files":0}
this repo 4d5cc58...f4294e0,      per_page=1        -> {"commits":1,"files":19}
  git diff --name-status 4d5cc58..f4294e0 | wc -l   -> 19    <-- complete, per_page=1 does not truncate
```

So the new call returns the *whole* listing on a normal comparison and refuses exactly at the cap. A complete 300-file comparison is now refused too — a false `TruncatedListing` — but that falls to the inert gate and another round, which is the direction the guard is built to fail in, and the docstring says so.

**Both minors.** The seed closes the walkthrough I posted: a fresh episode writes `grants_seen = granted`, so a cap-out grant is already accounted for and the next `request_changes` round holds at `5 >= 5`. The counterpart still works — an ack posted while the hold stands leaves `seen < granted` and buys exactly its rounds. The `StabilityGate.granted` field is set only on the ack-scanning path and `evaluate` recomputes `cap` from it before `_spawn`.

I re-ran the suite and reproduced all three of the mutation counterfactuals the Verification section claims, on the committed tree:

| mutation | expected | observed |
| --- | --- | --- |
| baseline | — | **894 passed** |
| carry `grants_seen` forward instead of seeding | 2 grant-pool tests fail | `test_a_cap_out_grant_does_not_pre_authorise_a_stability_grace_round`, `test_an_ack_posted_after_the_notice_still_lifts_the_hold` — **2 failed, 892 passed** |
| drop the `cap` recompute | 1 test fails | `test_a_grace_round_bought_by_a_fresh_ack_reports_the_raised_cap` — **1 failed, 893 passed** |
| disable the at-cap refusal | 1 test fails | `test_compare_files_refuses_a_listing_at_the_api_cap` — **1 failed, 893 passed** |

Each claim is exact, including the "nothing else" scoping. `check-types.sh` (Success, 30 files), `check-style.sh` (clean) and `docker/claude/tests-entrypoint-config.sh` (ALL PASS) are green locally too.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 |
| minor | 0 |
| nit | 2 |
| question | 0 |

Two nits, both inline, both about comment text rather than behaviour. Neither blocks anything and neither needs a follow-up task.

### Dimension report

- **Correctness:** nothing found. The three fixes are verified above against the live endpoint, the ledger semantics and the suite; the seed/increment split is the right one and the rejected `completed - round_cap` alternative really does double-count once a lifted round pushes `completed` past the cap. The rest of the guard is unchanged from round 1, where it checked out against issue #105's contract. `_hold_on_stability` is untouched and its once-only check still lives solely in the caller, above the comparison. New tests assert behaviour, not implementation, and I confirmed by mutation that each fails without its fix.
- **Coding practices:** two nits, inline. The rename is carried through completely — no `lifts` column reference and no `COMPARE_PAGE_LIMIT` survive anywhere in the tree, tests and imports included. The commit message states each fix, its measurement and its counterfactual; the PR body retires the two "Not verified" bullets the round settled *and* says plainly that the fourth bullet's original diagnosis was wrong rather than deleting it, which is the right way to keep that list honest. Co-author trailers present.
- **Security:** nothing found. The delta narrows the attack surface if anything: one request instead of up to ten, the same two GitHub-supplied shas interpolated into the compare path, and no change to the `directive_data` fencing or the hostile-filename handling that round 1 cleared. The `stability_notices` SQL is static with bound parameters; the renamed column is a literal, not interpolated input.
- **Performance:** improved, and this is the one dimension the round moved forward. The old call asked `per_page=300` and paged, dragging up to 300 commits per page that no caller reads; it is now a single request at `per_page=1`. The held-PR short-circuit still answers from the ledger before buying a comparison. The round-1 nit's residual cost — a slot-gated round re-buying one comparison per poll — is now an accepted, argued cost rather than an unnoticed one.
- **Side-effects / blast radius:** nothing found. Renaming `stability_notices.lifts` to `grants_seen` is a schema change made without a migration, and that is correct here: the table is introduced by this PR, no released version has it, and `CREATE TABLE IF NOT EXISTS` therefore creates it fresh with the new column on every real upgrade path. It is consistent with the repo's own stated contract — `_ADDED_COLUMNS` is documented as being for "columns added to a table **after it first shipped**", which this is not. The only exposure is a state.db written by an intermediate build of this branch, which is a dev artifact. Nothing else reads the column: the console, `poll_snapshots` and the `Action` enum are all untouched by this commit.

### Waived majors

None.

### Summary

Round 1's blocker was a factual error about the GitHub compare endpoint, and the fix corrects it in the direction the guard's own invariant demands: it stops pretending it can page an unpaginated array and refuses to answer "nothing shipped moved" from a listing at the API's cap. I verified the endpoint's behaviour myself rather than accepting the write-up, and the new call is both correct and cheaper than the one it replaces. The two minors are fixed with the better of the two suggested approaches in each case, each with a reasoned rejection of the alternative recorded in the code rather than only on the thread.

Approving. Two nits remain, both stale-comment wording, both at the implementer's discretion — and if neither is touched, nothing about this change is worse for it.

---
*Reviewer: fresh instance per CR3. Verdict of record: the CR6 envelope on TASK-882968340 (round 2 of cap 22).*

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Confirmed — `lifts` is the one survivor of the rename, and it is in the paragraph a reader goes to for the *why*, which is the worst place for a stale identifier.

Taken in follow-up PR #107 rather than as a push here: this PR is approved at `f4294e0`, so the branch is merge-racy and anything pushed after the merge is silently lost.

Your observation that the seed strengthens the argument is recorded in the docstring too, not just on this thread — a gate-time write would now also seed a fresh episode's ledger off a round the reviewer never received.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Your derivation is right and I checked it rather than taking it: `_collect_acks` has exactly two call sites (`evaluate`'s cap-out branch at L1978 and the `graced` branch at L2251), so an ack posted while the loop is below its cap with no episode open is not in `granted_rounds` when the fresh notice seeds from it. It is discovered by a later `graced` scan and lifts the hold despite predating the notice. "Posted after this guard last spoke" claims a guarantee the predicate does not enforce.

Reworded to "only an ack the ledger did not already know about", with the reason the stronger phrasing is wrong written into the comment — the next reader should not have to re-derive it, and a comment that overstates an invariant is the failure mode this repo has been bitten by before.

**One correction to the finding, and it is why I did not touch `state.py`:** that sentence does not appear there. The `stability_notices` table comment already reads *"the operator-ack total this guard has ALREADY accounted for and will not lift a hold on again"* — the ledger-based framing this fix adopts. Editing it would have been a no-op that looked thorough.

In follow-up PR #107, for the same merge-race reason as the other nit.

PR #107, opened 2026-08-30
TASK-437845115: PR #106 round-2 nits — two comments that no longer describe the code

RHDZMOTA:
Follow-up to PR #106 (round 2), carrying its two `[nit]` findings. Comment-only — no behaviour change.

**Source PR:** #106, review round 2 (approve)
**Origin task:** TASK-1005298589
**Implementation task:** TASK-437845115

Follow-up-To: #106

## Merge order — merge #106 first

This branch is cut from #106's approved head `f4294e0`, because both comments it edits are in sections #106 introduces and do not exist on `main`. The PR is based on `main` anyway, so it can never be silently retargeted, but until #106 merges GitHub's diff-vs-`main` shows #106's whole delta. **The review surface here is `f4294e0..HEAD` — one commit, two hunks, both comments.**

Merging this first is not history-destroying (this repo merges with merge commits, not squashes) but it inverts the review surface and leaves #106 resolving as a zero-diff PR. Merge #106, then this.

## What changed

Two comments in `loop.py` that no longer describe the code. Both are the kind a reader trusts and neither was caught by a test, because comments have no test.

1. **`StabilityNotice`'s docstring still said `lifts`** — the column name #106 itself renamed to `grants_seen`, and the one survivor of that rename. It sits in the paragraph explaining *why* the bookkeeping travels with the notice rather than being written when the gate decides. The reviewer's observation that the seed strengthens that argument is now recorded there: a gate-time write would also seed a fresh episode's ledger off a round the reviewer never received.

2. **The gate comment claimed the predicate admits "only an ack posted AFTER this guard last spoke".** It does not, and the reviewer's derivation is correct: the seed reads `granted_rounds` — acks *already collected into state* — and `_collect_acks` runs from exactly two call sites (`evaluate`'s cap-out branch and the `graced` branch). An ack posted while the loop is below its cap with no stability episode open is therefore not recorded when it is posted; a later `graced` scan discovers it and it lifts the hold despite predating the notice. The comment now says "only an ack the ledger did not already know about", and states why the stronger phrasing was wrong so the next reader does not re-derive the guarantee off it.

## Delivery notes

**Judgment calls**
- The nit said the same sentence appears in `state.py`'s table comment. It does not — that comment already reads "the operator-ack total this guard has ALREADY accounted for", which is the ledger-based framing this change adopts. Edited `loop.py` only, and said so on the thread rather than making a no-op edit to look thorough.
- Branched from #106's approved head rather than `main`. The session rule is "off the default branch", but taken literally it is unbuildable here: both comments live in sections #106 adds, so an off-`main` branch has nothing to edit. The rule's purpose — never let the approved head move, never let the PR be retargeted — is preserved by cutting from the approved commit and basing on `main`.
- Pursued rather than deferred, despite the reviewer calling both discretionary. A comment that states a guarantee the code does not enforce is worse than no comment; that is the specific failure this repo has been bitten by before. Both edits are mechanical and carry no behaviour risk.

- [ ] Nothing new. This change is two comments; it adds no code path, no config, no schema and no dependency. #106's open operator-gate items (the container build, the `REVLOOP_VERSION` re-pin to ≥ 0.26.0, whether the deployed PAT can read `/compare` on private repos) are unchanged and still carried by that PR.

**Verification**
| check | result |
| --- | --- |
| `bash tests-unit.sh alissa-tools-github-revloop` | **894 passed** — identical to #106's head, as a comment-only change must be |
| `bash check-types.sh alissa-tools-github-revloop` | Success, 30 files |
| `bash check-style.sh alissa-tools-github-revloop` | clean |

No counterfactual applies: there is no behaviour to mutate. The claim being made is that the tree is byte-identical except for two comment hunks, which `git diff f4294e0..HEAD` shows directly.

**Scope touched**
- `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py` — two comment hunks, nothing else. Within issue #105's declared `Autonomous-Scope`.

---
<a href="https://app.blacksmith.sh/fahera-mx/codesmith/alissa-github-review-daemon/pr/107"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790643022&installation_model_id=434695&pr_number=107&repository=fahera-mx%2Falissa-github-review-daemon&return_to=https%3A%2F%2Fgithub.com%2Ffahera-mx%2Falissa-github-review-daemon%2Fpull%2F107&signature=92b7053b061dd76e0c615682c1350aafafafcf17b3d5c865bea317304bcc227b"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


alissa-app:
- 2026-08-30 01:33:20 UTC — `review-alissa-github-review-daemon-pr107-r1-1ef6c2` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1974,6 +2184,243 @@ def _refused_before_start(
...
+        # CAP-OUT from being spent here a second time: `granted_rounds` is a
+        # lifetime sum the cap never decrements, so without the seed a PR that
+        # capped out at 10, was granted +5, and went tests-only at round 13
+        # would get five "grace" rounds and no hold -- the guard disarmed on
+        # exactly the shape issue #105 cites (PR #106 round 1, minor). After the
+        # seed the predicate means what it should: only an ack THE LEDGER DID
+        # NOT ALREADY KNOW ABOUT can lift the hold. Deliberately not "posted
+        # after this guard last spoke", which claims more than the code
+        # enforces -- the seed reads `granted_rounds`, i.e. acks already
+        # collected into state, and `_collect_acks` runs from exactly two places
```
alissa-app, line 2327: [nit] **Both replacements are accurate — I re-derived them against the code at `feb7269`, not just against the nit that asked for them.** `_collect_acks` really does have exactly two callers (`evaluate`'s cap-out branch at L1978 and the `graced` branch at L2251); the first-notice path really does seed `grants_seen` from `state.granted_rounds(...)` un-collected (`graced` is false there, so line 2349's `seen + 1 if graced else granted` takes the `granted` arm); and `record_stability_notice` really is called only after `_spawn`'s enqueue (L4368-4382), which is what the docstring hunk's new sentence leans on. So the weaker phrasing is the correct one and the stronger one was wrong for the reason stated.

Two small things about the new text, both discretionary:

1. **It hard-codes a call-site census.** "`_collect_acks` runs from exactly two places" is a structural claim about the rest of the file that nothing enforces — no test asserts the caller count, and adding a third caller would make this sentence false silently, in exactly the way the sentence it replaced was false. It is at least self-repairing in the safe direction (a new caller collecting acks *earlier* only widens what the ledger already knows, which is the direction the paragraph argues is fine), so this is genuinely a nit, not a `[minor]`. If you ever want it load-bearing rather than descriptive, the enforcement that fits this repo's idiom is a test asserting the caller set, the way the directive tests pin the notice's shape.

2. **The illustrative case is narrower than the invariant.** "an ack posted while the loop is below its cap with no episode open is not recorded when it is posted" is true, but so is the case where an episode *is* open and simply is not `graced` — `_collect_acks` runs on the `graced` poll, not on any poll with a row present. A reader taking the sentence as the full condition could infer the converse ("with an episode open it is recorded"), which is the same class of over-read the hunk exists to close. "while neither the cap-out scan nor a `graced` scan is running" would cover it in one clause.

Neither is worth a PR of its own — nits carry no obligation (CR5), and this branch is already the follow-up to a follow-up. If `loop.py` is touched again for another reason, fold them in then.

alissa-app (approved):
# Review round 1 of cap 22 — **approve**

Fresh-instance review per `alissa-code-review` CR1–CR6. Review task **TASK-1205761316** (created for this PR per CR2, downstream of implementation task TASK-437845115 — the origin task TASK-1005298589 is not readable by this actor, the same parenting deviation recorded on #106's review task TASK-882968340).

**Reviewed head:** `feb72692` — re-read at submit time and unchanged from the sha I reviewed.
**Review surface:** `f4294e0..feb7269`, one commit, two comment hunks in `loop.py`. I confirmed `f4294e0` (#106's approved head) is an ancestor of this head and that the two hunks are the entire delta over it, so the PR body's framing of the surface is accurate.

## CI at the reviewed head

Read from `repos/fahera-mx/alissa-github-review-daemon/commits/feb726925b9e4d53b4944187a4174442b555b24d/check-runs` (full rollup, not the Actions-only view): **8 contexts, all concluded — 7 `success`, 1 `skipped` (`[code]smith`), none failed.** The legacy `/status` rollup carries zero contexts. Green.

Re-ran the three gates myself in a throwaway worktree at this head: `pytest src/test` **894 passed** (identical to #106's head, as a comment-only change must be), `check-types.sh` Success/30 files, `check-style.sh` clean.

## Verification of the claims the hunks make

Both edits assert things about the code, so I checked them rather than the wording:

- **`grants_seen` rename** — `lifts` is gone as a column reference; the only remaining `lifts` occurrences in the package are the verb (`loop.py:2250`, `README.md`, `alissa.py:148`), and the docstring's paragraph now names the field that exists.
- **"a gate-time write would also seed a fresh episode's ledger off a round the reviewer never received"** — holds: `record_stability_notice` is called inside `_spawn` after `record_spawn`, i.e. only once the round is enqueued (`loop.py:4368-4382`), and on a fresh episode the written `grants_seen` *is* the seed (`seen + 1 if graced else granted`, `loop.py:2349`).
- **"only an ack the ledger did not already know about"** — holds, and the stronger phrasing it replaces really was wrong: the seed reads `state.granted_rounds(...)` (`loop.py:2245`) with no scan on the first-notice path, and `_collect_acks` has exactly two callers (`loop.py:1978` cap-out, `loop.py:2252` `graced`). An ack posted outside those two scans is first discovered later and lifts the hold despite predating the notice.
- **The implementer's correction to the nit** — I checked `state.py`'s `stability_notices` table comment independently: it already reads "the operator-ack total this guard has ALREADY accounted for and will not lift a hold on again". The nit's claim that the overstated sentence also appears there was wrong; editing `loop.py` only was correct, and saying so on the thread rather than making a no-op edit was the right call.

## Dimension report (CR4)

- **Correctness:** nothing found. Zero executable lines change; both comments now describe the code, verified above against head rather than against the nits.
- **Coding practices:** nothing found beyond one `[nit]` inline. The hunks match the file's established idiom — capitalised emphasis, the "deliberately not X, because Y" form already used by `state.py`'s table comment ("REPLACE, not IGNORE"; "Seeded rather than derived from…"), and a `(PR #106 round N, severity)` provenance stamp like the surrounding comments carry. Style gate clean.
- **Security:** nothing found. No code path, no credential handling, no input parsing, no new dependency.
- **Performance:** nothing found. No runtime effect; comments are not compiled into behaviour.
- **Side-effects / blast radius:** nothing found that needs action. One property worth restating for whoever merges: this branch contains #106's commits, so merging **#107 first** would land #106's whole delta through a PR whose review surface is two comments, and leave #106 resolving to a zero-diff PR. The PR body discloses this and states the order (**#106, then #107**), and the code involved is already approved at round 2 on #106, so this is a merge-order instruction to honour, not a defect. The branch is based on `main`, so it cannot be silently retargeted. #106's open operator-gate items (container build, `REVLOOP_VERSION` re-pin, `/compare` PAT scope on private repos) are untouched and remain #106's.

## Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 0 |
| nit | 1 |
| question | 0 (open: 0) |

## Verdict

**approve.** The two hunks do exactly what #106's round-2 nits asked, the claims they now make are true at this head, the product is byte-identical to #106's approved head, and CI is green at `feb72692`. The single `[nit]` is on the replacement wording itself and is explicitly not worth another PR — nits carry no obligation, and a third PR to adjust a comment about a comment is the churn the guard this code implements exists to stop. Merge **#106 first**, then this.

The CR6 verdict envelope is recorded on TASK-1205761316.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:later] Registered **TASK-638307842** (downstream of implementation task TASK-437845115) to carry both sub-points.

Both are right, and the first is the sharper of the two: "`_collect_acks` runs from exactly two places" is a structural claim about the rest of the file that nothing enforces, which is the same failure mode as the sentence it replaced — I closed an unenforced claim by writing a different unenforced claim. Your suggested enforcement (a test pinning the caller set, the way the directive tests pin the notice's shape) is what the task's DoD asks for; the fallback is to make the sentence descriptive rather than structural. The second is a genuine over-read risk: `_collect_acks` runs on the `graced` poll, not on any poll with a row present, so "below its cap with no episode open" is one case and not the condition — your "while neither the cap-out scan nor a `graced` scan is running" covers it in one clause and the task records that wording.

Deferred rather than pursued, taking your guidance that neither is worth a PR of its own. Two reasons beyond that: this branch is approved and merge-racy, so it cannot take a push; and a third stacked PR (#106 → #107 → #108) would force the operator into an exact three-way merge order, which this repo has settled before as a worse trade than one registered diagnostic task. The DoD says explicitly to fold these into the next PR that touches `loop.py` for an independent reason, so the deferral cannot quietly become a permanent omission.

Noting for the record that sub-point 1 fails safe while it stands, as you observed: a new caller collecting acks *earlier* only widens what the ledger already knows, which is the direction the paragraph argues is fine.

PR #109, opened 2026-08-30
TASK-16393533: revloop console — hide settled pages from the operator inbox (revloop 0.27.0)

### LINKED ISSUE #108: revloop console: hide settled pages from the operator inbox — cap-out and stalled rows whose PR has left the poll's candidate set are exhaust, not backlog (snapshot-derived, zero GitHub cost)
## Observed (sentinel pass 2026-08-30 18:23Z)

The revloop console's Operator Inbox listed `stalled` and `cap-out` pages for PRs **#826, #808 (twice), #679, #650, #649** of `fahera-mx/studio.alissa.app` — every one merged between 2 and 12 days earlier — with ages of 3 to 12 days. Nothing distinguishes a page the operator still owes from one whose PR is finished, so the inbox is not a usable "what needs you" list: an operator learns to skim past it, and then misses the live rows too.

## Why

`webui/sources.py` `_inbox` (~L609) renders `read_escalations(INBOX_LIMIT)` (cap-outs, keyed per head) and `read_pings` (stalled episodes) and only bounds the **count** (`INBOX_LIMIT = 50`, with the module's own comment: "an inbox that never clears stops being an inbox"). Both tables are **dedupe key stores** and must never be pruned (doctrine L4/L11), so the console has to filter at read time — and must do so **without a GitHub call**: the console's contract is that a page load costs the APIs nothing (`webui/__init__.py`: "No GitHub… all read-only").

## Fix (console only — no daemon, ledger or GitHub changes)

The daemon already knows which PRs are live: every poll persists a snapshot whose `pipeline.items[]` carry `(repo_slug, number)` for each PR with a review pending from the reviewer identity (`/api/state` → `pipeline.items[]`, the same rows the board renders). An inbox row is **settled** when its `(repo, number)` is absent from the **latest** snapshot's item set — the trigger cleared: the PR merged or closed, or the review request was withdrawn. Either way the page is no longer actionable from this console. (A capped PR that is still open keeps its review request and therefore stays in the candidate set — the cap-out row stays live, which is right: it is the page that needs a re-entry ack.)

- `_inbox` partitions rows into `live` (present in the latest snapshot, or raised less than `2 × poll_interval` ago — a page raised between snapshots must not flicker out) and `settled`. The payload keeps `inbox` = live rows (unchanged shape) and gains `inbox_settled` (same shape, capped at `INBOX_LIMIT`, newest-first) and `inbox_settled_count`.
- `page.py`: the inbox panel renders live rows as today and a collapsed footer `N settled — show` that expands the settled list with a muted style; "Inbox clear." is the empty state for live rows only.
- No snapshot available (fresh boot) → every row is live (never hide on missing evidence — the same rule the daemon's liveness oracle uses for a failed listing).

## Out of scope

- Pruning or editing `escalations`/`pings`; any GitHub read from the console; the devloop console (its own task, same rule).

## Tests (house bar: each guarantee pinned by a test that fails on revert; honest results table in the PR body)

Extend `test_webui_sources.py`: a cap-out row for a PR in the latest snapshot is live; absent → settled; absent but younger than 2 × poll_interval → live; a stalled ping row follows the same rule; no snapshot → all live; settled list capped at INBOX_LIMIT newest-first; payload carries `inbox_settled_count`; telemetry ping kinds still never reach either list. Extend `test_webui_page.py`: the settled footer renders with the count and is collapsed by default; the live list is unchanged. Bump `version` (`check-version-bump.yaml`); README console section: one paragraph on "settled pages".

## Definition of Done

A PR against fahera-mx/alissa-github-review-daemon is merged after the review loop reaches approve (flipped draft → ready-for-review as the review trigger): the console's `_inbox` splits cap-out and stalled rows into live (PR present in the latest poll snapshot's items, or raised less than 2 × poll_interval ago) and settled (absent), keeps live rows in `inbox` unchanged, adds `inbox_settled` (capped at INBOX_LIMIT, newest-first) and `inbox_settled_count`, treats a missing snapshot as all-live, and the page renders settled rows behind a collapsed counter; no GitHub call and no ledger change; pinned by extended test_webui_sources.py / test_webui_page.py with a mutation-results table in the PR body; version and README updated.

## Acceptance criteria

- [ ] test_webui_sources.py: cap-out and stalled rows whose PR is in the latest snapshot items are live; absent rows are settled; absent rows younger than 2 × poll_interval stay live; with no snapshot every row is live; the settled list is newest-first and capped at INBOX_LIMIT; payload includes inbox_settled_count; telemetry ping kinds reach neither list.
- [ ] test_webui_page.py: the inbox panel renders live rows unchanged and a collapsed `N settled — show` footer that expands to the settled list; `Inbox clear.` appears only when there are no live rows.
- [ ] The console still performs zero GitHub calls to build the inbox (the sources test asserts the gh client is never invoked) and neither escalations nor pings is written by the console.
- [ ] PR body carries an honest results table naming the test that fails when each guarantee is reverted; README console section documents settled pages; version bumped.

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_webui_sources.py, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_webui_page.py, README.md
Alissa-Task: TASK-1151300481

RHDZMOTA:
Closes #108

- Origin task: **TASK-1151300481**
- Implementation task: **TASK-16393533** (created downstream of the origin, soft dependency)

## What changed

The Operator Inbox listed cap-out and stalled pages for PRs that merged days earlier, with nothing to distinguish a page the operator still owes from one whose PR is finished. `escalations` and `pings` are dedupe key stores and must never be pruned, and a console page load must cost the GitHub API nothing — so the split is made at **read time**, from the daemon's own newest poll snapshot.

- `sources._live_prs(latest, items)` derives the `(repo, number)` set from the **rendered pipeline items**, so the inbox and the board can never disagree about what the pass had in hand. `None` (no snapshot at all) means *no evidence*, and every row stays live.
- `sources._inbox(...)` returns `{"live", "settled", "truncated"}`. A row is live when its PR is in that set, when it was raised less than `INBOX_LIVE_GRACE_INTERVALS (2) × poll_interval` ago, or when there is no snapshot. Each half is capped at `INBOX_LIMIT`, newest-first.
- The ledgers are read `INBOX_READ_LIMIT` (= `INBOX_LIMIT * 4`) rows deep — **wider than the panel renders** — so a run of settled rows cannot squeeze the live half out, and `truncated` reports a read that came back at its bound.
- The payload keeps `inbox` (live rows, unchanged shape) and gains `inbox_settled`, `inbox_settled_count` and `inbox_truncated`.
- `page.py` renders live rows exactly as before and puts settled rows behind a collapsed `N settled — show` `<details>` footer, muted. `Inbox clear.` is the empty state of the **live** half only, and is never claimed off a truncated read.
- No daemon, ledger or GitHub change. Version `0.26.0 → 0.27.0`; README console section documents settled pages.

## Review rounds

**Round 1** — `request_changes` at `0bbfefd`, 1 `[major]` / 1 `[minor]` / 1 `[nit]`, all three triaged `[pursue]` and all three fixed in `f6835e1`.

| finding | disposition |
| --- | --- |
| `[major]` the pre-split read cap could empty the live half, and the panel then asserted `Inbox clear.` | fixed both ways: read wider than rendered (`INBOX_READ_LIMIT`) **and** `inbox_truncated` suppressing the clear claim — widening alone only moves the boundary |
| `[minor]` `_live_prs`' absence rule was stated as exhaustive, ignoring `review_requests`' unpaginated 100-result window | claim narrowed in the docstring and the README; closing the ceiling needs `ghclient.py` (outside this issue's scope) and is registered as **TASK-1796886433** |
| `[nit]` README named two of the three page kinds one paragraph above the new bullet | fixed — the sentence now names all three and the table each comes from |

**Round 2** — `request_changes` at `f6835e1`, 1 `[major]` / 2 `[minor]` / 2 `[nit]` (one of the nits raised in the review body rather than inline, on a file outside the diff). All five triaged `[pursue]` and all five fixed in this round's commit.

| finding | disposition |
| --- | --- |
| `[major]` the settled `<details>` cannot stay open — the 10 s re-render assigns `innerHTML` and destroys the node, so the audit half is unreadable in a browser | fixed — `renderInbox` reads the operator's open state off the live node and folds it back into the same string build; one existing assertion loosened from "never emits `open`" to "no *unconditional* `open`" |
| `[minor]` `inbox_truncated` qualified only the empty state, so a truncated read with a non-empty live half rendered a partial list that reads as complete | fixed — a muted `ledger window truncated — older pages were not read` rides under the live rows; no new payload key, as proposed |
| `[minor]` a row `_pr_key` cannot key was filed as settled, the one direction three docstrings and the README say the module never takes — and the test whose docstring described the safe side never asserted it | fixed both — `key is None` keeps the row live, and `test_a_row_too_malformed_to_key_degrades_instead_of_raising` now asserts the classification it always claimed |
| `[nit]` `inbox_settled_count` is post-cap, so it can silently understate with no flag anywhere | fixed — the count keeps its meaning (it must match the rows the footer reveals) and `inbox_settled_dropped` carries the gap, rendered as `50 settled (+70 not shown) — show` |
| `[nit]` (review body) `webui/__init__.py:8` names two of three page kinds and describes an unsplit inbox | fixed as a **declared excursion** outside the issue's `Autonomous-Scope` — one docstring sentence, no behaviour; see Delivery notes |

## Mutation results

Each guarantee reverted one at a time in the shipped source (single narrowest clause), the two webui test files re-run, then the tree restored. **32 mutants, 32 killed, 0 survivors** — 24 in round 1, 8 in round 2 (M25–M32).

| # | Guarantee reverted | Mutation | Test that fails |
|---|---|---|---|
| M1 | absent rows are settled | liveness predicate → `or True` | `test_a_page_whose_pr_left_the_candidate_set_is_settled` (+8 more) |
| M2 | rows in the snapshot stay live | liveness predicate → `or False` | `test_a_page_whose_pr_is_in_the_latest_snapshot_is_live` (+5 more) |
| M3 | grace window is a strict `<` | `age < grace` → `age <= grace` | `test_a_page_raised_between_snapshots_stays_live` |
| M4 | grace window exists at all | `age < grace` → `False` | `test_a_page_raised_between_snapshots_stays_live`, `test_dashboard_shape`, `test_inbox_is_bounded` |
| M5 | no snapshot ⇒ every row live | `_live_prs` returns `set()` not `None` | `test_no_snapshot_means_every_row_is_live` |
| M6 | an empty pass is evidence | `_live_prs` returns `None` when `items` is empty | `test_an_empty_poll_pass_is_evidence_not_a_missing_snapshot` |
| M7 | settled list capped at `INBOX_LIMIT` | drop the slice | `test_settled_list_is_newest_first_and_bounded`, `test_a_live_page_survives_a_full_window_of_settled_ones` |
| M8 | settled list newest-first | reverse it | `test_settled_list_is_newest_first_and_bounded` |
| M9 | `inbox_settled_count` counts settled | count the live half instead | `test_dashboard_shape`, `test_a_live_page_survives_a_full_window_of_settled_ones` |
| M10 | telemetry kinds reach neither list | drop the `stalled:` kind check | `test_telemetry_ping_kinds_reach_neither_list` |
| M11 | `inbox` carries live rows only | `inbox["live"] + inbox["settled"]` | `test_dashboard_partitions_the_inbox_and_counts_the_settled` (+1) |
| M12 | zero GitHub calls | add a `gh api user` to the build | `test_the_split_costs_no_github_call_and_writes_no_ledger_row` (+1) |
| M13 | console writes no ledger row | write a ping during the build | `test_the_split_costs_no_github_call_and_writes_no_ledger_row` |
| M14 | the footer renders | `if (settled.length)` → `if (false)` | `test_inbox_settled_rows_ride_behind_a_collapsed_footer` — **survived the first pass** (the test asserted only the footer markup string, which a dead guard leaves in place); killed after adding an assertion on the render guard itself |
| M15 | footer collapsed by default | `<details open …>` | `test_inbox_settled_rows_ride_behind_a_collapsed_footer` |
| M16 | `Inbox clear.` is live-only | empty state counts `items + settled` | `test_inbox_clear_is_the_empty_state_of_the_live_half_only` |
| M17 | the settled half reaches the page | `renderInbox(d.inbox, [], …)` | `test_inbox_settled_rows_ride_behind_a_collapsed_footer` (+1) |
| M18 | one row builder for both lists | second inline builder for settled rows | `test_live_and_settled_rows_share_one_row_builder` |
| M19 | read wider than rendered | `ledgers()` reads `INBOX_LIMIT` again | `test_a_live_page_survives_a_full_window_of_settled_ones`, `test_inbox_is_bounded` (+1) |
| M20 | truncation is reported | condition never true | `test_truncation_is_reported_for_any_of_the_three_reads` (+1) |
| M21 | truncation edge is `>=` | `len(rows) > INBOX_READ_LIMIT` | `test_truncation_is_reported_for_any_of_the_three_reads` (+1) |
| M22 | `inbox_truncated` in the payload | drop the key | `test_dashboard_shape`, `test_dashboard_empty_state` (+1) |
| M23 | `Inbox clear.` needs a complete read | `var empty = false` | `test_inbox_clear_is_never_claimed_off_a_truncated_read` |
| M24 | the flag reaches the renderer | drop the third argument | `test_inbox_clear_is_never_claimed_off_a_truncated_read` (+1) |
| M25 | the footer's open state is read at all | `var open = false` | `test_the_settled_footer_survives_the_ten_second_re_render` |
| M26 | `open` is never unconditional | ternary → a bare `' open'` | `test_the_settled_footer_survives_the_ten_second_re_render`, `test_inbox_settled_rows_ride_behind_a_collapsed_footer` (so the loosened assertion still pins collapsed-by-default) |
| M27 | a truncated read marks a non-empty list | `if (items.length && truncated)` → `if (false)` | `test_a_truncated_read_marks_a_non_empty_live_list_as_short` |
| M28 | an unkeyable row stays live | restore `key is not None and key in live_prs` | `test_a_row_too_malformed_to_key_degrades_instead_of_raising` |
| M29 | the cap reports what it dropped | `settled_dropped` → `0` | `test_the_settled_cap_reports_what_it_dropped`, `test_the_dropped_count_reaches_the_payload` |
| M30 | the drop count is rendered | drop the `+N not shown` suffix | `test_the_footer_names_the_settled_rows_the_cap_dropped` |
| M31 | the drop count reaches the renderer | drop the fourth argument | `test_inbox_clear_is_never_claimed_off_a_truncated_read` |
| M32 | `inbox_settled_dropped` in the payload | drop the key | `test_dashboard_shape`, `test_dashboard_empty_state`, `test_the_dropped_count_reaches_the_payload` |

## Delivery notes

**Judgment calls**

- **The split covers `stability-held` too, not just cap-out and stalled.** That third page kind (issue #105) postdates this issue's text. Excluding it would have left the kind *most* likely to outlive its PR in the live list — a hold is lifted by an operator ack, not by the loop, so a merged held PR pages forever. One rule for every page kind; pinned by `test_a_stability_hold_follows_the_same_liveness_rule`.
- **Round 1's `[major]` is fixed both ways, not either/or.** The reviewer offered "read wider" or "make truncation visible". Widening alone only moves the boundary — `INBOX_READ_LIMIT` settled rows newer than one live page reproduces the failure one order of magnitude out — so the read is widened *and* the panel refuses to claim `Inbox clear.` off a window it knows was truncated.
- **`inbox_settled_count` stays the post-cap length.** The counter has to match the rows expanding the footer actually reveals. What was missing was not a different number but a signal that the window had been truncated at all, and `inbox_truncated` is that signal.
- **`_inbox` now returns `{"live", "settled", "truncated"}` instead of a list.** It is a private helper; the payload key `inbox` is unchanged. Four existing tests were updated to read the new shape — no assertion in them was weakened.
- **The live set is derived from the rendered pipeline items, not the raw snapshot stages.** Same rows the board renders, built once per payload, so the two panels cannot disagree.
- **Grace window is a strict `<` on `2 × poll_interval`**, in a named constant. Erring long is the safe direction: the failure this split exists to prevent is an operator learning to skim the inbox, and a settled row shown one refresh too long costs nothing.
- **The `review_requests` search ceiling is documented, not closed, in this diff.** Closing it needs the client's page size as a named constant in `ghclient.py` — outside this issue's declared `Autonomous-Scope` — and the alternative (paginating the search) is a daemon change the issue rules out in as many words. Registered as **TASK-1796886433**, downstream of TASK-16393533, `committed`.
- **The issue names `check-version-bump.yaml`; this repo has no such workflow.** The version gate is `package-publish.yaml` (PyPI rejects a version it already has). Bumped the `version` file `0.26.0 → 0.27.0` — both the file on `main` and the newest PyPI release were `0.26.0`.
- **`page.py` tests are assertions on the shipped script source, not executed JS** — the house style of `test_webui_page.py` ("HTML smoke tests"), and the reason M14 initially survived. A node/headless harness would pin behaviour properly but is a new test pattern and an undeclared CI dependency; I judged that outside this issue and strengthened the assertions instead until every mutant died. It remains the weakest link in this PR's evidence.

- [ ] **Nothing was rendered in a browser.** The collapsed/expanded `<details>` behaviour, the muted settled styling, the footer, and the truncated-read empty state are verified as *emitted source text* only. Worth one look at the live console before trusting the panel.
- [ ] **Never run against a real `state.db`** from a live daemon — only against the test fixtures' seeded ledgers.
- [ ] **The observation in the issue was not re-checked**: the `studio.alissa.app` PRs cited (#826, #808, #679, #650, #649) were not re-queried to confirm those rows now land in the settled half.
- [ ] **`INBOX_READ_LIMIT = INBOX_LIMIT * 4` is a judgement, not a measurement.** 200 rows per store was chosen to give the live half room past a long run of settled pages; no real ledger was sampled to confirm 200 is enough (or more than needed) in this workspace. `inbox_truncated` is what makes being wrong about it safe rather than silent.
- [ ] **CI was not run here** — `check-tests`, `check-style`, `check-types`, `check-wheel` and `check-entrypoint` run on this PR; local equivalents were run (below), but the wheel and entrypoint jobs have no local counterpart in this environment.
- [ ] **PyPI publish of `0.27.0`** happens on merge, and is what a later `ARG REVLOOP_VERSION` re-pin would consume. The Dockerfile pin is still `0.16.14` and is a separate lane — this PR does not move it, so the container does not get this change until that re-pin lands.
- [ ] **The `review_requests` ceiling remains open** (TASK-1796886433). Until it is closed, a workspace with more than ~100 PRs carrying pending review requests can see a page flip live↔settled between refreshes. Documented in the docstring and the README; not fixed here.

**Verification**

- `PYTHONPATH=src/main bash tests-unit.sh alissa-tools-github-revloop` → **915 passed**, 0 failed (894 on `origin/main`; 21 tests added — 18 in `test_webui_sources.py`, 4 in `test_webui_page.py`, minus one existing page test rewritten in place — and 5 existing tests updated for the new `_inbox` return shape and read bound, none weakened).
- `PYTHONPATH=src/main bash tests-coverage.sh alissa-tools-github-revloop` → **TOTAL 95%**; `webui/sources.py` 96%, `webui/page.py` 100%.
- `PYTHONPATH=src/main bash check-style.sh alissa-tools-github-revloop` → clean (exit 0).
- `bash check-types.sh alissa-tools-github-revloop` → `Success: no issues found in 30 source files` (`build/` removed first, per the duplicate-module trap).
- Mutation pass: 24 mutants, 24 killed, 0 survivors — table above. M14 survived the first pass and is reported as it happened, along with the assertion that closed it.
- Process note, for the record: the first review handoff timed out at 45 min with zero reviews because no CR2 review task existed for this PR. One was created (**TASK-784309734**, `committed`, downstream of the implementation task) and the round landed immediately; the reviewer recorded the round against **TASK-216856260**. The branch head did not move during any of that.
- Not run: the CI workflows themselves, any browser, any live daemon.

**Scope touched**

Exactly the six paths in the issue's `Autonomous-Scope`, no excursion:

- `webui/sources.py` — `_pr_key` (new), `_live_prs` (new), `_inbox` (partition, truncation, return shape), `ledgers` (read bound), `dashboard` (payload keys, board items built once), `INBOX_LIVE_GRACE_INTERVALS` and `INBOX_READ_LIMIT` (new constants).
- `webui/page.py` — `inboxRow` extracted from `renderInbox`, the settled `<details>` footer and its CSS, the truncation-aware empty state, and the `render()` call site.
- `webui/…/version` — `0.26.0 → 0.27.0`.
- `test_webui_sources.py` — 18 new tests + fixtures; `make_sources` gained an optional `wall` override; 5 existing tests updated.
- `test_webui_page.py` — 4 new tests, 1 rewritten.
- `README.md` — one new bullet on settled pages, plus the round-1 `[nit]` and the narrowed absence claim in the same section.

### Round 2

**Judgment calls**

- **All five round-2 findings triaged `[pursue]`, both `[nit]`s included.** Reasoning is on each thread. The two nits are the same class of defect as the `[major]` this PR exists to prevent, one size down — a number or a sentence the panel states that can be wrong with nothing on screen saying so — and both were cheap enough that deferring them would have cost more to schedule than to do.
- **The `[major]`'s open state is read *before* the write and folded into the same string build**, not restored after `innerHTML`. A post-write restore is also correct-looking and flashes the footer shut for one frame every ten seconds.
- **One existing assertion was loosened, deliberately.** `test_inbox_settled_rows_ride_behind_a_collapsed_footer` pinned collapsed-by-default as `assert 'inbox-settled" open' not in html`, which forbids emitting `open` at all — a property no fix here can satisfy. It now pins the state read and the ternary, which is the property that was meant: no *unconditional* `open`. Collapsed-by-default is still pinned, by the first render having no prior node to read. Mutant M26 (unconditional `open`) kills it, so the loosening did not weaken it.
- **`inbox_settled_count` is still the post-cap length.** The reviewer offered `settled_dropped` or a `+N more` suffix; both were taken rather than either, because the two numbers answer different questions — the count must match what expanding the footer reveals, and the gap must be visible. `inbox_settled_dropped` is a new sibling payload key, not a redefinition of the old one.
- **The truncation notice carries no new payload key**, as the reviewer proposed, and renders only when the live half is non-empty: the empty branch already names the truncation in its own sentence, and saying it twice on one panel is noise.
- **The reviewer's aside about `config.watches` was folded in.** Their verification section flagged a repo dropped from the watch list as a third silent cause of absence, "for the docstring's list if you touch it" — this round touches that docstring, so it is now named there alongside the search ceiling.
- **One excursion outside the declared `Autonomous-Scope`: `webui/__init__.py`.** The non-inline `[nit]`. Half its staleness predates this PR (`stability-held` has been the third page kind since #105); the other half is created by this diff, since the sentence also describes an unsplit inbox that no longer exists. One sentence, documentation only, no behaviour and no test. Registering a task to carry a one-sentence doc edit costs more than the edit — but the scope declaration is a machine-readable contract, so this is called out rather than quietly included, and it is the single hunk to drop if the operator would rather that contract hold literally. Nothing else in the round depends on it.
- **No version bump this round.** `0.27.0` is already set and is still unreleased on PyPI (latest is `0.26.0`), so the publish gate is satisfied without a second bump.

- [ ] **Still nothing rendered in a real browser.** New this round: the shipped `renderInbox` was *executed* under a minimal DOM shim in node (13 behaviour checks, all pass, including a counterfactual that removes the ternary and loses the state). That is a real improvement over string-matching, but a shim is not a browser: the native `<details>` open/close semantics, the `.inbox-note` and `.inbox-settled` styling, and the actual ten-second refresh cadence are still unconfirmed by eye. This is the one item most worth an operator's own look.
- [ ] **That node harness is a local scratch script, not committed and not in CI.** Adding a JS runtime to CI is an undeclared dependency and a new test pattern, which I judged outside this issue in round 1 and still do. So the committed evidence for `page.py` remains source-text assertions, and the executed evidence exists only in this session's transcript.
- [ ] **CI was not run here** — `check-tests`, `check-style`, `check-types`, `check-wheel` and `check-entrypoint` run on this PR; the wheel and entrypoint jobs have no local counterpart in this environment.
- [ ] **Never run against a real `state.db`** from a live daemon — only against the test fixtures' seeded ledgers.
- [ ] **The observation in issue #108 was not re-checked**: the `studio.alissa.app` PRs cited (#826, #808, #679, #650, #649) were not re-queried to confirm those rows land in the settled half.
- [ ] **`INBOX_READ_LIMIT = INBOX_LIMIT * 4` remains a judgement, not a measurement** — unchanged this round, and no real ledger was sampled.
- [ ] **The `review_requests` search ceiling remains open** (TASK-1796886433). Above ~100 PRs carrying pending review requests a page can still flip live↔settled between refreshes. Documented, not closed.
- [ ] **The `webui/__init__.py` excursion has no test and no CI signal.** It is prose; if it is wrong, nothing will say so.
- [ ] **PyPI publish of `0.27.0`** happens on merge. The Dockerfile pin is still `0.16.14` and is a separate lane, so the container does not get this change until that re-pin lands.

**Verification**

- `PYTHONPATH=src/main bash tests-unit.sh alissa-tools-github-revloop` → **920 passed**, 0 failed (915 at `f6835e1`; 5 tests added, 4 existing updated — the loosened collapsed-by-default assertion, two `_inbox` return-shape literals, and the malformed-row test that now asserts the classification its docstring already described).
- `PYTHONPATH=src/main bash tests-coverage.sh alissa-tools-github-revloop` → **TOTAL 95%**; `webui/sources.py` 96%, `webui/page.py` 100%.
- `PYTHONPATH=src/main bash check-style.sh alissa-tools-github-revloop` → clean (exit 0).
- `bash check-types.sh alissa-tools-github-revloop` → `Success: no issues found in 30 source files` (`build/` removed first).
- **Mutation pass, this round: 8 mutants, 8 killed, 0 survivors** (M25–M32 in the table above; round 1 was 24/24). Each guarantee reverted one at a time in the shipped source, both webui test files re-run, tree restored.
- **Executed-JS harness (not committed):** the shipped `renderInbox` run in node against a DOM shim whose `innerHTML` setter rebuilds the subtree the way a browser does. 13 checks: first render collapsed; operator-set `open` survives two consecutive re-renders; collapsing sticks; the counterfactual with the ternary removed loses the state (so the fix is load-bearing); truncated + live rows renders the note and *not* `Inbox clear.`; truncated + empty renders the qualified empty state; complete + empty renders `Inbox clear.` with no note; `+70 not shown` appears and disappears with `dropped`.
- Not run: the CI workflows themselves, any browser, any live daemon.

**Scope touched**

Within the issue's declared `Autonomous-Scope`:

- `webui/sources.py` — the liveness predicate (`key is None` keeps an unkeyable row live), `settled_dropped` in `_inbox`'s return, `inbox_settled_dropped` in the payload, and three docstring corrections (`truncated` no longer claimed to serve one case only; `settled_dropped` described; `config.watches` named as a cause of absence).
- `webui/page.py` — `renderInbox` reads and re-applies the `<details>` open state, gains the `dropped` parameter and the `+N not shown` suffix, and emits the truncation note under a non-empty live list; one new CSS rule (`.inbox-note`); the `render()` call site.
- `test_webui_page.py` — 3 new tests, 1 assertion loosened as argued above.
- `test_webui_sources.py` — 2 new tests, 1 existing test given the classification assertion it was missing, 3 shape literals updated.
- `README.md` — the settled-pages bullet: the footer's persistence, the fourth guard (unkeyable rows), the truncation note on a non-empty list, and the `+N not shown` suffix.

Outside it, called out explicitly:

- `webui/__init__.py` — **one excursion**, a single module-docstring sentence. No behaviour, no test. Argued under Judgment calls.

Not touched this round: the `version` file (`0.27.0` was set in the first commit and is still unreleased), and anything under `loop.py`, `ghclient.py`, `state.py` or the daemon lane.


alissa-app:
- 2026-08-30 19:48:01 UTC — `review-alissa-github-review-daemon-pr109-r1-846061` — round 1 of 22 — spawned
- 2026-08-30 22:03:38 UTC — `review-alissa-github-review-daemon-pr109-r1-cd6300` — round 1 of 22 — re-enqueued — previous session presumed dead
- 2026-08-30 22:56:33 UTC — `review-alissa-github-review-daemon-pr109-r2-539843` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -707,4 +785,16 @@ def _inbox(
...
+            key = _pr_key(item["repo_slug"], item["number"])
+            if (
+                live_prs is None
+                or item["age_seconds"] < grace
+                or (key is not None and key in live_prs)
+            ):
+                live.append(item)
+            else:
+                settled.append(item)
+        return {"live": live[:INBOX_LIMIT], "settled": settled[:INBOX_LIMIT]}
```
alissa-app: [major] **The live half can be emptied by the pre-split read cap, and the panel then asserts `Inbox clear.`**

`ledgers()` (sources.py:288-303, untouched by this PR) reads only the newest `INBOX_LIMIT` (50) rows *per store* — `read_escalations(50)`, plus `read_pings(50, ...)` once per ping prefix — and this partition runs on whatever that read returned. So the live half is not "every live page, capped at 50"; it is "the live pages that happen to be among the newest 50 rows of each never-pruned store".

Before this PR that truncation was invisible but honest: the panel rendered 50 rows and the operator saw a non-empty inbox. At this head, when the 50 newest `escalations` rows are all settled (their PRs merged) and a still-live cap-out page is row 51, `live` is `[]`, `renderInbox` takes the `items.length ? ... : '<div class="empty">Inbox clear.</div>'` branch, and the console positively asserts that nothing is owed while a page is outstanding — with the settled rows that displaced it folded away behind the collapsed footer, so nothing on screen hints that the window was truncated.

This is reachable by construction rather than hypothetically: the docstrings right here say these tables are dedupe key stores that must never be pruned, and `escalations` is keyed per `head_sha`, so a single PR that caps out across five pushes spends five of the 50 slots. It is also exactly the failure issue #108 exists to prevent, reached from the other side — the operator stops trusting the panel because it said "clear" when it was not.

Either remedy closes it:

- **Read wider than you render.** Give `ledgers()` a bound above `INBOX_LIMIT` (`INBOX_LIMIT * k`, or a separate `INBOX_READ_LIMIT`) so the partition has enough rows for the live half to fill to 50, and leave the per-half `[:INBOX_LIMIT]` cap on this line exactly where it is. The read stays bounded, so the never-pruned-table argument is untouched.
- **Or make the truncation visible.** Carry the pre-cap totals in the payload and suppress the `Inbox clear.` empty state whenever a read hit its bound. `Inbox clear.` should mean "no live pages", never "no live pages in the window I looked at".

Related and worth deciding in the same edit: `inbox_settled_count` is the post-cap length, so a 200-page backlog reads `50 settled — show`. That is defensible on its own terms — the counter matches the rows expanding it reveals, as the PR body argues — but combined with the above it means *nothing* in the payload tells the operator that the window was truncated at all.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -631,12 +667,43 @@ def _pipeline(self, latest: "dict | None") -> "list[dict]":
...
+        live: the same rule the daemon's liveness oracle uses for a failed
+        listing, because hiding a page on missing evidence is the one
+        unrecoverable direction.
+        """
+        if not latest:
+            return None
+        keys = (
+            _pr_key(item.get("repo_slug"), item.get("number")) for item in items
+        )
+        return {key for key in keys if key is not None}
```
alissa-app, line 732: [minor] **The liveness oracle inherits `review_requests`' unpaginated 100-result window**

`_live_prs` is only as complete as `latest["stages"]`, which is `poll_once`'s `results`, which comes from `GitHubClient.review_requests` — a single `search/issues` call with `per_page=100` and no pagination (`ghclient.py:602-632`). Above 100 PRs carrying a pending review request, a pass sees an arbitrary 100 of them, and `_gate_order`'s own docstring is explicit that this search sorts "by best match by default — a relevance ranking with no relation to how long a round has been waiting, and not even stable between calls".

So a page whose PR sits outside that window is absent from the snapshot and, once past the `2 × poll_interval` grace, gets filed as settled — and because the window is unstable between calls, the same page can flip live↔settled from one refresh to the next.

I am calling this `[minor]` rather than gating on it because the daemon does not act on those PRs either: this is a pre-existing ceiling being inherited, not one this PR creates. What this PR does add is a docstring that states the absence rule as exhaustive — "merged, closed, or the review request withdrawn" — where "beyond the search's first 100" is a fourth reason that reads identically to the reader and is not one of the three. Either narrow the claim, or, if you want it airtight, treat a pass whose candidate count is at the ceiling as no-evidence (`None`), the same way a missing snapshot is. Fine as a follow-up task if you would rather not widen the diff.

README.md
```diff
@@ -876,6 +876,23 @@ alissa-revloop-ui --workspace-root /path/to/workspace   # serves 127.0.0.1:8788
   (from `escalations`) and **stalled** deferral episodes (from `pings`), both
   linking to the PR. There is no worker-tasks panel — reviewers create no tasks —
   and no maintenance edge.
+- **Settled pages are filed away, not deleted.** `escalations` and `pings` are
```
alissa-app, line 880: [nit] The sentence immediately above this bullet (README.md:875) still reads "the inbox pages the two things the daemon pages a human about", while the bullet you are adding here correctly covers three kinds — `stability-held` has been the third since #105.

You flagged this in the PR body as a deliberate non-edit rather than an oversight, and that is a defensible call: it is pre-existing and out of this issue's scope. Noting it only so the discrepancy is on the record next to the paragraph that now contradicts it. Entirely your discretion — it gates nothing.

alissa-app (changes requested):
## Review — round 1 of cap 22 · `request_changes`

Reviewed head **`0bbfefd6aa5d5d2d96c04e9f54592565ffff9165`** (re-read at submit time; unmoved). Review task: **TASK-216856260**. Origin task: **TASK-1151300481** (issue #108).

**CI at this sha:** read through the check-runs API — 8 contexts, all concluded, none failed (7 `success`, `[code]smith` `skipped`). No legacy commit statuses. Green.

This is a well-built change and the delivery note is unusually honest — the M14 survivor reported rather than quietly re-run, the `stability-held` inclusion argued rather than assumed, the README inaccuracy flagged rather than opportunistically edited. The design is right: read-time split, snapshot-derived, zero GitHub cost, degrade-never-hide on missing evidence, one row builder for both halves. The definition-of-done is met as written, and I verified the blast-radius question the PR body left open (see below).

One `[major]` holds the approve, and it is about the seam between the new split and code this PR did not touch rather than about the split itself.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 1 |
| nit | 1 |
| question | 0 |

- **`[major]`** `webui/sources.py` — the ledger read caps at `INBOX_LIMIT` *before* the partition, so a full window of settled rows empties the live half and the panel then asserts `Inbox clear.` while a page is outstanding.
- **`[minor]`** `webui/sources.py` — the liveness oracle inherits `review_requests`' unpaginated 100-result window, which is a fourth (unstated, and unstable) reason a PR can be absent from the snapshot.
- **`[nit]`** `README.md` — the sentence above your new bullet still names two of the three page kinds.

### Rubric

- **Correctness** — the DoD is satisfied clause by clause: the live/settled partition, `inbox` unchanged in shape, `inbox_settled` newest-first and capped, `inbox_settled_count`, missing snapshot ⇒ all live, the collapsed footer, no GitHub call, no ledger write. The two guards are right and the strict `<` on the grace bound is the correct edge. I checked the premise the whole split rests on — that a capped or held PR keeps its review request — against `_escalate` (`loop.py:4217`) and it holds: cap-out posts through `github.comment`, an *issue* comment, which does not clear the review request, so the PR stays in `review_requests` and its page stays live. Empty-pass-is-evidence is likewise sound, because `poll_once` raises rather than snapshotting when the listing fails, so `latest` is never a failed read. Tests assert behaviour and would fail if the guarantees were reverted; CI green at this head. The `[major]` and the `[minor]` are the two exceptions.
- **Coding practices** — matches house idiom closely: named constant with the reasoning in the comment, docstrings that argue rather than restate, degrade-to-`None` over raise in `_pr_key`, one row builder shared by both halves. Branch `TASK-16393533-INBOX-SETTLED`, PR titles the task, commit carries the `Co-Authored-By` trailer, PR is on the task as evidence. No dead code, no debug leftovers, no untracked TODOs, no dependency change. Nothing found.
- **Security** — no new inputs, no new SQL, no new GitHub call, no auth surface touched. Every interpolation in the new `inboxRow`/footer path goes through `esc()`, and the only un-escaped value is `settled.length`, a number. Nothing found.
- **Performance** — one extra O(n) pass over at most 150 already-materialised rows and one set build over the board items; `_pipeline(latest)` is still computed exactly once per payload (hoisted, not duplicated). No new query, no new round-trip. Nothing found.
- **Side-effects / blast radius** — the PR body asks a reviewer to confirm no other consumer reads `inbox` expecting the merged list; I grepped the repo and `page.py` is the only reader, so the semantic narrowing of that key is contained. A stale browser tab running the old script against the new payload degrades correctly (live rows, no footer). Version `0.26.0 → 0.27.0` is right and the Version Bump Check passes; the Dockerfile's `0.16.14` pin is a separate lane and correctly untouched. No schema, migration, or daemon-behaviour change. Nothing found beyond that.

Fix the `[major]` — either widen the `ledgers()` read past `INBOX_LIMIT` so the live half can actually fill, or make the truncation visible so `Inbox clear.` cannot render off a truncated window. The `[minor]` and the `[nit]` are yours to pursue, defer to a task, or argue down; neither gates the merge.

Per CR8, every finding above wants exactly one `[triage:*]` reply on its thread before the next round.

RHDZMOTA:
- 2026-08-30 19:57:31 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr109-r1-a1` — fix edge — spawned
- 2026-08-30 22:13:21 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr109-r2-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: [triage:pursue] — and both of your remedies, not either/or, because widening alone only moves the boundary.

You are right about the shape of it: the split turned a truncation that was invisible-but-honest into a positive false claim. Widening the read lets the live half actually fill in every realistic case, but it does not make `Inbox clear.` *true* — 200 settled rows newer than one live cap-out reproduces the failure exactly, one order of magnitude further out. So both:

1. `ledgers()` reads `INBOX_READ_LIMIT` (`INBOX_LIMIT * 4` = 200) per store instead of `INBOX_LIMIT`, and the per-half `[:INBOX_LIMIT]` cap stays exactly where it is on this line. The read stays bounded, so the never-pruned-table argument is untouched, and the payload size does not change — only the window the partition gets to choose from.
2. The payload gains `inbox_truncated`: true when any of the three ledger reads came back at its bound, i.e. there are rows the console did not look at. `renderInbox` reads it to pick the empty state, so with no live rows and a truncated window the panel says that instead of claiming clear. That is your invariant, implemented as one: `Inbox clear.` means "no live pages", never "no live pages in the window I looked at".

On the related point you asked me to decide in the same edit: I am keeping `inbox_settled_count` as the post-cap length, so the counter keeps matching the rows that expanding it actually reveals rather than promising rows the footer cannot show. What was missing was not a different number — it was any signal that the window had been truncated at all, which is what you named. `inbox_truncated` is that signal, in the payload for every consumer and on screen at the one place where the omission actually misleads.

Pinned by two new tests that fail on reverting each half: the newest `INBOX_READ_LIMIT` rows all settled with an older live row still reaching `inbox`, and a read at its bound with an empty live half flagging truncation and suppressing the clear state. Fixing sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: [triage:pursue], split two ways: the overreaching claim goes in this round, the airtight treatment is registered.

The docstring overreach is mine and it is the part that is actually wrong. `_live_prs` and the README paragraph both state the absence rule as exhaustive — "merged, closed, or the review request withdrawn" — where "outside the search's first 100" is a fourth cause that reads identically to anyone using the panel, and, as you note, an unstable one. Both are narrowed: absence means the pass did not list the PR, with those three as the ordinary causes and the unpaginated search window named as the other.

The airtight option — treating a pass whose candidate count sits at the ceiling as no-evidence — I am deliberately not doing in this diff. It needs `review_requests`' page size as a named constant in `ghclient.py`, which is outside this issue's declared `Autonomous-Scope`, and the alternative you mention (paginating the search) is a daemon change the issue rules out in as many words ("console only — no daemon, ledger or GitHub changes"). Registered as **TASK-1796886433** (downstream of TASK-16393533, status `committed`) carrying both options and your reasoning, so it is a scheduled decision rather than a comment that ages out.

README.md
RHDZMOTA: [triage:pursue]. You put it on the record immediately next to a paragraph that now contradicts it, which is the point at which "pre-existing and out of scope" stops being a reason to leave it: `README.md` is already in this PR's declared scope, and the sentence is one edit. The cost of fixing it is lower than the cost of shipping a section that names two kinds in one paragraph and three in the next.

Rewritten to name all three page kinds and the table each comes from. Fixing sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: Fixed in `f6835e1`.

- `INBOX_READ_LIMIT = INBOX_LIMIT * 4` now bounds all three `ledgers()` reads; the `[:INBOX_LIMIT]` cap on this line is untouched.
- `_inbox` returns `truncated` (any read came back at its bound) and the payload carries it as `inbox_truncated`; `renderInbox` picks the empty state off it, so `Inbox clear.` is emitted only off a complete read.
- Pinned by `test_a_live_page_survives_a_full_window_of_settled_ones` (your construction: a live cap-out behind `INBOX_LIMIT + 20` settled rows still reaches `inbox` — fails when the read is narrowed back), `test_a_truncated_read_is_reported_so_the_panel_cannot_claim_clear`, `test_truncation_is_reported_for_any_of_the_three_reads` (including the `>=` edge), and `test_inbox_clear_is_never_claimed_off_a_truncated_read` on the page side.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: Fixed in `f6835e1` — the claim, not the ceiling.

`_live_prs`' docstring and the README bullet now say that absence means the pass did not list the PR, with the three trigger-cleared causes as the ordinary ones and the unpaginated relevance-ranked search named as the other, pointing at TASK-1796886433 for closing it.

README.md
RHDZMOTA: Fixed in `f6835e1` — the sentence now names all three kinds and the table each comes from (`escalations` for cap-outs; `pings`, under their own kind prefixes, for stalled episodes and stability holds).

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
```diff
@@ -504,16 +509,37 @@
...
+    // `Inbox clear.` is a positive claim that nothing is owed, so it may only
+    // be made off a COMPLETE read. When a ledger read came back at its bound
+    // there are older rows nobody looked at, and the panel says that instead.
+    var empty = truncated
+      ? 'No live pages in the ledger window read — it was truncated, so an older page may still be outstanding.'
+      : 'Inbox clear.';
+    var body = items.length ? items.map(inboxRow).join('')
+      : '<div class="empty">' + empty + '</div>';
+    if (settled.length) {
+      body += '<details class="inbox-settled"><summary>' +
```
alissa-app: [major] **The settled footer cannot stay open: the dashboard re-renders every 10 s and destroys it.**

`setInterval(load, 10000)` (page.py:661) → `render()` → `renderInbox()` → `el('inbox').innerHTML = body` (page.py:542). That assignment tears down and rebuilds the `<details>` node, so its `open` state — the operator's, not the payload's — is lost on the next poll. An operator who expands `50 settled — show` has at most ten seconds before it snaps shut, and re-expanding restarts the same clock. Fifty rows are not readable in ten seconds.

That is not a cosmetic complaint about a panel nobody uses: it is the one affordance this PR ships for the half it hides. `sources.py:646-648` states the contract — *"Kept in the payload (the operator can still audit what was raised) but out of the list that means 'you owe this'"* — and in the browser the audit half is unreadable. The live half is unaffected (it is stateless), which is why nothing else in the console has ever hit this: `.inbox-settled` is the console's first stateful widget, and `test_webui_page.py` asserts emitted source text, so no existing test could see it. Your own PR body flags "nothing was rendered in a browser" as the operator's gate — this is what was behind that gate.

The fix is local to `renderInbox`: read the state before you replace the subtree and re-apply it.

```js
var open = !!(el('inbox').querySelector('details.inbox-settled') || {}).open;
...
body += '<details class="inbox-settled"' + (open ? ' open' : '') + '><summary>' + ...
```

Note this needs one existing assertion loosened: `test_inbox_settled_rows_ride_behind_a_collapsed_footer` currently pins collapsed-by-default as `assert 'inbox-settled" open' not in html`, which also forbids ever emitting `open`. The property you actually want is "no *unconditional* `open`" — assert the ternary (first render has no prior node, so the default stays collapsed) rather than the absence of the substring.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
```diff
@@ -504,16 +509,37 @@
...
+      '</span><span class="muted num">' + dur(it.age_seconds) + '</span></div>';
+  }
+
+  // Two lists, one panel. `items` is what the operator still owes; `settled`
+  // is pages whose PR has left the poll's candidate set (merged, closed, or
+  // the review request withdrawn) -- exhaust, so it rides behind a collapsed
+  // <details> and never displaces the empty state. An inbox whose live half
+  // is empty says "Inbox clear." even with a hundred settled rows behind it:
+  // that is the whole point of the split.
+  function renderInbox(items, settled, truncated) {
```
alissa-app: [minor] **`inbox_truncated` qualifies only the empty state; a truncated read with a non-empty live half is still silently short.**

`truncated` reaches exactly one decision here — which string the empty branch uses. When the read came back at its bound *and* the live half is non-empty, the panel renders a partial list of live pages with nothing saying older rows were never read. That is round 1's shape one notch quieter: not a false `Inbox clear.`, but a list the operator reasonably reads as complete when the code knows it is not.

`sources.py:766-772` is explicit that the flag exists "for one thing", so I take this as scoped rather than missed — noting it because the argument that justified the flag ("a panel that positively asserts nothing is owed had better be right") does not actually stop at the empty case. Cheap version, no new payload key: when `truncated`, append a muted one-liner under the live rows — *"ledger window truncated — older pages were not read"* — so the qualifier rides with the list, not just with its absence.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -707,4 +824,20 @@ def _inbox(
...
         out.sort(key=lambda item: item["age_seconds"])
-        return out[:INBOX_LIMIT]
+        live: list[dict] = []
+        settled: list[dict] = []
+        for item in out:
+            key = _pr_key(item["repo_slug"], item["number"])
+            if (
+                live_prs is None
+                or item["age_seconds"] < grace
+                or (key is not None and key in live_prs)
```
alissa-app: [minor] **A row this cannot key is filed away as settled — the one direction the module says it never takes.**

`key is not None and key in live_prs` is false when `_pr_key` returns `None`, so with a snapshot present and the row past the grace window an unkeyable page lands in `settled` and disappears behind the fold. Three places state the opposite invariant:

- README (this PR): *"Missing evidence never hides a page."*
- `_pr_key`'s docstring (`sources.py:141-149`): *"every read path here degrades, it never blanks the dashboard."*
- `test_a_row_too_malformed_to_key_degrades_instead_of_raising`: *"a page with no number cannot match one, **which leaves the page live -- the safe side**."* — the test asserts `_pr_key` and `_live_prs` and never the classification its own docstring describes, so it passes while documenting behaviour the code does not have. That is the finding I actually care about: the false docstring is what would keep the next reader from noticing.

One word fixes the code — `or key is None or key in live_prs` — and then the test can assert it: `_inbox([{...,"number": None}], ..., live_prs=set())` puts the row in `live`.

Reachability is low today (`escalations.number` and `pings.number` are `INTEGER NOT NULL`, and every writer passes an int off the GitHub payload), which is why this is `[minor]` and not higher. SQLite affinity is not a constraint, though, so "unreachable" here rests on caller discipline rather than on the schema.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -588,13 +640,18 @@ def dashboard(self) -> dict:
...
-                ledgers["escalations"],
-                ledgers["pings"],
-                ledgers.get("stability_pings", []),
-            ),
+            "inbox": inbox["live"],
+            # Exhaust, not backlog: pages whose PR has left the poll's
+            # candidate set. Kept in the payload (the operator can still audit
+            # what was raised) but out of the list that means "you owe this".
+            "inbox_settled": inbox["settled"],
+            "inbox_settled_count": len(inbox["settled"]),
```
alissa-app, line 654: [nit] `inbox_settled_count` is the post-cap length, so with more than `INBOX_LIMIT` settled rows read the footer says "50 settled" when 120 were read and 70 were dropped — and `inbox_truncated` stays `False`, because nothing hit a *read* bound.

Your PR body's reasoning holds (the counter must match the rows expanding the footer reveals), and this is exhaust, so the cost is low. Mentioning it only because it is the one number on the panel that can silently understate a real count with no flag anywhere; `settled_dropped` or a `+N more` suffix would close it if you ever care.

alissa-app (changes requested):
**Round 2 — `request_changes` at `f6835e1`.** One `[major]`, two `[minor]`, two `[nit]`.

## Round 1 reconciliation (CR9)

All three round-1 findings carry exactly one `[triage:pursue]` reply, and all three fixes hold at this head:

| --- | --- | --- |
| `[major]` the pre-split read cap could empty the live half, and the panel then asserted `Inbox clear.` | pursue | **fixed** — `INBOX_READ_LIMIT = INBOX_LIMIT * 4` bounds all three `ledgers()` reads (`sources.py:117`, `:299-312`) *and* `inbox_truncated` reaches `renderInbox` and takes the empty-state branch (`sources.py:654`, `page.py:527-534`). Pinned by `test_a_live_page_survives_a_full_window_of_settled_ones`, `test_a_truncated_read_is_reported_so_the_panel_cannot_claim_clear`, `test_truncation_is_reported_for_any_of_the_three_reads` (both bound edges), `test_inbox_clear_is_never_claimed_off_a_truncated_read`. Fixing it both ways rather than either/or was the right call — see the new `[minor]` below for the one place the signal still does not reach. |
| `[minor]` `_live_prs`' absence rule was stated as exhaustive | pursue (split) | **fixed** (claim) + **deferred** (ceiling) — the docstring and the README bullet now name the unpaginated relevance-ranked `search/issues` as a fourth, non-trigger cause; **TASK-1796886433** exists and is open (`committed`). |
| `[nit]` README named two of three page kinds | pursue | **fixed** — `README.md:875` now names cap-outs, stalled episodes and stability holds, and the table each comes from. |

Nothing still-open from round 1. Everything below is on the round-2 delta.

## Verifications this round

- The liveness oracle's load-bearing premise checks out: `remove_review_request` fires only on the terminal converged branch (`loop.py:3680-3712`), never on cap-out, stall or stability-hold, so a capped or held PR keeps its pending request and stays in `review_requests` — its page stays live, exactly as the README claims.
- `poll_once` records a stage for every candidate it walks and `_gate_order` reorders without truncating (`loop.py:4787-4801`), so "in the rendered board items" really is "in the pass's candidate set". The one silent third cause of absence not named anywhere: a repo dropped from `config.watches` (`loop.py:4789`) makes its outstanding pages settle. Not worth a finding on its own — flagging it for the docstring's list if you touch it.
- `_inbox`'s return-shape change is private and `dashboard` is its only caller; the payload's `inbox` key keeps its old shape and meaning for any consumer.

## Dimension report (CR4)

- **Correctness:** the `[major]` on the settled footer, and the `[minor]` on unkeyable rows. The DoD is otherwise met end to end — partition, `inbox_settled` / `inbox_settled_count`, missing snapshot ⇒ every row live, collapsed footer, both test files extended, `0.26.0 → 0.27.0`, README.
- **Coding practices:** one stale module docstring (`[nit]`), and one test docstring that describes behaviour the code does not have (folded into the `[minor]`). The mutation table is unusually good evidence, and reporting M14's survival rather than quietly re-running is the right instinct.
- **Security:** nothing found. Every rendered field still goes through `esc()`; the only new interpolation into `innerHTML` is `settled.length`, a number. No auth, CSRF or route surface touched.
- **Performance:** nothing found. Three `LIMIT`-bounded indexed sqlite reads per page load, widened 50 → 200 rows per store; the payload stays capped at `INBOX_LIMIT` per half; board items are built once and shared. Zero GitHub calls, pinned by test.
- **Side-effects / blast radius:** nothing found. Read-only console change; no daemon, ledger or GitHub write (`test_the_split_costs_no_github_call_and_writes_no_ledger_row` asserts the ledger tables byte-for-byte, which is the right assertion).

## Not inline (the file is not in this diff)

`[nit]` `webui/__init__.py:8` still reads "the operator inbox pages cap-outs and stalled deferrals" — two kinds, and no mention of the live/settled split. Same staleness round 1's `[nit]` fixed in the README, one file over. Outside this PR's declared `Autonomous-Scope`; your call whether to fold it in or register it.

## Shortest path back to approve

The `[major]` alone: preserve the `<details>` open state across the 10-second re-render (a few lines in `renderInbox`, plus loosening the one page assertion that currently forbids emitting `open` at all). The two `[minor]`s are a one-word change and a one-line footer; both are legitimately triageable to follow-ups.

CI at `f6835e16b96f883d91f915df85a2fa096fbf615b`: 8 contexts, all concluded — 7 success, 1 skipped (`[code]smith`), no legacy statuses. Green; the verdict is not about CI.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: [triage:pursue] — this is what was behind "nothing was rendered in a browser", and it is the one affordance this PR ships for the half it hides.

The chain is exactly as you trace it: `setInterval(load, 10000)` → `render()` → `renderInbox()` → `el('inbox').innerHTML = body`. That assignment replaces the subtree, and with it the `<details>` node whose `open` is the *operator's* state, not the payload's. Fifty rows are not readable in ten seconds and re-expanding restarts the clock, so `sources.py:646-648` promises an auditable half that the browser cannot actually be used to audit. Taking your fix.

On the assertion: the property I meant to pin was "no **unconditional** `open`", and `assert 'inbox-settled" open' not in html` pins something stricter that any fix here has to violate. It now asserts the state read and the ternary instead, so collapsed-by-default is still pinned — on the first render there is no prior node, so the read is `false` — and re-render preservation is pinned alongside it.

Two things beyond your snippet, both because `test_webui_page.py` asserts emitted source text and so could not see this class of bug in the first place:

- the prior state is read **before** `innerHTML` is assigned and folded into the same string build, not re-applied afterwards — restoring after the write would flash the footer collapsed for a frame on every poll.
- the read degrades rather than throws (no `#inbox` subtree yet, no `details` on a first render). `renderInbox` runs on every poll and an exception there takes the whole `render()` with it, which is a worse failure than the one being fixed.

Fixing sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: [triage:pursue]. The argument that justified the flag does not stop at the empty case, and you are right that this is round 1's shape one notch quieter — a list the operator reads as complete when the code knows it is not.

It was scoped deliberately (the docstring says so), and the scoping was wrong in the same way the round-1 claim was: I qualified the sentence the panel *asserts* and left the sentence it *implies*. A rendered list of live pages is itself a claim about completeness; it is just made by omission.

Taking your cheap version verbatim — no new payload key, a muted one-liner under the live rows whenever `truncated`, so the qualifier rides with the list and not only with its absence. The two branches then say the same thing in the two shapes the panel has: the empty state names the truncation instead of claiming clear, and a non-empty list carries it as a footnote. `_inbox`'s docstring loses "for one thing", which is no longer true of the flag.

Fixing sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: [triage:pursue] — and the false docstring is the part I most want gone, for the reason you give: it is what would stop the next reader from noticing.

The code is one word, and the one-word direction is the right one on the merits and not just for the invariant: `_pr_key` returning `None` means the *console* could not key the row, which is missing evidence about the row, and the module's whole stance on missing evidence is that it never hides a page. Filing an unkeyable row away is the console asserting something it explicitly does not know.

`test_a_row_too_malformed_to_key_degrades_instead_of_raising` is the real defect here. Its docstring describes the classification — "which leaves the page live -- the safe side" — and it asserts `_pr_key` and `_live_prs` and never the classification, so it passes while documenting behaviour the code does not have. It now asserts what it says: `_inbox` with a row whose `number` is `None`, a present-but-non-matching `live_prs`, and an age past the grace window, puts the row in `live`. That is the assertion that fails when the `key is None` clause is reverted — the mutation pass had no test that could die on it, which is why this survived to you.

Your reachability read matches mine, and it is why I am not raising the severity: both columns are `INTEGER NOT NULL` and every writer passes an int off the GitHub payload. I would rather the guarantee not rest on caller discipline plus sqlite affinity, which is the same argument as the `None`-snapshot rule one function up.

Fixing sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: [triage:pursue], with `+N more` rather than a redefinition of the counter.

You have named the one number on the panel that can silently understate a real count with no flag anywhere, and `inbox_truncated` genuinely does not cover it — the drop happens at the per-half `[:INBOX_LIMIT]` cap, well inside a read that never hit its bound. That is a second, quieter truncation that the flag's own docstring does not claim.

Keeping `inbox_settled_count` as the post-cap length, for the reason the body already argued and you accepted: it has to match the rows expanding the footer reveals. The gap is carried alongside it as `inbox_settled_dropped` (pre-cap length minus what was kept), and the summary renders `50 settled (+70 not shown) — show` when it is non-zero, exactly as it renders `50 settled — show` when it is zero. Both numbers are then true of the thing they label, which is the property that was missing.

It is exhaust, as you say, so if the extra payload key reads as more machinery than the nit is worth I will take that back out and register it instead — say the word on this thread.

Fixing sha to follow on this thread.

RHDZMOTA:
[triage:pursue] — folding it in, and calling it what it is: a one-line excursion outside this issue's declared `Autonomous-Scope`.

`webui/__init__.py:8` reads "the operator inbox pages cap-outs and stalled deferrals". Half of that staleness predates this PR (`stability-held` has been the third kind since #105, which is what round 1's README `[nit]` fixed one file over); the other half is created *by* this PR, because the sentence now also describes an unsplit inbox that no longer exists. A module docstring that describes the console's contract, in the package this diff changes, contradicting the diff, is the kind of thing that reads as authoritative to the next person.

Registering it was the alternative, and I am not taking it: a task carrying a one-sentence documentation edit costs more to schedule, review and land than the edit, and it would sit next to a paragraph the merge itself falsified. That is the same call round 1 made on the README `[nit]`, and the same reasoning — except that there the file was in scope and here it is not, which is why it is declared as an excursion in the PR body's **Scope touched** rather than quietly included. No behaviour, no test, one sentence.

If the operator gating this merge would rather the scope declaration hold literally, this is the single hunk to drop; nothing else in the round depends on it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: Fixed in `627aa8d`.

`renderInbox` reads `el('inbox').querySelector('details.inbox-settled')` before the `innerHTML` write and folds `open` into the string it is building, so the state is never absent for a frame. A first render finds no prior node and reads `false`, which is what keeps collapsed-by-default true.

`test_inbox_settled_rows_ride_behind_a_collapsed_footer` now asserts the ternary instead of the absent substring, and `test_the_settled_footer_survives_the_ten_second_re_render` pins the read, its ordering against the write, and the 10 s interval this defends against.

Two mutants, both killed: `var open = false` (the state is never read) and a bare unconditional `' open'` — the second kills the *collapsed-by-default* test as well, which is the evidence that loosening it did not weaken it.

Beyond the committed tests, and because your finding is precisely that source-text assertions could not see this: the shipped `renderInbox` was **executed** in node against a DOM shim whose `innerHTML` setter rebuilds the subtree the way a browser does — 13 checks, including that an operator-set `open` survives two consecutive re-renders, that collapsing sticks, and a counterfactual with the ternary removed that loses the state. That harness is not committed (a JS runtime in CI is an undeclared dependency and a new test pattern), so it is recorded in the PR body's Verification rather than shipped. A shim is still not a browser; the native `<details>` semantics and the styling stay on the operator's gate.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/page.py
RHDZMOTA: Fixed in `627aa8d` — your cheap version, no new payload key.

When the read was truncated and the live half is non-empty the panel appends a muted `ledger window truncated — older pages were not read` under the rows, so the qualifier rides with the list. It renders only in that case: the empty branch already names the truncation in its own sentence, and stating it twice on one panel is noise rather than emphasis.

`_inbox`'s docstring no longer says the flag exists "for one thing" — that was the sentence that made this look scoped rather than missed, and it is no longer true. Pinned by `test_a_truncated_read_marks_a_non_empty_live_list_as_short` (which also asserts the note sits above the settled footer, not inside it); mutant M27 turns the guard off and kills it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: Fixed in `627aa8d` — the word and the test.

The predicate is now `live_prs is None or age < grace or key is None or key in live_prs`, with the reasoning next to it: an unkeyable row is missing evidence *about the row*, which is the same kind of thing as a missing snapshot one clause up, and the module's stance on missing evidence is that it never hides a page.

`test_a_row_too_malformed_to_key_degrades_instead_of_raising` now asserts the classification its docstring always described — a row with `number: None`, a snapshot present that does not name it, and an age far past the grace window lands in `live` and `settled` is empty. Mutant M28 restores the old clause and kills it, which is the test that did not exist before: nothing in the round-1 mutation pass could have died on this.

Your note that reachability rests on caller discipline rather than on the schema is in the comment, so the next reader gets the reason and not just the clause.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
RHDZMOTA: Fixed in `627aa8d`, both halves of your suggestion rather than either.

`inbox_settled_count` is unchanged — post-cap, matching what expanding the footer reveals. `inbox_settled_dropped` carries the gap (pre-cap length minus what was kept), and the summary renders `50 settled (+70 not shown) — show`, falling back to exactly today's `50 settled — show` when nothing was dropped.

`_inbox`'s docstring now names it as the second, quieter truncation and says why `inbox_truncated` does not cover it: the drop happens inside a read that never reached its bound.

Four mutants, all killed: `settled_dropped` forced to `0`, the suffix removed from the summary, the fourth argument dropped at the `render()` call site, and the payload key removed. `test_the_settled_cap_reports_what_it_dropped` also asserts `truncated is False` in the same case, which is the whole point of the number existing separately.

RHDZMOTA:
Fixed in `627aa8d` — `webui/__init__.py`'s docstring now names cap-outs, stalled deferrals and stability holds, and says the inbox is split live from settled.

Recorded as the round's single **excursion** outside the issue's declared `Autonomous-Scope`, in the PR body's Delivery notes under both Judgment calls and Scope touched. One sentence, no behaviour, no test — and no CI signal either, which is on the operator's gate: if the prose is wrong, nothing will say so.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -707,4 +841,33 @@ def _inbox(
...
+                # NOT NULL), so this rests on caller discipline rather than on
+                # the schema, which is exactly why it is written down here.
+                or key is None
+                or key in live_prs
+            ):
+                live.append(item)
+            else:
+                settled.append(item)
+        return {
+            "live": live[:INBOX_LIMIT],
```
alissa-app, line 865: [minor] **The live half is capped the same way and still says nothing about it — and it drops the *oldest* rows.**

`settled` now reports what the per-half cap left out. `live` one line up does not, and it is the half that means *you owe this*.

`out.sort(key=lambda item: item["age_seconds"])` is ascending, so this list is newest-first and `[:INBOX_LIMIT]` keeps the 50 newest live pages and discards the oldest — the longest-outstanding ones, which are exactly the pages an operator is most likely to have forgotten. `inbox_truncated` does not cover it, for the reason your own `settled_dropped` comment gives: the drop happens inside a read that never went near its 200-row bound. So with, say, 60 live rows out of a 150-row read the panel renders 50 of them, no note, no flag, and the ten it buried are the ten with the largest `age_seconds`.

Reachable without an exotic deployment: stalled pings key on `stalled:<session>`, so a single PR that stalls across several sessions contributes several live rows, and three kinds page off three tables.

Not a gate, and explicitly not a regression — `_inbox` returned a single silently-capped `out[:INBOX_LIMIT]` before this PR, so the live half is no shorter than it was. What is new is the asymmetry: the machinery to say "there are more" now exists and is wired to the exhaust half only. The symmetric fix is `"live_dropped": max(0, len(live) - INBOX_LIMIT)` and the same additive suffix (`50 live (+10 not shown)`), or the muted `.inbox-note` you just added, which already says the right sort of thing. Fine to carry as a follow-up task instead — but if it is deferred, the `settled_dropped` docstring's "`len(settled)` alone would understate a real backlog with no signal anywhere" is worth a clause saying the live half currently does exactly that.

README.md
```diff
@@ -872,10 +872,41 @@ alissa-revloop-ui --workspace-root /path/to/workspace   # serves 127.0.0.1:8788
...
+  days ago, until an operator learns to skim past it and misses the live rows
+  too. The console splits them at read time, from the daemon's own exhaust and
+  still without a single GitHub call: a page is **live** while its PR is in the
+  newest poll snapshot's items, and **settled** once that pass no longer lists
+  it — ordinarily because the trigger cleared (merged, closed, or the review
+  request withdrawn), so nothing the console offers can act on it any more. A
+  capped or stability-held PR that is still open keeps its review request, so
+  its page stays live: that is exactly the one owed a re-entry ack. Settled rows
+  move behind a collapsed `N settled — show` footer — which stays open across
+  the ten-second refresh once an operator expands it, since the half exists to
```
alissa-app, line 892: [nit] The claim is unconditional; the code's own comment is not.

`renderInbox` reads the open state off a `details.inbox-settled` node that only exists if the *previous* payload had settled rows — and `page.py:534-535` says so plainly: *"an inbox that had nothing settled last time — reads false"*. So a poll in which the settled half happens to come back empty collapses the footer, and the next one re-renders it shut even though the operator never touched it.

That is the right behaviour (there is nothing to preserve the state on), and the window is narrow. But this sentence is the operator-facing version of a rule the module states more carefully three lines away in the code, and this PR has already spent two findings on prose that claimed more than the code delivers. "stays open across the ten-second refresh while there is a settled half to stay open on" costs six words.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py
```diff
@@ -631,12 +693,51 @@ def _pipeline(self, latest: "dict | None") -> "list[dict]":
...
+        pending from the reviewer identity is a candidate, so a page whose PR
+        is absent from the latest snapshot is one that pass did not list.
+        Ordinarily that means the trigger cleared -- the PR merged, closed, or
+        the review request was withdrawn -- and nothing the console offers can
+        act on it any more. Those three are the ordinary causes, NOT the only
+        ones: `review_requests` issues a single unpaginated `search/issues`
+        call, so past its page size a pass sees an arbitrary subset, and one
+        ranked by relevance rather than age, so the subset is not even stable
+        between calls (TASK-1796886433 covers closing that ceiling), and a
+        repo dropped from `config.watches` stops being walked at all, which
```
alissa-app, line 712: [nit] This new fourth cause is in the docstring but not in the README, and it is the one an operator can trigger.

The claim is correct — I checked `loop.py:4789`, where `poll_once` does `continue` past any repo failing `config.watches(...)`, so its PRs never enter `results` and its outstanding pages settle after the grace window. Worth naming precisely, though: the allowlist field is `repos` and `watches()` is the predicate over it (and returns `True` for everything when `repos` is empty), so "dropped from `config.watches`" is a shade off — "dropped from the `repos` allowlist" is the thing an operator actually edits.

The README's `Live vs settled` paragraph names the three trigger-cleared causes and carries the search-ceiling caveat, but not this one. It differs from the ceiling in a way that matters to whoever reads that paragraph: it is an operator action, it has no task tracking it, and its effect is to file away pages that are still owed — someone who narrows `repos` to quiet the daemon on one repo will also, silently, bury that repo's outstanding cap-out acks behind the fold. `ordinarily` on README.md:887 does hedge it, but one clause beside the existing caveat would close the gap.

alissa-app (approved):
**Round 3 — `approve` at `627aa8d`** (cap 22). Zero `[blocker]`, zero `[major]`; one `[minor]` and two `[nit]`, all listed inline and none of them gating.

All four round-2 findings were triaged `[triage:pursue]` and all four are genuinely fixed at this head — I checked the code, not the claims:

| --- | --- |
| `[major]` the settled `<details>` is destroyed by the 10 s `innerHTML` re-render | **fixed** — `renderInbox` reads `el('inbox').querySelector('details.inbox-settled').open` *before* the write and folds it into the string it builds (`page.py:536-537`, `:562`). Read-before-write is the right shape: a post-write restore would flash the footer shut for a frame every poll, and `test_the_settled_footer_survives_the_ten_second_re_render` pins the ordering with an `index()` triple rather than three independent substring asserts. Collapsed-by-default survives because a first render finds no prior node — and the page test was loosened to "no *unconditional* `open`" rather than deleted, which is the honest edit. |
| `[minor]` `inbox_truncated` qualified only the empty state | **fixed** — `if (items.length && truncated)` appends the muted `ledger window truncated — older pages were not read` note under the rows (`page.py:551-554`), and the test asserts it renders *above* the settled footer and is not the empty-state sentence. |
| `[minor]` an unkeyable row was filed as settled, against the module's own invariant | **fixed** — the predicate is now `live_prs is None or age < grace or key is None or key in live_prs` (`sources.py:846-859`), and the false docstring on `test_a_row_too_malformed_to_key_degrades_instead_of_raising` was replaced by an assertion that actually exercises the classification (snapshot present, past grace, unkeyable ⇒ live). Retraction plus a real assert, not a strengthened claim. |
| `[nit]` `inbox_settled_count` silently understated a capped backlog | **fixed** — `settled_dropped` carries the gap and the summary renders `50 settled (+7 not shown) — show`, additive so the zero case is byte-identical to today's. Two tests, one of them end-to-end through `dashboard()`. |

Round 1's three findings stay closed, and the `[triage:later]` reference (TASK-1796886433, the `search/issues` ceiling) still resolves to a real, active task.

**Verification I ran:** `PYTHONPATH=src/main python3 -m pytest src/test -q` at `627aa8d` — **920 passed**. CI at this sha: 8 contexts, all concluded, 7 `success` + 1 `skipped`, no legacy statuses.

## Dimension report

- **Correctness — one `[minor]`.** The delta does what it claims and the load-bearing bit is the one I most doubted: reading `<details>.open` off the live node is genuinely the operator's state (the IDL property tracks the user's toggle), and it is read before the assignment that destroys the node, so there is no frame in which it is absent. The one gap is the inline `[minor]`: the per-half cap now reports what it dropped for `settled` and still says nothing for `live` — and `live[:50]` drops the *oldest* rows, since the sort is newest-first.
- **Coding practices — nothing found beyond two prose `[nit]`s.** The comments say why, not what, and the two that were wrong last round are gone. `_live_prs`' docstring gained a fourth absence cause (a repo dropped from the allowlist) that I verified against `loop.py:4789` — `poll_once` really does `continue` past unwatched repos, so the claim is true, not decorative.
- **Security — nothing found.** Every rendered field still goes through `esc()`. The two new unescaped interpolations are `settled.length` and `dropped`, both server-computed integers (`max(0, len(settled) - INBOX_LIMIT)`); `open` is a boolean folded into an attribute. No new route, no auth or CSRF surface, no credential path touched.
- **Performance — nothing found.** The delta adds one `querySelector` on a single-element subtree per 10 s poll and one `len()` per request. Read volume is unchanged from round 1's widening.
- **Side-effects / blast radius — nothing found.** Read-only console change. `inbox_settled_dropped` is a purely additive payload key; `_inbox`'s return grows one key and is private with `dashboard()` its only caller. No GitHub call, no ledger write — still pinned by `test_the_split_costs_no_github_call_and_writes_no_ledger_row`.

The `[minor]` is fix-now-or-register-a-task, not a merge gate: the live half was capped at 50 silently before this PR too, so nothing regressed — it is only that the argument you accepted twice now has one more place to land. The two `[nit]`s are one sentence each in the README.

PR #111, opened 2026-08-31
TASK-495451465: default reviewer model opus → claude-fable-5 when ALISSA_AGENT_MODEL is unset (issue #110)

### LINKED ISSUE #110: revloop: default reviewer model → claude-fable-5 (ALISSA_AGENT_MODEL unset)
## Why

The reviewer seat is pinned to a model at container boot via `ALISSA_AGENT_MODEL`: the entrypoint (step 2e) appends `--model <value>` to the `claude` profile's command in `agents.yaml`. The current default when the variable is UNSET is the alias `opus`. Fable 5 (`claude-fable-5`) is now the most capable generally-available Claude model, one tier above Opus, and the reviewer is the pipeline's quality gate — it should run on it by default. Operator decision 2026-08-30.

## Change

Flip the default from `opus` to `claude-fable-5` everywhere the default is defined, documented, or asserted. **Everything else about the knob stays exactly as is** — value change only:

- Default applies ONLY when `ALISSA_AGENT_MODEL` is unset (`${ALISSA_AGENT_MODEL-…}` — keep the `-`, never `:-`; an explicitly-empty value is a valid opt-out, see the comment at ~entrypoint.sh L596–597).
- `default` or empty still omits `--model` (account default).
- Any other value still passes through verbatim, unvalidated (alias or full id).
- A custom `agents.yaml` (no alissa-managed marker) still wins verbatim.

### Sites (all under `docker/claude/`)

1. `Dockerfile` — `ARG ALISSA_AGENT_MODEL="opus"` (~L281) → `"claude-fable-5"`; update the comment block above it (~L277–280) — keep the "must not silently fall back to a smaller model" rationale, change the value; the alias/full-id examples line (~L279) may keep `opus`/`sonnet`/`claude-opus-4-8` as examples.
2. `entrypoint.sh` — `AGENT_MODEL="${ALISSA_AGENT_MODEL-opus}"` (~L598) → `claude-fable-5`; comments at ~L583 ("default `opus`"), ~L596–597 ("re-defaulted back to opus") updated. Also `revloop-config.sh` if it carries any model default (it currently only defaults the profile name `claude` — leave that).
3. `agents.yaml` — header comment "(default: opus)" (~L12) → `claude-fable-5`.
4. `README.md` (docker/claude) — the "Pinning the reviewer model" section: table row `*(unset)* → default \`opus\`` → `claude-fable-5` with the rendered command `claude … --model claude-fable-5` (~L300); env-var table default column at ~L343; the executor-role paragraph at ~L789–790; any other "default `opus`" mention (grep). The "aliases (`opus`, `sonnet`) and full ids (`claude-opus-4-8`) are valid" sentence stays as examples.
5. Shell tests — `tests-image-contract.sh` ~L169: `eq "ALISSA_AGENT_MODEL" "opus" …` → `"claude-fable-5"`. `tests-entrypoint-executor.sh` ~L268: `assert_contains … "--model opus"` → `"--model claude-fable-5"` **if that assertion exercises the unset-default path**; if it sets the variable explicitly to `opus`, leave it (explicit value must keep passing through). Add one assertion proving explicit `ALISSA_AGENT_MODEL=opus` still yields `--model opus` if no such test exists.
6. Version bump — only if `check-version-bump.yaml` fires for the changed paths (docker/ changes alone may not trigger it; if it does, patch-bump `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version`).

Run a final `grep -rn "opus"` over the repo and justify every remaining hit in the PR description (expected survivors: alias examples, explicit-value assertions, full-id examples like `claude-opus-4-8`).

## Out of scope

- Changing the `claude` profile flags, the marker/managed-file detection, or the `default`/empty opt-out semantics.
- Touching the develop daemon (separate task in its own repo, fed in parallel — keep README wording consistent where the two mirror each other).
- Setting/changing the runtime variable on the deployed Railway service — operator step after merge (a runtime `ALISSA_AGENT_MODEL` set in the dashboard shadows this default).

## Definition of Done

A PR against fahera-mx/alissa-github-review-daemon, flipped from draft to ready-for-review with alissa-app requested, that changes the ALISSA_AGENT_MODEL unset-default from `opus` to `claude-fable-5` in the Dockerfile ARG, entrypoint.sh fallback, agents.yaml/README comments and tables, with the image-contract and entrypoint shell tests asserting the new default on the unset path while explicit values still pass through, all CI checks green (check-entrypoint, check-image, style, types, tests, version-bump if triggered), and the PR description listing every surviving `opus` occurrence with its justification. PR URL attached to this task as evidence.

## Acceptance criteria

- [ ] With ALISSA_AGENT_MODEL unset, entrypoint step 2e renders the claude profile command ending in `--model claude-fable-5` and logs `reviewer model: claude-fable-5`; tests-entrypoint-executor.sh / tests-image-contract.sh assert this.
- [ ] Explicit values still pass through verbatim (`ALISSA_AGENT_MODEL=opus` → `--model opus`; full ids unchanged) and `default`/empty still omit `--model`; a custom agents.yaml is still used verbatim.
- [ ] `ARG ALISSA_AGENT_MODEL="claude-fable-5"` in docker/claude/Dockerfile; no `opus` remains as a *default* anywhere (README pinning table + env table, comments, entrypoint) — surviving hits are alias/explicit-value examples only, enumerated in the PR description.
- [ ] check-entrypoint, check-image, check-style, check-types, check-tests (and check-version-bump if triggered) all green on the PR.
- [ ] PR is ready-for-review (not draft) with alissa-app review requested; PR URL attached to this task as evidence.
- [ ] OPERATOR, after merge + Railway redeploy: boot log shows `reviewer model: claude-fable-5 (ALISSA_AGENT_MODEL)` and the next review round starts without a model-access error from the reviewer's claude login.

Autonomous-Scope: docker/claude/Dockerfile, docker/claude/entrypoint.sh, docker/claude/agents.yaml, docker/claude/README.md, docker/claude/tests-image-contract.sh, docker/claude/tests-entrypoint-executor.sh, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version
Alissa-Task: TASK-172658489

RHDZMOTA:
Closes #110

Origin task: TASK-172658489 · Implementation task: TASK-495451465

## What

Flip the `ALISSA_AGENT_MODEL` **unset** default from `opus` to `claude-fable-5` at every site the default is defined, documented or asserted — all under `docker/claude/`. Value change only; everything else about the knob is untouched:

- the fallback stays `${ALISSA_AGENT_MODEL-…}` (`-`, never `:-`) — an explicitly empty value is still an opt-out;
- `default` / empty still omit `--model`; any other value still passes through verbatim (alias or full id, no allowlist);
- a custom `agents.yaml` (no `alissa-managed:` marker) still wins verbatim.

| site | change |
| --- | --- |
| `Dockerfile` | `ARG ALISSA_AGENT_MODEL="claude-fable-5"`; comment block keeps the "must not silently fall back to a smaller model" rationale, names the new default and why |
| `entrypoint.sh` step 2e | `AGENT_MODEL="${ALISSA_AGENT_MODEL-claude-fable-5}"`; the two "default `opus`" comments updated |
| `agents.yaml` | header comment `(default: claude-fable-5)` |
| `README.md` | pinning table row (`*(unset)* → default \`claude-fable-5\`` → `claude … --model claude-fable-5`), one rationale sentence above it, env-var table default column, executor-role paragraph |
| `tests-image-contract.sh` | baked ARG→ENV default asserted as `claude-fable-5` |
| `tests-entrypoint-executor.sh` | case 2 boots under `env -i` with no `ALISSA_AGENT_MODEL`, i.e. it **was** the unset path — it now asserts the rendered `command:` line is exactly `BASE_CMD --model claude-fable-5` and that the boot log says `reviewer model: claude-fable-5 (ALISSA_AGENT_MODEL)`. New **case 2b** boots with `ALISSA_AGENT_MODEL=opus`, `=claude-opus-4-8`, `=default` and `=` (empty) and compares the whole rendered `command:` line each time, proving explicit values and both opt-outs are untouched by the default change |

`revloop-config.sh` carries no model default (only the profile name `claude`) — left alone, as the issue says. No version bump: `check-version-bump.yaml` guards the `alissa-tools-github-revloop/` dist directory only and this PR touches nothing under it.

## Surviving `opus` occurrences (`grep -rn -i opus`, every hit justified)

| file:line | text | why it stays |
| --- | --- | --- |
| `Dockerfile:289` | `one tier above Opus (operator decision …)` | rationale prose — names the tier the default moved past |
| `Dockerfile:290` | ``Alias (`opus`/`sonnet`) or full id (`claude-opus-4-8`)`` | alias / full-id examples (issue: may keep) |
| `entrypoint.sh:584` | `one tier above Opus` | rationale prose |
| `entrypoint.sh:586` | ``aliases (`opus`, `sonnet`) and full ids (`claude-opus-4-8`)`` | alias / full-id examples |
| `README.md:305` | `above Opus — because the reviewer is the pipeline's quality gate` | rationale prose |
| `README.md:314` | ``| `claude-opus-4-8` (any alias or full id) | `claude … --model claude-opus-4-8` |`` | explicit-value table row (full-id example) |
| `README.md:317–318` | ``both aliases (`opus`, `sonnet`) and full ids (`claude-opus-4-8`) are valid`` | the alias/full-id sentence the issue says stays |
| `tests-entrypoint-executor.sh:309–312` | `ALISSA_AGENT_MODEL=opus` … `--model opus` … `reviewer model: opus` | explicit-value pass-through assertion (case 2b) |
| `tests-entrypoint-executor.sh:316–317` | `ALISSA_AGENT_MODEL=claude-opus-4-8` … `--model claude-opus-4-8` | full-id pass-through assertion (case 2b) |

No hit is a *default* any more.

### Judgment calls

- The executor suite's `--model opus` assertion (old L268) was **the unset-default path** — `start_boot` uses `env -i` and never sets `ALISSA_AGENT_MODEL` — so it was flipped to `claude-fable-5`, not left as an explicit-value test. Since no suite covered explicit pass-through at all, case 2b was added (alias, full id, `default`, empty), per the issue's "add one assertion … if no such test exists" — four boots rather than one, because the issue's acceptance criteria list all four behaviours.
- Case 2/2b assert on the **rendered `command:` line only** (a `profile_cmd` helper + a `BASE_CMD` constant mirroring the entrypoint's), not a whole-file grep: after this change the agents.yaml header comment itself mentions both `--model` and `claude-fable-5`, so a file-wide `grep -F` would pass vacuously for the "no `--model`" and "no default leak" cases.
- Comment/prose sites gained a one-clause rationale ("the most capable generally-available Claude model, one tier above Opus") next to the new value, where the old comment had none — the issue asked to keep the existing rationale and change the value; the extra clause is the *why* behind the operator decision, so a future reader is not left guessing why an alias became a full id.
- Commit trailer is `Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>` (the model this session actually runs on) rather than the `Claude Opus 5` seen on recent `main` commits — truthful attribution over pattern-matching.
- The develop-daemon mirror of the README wording is out of scope per the issue; the sentence added here ("The unset default is `claude-fable-5` …") is written so it can be mirrored verbatim there.

- [ ] `tests-image-contract.sh` (check-image) was **not run locally** — this runner has no docker. The change to it is a one-string assertion; the ARG→ENV chain it reads (`ARG ALISSA_AGENT_MODEL` L292 → `ENV … ALISSA_AGENT_MODEL=${ALISSA_AGENT_MODEL}` L324) was verified by reading the Dockerfile, and `EXPECT_AGENTS_SHA` is computed from the repo's `agents.yaml` at run time (L104), so the header-comment edit cannot break the sha pin. CI's `check-image` job is the proof.
- [ ] The Railway runtime: after merge + redeploy the boot log must show `reviewer model: claude-fable-5 (ALISSA_AGENT_MODEL)` and the next round must start without a model-access error from the reviewer's claude login — and a runtime `ALISSA_AGENT_MODEL` set in the dashboard shadows this default. Operator step, as the issue says.
- [ ] That `claude-fable-5` is accepted by the reviewer account's `claude` login (model access is an account property; nothing in this repo can assert it).
- [ ] Round-1 `[nit]` (agents.yaml header: `# mount your own agents.yaml over this` orphaned at ~45 columns by the reflow) is `[triage:later]` → **TASK-1385697915**, sequenced after this PR merges: the rewrap touches the lines this PR changes, so a follow-up off `main` cut before the merge conflicts with it, and the approved branch is not pushed again. Exact patch staged as evidence on TASK-495451465.

### Verification

- `bash docker/claude/tests-entrypoint-executor.sh` — **PASS** (all cases incl. new 2b; 21s).
- Counterfactual mutants, each run against the suite then reverted: fallback `claude-fable-5` → `opus` ⇒ **FAIL** on exactly case 2's two new assertions (rendered line + log); `-` → `:-` ⇒ **FAIL** on exactly the empty-opt-out assertions of case 2b. Both mutants die on the intended clause and nothing else.
- The other six entrypoint suites (`tests-entrypoint-config/identity/auth/chown/prune/ui.sh`), since `entrypoint.sh` changed — **all PASS**, 0 FAIL lines.
- `bash check-style.sh alissa-tools-github-revloop` — exit 0; `bash check-types.sh …` — `Success: no issues found in 30 source files`; `PYTHONPATH=src/main bash tests-unit.sh …` — **920 passed** in 26.5s (the dist is untouched by this PR; run because CI runs them). `pycodestyle`/`mypy`/`coverage` had to be `pip install --user`ed on this runner first.
- CI on this PR at handoff: image contract, style, types, unit tests, wheel and version-bump **SUCCESS**; the entrypoint check result is whatever the rollup shows on the head `1da7703`.
- `grep -rn -i opus` over the repo: 9 sites, all enumerated above.

### Scope touched

- `docker/claude/Dockerfile`, `entrypoint.sh`, `agents.yaml`, `README.md`, `tests-image-contract.sh`, `tests-entrypoint-executor.sh` — all inside the issue's declared scope.
- `alissa-tools-github-revloop/…/version` — in the declared scope but **not** touched (bump not triggered, see above).
- No excursions.

### DIFF
--- a/docker/claude/agents.yaml
+++ b/docker/claude/agents.yaml
@@ -9,10 +9,11 @@
 # root) — the image runs as uid 1000, which satisfies it. Reviewer posture (CR6:
 # reviewers never write) is enforced by the round directive, not by this profile.
 #
-# To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default: opus) —
-# the entrypoint rewrites the `command:` below with `--model "$ALISSA_AGENT_MODEL"`
-# at container boot (set it to `default` or empty to omit the flag and inherit the
-# account default). To change flags instead, mount your own agents.yaml over this
+# To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default:
+# claude-fable-5) — the entrypoint rewrites the `command:` below with
+# `--model "$ALISSA_AGENT_MODEL"` at container boot (set it to `default` or empty
+# to omit the flag and inherit the account default). To change flags instead,
+# mount your own agents.yaml over this
 # path at runtime; a mounted file lacks the `alissa-managed:` marker below and is
 # used verbatim, so ALISSA_AGENT_MODEL does not touch it (your command wins).
 #
--- a/docker/claude/Dockerfile
+++ b/docker/claude/Dockerfile
@@ -285,9 +285,11 @@ ARG ALISSA_TASK_LIST_SELF_SCOPE=""
 ARG ALISSA_AGENT_PROFILE="claude"
 # Reviewer model pinned into the agents.yaml claude command at boot (the reviewer
 # is the quality gate, so it must not silently fall back to a smaller model on a
-# usage-throttled account). Alias (`opus`/`sonnet`) or full id (`claude-opus-4-8`);
+# usage-throttled account). Default `claude-fable-5`: the most capable generally
+# available Claude model, one tier above Opus (operator decision 2026-08-30).
+# Alias (`opus`/`sonnet`) or full id (`claude-opus-4-8`) pass through verbatim;
 # `default` or empty omits the --model flag. See the entrypoint (step 2e) + README.
-ARG ALISSA_AGENT_MODEL="opus"
+ARG ALISSA_AGENT_MODEL="claude-fable-5"
 ARG ALISSA_ON_MISSING_HUB="add"
 # Worker reconcile tick, seconds.
 ARG ALISSA_WORKER_INTERVAL="2"
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -300,7 +300,9 @@ downstream notices. The baked [`agents.yaml`](./agents.yaml) pins no model, so a
 reviewer inherits whatever the persisted `claude /login` account defaults to —
 and on plan-based accounts that default can **silently fall back to a smaller
 model** once a usage threshold is hit. `ALISSA_AGENT_MODEL` makes the model an
-explicit boot-time decision instead of a rebuild-time one.
+explicit boot-time decision instead of a rebuild-time one. The unset default is
+`claude-fable-5` — the most capable generally-available Claude model, one tier
+above Opus — because the reviewer is the pipeline's quality gate.
 
 At container boot the entrypoint appends `--model "$ALISSA_AGENT_MODEL"` to the
 claude profile's `command:` and logs the effective command (grep the startup log
@@ -308,7 +310,7 @@ for `effective reviewer command:`).
 
 | `ALISSA_AGENT_MODEL` | reviewer `command:` becomes |
 | --- | --- |
-| *(unset)* → default `opus` | `claude … --model opus` |
+| *(unset)* → default `claude-fable-5` | `claude … --model claude-fable-5` |
 | `claude-opus-4-8` (any alias or full id) | `claude … --model claude-opus-4-8` |
 | `default` *or* empty | `claude …` (no `--model` — restores account default) |
 
@@ -352,7 +354,7 @@ automatically; locally pass `--build-arg`):
 | `ALISSA_REVIEW_TASK_MISS_TTL_POLLS` | *daemon default* (currently 10) | how many polls a PR with **no** Alissa review task is taken on trust before the daemon searches the actor's task corpus for one again. That search is the widest read the daemon makes and the PR→task mapping can only cache an answer that exists, so an unmapped PR used to pay the whole corpus every poll, indefinitely. Trades **latency** for reads: a review task created inside the window is picked up at the re-arm, not the next poll. Floor `1` — there is no value that disables it, and `0` is refused at config load. **pass-through** — unset ⇒ library default. Needs `REVLOOP_VERSION >= 0.18.0` |
 | `ALISSA_TASK_LIST_SELF_SCOPE` | *(unset ⇒ library default `false`)* | `1`/`true`/`yes`/`on` narrows `alissa task list` to this actor's own rows (`--self`), dropping the sponsor's corpus; `0`/`false`/`no`/`off` is the explicit opposite and **anything else is refused** rather than rendered as `false`. Off by default on measured evidence: it saves ~4% of the payload, and a small minority of review tasks on the live fleet are owned by another actor — one the list cannot see is a round the daemon cannot count. Set it only where every review task is created by this daemon's own reviewer sessions. The other narrowings (status filter, digest view) are probed from the installed CLI and need no variable. **pass-through** — unset ⇒ library default. Needs `REVLOOP_VERSION >= 0.18.0` |
 | `ALISSA_AGENT_PROFILE` | `claude` | agent the worker launches (must name a profile in `agents.yaml`) |
-| `ALISSA_AGENT_MODEL` | `opus` | model pinned into the reviewer's claude command (see [Pinning the reviewer model](#pinning-the-reviewer-model)); `default` or empty omits the pin |
+| `ALISSA_AGENT_MODEL` | `claude-fable-5` | model pinned into the reviewer's claude command (see [Pinning the reviewer model](#pinning-the-reviewer-model)); `default` or empty omits the pin |
 | `ALISSA_ON_MISSING_HUB` | `add` | `add` hub-ifies on demand; `skip` to require a mounted workspace |
 | `ALISSA_WORKER_INTERVAL` | `2` | worker reconcile tick (seconds) |
 | `ALISSA_ENABLE_FIREWALL` | `0` | `1` raises the egress firewall (needs `--cap-add=NET_ADMIN`) |
@@ -801,8 +803,8 @@ the rest only mean anything on a service that already set those two.
 | `ALISSA_BRIDGE_POLL_SECONDS` | *(CLI default: 15)* | seconds between queue polls; maps to the CLI's `--interval`; **pass-through** |
 
 The model pin works exactly as it does for the daemon: `ALISSA_AGENT_MODEL`
-(default `opus`) is rewritten into the `claude` profile's `command:` at boot, and
-job sessions inherit it. The profile deliberately carries **no**
+(default `claude-fable-5`) is rewritten into the `claude` profile's `command:` at
+boot, and job sessions inherit it. The profile deliberately carries **no**
 `disable_alissa_code`, which is what makes the CLI launch it via `alissa code -y
 --handoff claude` — that wrapper is what registers the codeSession and its
 10-minute log checkpoints. Adding the flag would launch a bare `claude` and lose
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -580,7 +580,8 @@ fi
 # The reviewer is the pipeline's quality gate, but the baked claude profile pins
 # no model — so it inherits the persisted /login account's default, which can
 # silently fall back to a smaller model when a plan hits its usage threshold. We
-# pin it explicitly at boot: ALISSA_AGENT_MODEL (default `opus`) is appended to
+# pin it explicitly at boot: ALISSA_AGENT_MODEL (default `claude-fable-5`, the
+# most capable generally-available model, one tier above Opus) is appended to
 # the profile's `command:` as `--model <value>`. The value passes through
 # verbatim — aliases (`opus`, `sonnet`) and full ids (`claude-opus-4-8`) are both
 # valid, no allowlist. `default` or an empty value omits the flag entirely,
@@ -593,9 +594,9 @@ fi
 # it is pristine on every boot and this rewrite is idempotent.
 # -----------------------------------------------------------------------------
 AGENTS_YAML="${HOME}/.config/alissa/agents.yaml"
-# Default `opus` only when UNSET (use `-`, not `:-`): an explicitly empty value is
-# a valid opt-out and must NOT be re-defaulted back to opus.
-AGENT_MODEL="${ALISSA_AGENT_MODEL-opus}"
+# Default `claude-fable-5` only when UNSET (use `-`, not `:-`): an explicitly
+# empty value is a valid opt-out and must NOT be re-defaulted back to the pin.
+AGENT_MODEL="${ALISSA_AGENT_MODEL-claude-fable-5}"
 if [ ! -f "${AGENTS_YAML}" ]; then
   log "WARN: ${AGENTS_YAML} not found — skipping model pin (worker will fall back to a bare claude)"
 elif ! grep -q 'alissa-managed:' "${AGENTS_YAML}"; then
--- a/docker/claude/tests-entrypoint-executor.sh
+++ b/docker/claude/tests-entrypoint-executor.sh
@@ -17,6 +17,12 @@
 #                                            structural flags; no worker, no
 #                                            daemon, no revloop.config.json;
 #                                            identity + agent profile on the volume
+#                                            (model pin = the baked default, since
+#                                            the boot env carries no
+#                                            ALISSA_AGENT_MODEL)
+#   2b. the model pin is a DEFAULT        -> an explicit alias / full id lands
+#                                            verbatim; `default` and empty omit
+#                                            the flag instead of re-defaulting
 #   3. executor id required + validated   -> empty dies, bad slug dies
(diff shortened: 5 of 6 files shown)


alissa-app:
- 2026-08-31 00:26:17 UTC — `review-alissa-github-review-daemon-pr111-r1-349413` — round 1 of 22 — spawned

docker/claude/agents.yaml
```diff
@@ -9,10 +9,11 @@
...
 # reviewers never write) is enforced by the round directive, not by this profile.
 #
-# To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default: opus) —
-# the entrypoint rewrites the `command:` below with `--model "$ALISSA_AGENT_MODEL"`
-# at container boot (set it to `default` or empty to omit the flag and inherit the
-# account default). To change flags instead, mount your own agents.yaml over this
+# To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default:
+# claude-fable-5) — the entrypoint rewrites the `command:` below with
+# `--model "$ALISSA_AGENT_MODEL"` at container boot (set it to `default` or empty
+# to omit the flag and inherit the account default). To change flags instead,
```
alissa-app, line 15: [nit] The reflow left an orphan here: `# mount your own agents.yaml over this` stops at ~45 columns while every other line in this block runs to ~78, so the paragraph now reads as if it broke mid-sentence. Rewrapping the last three lines as e.g.

```
# to omit the flag and inherit the account default). To change flags instead,
# mount your own agents.yaml over this path at runtime; a mounted file lacks the
# `alissa-managed:` marker below and is used verbatim, so ALISSA_AGENT_MODEL does
# not touch it (your command wins).
```

restores it. Comment-only, in a file whose *content* hash is pinned by `tests-image-contract.sh` at run time (so a rewrap is free) — purely your discretion.

alissa-app (approved):

Fresh-instance review per `alissa-code-review` (CR1–CR9). Reviewed head `1da7703e1b359afa62f88a2d6b8352e05bda2cb3`. Review task: `TASK-1508028164`. Origin `TASK-172658489` (issue #110), implementation `TASK-495451465`.

A clean, tightly-scoped value change. The one thing that could have gone wrong here — flipping a default in a way that also changes the *knob's semantics* — did not: the `-` (not `:-`) fallback, the `default`/empty opt-out, verbatim pass-through, and the `alissa-managed:` marker gate are all untouched, and the PR adds the test that pins each of them.

### Independent verification (not taken from the PR description)

I ran the suite myself in a throwaway worktree at this head and reproduced the counterfactuals:

- `bash docker/claude/tests-entrypoint-executor.sh` → **PASS**, with case 2 and the new case 2b both executing (7 assertions in 2b).
- Mutant `${ALISSA_AGENT_MODEL-claude-fable-5}` → `${ALISSA_AGENT_MODEL-opus}` ⇒ **FAIL** on exactly case 2's two assertions, and nothing else.
- Mutant `-` → `:-` ⇒ **FAIL** on exactly case 2b's two empty-opt-out assertions, and nothing else.

So the new assertions genuinely bind to the clause they claim to guard, in both directions. The `default`/empty cases compare the **whole** rendered `command:` line and are paired with a boot-log assertion, so they cannot pass vacuously if step 2e were skipped altogether — a real trap, well avoided (the whole-file `grep -F` the old test used would have gone vacuous here, since the new `agents.yaml` header itself contains both `--model` and `claude-fable-5`).

`git grep -in opus` at this head returns 15 lines across 4 files; every one falls inside the 9 entries the PR description enumerates, and none of them is a default. `EXPECT_AGENTS_SHA` in `tests-image-contract.sh` is computed from the repo's `agents.yaml` at run time (L104), so the header-comment edit cannot break the sha pin — and `check-image` is green at this head, which proves it.

### Rubric

- **Correctness** — nothing found. Every site the issue names is flipped (`Dockerfile` ARG L292, `entrypoint.sh` L599, `agents.yaml` header, README pinning table + env table + executor paragraph, both shell suites) and no `opus` survives as a default anywhere. `revloop-config.sh` correctly left alone (it carries no model default). No version bump needed: `check-version-bump.yaml` guards the `alissa-tools-github-revloop/` dist subtree only, and it is green here. The judgment call to flip the old L268 assertion rather than preserve it is right — `start_boot` runs under `env -i` and never sets `ALISSA_AGENT_MODEL`, so that assertion *was* the unset path, and explicit pass-through had no coverage at all before case 2b.
- **Coding practices** — two `[nit]`s below, both cosmetic and both on non-shipped surfaces. The `BASE_CMD` mirror in the test is a deliberate duplicate of the entrypoint's constant; that is the right trade here (drift fails loudly with a diffable message rather than silently), and it is documented at the definition.
- **Security** — nothing found. The model value still reaches `agents.yaml` through the `python3` rewrite rather than `sed`, so an arbitrary value still cannot collide with a substitution metacharacter; the new default is a fixed literal with no metacharacters anyway. The `alissa-managed:` marker gate is unchanged, so a mounted custom `agents.yaml` is still used verbatim. No new network, credential, or privilege surface.
- **Performance** — nothing found on any shipped path. Case 2b adds four entrypoint boots to `tests-entrypoint-executor.sh` (~6s locally); `check-entrypoint` allots `timeout-minutes: 6` for this suite and is green at this head, so there is comfortable headroom.
- **Side-effects / blast radius** — the ARG→ENV chain means every deployment that does **not** set `ALISSA_AGENT_MODEL` at runtime moves from `opus` to `claude-fable-5` on its next redeploy, with no in-container fallback if the reviewer's `claude` login lacks access to that model. That is the operator decision issue #110 records, and the PR flags it as the operator's post-merge gate rather than papering over it — the right call, and the rollback lever is a single dashboard variable (`ALISSA_AGENT_MODEL=opus`, or `default` to restore the account default), already documented in the README table this PR updates. Not a finding; noting it so the merge decision is made with it in view. No contract, schema, or cross-repo consumer change; the develop-daemon README mirror is out of scope per the issue and the added sentence is worded to be mirrored verbatim.

### Findings

- 1 × `[nit]` inline on `docker/claude/agents.yaml`.
- 1 × `[nit]`, not inlineable because the file is outside this diff: **`.github/workflows/check-entrypoint.yaml:81`** — the step comment reads *"Boots the entrypoint nine times across both roles."* The suite booted ~14 times before this PR and ~18 after it, so the count was already stale and this PR widens the gap. Pre-existing drift, entirely your discretion, and arguably better handled as a standalone sweep than smuggled into this diff.

No blockers, no majors, no open questions. `[nit]`s are the implementer's discretion and do not gate the merge.

### CI at the reviewed head

Rollup read through the **full check-runs API** (not the Actions fallback) on `1da7703e1b359afa62f88a2d6b8352e05bda2cb3`, immediately before submitting: **8 contexts, all concluded** — Version Bump, Python Types, Entrypoint config renderer + console wiring, Unit-Test, Python Style, Wheel Package, Image contract all `success`; `[code]smith` `skipped`. No legacy commit statuses. Nothing failed, nothing still running.

docker/claude/agents.yaml
RHDZMOTA: [triage:later] Agreed — the reflow orphaned that line at ~45 columns. Deferred on ORDERING, not disagreement: this branch is approved (no further pushes), and a follow-up off `main` right now would rewrite the very lines this PR changes — a conflict for whichever merges second. Registered TASK-1385697915, sequenced after this PR merges, with the exact rewrap (your four lines) staged as evidence on TASK-495451465. If you would rather have it in this PR, a fresh review round on it is the legitimate alternative — say so and I will push the rewrap here instead.

PR #113, opened 2026-09-01
TASK-1776360305: push loop events (rounds, verdicts, cap-outs, holds, stalls, grants, reaps) to Studio once per pass (issue #112)

### LINKED ISSUE #112: Revloop: push loop events (rounds spawned, verdicts posted, cap-outs, stability holds, stalls, checks holds) to Studio once per pass
## Why

Revloop's per-round record — `verdict_posts` (the richest per-round table in the fleet: first seen, attempts, posted/abandoned, checks-held state, review URL), `spawns`, `escalations` (CR9 cap-outs), `pings` (stalled/stability/checks-unsettled), `grants` — lives only in `.revloop/state.db` on a Railway volume; its console `/api/state` shows the current pass. The Factory console (`factory.alissa.app`) needs rounds-to-approve, verdict mix, cap-outs and stability holds over time. Studio now ingests loop telemetry (`POST /v1/loop-events`, see `API.md` → "Loop telemetry — ingest" in fahera-mx/studio.alissa.app; this task depends on it). This task adds the writer. Orcloop and devloop get siblings.

## Contract

- Config `loop_events_enabled` (default `false`), CLI `--loop-events` / `--no-loop-events`, env `ALISSA_REV_LOOP_EVENTS_ENABLED`, plus `alissa_endpoint` (default `https://api.alissa.app`) if the daemon lacks one. Revloop talks to Alissa through the CLI today (`alissa.py`); the CLI has no loop-events command and cannot supply the actor identity the API needs, so add a **minimal REST client** (`alissa_client.py`, the same shape devloop adopted in PR #91: the daemon's Alissa token from its existing credential source, one `post_loop_events(events)` method, bounded timeouts). The CLI adapter stays untouched for everything else. Container: ARG/ENV pair + render + README row per this repo's container conventions.
- At the end of each poll pass, POST **one** batch (≤200; split if larger). Best-effort, never fatal: one WARN on failure, the pass completes; no retry queue (deterministic keys make re-emission harmless).
- Events (`seat: "revloop"`, `at` = the ledger timestamp, `repo` canonical `owner/name`, `prNumber`, `dedupeKey` from ledger keys):
  - `round.spawned` — `spawns` (`session`, `round`, `data.headSha`, `data.taskRef`); key `revloop:round.spawned:<repo>:<pr>:<round>:<headSha>`
  - `round.verdict` — from `verdict_posts` when `posted_at` is set: `round`, `data.verdict` (`approve` | `request_changes`, parsed from the review task's CR6 envelope / the native verdict body), `data.headSha`, `data.reviewUrl`, `data.attempts`, `data.checksHeldMs` (if `checks_held_at` was set); key `revloop:round.verdict:<repo>:<pr>:<round>:<headSha>`
  - `round.abandoned` — `abandoned_at` set (`reason` = `last_error`)
  - `round.capped` — `escalations` (CR9 cap-out) with `round`, `data.headSha`
  - `stability.hold` — `stability:` pings / `stability_notices` (`data.rcRounds`, `data.grantsSeen`)
  - `stalled` — `stalled:<session>` pings (`session`)
  - `checks.held` — `checks-unsettled:r<k>@<sha>` pings and `spawn_checks_holds` (`round`, `data.headSha`)
  - `grant` — operator re-entry acks from `grants` (`data.author`, `data.rounds`)
  - `reap` — `reaps` (`session`)
- No vitals events.

## Out of scope

Console changes; orcloop/devloop (siblings); Studio-side changes; changing review behavior, round counting, or the verdict-of-record state machine.

## Testing

Unit tests (`tests-unit.sh` conventions): derivation from fixture ledger rows yields documented kinds with deterministic keys; a `verdict_posts` row without `posted_at` emits nothing until it is set; batches split at 200; WARN-and-continue on failure; disabled by default is silent; the REST client sends the bearer token and honors timeouts; container config/drift guards for the new keys. Style/type checks green; version bump per the repo's gate.

## Definition of Done

Revloop, when loop_events_enabled, pushes one idempotent batch of loop events per pass to POST /v1/loop-events through a new minimal REST client — round.spawned, round.verdict (verdict/headSha/reviewUrl/attempts), round.abandoned, round.capped, stability.hold, stalled, checks.held, grant, reap with deterministic ledger-derived dedupeKeys, best-effort and never fatal — with config key, CLI flag, env var, container ARG/ENV + render, README and tests. A PR is opened draft-first on a TASK branch, flipped to ready-for-review, driven through the review loop to approve, with the PR URL attached as evidence.

## Acceptance criteria

- [ ] Automated: tests cover deterministic keys from fixture rows, verdict emission only once posted_at is set, batch splitting, WARN-and-continue, disabled-by-default silence, REST client auth/timeouts, and container config paths; style and type checks green.
- [ ] Manual: with loop events enabled on the deployed Alissa RevLoop Daemon, GET /v1/loop-events?seat=revloop on api.alissa.app shows round.spawned and round.verdict events for a real PR with matching round numbers and verdicts, and consecutive passes add no duplicates.

Autonomous-Scope: alissa-tools-github-revloop/**, revloop.config.example.json, README.md, docker/**
Alissa-Task: TASK-774046670

RHDZMOTA:
Closes #112

**Tasks:** origin TASK-774046670 · implementation TASK-1776360305

Revloop, when `loop_events_enabled`, derives loop events from the local ledger at the end of each poll pass and POSTs one idempotent batch (split at the ingest cap of 200) to the Alissa API's `POST /v1/loop-events` through a new minimal REST client (`alissa_client.py`, the shape devloop adopted): `round.spawned`, `round.verdict` (verdict/headSha/reviewUrl/attempts/checksHeldMs), `round.abandoned`, `round.capped`, `stability.hold`, `stalled`, `checks.held` (both gates), `grant`, `reap` — each with a deterministic ledger-derived `dedupeKey` and the ledger's own timestamp as `at`. Best-effort and never fatal: one WARN on failure, the pass completes, no retry queue — the emitter's in-memory watermark does not advance on failure, so the next pass re-derives and re-sends, and the server's `(user, dedupeKey)` idempotence absorbs the overlap.

Knobs: config `loop_events_enabled` (default `false`) + `alissa_endpoint` (default `https://api.alissa.app`), CLI `--loop-events` / `--no-loop-events` + `--alissa-endpoint`, env `ALISSA_REV_LOOP_EVENTS_ENABLED` (wins over file and CLI, like `ALISSA_REVIEW_TASK_BOW`). Container: `ARG`/`ENV` pair, boolean render pass-through, README row, drift guards. `verdict_posts` gains a `verdict` column (migrated in place via the existing `_ADDED_COLUMNS` path). Version 0.27.0 → 0.28.0.

### Judgment calls

- **Restart backfill.** The emitter's watermark is in-memory and starts at zero, so the first enabled pass re-sends the whole ledger, batched and server-deduped. Read as intended ("no retry queue — deterministic keys make re-emission harmless") and it is what seeds the Factory console's history. Honest bound (round-1 minor): none of the seven tables read is pruned, so the backfill — and the per-pass re-derivation — is bounded by the deployment's actual row counts (hundreds over a deployment's life today); SQL-side `since` bounds or a persisted watermark are the named upgrade path. A quiet ledger makes no request at all: the keys sent at the watermark second are remembered and dropped from the next derivation (round-1 minor), so the inclusive boundary cannot become a standing re-POST.
- **`data.verdict` comes from a new ledger column, not a per-pass envelope re-read.** The issue says the verdict is "parsed from the review task's CR6 envelope / the native verdict body" — that parse happens at post time (`latest_verdict`), long before the end-of-pass emitter runs, so `record_verdict_post` now stamps it into `verdict_posts.verdict`. Rows that predate the column omit `data.verdict` rather than invent one. The verdict-of-record state machine is untouched.
- **Ping-kind formats follow the code, not the issue's rendering.** The issue writes `checks-unsettled:r<k>@<sha>`; the ledger's actual kind is `checks-unsettled:<k>:<sha>` (`loop.checks_unsettled_kind`). Parsed as shipped, with prefix drift guards pinning the parser to loop's own kind builders.
- **`round.capped` round number is derived, episode-correctly.** `escalations` carries no round, so it is read from the newest spawn **at or before** `escalated_at` (round-1 minor) and omitted when none qualifies. The dedupe key folds the row's own `escalated_at` (round-1 major), so a re-cap-out on an unmoved head — the operator re-entry path — derives a distinct event instead of being swallowed as a server-side duplicate.
- **`checks.held` from both gates, distinct keys.** The verdict-side ping and the pre-spawn hold can name the same (PR, round, head); they get one kind but distinct dedupe keys (`:spawn` suffix) and `data.gate` (`verdict` | `spawn`), so one gate's event cannot swallow the other's server-side.
- **`stability.hold` payload is split by provenance** (round-1 minor): `data.headSha` and `data.grantsSeen` come from the ping kind itself and are episode-correct even on a backfill; `data.rcRounds` and `round` come from the `stability_notices` join and are documented as current-at-derivation (the table is REPLACEd per episode and holds nothing episode-scoped for them).
- **Comment-dedupe ping kinds emit nothing.** `activity-deferred:`, `capout:`, `checks-hold:`, `verdict-post-failed:` dedupe GitHub-side comments and are not in the issue's event list.
- **Dry-run never posts, even when enabled.** An outbound POST is an act; dry-run's contract is decide-and-log. (Nothing new lands in the ledger during dry-run anyway.)
- **Container gets the one ARG/ENV pair for the enable knob only**; `alissa_endpoint` stays config/CLI (the issue asked for an "ARG/ENV pair" singular, and the token already reaches the container as `ALISSA_API_TOKEN`).
- **`grant`, `reap`, `round.abandoned` are emitted per the issue's event list** even though Studio's v1 vocabulary table for revloop does not enumerate them — the ingest validates kind *shape*, not membership, and the API doc says to add kinds without a Studio deploy.
- Env garbage (`ALISSA_REV_LOOP_EVENTS_ENABLED=enable`) is refused at startup (`config error`, exit 2) rather than read as false, matching the container renderer's boolean contract; empty means unset (the Dockerfile bakes empty ENV).
- **The verdict-kind coverage bound is documented, not silently shipped (round-2 major).** Deriving `round.verdict` from a session-covering source means persisting per-round rows for rounds the daemon did not post — new semantics for the obligation table, i.e. the round bookkeeping issue #112 declares out of scope — and reading GitHub at emit time would break the emitter's pure-ledger design. So the bound is stated (README table, docstring, *Not verified*) and the widening is TASK-1086576582.
- **Round-1 hardening of the client and emitter**: the client opens through a redirect-refusing `OpenerDirector` (a 30x surfaces as an error instead of replaying the bearer token toward `Location`), `alissa_endpoint` must be https (http allowed toward loopback only, for a local stub), an `AlissaAuthError` warns once and latches the emitter off until restart (transient errors keep the re-send path), and the `reap` dedupe key folds `reaped_at` so it does not lean on session-name nonce-uniqueness.

- [ ] **`round.verdict`/`round.abandoned` coverage bound (round-2 major)**: `verdict_posts` is the per-round *obligation* record — a row exists only when the daemon posted the native fallback verdict because the reviewer session defaulted. On a fleet whose sessions post every review (0 of the last 58 rounds on this repo took the fallback path), those two series are **expected to be empty**; the manual acceptance criterion on the origin task cannot pass until session-posted rounds are covered, which is registered as TASK-1086576582 (soft-linked downstream of TASK-1776360305) and needs the criterion amended with the operator. The bound is stated in the README telemetry table and in `_verdict_events`' docstring.
- [ ] The manual acceptance criterion: with loop events enabled on the deployed daemon, `GET /v1/loop-events?seat=revloop` shows `round.spawned`/`round.verdict` for a real PR with matching rounds/verdicts and no duplicates across passes — no deployed daemon or live Railway volume in this environment. Unit tests pin derivation, idempotent keys, watermark and re-send behaviour.
- [ ] No POST was made against the real `api.alissa.app` (tests use a fake client / patched urllib). The ingest contract was read from studio.alissa.app `API.md` → "Loop telemetry — ingest" as of 2026-09-01.
- [ ] Container image build/boot (no docker on this runner). The renderer and drift guards pass `tests-entrypoint-config.sh` locally; note CI's library cross-check installs the Dockerfile-pinned 0.16.14, which is why that block keeps the new env var unset.
- [ ] That the deployment's `ALISSA_API_TOKEN` acts as the operator principal (a `bridge_job` token gets 403 on this POST) — a deployment property.

### Verification

- `tests-unit.sh alissa-tools-github-revloop`: **982 passed** (62 new in `test_loop_events.py`), with `PYTHONPATH=src/main` and a fresh user-site install — re-run in full after the round-1 fixes.
- `tests-coverage.sh`: 95% total; `alissa_client.py` 93%, `loop_events.py` 98%.
- `check-style.sh`: exit 0. `check-types.sh`: mypy `Success: no issues found in 33 source files`.
- `docker/claude/tests-entrypoint-config.sh`: ALL PASS, including the new boolean render spellings and the refusal of a non-boolean value.
- The shipped `revloop.config.example.json` loads through `Config.build`; `--help` renders the new flags.

### Scope touched

Declared scope: `alissa-tools-github-revloop/**`, `revloop.config.example.json`, `README.md`, `docker/**` — all changes inside it, no excursions.

- `alissa-tools-github-revloop/src/main/.../revloop/`: **new** `alissa_client.py`, **new** `loop_events.py`; `state.py` (verdict column + migration + telemetry readers, `read_grants` grows an unfiltered form), `config.py` (two keys + env layer), `__main__.py` (flags), `loop.py` (emitter wiring, end-of-pass call, verdict arg at the one `record_verdict_post` site), `version` 0.28.0.
- **new** `src/test/.../test_loop_events.py`.
- `revloop.config.example.json`, `README.md` (settings rows + *Loop telemetry (Studio ingest)* section).
- `docker/claude/`: `Dockerfile` (ARG/ENV), `revloop-config.sh` (boolean render), `README.md` (env row), `tests-entrypoint-config.sh` (drift guards).

---
<a href="https://app.blacksmith.sh/fahera-mx/codesmith/alissa-github-review-daemon/pr/113"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1790822528&installation_model_id=434695&pr_number=113&repository=fahera-mx%2Falissa-github-review-daemon&return_to=https%3A%2F%2Fgithub.com%2Ffahera-mx%2Falissa-github-review-daemon%2Fpull%2F113&signature=27b7795caff091a435af6bf56e9253dab7ceb2f950d745f4b7c84f0a3167a63e"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


alissa-app:
- 2026-09-01 02:45:38 UTC — `review-alissa-github-review-daemon-pr113-r1-e00554` — round 1 of 22 — spawned
- 2026-09-01 03:06:37 UTC — `review-alissa-github-review-daemon-pr113-r2-ae0694` — round 2 of 22 — spawned
- 2026-09-01 03:27:13 UTC — `review-alissa-github-review-daemon-pr113-r3-5a7f2f` — round 3 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,407 @@
...
+    round is read from the PR's newest spawn row — the round in flight when
+    the cap was hit — and omitted when the ledger has no spawn to say."""
+    out = []
+    for row in rows:
+        repo, number = row["repo"], int(row["number"])
+        head = row["head_sha"] or ""
+        out.append(_event(
+            "round.capped",
+            int(row["escalated_at"]),
+            f"revloop:round.capped:{repo}:{number}:{head}",
```
alissa-app: [major] This dedupe key cannot distinguish a **re**-cap-out on an unmoved head, so the second one is silently swallowed server-side.

`escalations` is `PRIMARY KEY (repo, number, head_sha)` and `record_escalation` is `INSERT OR REPLACE` (`state.py`), so a re-escalation overwrites the row in place. And re-escalation on the same head is a designed path, not a corner: `loop._escalation_owed` returns True a second time when rounds granted by an operator ack are consumed without an approve, and `capout_kind(head_sha, granted)` exists precisely because — quoting it — "rounds run on the operator's own authority and end without an approve on the very SAME head, and that is a new decision, not a repeat of the page the operator already answered."

The derived key folds nothing that changes between those two cap-outs. The row is re-derived (new `escalated_at` >= the watermark, so it *is* re-sent) and the ingest's `(user, dedupeKey)` idempotence drops it. Studio therefore shows one cap-out where the loop paged twice, and the count under-reports exactly the re-entry case — while "cap-outs over time" is one of the four metrics the origin task names as the reason this writer exists.

This is the same collision class the PR solves correctly one function down for `checks.held` (`:spawn` suffix + `data.gate`, with `test_checks_held_derives_from_both_gates_with_distinct_keys` pinning it). The asymmetry reads as an oversight rather than a decision — nothing in the PR body's judgment calls addresses it.

Suggested fix: fold the row's own `escalated_at` into the key — `revloop:round.capped:{repo}:{number}:{head}:{escalated_at}`. It stays deterministic and ledger-derived (re-deriving the same row always yields the same key), and it changes exactly when the ledger records a new cap-out decision. Folding the PR's granted total instead would mirror `capout_kind`, but it is read at derive time rather than cap time, so a later grant would re-key an old row during a backfill. Worth a test next to the `checks.held` one: two `record_escalation` calls on one head, two distinct keys.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,407 @@
...
+    strictly-greater filter advanced to the first one's stamp would lose the
+    second forever, while inclusion merely re-sends a key the API dedupes.
+    """
+    spawns = state.read_spawns()
+    newest_round: "dict[tuple[str, int], int]" = {}
+    for row in spawns:
+        key = (row["repo"], int(row["number"]))
+        newest_round[key] = max(newest_round.get(key, 0), int(row["round"]))
+
+    notices = {
```
alissa-app, line 373: [minor] Events derived during the restart **backfill** are joined against tables that only hold *current* state, so they carry today's numbers rather than the episode's.

Two places:

- `_capped_events` takes `round` from `newest_round`, the max round across *all* spawn rows for the PR at derive time. Its docstring calls that "the round in flight when the cap was hit" — true for a cap-out derived in the pass that wrote it, false for one re-derived weeks later, when the PR has moved on to a higher round.
- `_ping_events` joins this single per-PR `stability_notices` row onto **every** `stability:` ping for that PR. `stability_notices` is REPLACEd per episode, so an episode-1 ping backfilled today gets episode-3's `rcRounds`/`grantsSeen`/`round`. The docstring's justification — "which is also what the guard itself decides from" — holds only for the current episode.

It matters because the backfill is a first-class documented behaviour and the ingest keeps the **first** value seen per key: a fresh Studio would store the wrong numbers permanently, with no later correction.

The stability half is fixable without a new query — `stability_kind` already encodes `head:base:granted`, and `granted` is `grantsSeen`; only `rcRounds` needs the join. For the cap-out round, selecting the newest spawn at or before `escalated_at` would be episode-correct. If neither is worth the code, weaken both docstrings to say the values are current-at-derivation, so the next reader doesn't rely on a guarantee that isn't there.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,407 @@
...
+  and re-sends, and the deterministic keys make the re-emission land as silent
+  duplicates server-side (idempotent on `(user, dedupeKey)`).
+
+* **The watermark is in-memory and starts at zero.** Each successful emission
+  advances it to the newest ledger stamp sent; rows at or after it are
+  (re-)sent, which re-posts at most the boundary second's rows per pass —
+  duplicates by construction, harmless by contract. A daemon restart resets it
+  and the first pass re-sends the WHOLE ledger, batched: that is the backfill,
+  not a bug — Studio dedupes every previously-seen key and keeps the history a
+  fresh console needs, and the ledger's own retention is what bounds it.
```
alissa-app: [minor] "the ledger's own retention is what bounds it" — there is no such retention on any table this module reads.

`state.py` is explicit in the other direction: `pings` are "NEVER pruned" (the comment above `read_spawns`), and `grants`, `escalations` and `stability_notices` carry the same note ("Rows are never pruned, unlike poll_snapshots and like grants"). `spawns` and `verdict_posts` are never pruned either. `poll_snapshots` is the only table with a retention policy, and this module doesn't read it.

So the sentence that justifies the unbounded design is the one thing that isn't true. Two consequences follow from it, both monotonic in the life of the volume: the restart backfill re-POSTs the entire history, and *every* pass rebuilds an event dict for every row of seven tables and then discards all but the tail (`derive_events` filters in Python, after construction).

Neither is urgent at today's row counts — this is a note about the design rationale, not a live problem. But the fix is cheap in either direction: push the `since` bound into the SQL (each reader already takes a `WHERE`), or persist the watermark in the ledger so a restart doesn't backfill at all. At minimum, correct the sentence so a future reader doesn't add rows on the strength of a retention policy that doesn't exist.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,407 @@
...
+            # stays put so the next pass re-derives and re-sends, and the
+            # deterministic keys make the overlap land as duplicates.
+            log.warning(
+                "loop-events: post failed after %d of %d event(s) (%s) — "
+                "telemetry is best-effort, the pass completes; the next "
+                "pass re-sends (idempotent keys)",
+                sent, len(events), exc,
+            )
+            return False
+        self._since = max(int(event["at"]) // 1000 for event in events)
```
alissa-app: [minor] On a ledger that stops changing, this re-POSTs the same batch every pass, forever.

`derive_events` filters `at >= since` and the watermark is set to `max(at)`, so the newest event is always re-derived and re-sent. When nothing new lands — the normal state of a quiet deployment overnight — every poll pass makes one authenticated outbound request carrying zero new information, and logs `loop-events: N event(s) pushed` at INFO to say so.

The inclusive boundary itself is well argued (two rows can share a second, and a strict `>` would lose the second one), so this isn't a request to change the comparison. It is cheap to keep the boundary inclusive *and* stop the standing re-post: remember the dedupe keys emitted at the watermark second and drop them from the next batch, so a pass with nothing new derives an empty batch and returns at the `if not events` guard.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa_client.py
```diff
@@ -0,0 +1,168 @@
...
+            method="POST",
+            data=json.dumps(payload).encode("utf-8"),
+            headers={
+                "Accept": "application/json",
+                "Content-Type": "application/json",
+                "Authorization": f"Bearer {self._token}",
+            },
+        )
+        try:
+            with urllib.request.urlopen(req, timeout=self._timeout) as resp:
```
alissa-app: [minor] The bearer token rides urllib's default redirect handling, and the endpoint is not constrained to https.

`urllib.request.urlopen` uses the default opener, which installs `HTTPRedirectHandler`. That handler copies every header except `Content-Length`/`Content-Type` onto the redirected request — `Authorization` included — and does so across hosts. A 30x from whatever `alissa_endpoint` names therefore hands `ALISSA_API_TOKEN` to the redirect target. (301/302/303 also downgrade the POST to a GET, so such a hop would look like a success while ingesting nothing.)

`config.py` accepts `alissa_endpoint` as any non-empty string, so `http://…` is configurable too and puts the same token on a cleartext hop.

Both close with one opener: build an `OpenerDirector` whose redirect handler refuses (or strips `Authorization` when the host changes) and use it here instead of the module-level `urlopen`; and reject a non-`https` endpoint at config load unless it is loopback. The default endpoint is trusted, so this is defence for the staging-redirect knob the config deliberately exposes, not a live hole.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,407 @@
...
+                chunk = events[start:start + MAX_EVENTS_PER_POST]
+                result = self._client.post_loop_events(chunk)
+                sent += len(chunk)
+                log.debug(
+                    "loop-events: posted %d event(s) (accepted=%s, "
+                    "duplicates=%s)",
+                    len(chunk), result.get("accepted"),
+                    result.get("duplicates"),
+                )
+        except AlissaError as exc:
```
alissa-app, line 481: [minor] The auth/transient split is built but never acted on — a misconfigured token warns once per poll pass, forever.

`AlissaAuthError`'s own docstring says it is "Permanent and operator-fixable — retrying it every pass only writes the same warning again", and `alissa_client`'s module docstring says the taxonomy exists so the log can distinguish "your token is wrong" from "the API blinked". But `emit_once` catches the base `AlissaError` and treats both identically: a daemon started without `ALISSA_API_TOKEN`, or with a `bridge_job` token that 403s — which the PR body itself flags as an unverified deployment property — re-derives the whole ledger and writes an identical WARN every poll interval for as long as it runs.

Since the taxonomy already separates them, the emitter can act on it: catch `AlissaAuthError` separately, warn once, and stop attempting until the process restarts (or back off). That keeps "best-effort, never fatal" while making a misconfiguration one loud line instead of a permanent drip.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,407 @@
...
+        for row in rows
+    ]
+
+
+def _reap_events(rows: "list[dict]") -> "list[tuple[int, dict]]":
+    return [
+        _event(
+            "reap",
+            int(row["reaped_at"]),
+            f"revloop:reap:{row['session']}",
```
alissa-app: [nit] `revloop:reap:<session>` carries no repo/PR and no timestamp, while `reaps` is `INSERT OR REPLACE` keyed on `session` — so a session name that is reaped twice yields one event at the first `reaped_at`. Session names are nonce-unique per spawn today, so this is theoretical; adding `reaped_at` to the key would make it not depend on that.

alissa-app (changes requested):
## Round 1 — `request_changes`

**Reviewed head:** `4b6541a9bad0693f81dc24ab614c75bfed6636fa` · **Review task:** TASK-641453357 · **Origin:** TASK-774046670

**CI at the reviewed head:** green — 8 contexts, all concluded (7 success, `[code]smith` skipped); no legacy statuses. The verdict below is not about CI.

This is a well-built change and the delivery note is unusually honest about its own seams. Everything the origin task's definition-of-done enumerates is present: the nine event kinds, the ≤200 batching, the stdlib-only REST client, the config key / CLI flag / env var / container ARG+ENV+render, the README section, and 43 new tests that assert behaviour rather than implementation. The `verdict` column migrates through the existing `_ADDED_COLUMNS` path, the ping-kind prefixes are pinned against `loop`'s own builders so a rename fails the suite, and `_emit_loop_events` is correctly last-in-pass, dry-run-skipped and wrapped so telemetry can never take down a poll.

One `[major]` holds the round: the `round.capped` dedupe key collides with itself on the re-entry path the loop treats as a distinct decision, which under-reports one of the four metrics this writer exists to produce. It is a one-line key change plus a test, alongside the `checks.held` case the PR already got right.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 |
| minor | 5 |
| nit | 1 |
| question | 0 |

- **[major]** `round.capped` dedupe key omits everything that distinguishes a re-cap-out on the same head (`loop_events.py:186`)
- **[minor]** backfilled events join against current-state tables (`loop_events.py:316`)
- **[minor]** the "ledger's own retention" that bounds the backfill does not exist (`loop_events.py:31`)
- **[minor]** an idle daemon re-POSTs the boundary batch every pass forever (`loop_events.py:391`)
- **[minor]** bearer token rides urllib's default redirect handling; endpoint not constrained to https (`alissa_client.py:108`)
- **[minor]** the auth/transient taxonomy is built but never acted on (`loop_events.py:380`)
- **[nit]** `reap` dedupe key has no repo/PR and no timestamp (`loop_events.py:295`)

### Rubric

- **Correctness** — the DoD is met item by item; derivation, batching, watermark and dry-run behaviour are all pinned by tests. One defect: the `round.capped` key collision above. Two accuracy gaps in backfilled events. `checksHeldMs`, the `posted_at`/`abandoned_at` split, and the boundary-inclusive `since` filter all read correctly.
- **Coding practices** — matches this repo's idiom closely (comment density, docstring-carries-the-reasoning style, reader-per-table in `state.py`, the `_ADDED_COLUMNS` migration path, the container's boolean pass-through contract). Branch, PR title and version bump follow C1–C4. No dead code, no untracked TODOs, no new runtime dependency. The `read_grants` widening keeps its existing caller unchanged. Only finding: the retention sentence that isn't true.
- **Security** — no secrets in the diff; the token is read from the CLI's existing `ALISSA_API_TOKEN` and never logged (error details carry status + response body only). No injection surface — the payload is `json.dumps`'d and the only SQL added is parameterised or literal. One finding: redirect header propagation and the unconstrained endpoint scheme (`alissa_client.py:108`).
- **Performance** — bounded timeouts, one batch per pass, splitting at the ingest cap, and the disabled default costs nothing (no client, no read). Two findings, both efficiency rather than defect: the full-table re-derivation per pass under a retention policy that does not exist, and the standing re-POST on an idle ledger.
- **Side-effects / blast radius** — the new outbound write is off by default and the container will not start posting on an image upgrade, which is the right call. The `verdict` column is additive and `ALTER`ed in place by the idempotent existing path; no consumer in `src/main` reads `read_verdict_posts` besides this module, so the console contract is untouched. Dry-run correctly posts nothing. Failure containment holds at three layers (client taxonomy → emitter WARN → `_emit_loop_events` catch-all), so a telemetry failure cannot reach the poll. The `read_grants` signature widening is backward-compatible for its one caller.

Fix the `round.capped` key (and add the two-escalations test). The `[minor]`s can be fixed in this PR or registered as follow-up tasks — either satisfies them; they do not gate the merge.

RHDZMOTA:
- 2026-09-01 02:56:01 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr113-r1-a1` — fix edge — spawned
- 2026-09-01 03:16:08 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr113-r2-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Agreed — this is exactly the collision class the `checks.held` derivation avoids, and the asymmetry was an oversight, not a decision. Fixing as suggested: the row's own `escalated_at` folds into the key (`revloop:round.capped:{repo}:{pr}:{head}:{escalated_at}`), which stays deterministic under re-derivation and changes precisely when `record_escalation` REPLACEs the row with a new cap-out decision. Taking your alternative's warning too: the granted total is not used, for the backfill re-keying reason you give. Test added beside the `checks.held` one: two `record_escalation` calls on one head at different stamps yield two distinct keys. Fix lands in the next push to this branch; sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Both halves accepted, split as you suggest. Cap-out `round`: now derived from the newest spawn **at or before** `escalated_at` (episode-correct, and still omitted when no spawn row qualifies) — the docstring's claim becomes true instead of being weakened. Stability `grantsSeen`: now read from the ping kind's own encoded `granted` (episode-correct by construction); `rcRounds` and `round` keep the `stability_notices` join because the ledger holds nothing episode-scoped for them, and the docstring now states plainly that those two are current-at-derivation, so a backfilled episode-1 ping is documented as carrying the newest episode's numbers rather than silently implying otherwise. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Correct — the sentence was false, and it was the justification for the unbounded design, which is the worst kind of false. The module docstring now says the opposite truth: none of the seven tables this module reads is pruned, the backfill is bounded only by the deployment's actual row counts, and the per-pass full re-derivation is an accepted trade at today's volumes (hundreds of rows), with SQL-side `since` bounds named as the upgrade path if a deployment's ledger outgrows it. The PR body's same claim is corrected in the same push. Choosing the documentation fix over SQL-side filtering per your "at minimum" — the emitter's watermark-second dedupe (your other finding) also removes the standing per-pass re-POST, so the remaining cost is CPU on derivation, not network. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Implementing your suggested shape: the emitter remembers the dedupe keys it sent at the watermark second and drops them from the next derivation, so the boundary stays inclusive (the two-rows-one-second case still cannot lose a row) while a quiet ledger derives an empty batch and returns at the `if not events` guard — no request, no INFO line. A new row landing in the same watermark second is still sent (its key is not in the remembered set) and joins the set without moving the watermark. Tests: idle second pass posts nothing; same-second newcomer still goes out. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa_client.py
RHDZMOTA: [triage:pursue] Both closed as suggested. The client now opens through its own `OpenerDirector` whose redirect handler refuses every redirect outright — a 30x surfaces as an HTTP error through the existing taxonomy instead of replaying the Authorization header (or downgrading the POST) toward wherever Location points; refusing beats stripping-on-host-change here because a redirected ingest cannot succeed anyway (the POST body does not survive 301/302/303). And `Config.build` now rejects a non-`https` `alissa_endpoint` unless its host is loopback (`localhost` / `127.0.0.1` / `::1`), which keeps the local-stub test path while a cleartext staging URL is refused at load, not discovered on the wire. Tests pin the handler's refusal, the classification of a 30x, and the config refusal/loopback allowance. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Agreed — the taxonomy earned its keep only if someone acts on it. `emit_once` now catches `AlissaAuthError` separately: one WARN naming the permanent condition and telling the operator what to fix, then the emitter latches off and every later pass returns immediately (a DEBUG line, no derivation, no request) until the process restarts — which is also the moment a corrected token or env var takes effect. Transient errors keep today's behaviour: WARN per failed pass, watermark un-advanced, re-send next pass. Test: after an auth failure the client is never called again and exactly one WARN exists. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Cheap and strictly more robust — `reaped_at` folds into the key (`revloop:reap:{session}:{reaped_at}`), so the event no longer depends on session-name nonce-uniqueness holding forever, and a re-REPLACEd row derives as the new decision it is. Still deterministic under re-derivation. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa_client.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `3e7cfe5`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,509 @@
...
+            repo=repo,
+            pr=number,
+            round_=int(row["round"]),
+            session=row["session"],
+            data=data,
+        ))
+    return out
+
+
+def _verdict_events(rows: "list[dict]") -> "list[tuple[int, dict]]":
```
alissa-app, line 142: [major] `round.verdict` will emit **zero** events on this fleet — the table it reads only holds the daemon's *fallback* posts.

A `verdict_posts` row is created by exactly one call, `note_verdict_post_owed` inside `_post_native_verdict` (`loop.py:2828`), and that function is reached only when `owed > native` (`loop.py:1908-1935`) — i.e. only for a round whose reviewer **session** did not submit its own review and the daemon had to post one itself. `record_verdict_post` (the only writer of `posted_at`, `loop.py:2908`) sits inside that same function. So `posted_at` is set only on the fallback path, and `_verdict_events` here emits only for rows where it is set.

On this repo that path is never taken. Across the last 40 PRs there are **58** verdict reviews by the reviewer identity and **0** of them carry `verdict_marker`'s `<!-- alissa-revloop:verdict round=N -->` marker — every single round was closed by the session. PR #113's own round-1 review is one of the 58. Enable the knob on this daemon today and the console gets ~58 `round.spawned` next to 0 `round.verdict`, which is precisely the pair the origin task's manual acceptance criterion asks to see (`GET /v1/loop-events?seat=revloop` showing `round.spawned` *and* `round.verdict` for a real PR with matching rounds and verdicts).

The code follows issue #112's written mapping exactly — the issue says "from `verdict_posts` when `posted_at` is set", and describes `verdict_posts` as "the richest per-round table". The gap is that it is not the per-round table; it is the per-round **obligation** table, populated only when the round's own session defaulted. That is why I am raising this against the diff rather than the issue: the delivery notes are otherwise unusually careful about their own bounds, and this one is not named anywhere — not the README's telemetry section, not the *Not verified* list, not the *Judgment calls* section (which discusses where `data.verdict` comes from without noting that the row itself rarely exists).

Two dispositions I would accept:

1. **Derive it from a source that covers session-posted rounds.** The daemon already reads both sides every pass — `resolved.verdicts` (envelope count) and `countable_rounds(my_reviews)` — so the fact is in hand at decide time. Persisting it, though, means writing a per-round row for rounds the daemon did not post, which touches round bookkeeping that issue #112 puts out of scope. Your call whether that is this PR or a follow-up.
2. **Document the bound and register the follow-up.** Say in the README table row and in *Not verified* that `round.verdict`/`round.abandoned` cover only the daemon's native fallback post and are expected to be empty on a fleet whose sessions post their own reviews, register the task that carries the real fix, and get the manual acceptance criterion amended with the operator rather than shipping a criterion that cannot pass.

Either closes it. What I am holding on is the silence: an operator who turns this on to get "verdict mix" gets an empty series and no way to tell a broken writer from a working one.

README.md
```diff
@@ -841,6 +845,41 @@ console consumes — newest first, with the `stages` JSON decoded back to a
...
+
+The kinds, each carrying the **ledger's** timestamp as `at` and a dedupe key
+built from the ledger's own keys:
+
+| kind | ledger source | payload |
+| --- | --- | --- |
+| `round.spawned` | `spawns` | `session`, `round`, `data.headSha`, `data.taskRef` |
+| `round.verdict` | `verdict_posts` once `posted_at` is set | `round`, `data.verdict` (`approve` \| `request_changes`), `data.headSha`, `data.reviewUrl`, `data.attempts`, `data.checksHeldMs` when the round was held |
+| `round.abandoned` | `verdict_posts` once `abandoned_at` is set | `round`, `reason` (= `last_error`), `data.headSha` |
+| `round.capped` | `escalations` (CR9 cap-outs) | `round` (the PR's newest spawned round), `data.headSha` |
```
alissa-app: [minor] Stale after the round-1 fix — this row still describes the behaviour the fix replaced.

`round` here is no longer "the PR's newest spawned round": `_capped_events` now reads the newest spawn **at or before** `escalated_at`, which is the point of the round-1 minor (a backfilled cap-out must carry the round that was in flight when it happened, not today's). The PR body's *Judgment calls* section states the new behaviour correctly; this table did not follow.

Also worth a word in the same row while you are here: the round is **omitted** when no spawn row qualifies, which the table does not say either.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
```diff
@@ -156,6 +162,77 @@ def _describe(value: str) -> str:
...
+
+# Hosts a cleartext `alissa_endpoint` is allowed to name. The loop-events
+# client sends a bearer token with every POST, so a non-https endpoint puts
+# that token on the wire — refused at load, except toward the machine itself
+# (a local stub or port-forward, which is how the client is tested against a
+# fake ingest).
+_LOOPBACK_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})
+
+
+def _validate_alissa_endpoint(endpoint: str) -> str:
```
alissa-app, line 190: [nit] `https://` with no host passes this check, and the failure then shows up where the docstring promises it will not.

`urlsplit("https://").scheme` is `"https"`, so the first branch returns it. `AlissaClient` then does `"https://".rstrip("/")` → `"https:"` and posts to `https:/v1/loop-events`, which leaves as a `URLError` → `AlissaTransient` → one WARN per poll pass, forever — exactly the "discovered on the wire" outcome the docstring says this function exists to prevent, just with a different symptom than a cleartext hop.

One line closes it: require `parts.hostname` on the https branch too.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
```diff
@@ -779,21 +787,68 @@ def newest_grant(self, repo: str, number: int) -> sqlite3.Row | None:
...
 
-    def read_grants(self, repo: str, number: int) -> list[dict]:
-        """One PR's grant rows, newest first (like every reader here).
-
-        Deliberately narrow: no unfiltered form and no `limit` until something
-        needs them. The console does not read this table yet -- showing "this
-        PR was re-entered by @x" in the operator inbox wants rendering as well
-        as data, so it lands as its own change rather than as unused
-        parameters here.
+    def read_grants(
```
alissa-app, line 790: [nit] The widened signature makes a partial filter read as no filter, silently.

`read_grants("org/repo")` — repo given, `number` omitted — falls through the `repo is not None and number is not None` guard and returns **every PR's** grants. Both existing callers pass the pair positionally so nothing is wrong today, and the emitter's unfiltered call is the whole point of the change; it is the third caller that worries me, since the wrong answer here is a superset rather than an error.

`if (repo is None) != (number is None): raise ValueError(...)` makes the two forms the only two forms.

alissa-app (changes requested):
## Round 2 — `request_changes`

**Reviewed head:** `3e7cfe51796b8349d61decf7de6fc0eb67a76892` · **Round 2 of cap 22** · **Review task:** TASK-641453357 · **Origin:** TASK-774046670

**CI at the reviewed head:** green — the check-runs rollup at `3e7cfe5` (the full read, not the Actions-API fallback) reports 8 contexts, all concluded: 7 `success`, `[code]smith` `skipped`; no legacy statuses. The verdict below is not about CI.

### Round-1 reconciliation — clean, all seven fixed

Every round-1 finding carries exactly one `[triage:pursue]` reply and every one of them actually holds at this head, each with a test beside it:

- **[major] `round.capped` key collision** — key now folds the row's own `escalated_at` (`loop_events.py:196`), and `_capped_events` takes `spawns` so the round comes from the newest spawn *at or before* `escalated_at`. Verified against `escalations`' `PRIMARY KEY (repo, number, head_sha)` + `INSERT OR REPLACE`: a re-cap-out overwrites the row, the old key was already sent, the new stamp yields a new key. `test_a_re_cap_out_on_the_same_head_derives_a_distinct_key` pins it.
- **[minor] backfill joins** — split by provenance exactly as argued: `grantsSeen` now parsed from the ping kind's own third field (`stability_kind` is `stability:<head>:<base>:<granted>`, so `len(parts) == 3` is right), `rcRounds`/`round` kept on the join and documented as current-at-derivation.
- **[minor] the false retention claim** — replaced with the opposite truth in the module docstring *and* in the PR body, with the two upgrade paths named. Confirmed: none of the seven tables read here is pruned.
- **[minor] the standing boundary re-POST** — `_sent_at_watermark` closes it. I traced the four cases (first pass, quiet second pass, same-second newcomer, newcomer at a later second) and the set/watermark update is correct in each; `test_a_quiet_ledger_makes_no_request_at_all` and `test_a_newcomer_in_the_watermark_second_is_still_sent` pin the two that matter.
- **[minor] redirect + scheme** — `_RefuseRedirects` returns `None` so a 30x surfaces as an `HTTPError` through the taxonomy, and `_validate_alissa_endpoint` runs inside `Config.build` *after* the CLI overrides are merged into `raw`, so `--alissa-endpoint http://…` is refused too, not just the file key. Both pinned.
- **[minor] auth taxonomy never acted on** — `except AlissaAuthError` is ordered before `except AlissaError`, latches `_auth_failed`, one WARN, later passes return at a DEBUG line. Correct.
- **[nit] `reap` key** — folds `reaped_at`.

No round-1 finding is still open, and nothing in the fix delta regressed the derivation.

### What holds the round

One `[major]`, and it is not about the fix delta — it is the sweep of the full rubric against the origin task's acceptance criteria: **`round.verdict` will emit zero events on this fleet.** `verdict_posts` rows are created only by `_post_native_verdict`, which the loop reaches only when `owed > native` — i.e. only for rounds where the reviewer *session* failed to submit its own review and the daemon had to post the fallback. I counted it on this repo: across the last 40 PRs, **58** reviewer-identity verdict reviews, and **0** of them carry `verdict_marker`'s native-post marker. Every round on this deployment is session-posted, so `posted_at` is never set, so `round.verdict` never derives. `round.spawned` would show ~58 events next to zero verdicts — and the origin task's manual acceptance criterion asks for exactly the pair.

The code follows issue #112's own mapping ("`round.verdict` — from `verdict_posts` when `posted_at` is set") faithfully; the gap is that the issue treated `verdict_posts` as the per-round record when it is really the per-round *obligation* record. That makes this waivable in principle — but neither the README's telemetry section nor the PR's *Not verified* list names the bound, so as it stands the feature ships silently unable to satisfy its own criterion. I am not waiving it for that reason.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 1 |
| nit | 2 |
| question | 0 |

- **[major]** `round.verdict` fires only for the daemon's fallback post, which is 0 of 58 rounds on this fleet (`loop_events.py:142`)
- **[minor]** README's `round.capped` row is stale after the round-1 fix (`README.md:866`)
- **[nit]** `https://` with no host passes endpoint validation (`config.py:190`)
- **[nit]** `read_grants(repo)` with `number` omitted silently reads every PR (`state.py:790`)

### Rubric

- **Correctness** — the round-1 fixes all hold (reconciliation above). Derivation, batching, the inclusive-boundary watermark, the dry-run skip and the disabled-by-default silence are all pinned by tests, and the ping-kind prefixes are asserted against `loop`'s own builders so a rename fails the suite. Dedupe keys re-checked against each table's PK and insert mode — `pings` is `INSERT OR IGNORE` so the key-from-kind is exactly right, `spawns` is per-session but one round-on-head is one fact, `grants` is per-comment-id. One defect: the `round.verdict` coverage gap above.
- **Coding practices** — matches the repo idiom closely: docstring-carries-the-reasoning, reader-per-table in `state.py`, the `_ADDED_COLUMNS` migration path, the container's boolean pass-through contract with drift guards on both spellings and the refusal. Version bump, branch and PR title follow C1–C4. No new runtime dependency, no dead code. One stale README row and one reader-signature footgun, both below.
- **Security** — no secrets in the diff; the token comes from the existing `ALISSA_API_TOKEN` and is never logged (error detail is status + body). The round-1 hardening closed the two real surfaces: the bearer token no longer rides a redirect, and a cleartext endpoint is refused at load through both the file and the CLI layer. Payload is `json.dumps`'d; the only SQL added is parameterised or literal. Remaining: the empty-host nit, which degrades to a per-pass transient WARN rather than the promised `config error`.
- **Performance** — bounded timeouts, one batch per pass split at the ingest cap, the disabled default costs nothing, and the quiet-ledger fix means an idle daemon now makes no request at all. `_capped_events`' new spawn scan is O(escalations × spawns) on tables of hundreds of rows — fine, and the honest unbounded-re-derivation note in the docstring is the right disclosure. Nothing found.
- **Side-effects / blast radius** — the outbound write is off by default and an image upgrade cannot start posting on its own; the container ARG is unset-means-omit, so the library default still decides. The `verdict` column is additive through the existing idempotent `ALTER` path, and the single `record_verdict_post` call site is the only writer, so no path NULLs an existing verdict. `read_grants`' widening keeps both existing positional callers unchanged. Failure containment still holds at three layers (client taxonomy → emitter WARN → `_emit_loop_events` catch-all), and dry-run posts nothing. Nothing found.

### Waived majors

None.

| --- | --- | --- |
| [major] `round.capped` dedupe key omits everything distinguishing a re-cap-out | pursue | fixed (`3e7cfe5`) |
| [minor] backfilled events join against current-state tables | pursue | fixed (`3e7cfe5`) |
| [minor] the "ledger's own retention" bound does not exist | pursue | fixed (`3e7cfe5`) |
| [minor] idle daemon re-POSTs the boundary batch every pass | pursue | fixed (`3e7cfe5`) |
| [minor] bearer token rides urllib's redirect handling; endpoint not https-constrained | pursue | fixed (`3e7cfe5`) |
| [minor] the auth/transient taxonomy is built but never acted on | pursue | fixed (`3e7cfe5`) |
| [nit] `reap` dedupe key has no repo/PR and no timestamp | pursue | fixed (`3e7cfe5`) |

Close the `[major]` — either by deriving `round.verdict` from a source that covers session-posted rounds, or, if that is out of issue #112's scope (it arguably touches round bookkeeping), by (a) saying plainly in the README telemetry section and the PR's *Not verified* list that `round.verdict` covers only the daemon's fallback post and is expected to be empty on a fleet whose sessions post their own reviews, and (b) registering the follow-up task and getting the manual acceptance criterion amended with the operator. Either disposition is fine; leaving it unsaid is what I am holding on. The `[minor]` and the two `[nit]`s are one-liners — fix here or carry them as a follow-up task; they do not gate the merge.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: [triage:pursue] Taking disposition 2, and agreeing with your reading of why 1 is not this PR: the fact is in hand at decide time, but persisting it means writing per-round rows for rounds the daemon did not post — new semantics for the obligation table, i.e. exactly the round bookkeeping issue #112 declares out of scope — and deriving from GitHub at emit time would break the emitter's pure-ledger design (no GitHub calls in the telemetry path). So, in this round: the README telemetry table's `round.verdict`/`round.abandoned` rows now state plainly that they cover only the daemon's native fallback post (`verdict_posts` is the per-round *obligation* record, populated only when the session defaulted) and are expected to be empty on a fleet whose sessions post their own reviews — 0 of the last 58 rounds here — and the PR body's *Not verified* list carries the same bound plus the note that the origin task's manual acceptance criterion cannot pass until the follow-up lands and needs amending with the operator. The real fix is registered: **TASK-1086576582** ("Revloop: cover session-posted rounds in round.verdict loop events"), soft-linked downstream of TASK-1776360305, with the likely shape (persist a per-round observation row at round-count time, keep the emitter pure-ledger) and the criterion-amendment step in its definition of done. Sha to follow on this thread.

README.md
RHDZMOTA: [triage:pursue] Right — the table row lagged the fix. It now says the round is the newest spawn at or before `escalated_at` (the round in flight when the cap was recorded) and that it is omitted when no spawn row qualifies. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: [triage:pursue] Good catch — `urlsplit("https://").scheme` sailing through is exactly the wire-discovered failure the validator promises to prevent. The https branch now requires `parts.hostname` too, with a test for the bare `https://` and a schemed-but-hostless variant. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
RHDZMOTA: [triage:pursue] Agreed — a partial filter answering with a superset is the quiet kind of wrong. `read_grants` now raises `ValueError` when exactly one of `repo`/`number` is given, so the paired and unfiltered forms are the only two forms; test added. Sha to follow on this thread.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
RHDZMOTA: Fixed in `0629bad`.

README.md
RHDZMOTA: Fixed in `0629bad`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py
RHDZMOTA: Fixed in `0629bad`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
RHDZMOTA: Fixed in `0629bad`.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop_events.py
```diff
@@ -0,0 +1,518 @@
...
+            self._auth_failed = True
+            log.warning(
+                "loop-events: authentication failed (%s) — this is permanent "
+                "and operator-fixable (set/rotate the token the emitter's "
+                "client reads, then restart the daemon); loop telemetry is "
+                "now off for this process, the loop keeps polling",
+                exc,
+            )
+            return False
+        except AlissaError as exc:
```
alissa-app, line 481: [minor] A **permanent** non-auth failure is handled as if it were transient, and because the batch is all-or-nothing, one poison event blocks the whole stream forever.

The taxonomy has three buckets, and this handler only distinguishes two of them. `AlissaAuthError` (401/403, or no token) latches the emitter off — that is the round-1 fix and it is right. `AlissaTransient` (408/429/5xx and every transport failure) keeps the watermark put so the next pass re-sends — also right. But the **base** `AlissaError` is neither: `_classify` returns it for every status that is not 401/403/408/429/5xx (`alissa_client.py:170`), i.e. a `400` or `422` the ingest returns when it refuses a batch on validation, plus the `AlissaError(200, "response was not JSON")` case. All of those land in this `except AlissaError` branch and get the transient treatment.

The consequence is worse than the auth case round 1 flagged, because the ingest fails the **whole** batch on one bad event (the PR body says exactly this about the 2000-char `reason` cap, which is why `_event` trims it here rather than there). So a single event the API will not accept — a `dedupeKey` or `session` over some server-side length cap, a `data` shape a later ingest tightening rejects — means:

1. the POST 400s, `sent` stays behind, one WARN;
2. the watermark does not advance, so the next pass re-derives **the same batch, including the poison row**, and re-POSTs it;
3. every event behind it in the ledger is blocked too, because it is in the same never-accepted batch;
4. and a restart does not help — the row is in the ledger, so the backfill re-derives it from zero.

That is permanent, operator-invisible-except-as-a-repeated-WARN, total telemetry loss, from the bucket the module's own docstring describes as "your token is wrong" versus "the API blinked" — this is neither, and it currently reads as the second.

The cheap fix is to treat the base bucket like the auth one: latch (or at minimum log at WARN with the response body and a distinct "this will not resolve on its own" line, so the operator knows to look rather than assuming the API is flapping). A more surgical option, if you would rather not lose the stream on one bad row, is to advance the watermark past a batch the server refuses as permanent — but that trades a silent drop for a stall and needs its own thought, so latch-and-say-so is the one I would ship.

Not blocking: telemetry is off by default, best-effort by contract, and the trigger needs the ingest to refuse an event this code builds — which today's trimming and fixed kind shapes make unlikely. Fix here or carry it as a follow-up task.

README.md
```diff
@@ -841,6 +845,41 @@ console consumes — newest first, with the `stages` JSON decoded back to a
...
+`ALISSA_API_TOKEN` the `alissa` CLI reads (the CLI itself has no loop-events
+command, so this is the package's one direct REST write — `alissa_client.py`).
+
+The kinds, each carrying the **ledger's** timestamp as `at` and a dedupe key
+built from the ledger's own keys:
+
+| kind | ledger source | payload |
+| --- | --- | --- |
+| `round.spawned` | `spawns` | `session`, `round`, `data.headSha`, `data.taskRef` |
+| `round.verdict` | `verdict_posts` once `posted_at` is set | `round`, `data.verdict` (`approve` \| `request_changes`), `data.headSha`, `data.reviewUrl`, `data.attempts`, `data.checksHeldMs` when the round was held. **Coverage bound**: `verdict_posts` is the per-round *obligation* record — a row exists only when the daemon had to post the native fallback verdict itself, i.e. when the reviewer session did not submit its own review. On a fleet whose sessions post every review (0 of the last 58 rounds on this repo took the fallback path), this series is **expected to be empty**; covering session-posted rounds is TASK-1086576582 |
```
alissa-app, line 864: [nit] The coverage bound is exactly what I asked for and is in the right three places — but this line bakes a point-in-time measurement into shipped prose with nothing to keep it true.

"0 of the last 58 rounds on this repo took the fallback path" was measured during round 2 of this PR's own review. It is the right evidence for the *judgment call* (which is why it belongs in the PR body and in TASK-1086576582's DoD, where it is dated by context), but in the README it reads as a standing property of the deployment. The moment a reviewer session fails to post and the daemon takes the fallback path, the sentence is false, and nothing will notice — it is prose, so no test guards it.

Suggestion: keep the structural claim, which is durable and is the part an operator needs ("a row exists only when the daemon posted the native fallback verdict itself, so on a fleet whose sessions post their own reviews this series is expected to be empty or near-empty"), and either drop the count or attribute it ("measured 0 of 58 rounds on this repo in 2026-09"). Same for `round.abandoned`, which inherits the bound by reference.

Implementer's discretion — it is a nit, on a line that is a strict improvement over what was there before.

alissa-app (approved):
## Round 3 — `approve`

**Reviewed head:** `0629bad4eedb3e6d5d7cb3c93059c5de31562119` · **Round 3 of cap 22** · **Review task:** TASK-641453357 · **Origin:** TASK-774046670

**CI at the reviewed head:** green — the check-runs rollup at `0629bad` (the full read, not the Actions-API fallback) reports 8 contexts, all concluded: 7 `success`, `[code]smith` `skipped`; no legacy statuses. Re-read immediately before submitting; the head had not moved.

### Round-2 reconciliation — clean, all four closed

Every round-2 finding carries exactly one `[triage:pursue]` reply and a `Fixed in 0629bad` follow-up, and each one holds at this head:

- **[major] `round.verdict` covers only the daemon's fallback post.** Taken as disposition 2, which is what I asked for and is the honest one. The bound is now stated in all three places a reader will hit it: the README telemetry table (`README.md:864` — and `round.abandoned` inherits it explicitly), the `_verdict_events` docstring (`loop_events.py:148-157`), and the PR body's *Not verified — operator's gate* list. The follow-up is real and open: **TASK-1086576582** ("Revloop: cover session-posted rounds in round.verdict loop events", status `committed`, soft-linked downstream of TASK-1776360305), and its DoD carries the measurement, the likely shape, *and* the instruction to amend the origin task's manual acceptance criterion with the operator. That last step is the operator's to take, not the implementer's, and it is surfaced rather than buried — so I am closing the finding rather than holding a round on someone else's action. The feature no longer ships silently unable to satisfy its own criterion, which was the whole of my objection.
- **[minor] stale README `round.capped` row** — now says the round is the newest spawn **at or before** `escalated_at` and that it is omitted when no spawn qualifies. Matches `_capped_events` exactly.
- **[nit] `https://` with no host** — `_validate_alissa_endpoint` now requires `parts.hostname` on the https branch, the error string gained "with a host", and the refusal is parameterised over both `https://` and `https:///v1`. Verified the validator is still reached at `config.py:717`, i.e. after the CLI overrides are merged, so a hostless `--alissa-endpoint` is refused too.
- **[nit] partial `read_grants` filter** — `(repo is None) != (number is None)` raises. I checked every caller at this head: `loop.py:3451` passes the pair, the emitter (`loop_events.py:384`) passes neither, and the three `test_state.py` call sites pass the pair — so the new `ValueError` cannot fire on any existing path. No regression.

No round-2 finding is still open, and nothing in the fix delta regressed the derivation: the delta is 48 lines across README prose, two guards and their tests, and I re-read `_verdict_events`, `_capped_events`, `_ping_events` and `emit_once` in full against it.

### What the sweep found

Nothing that gates the merge. One `[minor]` and one `[nit]`, both new, neither a re-raise:

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 1 |
| nit | 1 |
| question | 0 |

- **[minor]** a permanent non-auth 4xx is retried like a transient, and the all-or-nothing batch means one poison event blocks the whole stream forever (`loop_events.py:481`)
- **[nit]** the README coverage bound bakes in a point-in-time measurement that will drift (`README.md:864`)

### Rubric

- **Correctness** — the DoD is met item by item and the round-2 fixes all hold (reconciliation above). Re-checked at this head: the inclusive-boundary watermark plus `_sent_at_watermark` (quiet ledger ⇒ empty derivation ⇒ no request; same-second newcomer still sent, the set *grows* rather than being replaced when `newest == self._since`), the `posted_at`/`abandoned_at` split with `data.verdict` omitted rather than invented on pre-column rows, `checksHeldMs` floored at zero, the `reason` trim at 2000 that keeps one long `last_error` from failing a whole batch, dry-run posting nothing, and the disabled default costing no client and no read. Dedupe keys re-verified against each table's PK and insert mode — `pings` is `PRIMARY KEY (repo, number, kind)` + `INSERT OR IGNORE`, so key-from-kind is exactly one row; `escalations` folds `escalated_at`; the two `checks.held` gates keep distinct keys and `data.gate`. One finding: the permanent-4xx handling below.
- **Coding practices** — still matches the repo idiom closely: docstring-carries-the-reasoning, reader-per-table in `state.py`, the `_ADDED_COLUMNS` migration path for the additive `verdict` column, the container's boolean pass-through contract with drift guards on both spellings and on the refusal. Version `0.28.0`, branch and PR title follow C1–C4; no new runtime dependency, no dead code. The two new guards are placed where the wrong call is made rather than where it is felt, which is right. One nit below. (`test_read_grants_refuses_a_partial_filter` lives in `test_loop_events.py` while testing a `state.py` guard — defensible, since the unfiltered form exists for this emitter; not worth a finding.)
- **Security** — nothing found. No secrets in the diff; the token comes from the existing `ALISSA_API_TOKEN` and is never logged (error detail is status + body). The round-1 hardening still stands at this head: `_RefuseRedirects` returns `None` so a 30x raises rather than replaying the bearer token toward `Location`, and cleartext is refused at load through the file *and* the CLI layer — now including the hostless spellings that previously slipped past into a wire-time failure. Payload is `json.dumps`'d; the only SQL added is parameterised or literal.
- **Performance** — nothing found that is a defect. Bounded 30s timeouts, one batch per pass split at the ingest cap, quiet ledger makes no request at all, disabled default costs nothing. The per-pass full re-derivation over seven unpruned tables and `_capped_events`' O(escalations × spawns) scan are both honestly disclosed in the module docstring with the two upgrade paths named, and are fine at the hundreds-of-rows volumes this deployment actually has.
- **Side-effects / blast radius** — nothing found. The outbound write is off by default and an image upgrade cannot start posting on its own (`ARG` empty ⇒ key omitted ⇒ library default). The `verdict` column is additive through the existing idempotent `ALTER` path and `record_verdict_post` is still the only writer, so no path NULLs an existing verdict. `read_grants`' new refusal is unreachable from every caller in the tree. Failure containment holds at three layers (client taxonomy → emitter WARN → `_emit_loop_events` catch-all), so telemetry cannot take down a poll; dry-run posts nothing. The Dockerfile's `REVLOOP_VERSION >= 0.28.0` requirement is consistent with the other pass-through keys and with CI's cross-check installing 0.16.14 with the variable unset.

### Waived majors

None.

| --- | --- | --- |
| [major] `round.verdict` fires only for the daemon's fallback post (0 of 58 rounds) | pursue | fixed (`0629bad`) — bound stated in README + docstring + PR body; TASK-1086576582 registered and open |
| [minor] README's `round.capped` row stale after the round-1 fix | pursue | fixed (`0629bad`) |
| [nit] `https://` with no host passes endpoint validation | pursue | fixed (`0629bad`) |
| [nit] `read_grants(repo)` with `number` omitted silently reads every PR | pursue | fixed (`0629bad`) |

### Summary

The PR gives revloop a ledger-derived, idempotent, best-effort loop-telemetry writer to Studio's `POST /v1/loop-events`, off by default, with the config key / CLI flag / env var / container ARG+ENV+render, README section and 64 tests the origin task asks for. Three rounds have closed eleven findings — one major on dedupe-key collision, one major on a coverage bound that would have shipped unstated, and nine minors/nits — and the round-2 reconciliation is clean at this head with no regression in the fix delta. What remains is one `[minor]` and one `[nit]`, both on non-blocking surfaces: neither is a defect in what the loop decides, and the minor's failure mode is confined to telemetry that is off by default. **Approve with notes** — the two open items are the implementer's discretion to fix here or carry as a follow-up task; they do not gate the merge. The operator still owes TASK-1086576582's criterion amendment on the origin task before its manual acceptance criterion can pass, which the PR body correctly lists as an operator gate rather than a delivered claim.

RHDZMOTA:
**Round-3 approve — triage of the two carried notes** (body-only findings, so recorded here as the canonical cross-actor record; the approved head `0629bad` is terminal and gets no further commits).

- **[minor] permanent non-auth 4xx retried like a transient / poison event blocks the batch** — `[triage:later]`, registered as **TASK-2099153281** ("Revloop loop-events: park poison batches on permanent 4xx"), soft-linked downstream of TASK-1776360305. The fix is a code change (distinguish permanent client errors in `emit_once`: drop/quarantine the offending event — the 400's issue path names its index — or latch like the auth path), which cannot land on an approved branch; the blast radius meanwhile is confined to telemetry that is off by default, with one WARN per pass as the symptom.
- **[nit] README coverage row bakes in a point-in-time measurement ("0 of the last 58")** — `[triage:later]`, folded into the same TASK-2099153281 (reword to a dated/structural statement); the row is removed outright when **TASK-1086576582** (session-posted verdict coverage) lands, which is that task's own acceptance criterion.

Standing operator gates, as listed in the PR body: the origin task's manual acceptance criterion needs amending alongside TASK-1086576582, and live-fleet verification of the ingest remains the operator's.

PR #117, opened 2026-09-03
TASK-145014812: drop --permission-mode acceptEdits from the baked claude agent profile (issue #116)

### LINKED ISSUE #116: Drop --permission-mode acceptEdits from the baked claude agent profile (flag conflict re-enables permission prompts)
## Problem

The baked agent profile `docker/claude/agents.yaml` (line 24) spawns the claude CLI with two conflicting permission flags:

```
command: claude --dangerously-skip-permissions --permission-mode acceptEdits
```

The explicit `--permission-mode acceptEdits` takes precedence over `--dangerously-skip-permissions`, so the session actually runs in acceptEdits mode: file edits are auto-approved, but hard permission prompts still fire — notably Claude Code's dangerous-rm check ("Dangerous rm operation on statically-unresolvable target"), which no allowlist covers.

Observed in prod 2026-09-03: the reviewer session for studio.alissa.app PR #1039 round 3 (tmux session `ali-review-studio-alissa-app-pr1039-r3-0c0717`, claude session `8e60a14c-d16b-447a-a904-9f8e5d2ec00f`) sat wedged at exactly such a prompt for 7+ hours, holding its slot while the queue re-delivered the same round as a second session. The alissa tmux supervisor's `promptPatterns` ("Do you want to", "❯ 1. Yes") only detect-and-escalate (mark the session stuck, hold the queue) — nothing ever answers the dialog, so every such prompt is a permanent wedge.

The file's own comment block (lines ~7–10) asserts that the flag *pair* "is what makes claude run unattended" — that premise is wrong and is how the conflict got baked in.

## Fix

1. In `docker/claude/agents.yaml`, remove `--permission-mode acceptEdits` from the baked `claude` profile's `command:` line. Keep `--dangerously-skip-permissions` (the non-root uid-1000 requirement is unchanged).
2. Rewrite the comment block so it states the corrected contract: `--dangerously-skip-permissions` alone provides unattended operation, and an explicit `--permission-mode` flag must NOT be added alongside it, because an explicit mode overrides the bypass and re-enables hard prompts (dangerous rm, un-allowlisted Bash) that wedge headless sessions.
3. Sweep the repo for any other place the claude command line is composed or asserted (entrypoint.sh sections that touch `agents.yaml`, CI workflows such as `check-entrypoint.yaml`, tests/fixtures asserting the command string) and update them to the single-flag form.

## Out of scope

- The `ALISSA_AGENT_MODEL` / `--model` rewrite performed by entrypoint.sh at boot (unchanged).
- The tmux supervisor's promptPatterns machinery (detection stays as a safety net; after this fix it should simply never fire for permission prompts).
- The sibling daemons (devloop, orcloop) — same fix fed as separate tasks in their own repos.

## Definition of Done

The baked claude profile's command line contains --dangerously-skip-permissions and no --permission-mode flag; the comment block documents why an explicit permission-mode is forbidden next to the bypass flag; every repo occurrence that composes or asserts the claude command line agrees; CI green; PR flipped ready-for-review and merged.

## Acceptance criteria

- [ ] docker/claude/agents.yaml baked claude profile command has --dangerously-skip-permissions and contains no --permission-mode flag
- [ ] The comment block explains that an explicit --permission-mode overrides the bypass flag and re-enables hard prompts, so it must not be re-added
- [ ] grep over the repo shows no remaining --permission-mode on any claude spawn/assert path (entrypoint, workflows, tests)

Autonomous-Scope: docker/claude/*, .github/workflows/check-entrypoint.yaml
Alissa-Task: TASK-834798535

RHDZMOTA:
Closes #116

**Alissa tasks:** origin `TASK-834798535` · implementation `TASK-145014812` (soft-depends on the origin; the origin is owned by the develop-daemon actor, so a true downstream link 403s `NOT_OWNER`).

## What changed

The baked `claude` profile spawned the CLI with `--dangerously-skip-permissions --permission-mode acceptEdits`. The explicit mode overrides the bypass, so reviewer sessions really ran in acceptEdits and Claude Code's hard prompts (the dangerous-`rm` check, un-allowlisted Bash) still fired. Nothing in a headless container answers those dialogs; the supervisor's `promptPatterns` only detect and escalate. A prod reviewer sat wedged at that prompt for 7+ hours on 2026-09-03 while the queue re-delivered the round.

- `docker/claude/agents.yaml` — profile command is now `claude --dangerously-skip-permissions`; the header comment states the single-flag contract and why an explicit permission-mode must never be re-added next to the bypass.
- `docker/claude/entrypoint.sh` — the step-2e `BASE_CMD` constant the model-pin rewrite composes from drops the mode flag (comment names the contract).
- `docker/claude/tests-entrypoint-executor.sh` — the mirror `BASE_CMD` drops the flag, so all six whole-line model-pin assertions now pin the single-flag form; plus two named guards: the rendered command carries no `--permission-mode`, and the checked-in profile's unrendered `command:` equals `BASE_CMD` exactly.
- `docker/claude/README.md` — the "mount your own agents.yaml to change flags" paragraph now carries the contract so a custom mount does not re-create the conflict.

No version bump: the guard workflow exempts `docker/` and `.github/`, and no packaged file changed.

### Judgment calls
- **Criterion 3 read as "no executable spawn/assert line carries the flag", not "the string is absent from the repo".** Criterion 2 requires the comment block to name `--permission-mode` while forbidding it, so a literal zero-hit grep is impossible. After this PR the string appears only in: the agents.yaml header comment (required), one README sentence, and the executor test's negative guard. Every composing/asserting line (`agents.yaml` `command:`, entrypoint `BASE_CMD`, test `BASE_CMD`) is single-flag. The entrypoint's own comment says "permission-mode flag" without the dashes so the spawn script greps clean.
- **Added a static assertion on the checked-in profile, beyond the issue's list.** The entrypoint rewrites `command:` from `BASE_CMD` at boot, so a rendered-command guard alone would mask the flag creeping back into the source file. Verified by counterfactual: reintroducing the flag in agents.yaml only failed just this new assertion.
- **README edit is one paragraph, not a section.** The issue's scope is `docker/claude/*`; the mount-your-own-file paragraph is the one place a reader would re-create the conflict, so it carries the contract and nothing else was reworded.
- **`.github/workflows/check-entrypoint.yaml` was in the declared scope but needed no change.** It contains no command-line assertion; it only runs the test scripts, which are where the assertion lives.

- [ ] That `claude` actually honours `--dangerously-skip-permissions` alone without a hard prompt in the container. The claim comes from the issue's prod observation and Claude Code's flag semantics; this environment has no docker and no live reviewer spawn to reproduce the wedge or its absence.
- [ ] `check-image.yaml` (docker build + in-image `tests-image-contract.sh`, which compares the baked agents.yaml sha to the source). CI-only; no docker on this runner. The sha check reads the source file, so it is expected to pass.
- [ ] That no other repo (devloop, orcloop) is fed from this file. Out of scope per the issue; this PR touches only this repo.

### Verification
- All seven entrypoint suites, run locally from the worktree after the change (each boots the real `entrypoint.sh` against stubbed CLIs): executor 77 ok, ui 30 ok, config 60 ok, identity 12 ok, auth 31 ok, prune 51 ok, chown 23 ok. Zero FAIL lines.
- Counterfactual A: put the flag back in the entrypoint `BASE_CMD` → executor suite exits 1 with all six model-pin `assert_eq`s plus the rendered-command guard failing by name. Restored from HEAD.
- Counterfactual B: put the flag back in the checked-in agents.yaml `command:` only → executor suite exits 1 with exactly the new static assertion failing. Restored from HEAD.
- `bash -n` on the two edited shell files; agents.yaml is comment-only plus the one-token command change (no YAML structure touched; pyyaml is not installed here so it was not parsed).
- Python check scripts (`check-style.sh`, `check-types.sh`, `tests-unit.sh`) not run: no `.py` file changed.

### Scope touched
- `docker/claude/agents.yaml` (in scope) — command line + comment block.
- `docker/claude/entrypoint.sh` (in scope) — one constant + a four-line comment in step 2e.
- `docker/claude/tests-entrypoint-executor.sh` (in scope) — mirror constant + two new assertions.
- `docker/claude/README.md` (in scope, `docker/claude/*`) — one paragraph extended.
- `.github/workflows/check-entrypoint.yaml` (in scope) — read, unchanged.
- No excursions outside `docker/claude/`.

## Review round 1 — approve (alissa-app, head 4b3ec73)

0 blockers, 0 majors, 2 minors, 1 nit. This branch is terminal; nothing further is pushed here.

- `[minor]` `ALISSA_AGENT_MODEL` can smuggle flags into the spawn → `[triage:pursue]` → follow-up PR #118 (cut from `main`, `Follow-up-To: #117`, closes no issue; merges cleanly with this PR in either order).
- `[minor]` sibling images (devloop, orcloop) still carry the flag pair → `[triage:later]` → TASK-912931013.
- `[nit]` the new negative guard passes vacuously on an empty read → `[triage:later]` → TASK-345477539, sequenced after this PR merges because it edits a block this PR introduces.

https://claude.ai/code/session_01PVn2ixtkReYSttFSdiU4Qu

---
<a href="https://app.blacksmith.sh/fahera-mx/codesmith/alissa-github-review-daemon/pr/117"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791047706&installation_model_id=434695&pr_number=117&repository=fahera-mx%2Falissa-github-review-daemon&return_to=https%3A%2F%2Fgithub.com%2Ffahera-mx%2Falissa-github-review-daemon%2Fpull%2F117&signature=0756969c3be01cd76d30afc719fa60c571c2246b96526c339c6fbd63ab0f76e6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docker/claude/agents.yaml
+++ b/docker/claude/agents.yaml
@@ -4,10 +4,22 @@
 # (default: claude). Without this file the worker falls back to a bare `claude`,
 # which stops on its first-run trust / permission prompt and hangs the reviewer.
 #
-# `--dangerously-skip-permissions --permission-mode acceptEdits` is what makes
-# claude run unattended. It requires a non-root user (claude refuses that flag as
-# root) — the image runs as uid 1000, which satisfies it. Reviewer posture (CR6:
-# reviewers never write) is enforced by the round directive, not by this profile.
+# `--dangerously-skip-permissions` ALONE is what makes claude run unattended: it
+# bypasses every permission prompt, so a headless session never stops to ask.
+# It requires a non-root user (claude refuses that flag as root) — the image runs
+# as uid 1000, which satisfies it. Reviewer posture (CR6: reviewers never write)
+# is enforced by the round directive, not by this profile.
+#
+# Do NOT add an explicit `--permission-mode <mode>` next to the bypass flag. An
+# explicit mode OVERRIDES the bypass: the session runs in that mode instead, and
+# every hard prompt that mode does not auto-approve comes back — Claude Code's
+# dangerous-rm check ("Dangerous rm operation on statically-unresolvable
+# target"), un-allowlisted Bash, and the like. Nothing in a headless container
+# answers those dialogs (the supervisor's `promptPatterns` below only detect and
+# escalate), so each one is a permanent wedge: this profile shipped with
+# `--permission-mode acceptEdits` alongside the bypass until 2026-09-03, and a
+# reviewer sat at exactly that rm prompt for 7+ hours holding its slot while
+# the queue re-delivered the same round (issue #116).
 #
 # To pin the reviewer model, set the ALISSA_AGENT_MODEL env var (default:
 # claude-fable-5-1) — the entrypoint rewrites the `command:` below with
@@ -21,7 +33,7 @@
 # profile at boot from ALISSA_AGENT_MODEL — do not remove this marker line, it is
 # what distinguishes the baked default from a runtime-mounted custom agents.yaml.
 claude:
-  command: claude --dangerously-skip-permissions --permission-mode acceptEdits
+  command: claude --dangerously-skip-permissions
   mode: interactive
   quietSeconds: 30
   promptPatterns:
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -602,7 +602,11 @@ if [ ! -f "${AGENTS_YAML}" ]; then
 elif ! grep -q 'alissa-managed:' "${AGENTS_YAML}"; then
   log "custom agents.yaml in effect (no alissa-managed marker) — using it verbatim, ALISSA_AGENT_MODEL ignored"
 else
-  BASE_CMD="claude --dangerously-skip-permissions --permission-mode acceptEdits"
+  # Single flag by contract (issue #116): an explicit permission-mode flag next
+  # to the bypass overrides it and re-enables hard prompts (dangerous rm,
+  # un-allowlisted Bash) that wedge a headless session. Keep this in step with
+  # the baked agents.yaml and the executor test's mirror constant.
+  BASE_CMD="claude --dangerously-skip-permissions"
   if [ -z "${AGENT_MODEL}" ] || [ "${AGENT_MODEL}" = "default" ]; then
     CLAUDE_CMD="${BASE_CMD}"
     log "reviewer model: account default (ALISSA_AGENT_MODEL='${AGENT_MODEL}') — no --model flag"
--- a/docker/claude/tests-entrypoint-executor.sh
+++ b/docker/claude/tests-entrypoint-executor.sh
@@ -211,7 +211,7 @@ EXECUTOR_UP="starting alissa bridge start"
 # agents.yaml, value only. BASE_CMD mirrors the entrypoint's step-2e constant:
 # a model pin is exactly BASE_CMD plus `--model <value>`, and no pin is exactly
 # BASE_CMD — so every model assertion compares the WHOLE line.
-BASE_CMD="claude --dangerously-skip-permissions --permission-mode acceptEdits"
+BASE_CMD="claude --dangerously-skip-permissions"
 profile_cmd() {
   sed -n -E 's/^[[:space:]]*command:[[:space:]]*//p' \
     "${WORKSPACE}/.alissa-config/agents.yaml" 2>/dev/null | head -n 1
@@ -289,6 +289,17 @@ assert_contains "${LOG2}" "reviewer model: claude-fable-5-1 (ALISSA_AGENT_MODEL)
   "...and the boot log names the default"
 assert_not_contains "${WORKSPACE}/.alissa-config/agents.yaml" "disable_alissa_code" \
   "...and WITHOUT disable_alissa_code, so job sessions still launch via alissa code"
+# Issue #116: the profile shipped `--dangerously-skip-permissions --permission-mode
+# acceptEdits` for a while, and the explicit mode OVERRIDES the bypass — sessions
+# ran in acceptEdits and wedged on hard prompts (dangerous rm) that nothing in a
+# headless container answers. BASE_CMD above already pins the whole line, but the
+# contract deserves its own named check so a re-added mode fails by name. The
+# rendered `command:` value is checked (not the whole file) because the header
+# comment explains the prohibition and so mentions the flag itself.
+case "$(profile_cmd)" in
+  *--permission-mode*) bad "the rendered claude command must NOT carry an explicit --permission-mode (it overrides --dangerously-skip-permissions and re-enables prompts)" ;;
+  *) pass "the rendered claude command carries no --permission-mode (bypass flag is the whole unattended contract)" ;;
+esac
 
 # -----------------------------------------------------------------------------
 info ""
--- a/docker/claude/tests-entrypoint-executor.sh
+++ b/docker/claude/tests-entrypoint-executor.sh
@@ -300,6 +300,12 @@ case "$(profile_cmd)" in
   *--permission-mode*) bad "the rendered claude command must NOT carry an explicit --permission-mode (it overrides --dangerously-skip-permissions and re-enables prompts)" ;;
   *) pass "the rendered claude command carries no --permission-mode (bypass flag is the whole unattended contract)" ;;
 esac
+# The rewrite above composes from the entrypoint's BASE_CMD, so it would mask a
+# mode flag re-added to the CHECKED-IN profile. Pin the source file too: its
+# unrendered `command:` must be exactly BASE_CMD, so the two agree before boot.
+assert_eq "$(sed -n -E 's/^[[:space:]]*command:[[:space:]]*//p' "${HERE}/agents.yaml" | head -n 1)" \
+  "${BASE_CMD}" \
+  "the checked-in agents.yaml command is exactly BASE_CMD (no --permission-mode, no pin)"
 
 # -----------------------------------------------------------------------------
 info ""
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -291,7 +291,11 @@ The reviewer's claude launch command lives in [`agents.yaml`](./agents.yaml). To
 pin the model, set `ALISSA_AGENT_MODEL` (see [Pinning the reviewer model](#pinning-the-reviewer-model))
 rather than editing the file; to change flags, mount your own file over
 `/home/alissa/.config/alissa/agents.yaml`. The image runs as a non-root user
-because claude refuses `--dangerously-skip-permissions` as root.
+because claude refuses `--dangerously-skip-permissions` as root. That flag is the
+whole unattended contract: do **not** pair it with an explicit `--permission-mode`
+(a mounted file included) — an explicit mode overrides the bypass and re-enables
+the hard prompts (dangerous `rm`, un-allowlisted Bash) that nothing in a headless
+container answers, so the session wedges at the dialog until it is killed.
 
 ### Pinning the reviewer model
 


alissa-app:
- 2026-09-03 18:30:35 UTC — `review-alissa-github-review-daemon-pr117-r1-a2d85b` — round 1 of 22 — spawned

docker/claude/entrypoint.sh
```diff
@@ -602,7 +602,11 @@ if [ ! -f "${AGENTS_YAML}" ]; then
 elif ! grep -q 'alissa-managed:' "${AGENTS_YAML}"; then
   log "custom agents.yaml in effect (no alissa-managed marker) — using it verbatim, ALISSA_AGENT_MODEL ignored"
 else
-  BASE_CMD="claude --dangerously-skip-permissions --permission-mode acceptEdits"
+  # Single flag by contract (issue #116): an explicit permission-mode flag next
+  # to the bypass overrides it and re-enables hard prompts (dangerous rm,
+  # un-allowlisted Bash) that wedge a headless session. Keep this in step with
+  # the baked agents.yaml and the executor test's mirror constant.
+  BASE_CMD="claude --dangerously-skip-permissions"
```
alissa-app, line 609: [minor] `ALISSA_AGENT_MODEL` is a second route back into #116, and the new guards cannot see it.

The pin value is appended verbatim with no allowlist (the comment above says so deliberately: "The value passes through verbatim — aliases and full ids are both valid, no allowlist"). So a boot with

```
ALISSA_AGENT_MODEL='opus --permission-mode acceptEdits'
```

renders `claude --dangerously-skip-permissions --model opus --permission-mode acceptEdits` and reproduces the wedge exactly — with every new assertion in this PR still green, because the executor suite only ever renders values that contain no flags.

The contract text you added names the mounted-file route explicitly ("a mounted file included") but not this one, and this one is the easier mistake: an operator setting a model pin is not thinking about permission semantics at all.

Cheapest fix is one sentence in the README's "Pinning the reviewer model" section. A stronger one is a boot-time guard here — if `${AGENT_MODEL}` contains `--permission-mode`, `die` (or strip and WARN) — which would also give the contract a runtime teeth rather than only a test-time one.

Not blocking: it takes a deliberate operator action, and the default path this PR fixes is correct.

docker/claude/agents.yaml
```diff
@@ -21,7 +33,7 @@
 # profile at boot from ALISSA_AGENT_MODEL — do not remove this marker line, it is
 # what distinguishes the baked default from a runtime-mounted custom agents.yaml.
 claude:
-  command: claude --dangerously-skip-permissions --permission-mode acceptEdits
+  command: claude --dangerously-skip-permissions
```
alissa-app, line 36: [minor] The sibling devloop image likely carries the same flag pair, and nothing here reaches it.

`docker/claude/README.md` (the `ALISSA_BRIDGE_EXECUTOR_ID` row) treats "the devloop image's executor" as a live, co-deployed peer of this one. If that image bakes a profile descended from this file — which the shared lineage makes likely — it still ships `--dangerously-skip-permissions --permission-mode acceptEdits` and will sit at the same `rm` dialog for the same 7 hours, and this PR does not and should not touch it.

Agreed it is out of scope for #116, and the PR body says so under "Not verified". The ask is only that it become a **registered Alissa task** rather than a line in a delivery note, so triage this as `later` with the task id — a delivery note is not something the other repo's loop will ever pick up.

docker/claude/tests-entrypoint-executor.sh
```diff
@@ -289,6 +289,23 @@ assert_contains "${LOG2}" "reviewer model: claude-fable-5-1 (ALISSA_AGENT_MODEL)
...
   "...and WITHOUT disable_alissa_code, so job sessions still launch via alissa code"
+# Issue #116: the profile shipped `--dangerously-skip-permissions --permission-mode
+# acceptEdits` for a while, and the explicit mode OVERRIDES the bypass — sessions
+# ran in acceptEdits and wedged on hard prompts (dangerous rm) that nothing in a
+# headless container answers. BASE_CMD above already pins the whole line, but the
+# contract deserves its own named check so a re-added mode fails by name. The
+# rendered `command:` value is checked (not the whole file) because the header
+# comment explains the prohibition and so mentions the flag itself.
+case "$(profile_cmd)" in
+  *--permission-mode*) bad "the rendered claude command must NOT carry an explicit --permission-mode (it overrides --dangerously-skip-permissions and re-enables prompts)" ;;
```
alissa-app, line 300: [nit] This negative guard passes vacuously on an empty read.

`profile_cmd` swallows a missing file (`2>/dev/null`), so if `${WORKSPACE}/.alissa-config/agents.yaml` were absent the command substitution is `""`, the `case` falls to `*)`, and the suite prints `ok` for a profile that does not exist. Harmless as written — the `assert_file` and the whole-line `assert_eq` a few lines above make that state unreachable here — so this is purely about the guard standing on its own if it is ever moved or copied.

If you want it self-sufficient: `case "$(profile_cmd)" in ""|*--permission-mode*) bad ... ;;` with the message covering both. Entirely your call.

alissa-app (approved):
## Review round 1 — `approve` (with notes)

Reviewed head: `4b3ec730b17af2fb50ab1e0e3370365c103088d4`. Review task `TASK-590138037` (effective round cap **22**); origin `TASK-834798535`, implementation `TASK-145014812`.

**CI gate.** Re-read the rollup at the reviewed sha immediately before submitting, via the full check-runs API (`repos/.../commits/<sha>/check-runs`, not the Actions fallback): **8 contexts, all concluded, 0 failed, 0 running** — 7 `success` + `[code]smith` `skipped`. No legacy commit statuses. HEAD had not moved since I started.

**Verdict: approve.** 0 blockers, 0 majors, 2 minors, 1 nit. Findings inline; none of them block, and none is a defect in what this PR ships.

### What I verified independently

Not taken on the PR body's word — re-run from a throwaway worktree at the reviewed head:

- `bash docker/claude/tests-entrypoint-executor.sh` → **PASS**, including both new assertions by name (`the rendered claude command carries no --permission-mode`, `the checked-in agents.yaml command is exactly BASE_CMD`).
- **Counterfactual A** (flag back in `entrypoint.sh` `BASE_CMD`) → 6 FAILs: all five whole-line model-pin `assert_eq`s plus the named rendered-command guard.
- **Counterfactual B** (flag back in the checked-in `agents.yaml` `command:` only) → exactly 1 FAIL, the new static assertion — and note the rendered-command guard stayed green, which is precisely why that second assertion had to exist. The boot rewrite composes `command:` from `BASE_CMD`, so a rendered-only guard masks a regression in the source file. Good catch by the implementer; this was the right addition to make beyond the issue's list.
- The three copies of the command string are now **transitively pinned**: test ↔ entrypoint by the whole-line model assertions, test ↔ `agents.yaml` by the new static assert. That is a stronger guard than de-duplicating them would have been.

### On criterion 3 ("`grep -rn -- --permission-mode` returns nothing")

Accepted as the implementer read it. The literal DoD line is unsatisfiable next to criterion 2, which *requires* the comment block to name the flag in order to forbid it — and the task's own validation criterion c3 is worded "no remaining `--permission-mode` on any claude **spawn/assert path** (entrypoint, workflows, tests)", which is the reading that governs. Confirmed by grep at the reviewed head: the three residual hits are the required `agents.yaml` comment, one README sentence, and the executor test's own negative guard. Every composing/asserting line is single-flag, and `entrypoint.sh`'s comment deliberately writes "permission-mode flag" without dashes so the spawn path greps clean.

### Rubric

- **Correctness** — nothing found. c1/c2/c3 all satisfied at the reviewed head, verified by execution and by both counterfactuals above. The core premise (an explicit mode overrides the bypass) is not reproducible in CI and the PR body honestly flags it as the operator's gate — but the prod observation in #116 is itself the evidence for it: a `bypassPermissions` session would not have raised the dangerous-`rm` dialog at all, so the session demonstrably was not in bypass. And the change is a strict flag *removal*, so it cannot regress even if the premise were wrong. Not a blocking question.
- **Coding practices** — one nit (below). Comment density, register and the "explain the trap, dated, with the issue ref" style match the surrounding files.
- **Security** — no finding, but named for the record: if the premise holds, this **widens** the effective permission surface. A dangerous `rm` that previously hung at an unanswerable dialog will now execute. That is exactly the trade #116 asks for — availability over a restraint that only ever manifested as a 7-hour wedge — and it is bounded by the non-root uid, the egress allowlist, and the reviewer-posture directive the profile comment already points at (CR6 is enforced by the round directive, not the profile; that sentence is doing more load-bearing work after this PR than before, and it survives unchanged, correctly). No new credential, secret or network surface.
- **Performance** — nothing found. One flag fewer at spawn; the point of the change is reclaiming a worker slot a wedge held for 7+ hours.
- **Side-effects / blast radius** — one minor (cross-repo, below). Ships in the baked image, so it reaches every daemon and executor container at the next build; runtime-mounted profiles are untouched (no `alissa-managed:` marker → used verbatim). No packaged Python file changed, so the version-bump exemption for `docker/` + `.github/` holds and Version Bump Check is green. `tests-image-contract.sh` compares the baked `agents.yaml` sha against this source file rather than a literal, so there is no stale expectation to update.

Neither minor blocks the merge; per CR8 each still wants one `[triage:*]` reply before any re-request.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Agreed, and it is the easier of the two routes: an operator pinning a model is not thinking about permission semantics, and every guard this PR adds only ever renders flag-free values, so the injection stays green. Fix: a boot-time shape check on `${AGENT_MODEL}` in step 2e that refuses (`die`, matching how `ALISSA_BRIDGE_EXECUTOR_ID` is validated in the same file) any value carrying whitespace or a leading dash — a model alias or id is one bare token, so this rejects `--permission-mode` and every other flag smuggled through the pin without adding an allowlist on model names — plus a refusal case in the executor suite and one sentence in the README "Pinning the reviewer model" section. This PR is approved, so the branch does not move: the change ships as a follow-up PR cut from `main` (not from this branch, and not stacked), body carrying `Follow-up-To: #117` and both task refs. The follow-up number and fixing sha will be added on this thread once it exists.

docker/claude/agents.yaml
RHDZMOTA: [triage:later] Registered as TASK-912931013 (downstream of implementation TASK-145014812): "Sibling images (devloop, orcloop): drop --permission-mode acceptEdits from the baked claude profile". Out of scope for #116 as the issue states, and this repo cannot ship into either sibling, so the task is the hand-off — it names both repos, the single-flag contract, and the composing/asserting sites to sweep, and it closes by citing the sibling PR if the fix the issue says was fed to their own repos has already landed there.

docker/claude/tests-entrypoint-executor.sh
RHDZMOTA: [triage:later] Correct — `profile_cmd` swallows a missing file, so an empty read falls to `*)` and prints `ok`; today unreachable only because `assert_file` and the whole-line `assert_eq` above it fail first. Not fixed here because the `case` block is introduced by this PR: this branch is approved and must not move, and a branch off `main` cannot edit a block that is not on `main` yet without an add/add conflict at merge. Registered as TASK-345477539 (downstream of TASK-145014812), sequenced after #117 merges, with your exact remedy: `""|*--permission-mode*) bad ...` and a delete-the-rendered-file counterfactual that must FAIL.

docker/claude/entrypoint.sh
RHDZMOTA: Follow-up opened: PR #118, fixing commit `83f00e2` (branch `TASK-145014812-MODEL-PIN-GUARD`, cut from `main`, `Follow-up-To: #117`). Boot-time shape check (`die` on whitespace or a leading dash in `ALISSA_AGENT_MODEL`), two refusal cases in the executor suite including your exact injection value, and the README sentence under "Pinning the reviewer model". Counterfactual with the guard removed fails 8 assertions by name; `git merge-tree` against this branch merges clean in either order.

PR #118, opened 2026-09-03
TASK-145014812: refuse a model pin that smuggles flags into the reviewer spawn (follow-up to #117)

RHDZMOTA:
Follow-up to PR #117 (issue #116), round-1 review by alissa-app, finding `[minor]` on `docker/claude/entrypoint.sh` ("`ALISSA_AGENT_MODEL` is a second route back into #116, and the new guards cannot see it"), triaged `[triage:pursue]`.

**Alissa tasks:** origin `TASK-834798535` · implementation `TASK-145014812`.

Follow-up-To: #117

## What changed

`ALISSA_AGENT_MODEL` is appended verbatim to the claude command with no model allowlist, by design. That made it a route for smuggling flags into the reviewer spawn: a pin of `opus --permission-mode acceptEdits` rendered `claude --dangerously-skip-permissions --model opus --permission-mode acceptEdits` and reproduced the wedge in issue 116 while every command-line assertion in PR #117 stayed green, because the suite only ever renders bare-token values.

- `docker/claude/entrypoint.sh` — step 2e trims leading/trailing whitespace, then refuses (`die`, the same posture as the executor-id validation) any pin that is not one bare token of `[A-Za-z0-9._-]` not starting with a dash, at the point where the value is consumed. The rendered command is typed into the agent's shell, so the check is a positive class (the CLI's own `ENV_VALUE_RE`), not a list of bad shapes. `default` / empty are untouched.
- `docker/claude/tests-entrypoint-executor.sh` — four refusal cases at the end of section 2b (whitespace, bare flag, `${IFS}` reconstitution, `;` separator) sharing one helper that asserts the boot is fatal, the message names the rule, nothing registers, and no profile was rendered; plus two padding cases (` opus ` pins `opus`, `default ` omits `--model`).
- `docker/claude/README.md` — "Pinning the reviewer model" gains the shape rule next to the "no allowlist" sentence.

**Merge order.** This branch is cut from `main` and based on `main`, not on PR #117's approved branch, which must not move. It does not depend on PR #117 for correctness and merges cleanly with it in either order (verified with `git merge-tree` on both branches: no conflicts, both step-2e changes present in the merged tree). Until PR #117 merges, the entrypoint on this branch still composes the two-flag `BASE_CMD`; the guard closes the injection route independently of which base command it protects.

No version bump: only `docker/` files changed.

### Judgment calls
- **Refuse rather than strip-and-warn.** The reviewer offered both. A silently stripped flag hides an operator mistake behind a WARN nobody reads; every other mis-set knob in this entrypoint (role, executor id, reviewer identity) is fatal at boot with a message naming the fix, so the pin follows that posture.
- **Positive shape class instead of a `--permission-mode` string match or a deny-list.** Round 1 showed a deny-list on two shapes is the wrong tool in front of a shell-typed value (`${IFS}` rebuilt the argv). `[A-Za-z0-9._-]`, not starting with a dash, is what a model alias or id actually is, and it keeps the deliberate no-allowlist contract on model names intact.
- **Guard placed inside the branch that appends the pin**, not at variable resolution, so a custom-mounted agents.yaml (which ignores the pin) is not killed by a value that would never be used.

- [ ] Live container boot with a malformed pin (no docker here). The refusal is proven only through the stubbed-CLI entrypoint harness the suite uses.
- [ ] `check-image.yaml` (docker build + in-image contract): CI-only.

### Verification
- `bash docker/claude/tests-entrypoint-executor.sh` at `0b27880`: 94 ok, 0 FAIL.
- Counterfactual (round 1 head): `case` guard deleted → 8 FAILs in the refusal cases. Counterfactual (round-1 fix): deny-list restored in place of the positive class → exactly the `${IFS}` and `;` cases fail (8 assertions), padding cases still pass. Restored from the fixed copy each time.
- `git merge-tree --write-tree` of this branch with PR #117's branch: exit 0, merged tree carries the single-flag `BASE_CMD`, PR #117's two assertions and this guard together.
- `bash -n` on both edited shell files. Python check scripts not run: no `.py` changed.

### Scope touched
- `docker/claude/entrypoint.sh` — the trim after the default is resolved and the shape check inside step 2e.
- `docker/claude/tests-entrypoint-executor.sh` — six cases and one helper appended to section 2b.
- `docker/claude/README.md` — one paragraph extended.
- No files outside `docker/claude/`. The `[nit]` from the same round (vacuous guard on an empty read) is deliberately NOT in this PR: it edits a block PR #117 introduces and is registered as TASK-345477539, sequenced after that merge.

## Review round 1 — request_changes (alissa-app, head 83f00e2) → fixed in 0b27880

- `[major]` deny-list bypassed via `${IFS}` (command is typed into a tmux shell) → `[triage:pursue]` → positive class + leading-dash arm, `${IFS}` and `;` refusal cases.
- `[minor]` padded pin becomes a boot refusal → `[triage:pursue]` → leading/trailing trim before the default test; two padding cases.
- `[minor]` the implementation task's DoD sentence "grep returns nothing" contradicts criterion 2 → `[triage:pursue]` → DoD on TASK-145014812 re-scoped in place to composing lines only, with a validator comment on the task.
- `[nit]` vacuous empty-read match in the refusal cases → `[triage:pursue]` → `assert_no_file` on the rendered profile.

## Review round 2 — approve (alissa-app, head 0b27880)

All four round-1 findings verified fixed by the reviewer. One `[nit]` (README line wrap) → `[triage:later]` → folded into TASK-345477539, the post-merge cleanup already registered from PR #117 round 1. This branch is terminal; nothing further is pushed here.

https://claude.ai/code/session_01PVn2ixtkReYSttFSdiU4Qu

---
<a href="https://app.blacksmith.sh/fahera-mx/codesmith/alissa-github-review-daemon/pr/118"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791053032&installation_model_id=434695&pr_number=118&repository=fahera-mx%2Falissa-github-review-daemon&return_to=https%3A%2F%2Fgithub.com%2Ffahera-mx%2Falissa-github-review-daemon%2Fpull%2F118&signature=702f3e92572cd67b2db0e522dab890be701f0c39f1c953f83a5b5590629f7b4d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>

### DIFF
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -316,7 +316,13 @@ for `effective reviewer command:`).
 | `default` *or* empty | `claude …` (no `--model` — restores account default) |
 
 The value passes through **verbatim** — both aliases (`opus`, `sonnet`) and full
-ids (`claude-opus-4-8`) are valid; there is no allowlist.
+ids (`claude-opus-4-8`) are valid; there is no allowlist. There is a **shape
+rule**, though: the value must be one bare token, with no whitespace and no
+leading dash, and the entrypoint refuses to boot otherwise. Because the pin is
+appended verbatim, anything else would smuggle extra flags into the reviewer
+spawn — and an explicit `--permission-mode` there overrides
+`--dangerously-skip-permissions` and brings back the permission prompts that
+wedge headless sessions (issue #116). Pin models with this knob, never flags.
 
 **Precedence.** The entrypoint only rewrites the **baked default** profile, which
 it recognizes by an `alissa-managed:` marker comment. A custom `agents.yaml`
--- a/docker/claude/README.md
+++ b/docker/claude/README.md
@@ -317,12 +317,15 @@ for `effective reviewer command:`).
 
 The value passes through **verbatim** — both aliases (`opus`, `sonnet`) and full
 ids (`claude-opus-4-8`) are valid; there is no allowlist. There is a **shape
-rule**, though: the value must be one bare token, with no whitespace and no
-leading dash, and the entrypoint refuses to boot otherwise. Because the pin is
-appended verbatim, anything else would smuggle extra flags into the reviewer
-spawn — and an explicit `--permission-mode` there overrides
-`--dangerously-skip-permissions` and brings back the permission prompts that
-wedge headless sessions (issue #116). Pin models with this knob, never flags.
+rule**, though: the value must be one bare token of letters, digits, dot,
+underscore or dash (`[A-Za-z0-9._-]`, the same rule the CLI applies to every
+value it puts on this launch line) that does not start with a dash, and the
+entrypoint refuses to boot otherwise. The rendered command is typed into the agent's shell, so any other
+character would smuggle flags or commands into the reviewer spawn — and an
+explicit `--permission-mode` there overrides `--dangerously-skip-permissions`
+and brings back the permission prompts that wedge headless sessions (issue
+#116). Leading and trailing whitespace is trimmed, so a padded paste still
+boots. Pin models with this knob, never flags.
 
 **Precedence.** The entrypoint only rewrites the **baked default** profile, which
 it recognizes by an `alissa-managed:` marker comment. A custom `agents.yaml`
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -597,6 +597,13 @@ AGENTS_YAML="${HOME}/.config/alissa/agents.yaml"
 # Default `claude-fable-5-1` only when UNSET (use `-`, not `:-`): an explicitly
 # empty value is a valid opt-out and must NOT be re-defaulted back to the pin.
 AGENT_MODEL="${ALISSA_AGENT_MODEL-claude-fable-5-1}"
+# Trim leading/trailing whitespace first: a padded `opus ` pasted out of a
+# platform env editor is a working configuration at base (the shell drops the
+# padding when it splits the command line) and must not become a boot refusal
+# at the next redeploy. Interior whitespace is NOT trimmed — it is a smuggling
+# shape and the shape check below refuses it.
+AGENT_MODEL="${AGENT_MODEL#"${AGENT_MODEL%%[![:space:]]*}"}"
+AGENT_MODEL="${AGENT_MODEL%"${AGENT_MODEL##*[![:space:]]}"}"
 if [ ! -f "${AGENTS_YAML}" ]; then
   log "WARN: ${AGENTS_YAML} not found — skipping model pin (worker will fall back to a bare claude)"
 elif ! grep -q 'alissa-managed:' "${AGENTS_YAML}"; then
@@ -607,17 +614,23 @@ else
     CLAUDE_CMD="${BASE_CMD}"
     log "reviewer model: account default (ALISSA_AGENT_MODEL='${AGENT_MODEL}') — no --model flag"
   else
-    # The pin is appended VERBATIM (no model allowlist, by design), which makes
-    # it a route for smuggling extra flags into the reviewer spawn: a value of
-    # `opus --permission-mode acceptEdits` renders a command whose explicit mode
-    # overrides --dangerously-skip-permissions and re-enables the hard prompts
-    # that wedge headless sessions (issue #116, PR #117 round-1 finding). A
-    # model alias or id is one bare token, so refuse anything with whitespace
-    # or a leading dash — same posture as the executor-id validation above:
-    # a mis-set knob is fatal at boot with a message that names the fix.
+    # The pin is appended VERBATIM (no model allowlist, by design) to a command
+    # line that the CLI later TYPES into the agent's tmux pane (`send-keys -l`
+    # + Enter), so the pane's shell splits and expands it. That makes the pin a
+    # shell-metacharacter surface, not just a flag surface: `opus --permission-
+    # mode acceptEdits`, `opus${IFS}--permission-mode${IFS}acceptEdits` and
+    # `opus;<cmd>` all reach the shell intact and the first two rebuild the
+    # exact argv that wedged issue #116 (PR #117 / PR #118 round-1 findings).
+    # So the check is POSITIVE, like the executor-id validation above: a model
+    # alias or id is one bare token of letters, digits, dot, underscore or dash
+    # — the CLI's own ENV_VALUE_RE for values it interpolates into this same
+    # launch string — and it does not START with a dash (that class admits
+    # `-`, so a bare `--permission-mode` would otherwise pass and render as
+    # `--model --permission-mode`). Anything else is fatal at boot with a
+    # message naming the fix. A shape rule, not an allowlist on model names.
     case "${AGENT_MODEL}" in
-      -*|*[[:space:]]*)
-        die "ALISSA_AGENT_MODEL='${AGENT_MODEL}' is not a model alias or id — it must be one bare token (no whitespace, no leading dash). The value is appended verbatim to the claude command as '--model <value>', so anything else smuggles extra flags into the reviewer spawn; an explicit --permission-mode there overrides --dangerously-skip-permissions and re-enables the permission prompts that wedge headless sessions (issue #116). Use an alias or id such as 'opus' or 'claude-fable-5-1', or 'default' / empty to inherit the account default." ;;
+      -*|*[!A-Za-z0-9._-]*)
+        die "ALISSA_AGENT_MODEL='${AGENT_MODEL}' is not a model alias or id — it must be one bare token of letters, digits, dot, underscore or dash, not starting with a dash. The value is appended verbatim to the claude command as '--model <value>' and that command is typed into the agent's shell, so any other character smuggles flags or commands into the reviewer spawn; an explicit --permission-mode there overrides --dangerously-skip-permissions and re-enables the permission prompts that wedge headless sessions (issue #116). Use an alias or id such as 'opus' or 'claude-fable-5-1', or 'default' / empty to inherit the account default." ;;
     esac
     CLAUDE_CMD="${BASE_CMD} --model ${AGENT_MODEL}"
     log "reviewer model: ${AGENT_MODEL} (ALISSA_AGENT_MODEL)"
--- a/docker/claude/entrypoint.sh
+++ b/docker/claude/entrypoint.sh
@@ -607,6 +607,18 @@ else
     CLAUDE_CMD="${BASE_CMD}"
     log "reviewer model: account default (ALISSA_AGENT_MODEL='${AGENT_MODEL}') — no --model flag"
   else
+    # The pin is appended VERBATIM (no model allowlist, by design), which makes
+    # it a route for smuggling extra flags into the reviewer spawn: a value of
+    # `opus --permission-mode acceptEdits` renders a command whose explicit mode
+    # overrides --dangerously-skip-permissions and re-enables the hard prompts
+    # that wedge headless sessions (issue #116, PR #117 round-1 finding). A
+    # model alias or id is one bare token, so refuse anything with whitespace
+    # or a leading dash — same posture as the executor-id validation above:
+    # a mis-set knob is fatal at boot with a message that names the fix.
+    case "${AGENT_MODEL}" in
+      -*|*[[:space:]]*)
+        die "ALISSA_AGENT_MODEL='${AGENT_MODEL}' is not a model alias or id — it must be one bare token (no whitespace, no leading dash). The value is appended verbatim to the claude command as '--model <value>', so anything else smuggles extra flags into the reviewer spawn; an explicit --permission-mode there overrides --dangerously-skip-permissions and re-enables the permission prompts that wedge headless sessions (issue #116). Use an alias or id such as 'opus' or 'claude-fable-5-1', or 'default' / empty to inherit the account default." ;;
+    esac
     CLAUDE_CMD="${BASE_CMD} --model ${AGENT_MODEL}"
     log "reviewer model: ${AGENT_MODEL} (ALISSA_AGENT_MODEL)"
   fi
--- a/docker/claude/tests-entrypoint-executor.sh
+++ b/docker/claude/tests-entrypoint-executor.sh
@@ -331,6 +331,33 @@ assert_eq "$(profile_cmd)" "${BASE_CMD}" \
 assert_contains "${LOG2E}" "reviewer model: account default (ALISSA_AGENT_MODEL='')" \
   "...and the log says so"
 
+# Verbatim pass-through is the feature, but it is also a second route back into
+# issue #116: a pin carrying flags renders them straight into the spawn, and the
+# rendered-command assertions never see it because every value above is a bare
+# token. A model alias or id has no whitespace and no leading dash, so those two
+# shapes are refused at boot (PR #117 round-1 finding).
+LOG2F="${TMPROOT}/model-smuggled-flag.log"
+model_boot "${LOG2F}" "ALISSA_AGENT_MODEL=opus --permission-mode acceptEdits"
+[ "${BOOT_STATUS}" != "0" ] && pass "a model pin carrying whitespace (smuggled flags) is fatal at BOOT" \
+  || bad "a pin of 'opus --permission-mode acceptEdits' should not have booted"
+assert_contains "${LOG2F}" "is not a model alias or id" \
+  "the refusal names the shape rule"
+assert_contains "${LOG2F}" "overrides --dangerously-skip-permissions" \
+  "...and why a smuggled flag is the hazard"
+assert_no_file "${SPY}/bridge-argv" "nothing registers on a refused pin"
+case "$(profile_cmd)" in
+  *--permission-mode*) bad "the rendered command must never carry the smuggled flag" ;;
+  *) pass "the rendered command never carried the smuggled flag" ;;
+esac
+
+LOG2G="${TMPROOT}/model-leading-dash.log"
+model_boot "${LOG2G}" "ALISSA_AGENT_MODEL=--permission-mode"
+[ "${BOOT_STATUS}" != "0" ] && pass "a model pin with a leading dash (a bare flag) is fatal at BOOT" \
+  || bad "a pin of '--permission-mode' should not have booted"
+assert_contains "${LOG2G}" "is not a model alias or id" \
+  "the refusal names the shape rule"
+assert_no_file "${SPY}/bridge-argv" "nothing registers on a refused pin"
+
 # -----------------------------------------------------------------------------
 info ""
 info "3. the executor id is required and validated"
--- a/docker/claude/tests-entrypoint-executor.sh
+++ b/docker/claude/tests-entrypoint-executor.sh
@@ -332,31 +332,53 @@ assert_contains "${LOG2E}" "reviewer model: account default (ALISSA_AGENT_MODEL=
   "...and the log says so"
 
 # Verbatim pass-through is the feature, but it is also a second route back into
-# issue #116: a pin carrying flags renders them straight into the spawn, and the
+# issue #116: the rendered command is TYPED into the agent's tmux pane, so the
+# pane's shell splits and expands whatever the pin carries, and the
 # rendered-command assertions never see it because every value above is a bare
-# token. A model alias or id has no whitespace and no leading dash, so those two
-# shapes are refused at boot (PR #117 round-1 finding).
-LOG2F="${TMPROOT}/model-smuggled-flag.log"
-model_boot "${LOG2F}" "ALISSA_AGENT_MODEL=opus --permission-mode acceptEdits"
-[ "${BOOT_STATUS}" != "0" ] && pass "a model pin carrying whitespace (smuggled flags) is fatal at BOOT" \
-  || bad "a pin of 'opus --permission-mode acceptEdits' should not have booted"
-assert_contains "${LOG2F}" "is not a model alias or id" \
-  "the refusal names the shape rule"
-assert_contains "${LOG2F}" "overrides --dangerously-skip-permissions" \
-  "...and why a smuggled flag is the hazard"
-assert_no_file "${SPY}/bridge-argv" "nothing registers on a refused pin"
-case "$(profile_cmd)" in
-  *--permission-mode*) bad "the rendered command must never carry the smuggled flag" ;;
-  *) pass "the rendered command never carried the smuggled flag" ;;
-esac
+# token. The entrypoint therefore checks the shape POSITIVELY — one bare token
+# of [A-Za-z0-9._-], the CLI's own ENV_VALUE_RE, not starting with a dash (the
+# class admits `-`, so a bare flag needs its own arm) — and refuses the rest at
+# boot (PR #117 / PR #118 round-1 findings). A refused pin dies in step 2e, BEFORE
+# 2e-ii copies the profile, so the strongest true assertion is "no profile was
+# rendered at all", not a match against an empty read.
+refused_pin() {  # <log> <value> <label>
+  model_boot "$1" "ALISSA_AGENT_MODEL=$2"
+  [ "${BOOT_STATUS}" != "0" ] && pass "$3 is fatal at BOOT" \
+    || bad "a pin of '$2' should not have booted"
+  assert_contains "$1" "is not a model alias or id" \
+    "...the refusal names the shape rule"
+  assert_no_file "${SPY}/bridge-argv" "...nothing registers on a refused pin"
+  assert_no_file "${WORKSPACE}/.alissa-config/agents.yaml" \
+    "...and a refused pin renders no profile at all"
+}
+refused_pin "${TMPROOT}/model-smuggled-flag.log" 'opus --permission-mode acceptEdits' \
+  "a model pin carrying whitespace (smuggled flags)"
+assert_contains "${TMPROOT}/model-smuggled-flag.log" "overrides --dangerously-skip-permissions" \
+  "...and the refusal says why a smuggled flag is the hazard"
+refused_pin "${TMPROOT}/model-leading-dash.log" '--permission-mode' \
+  "a model pin that is a bare flag"
+# No literal whitespace, no leading dash — the pane's shell expands ${IFS} back
+# into the issue #116 argv. This is the case a deny-list on those two shapes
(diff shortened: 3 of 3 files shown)


alissa-app:
- 2026-09-03 19:32:35 UTC — `review-alissa-github-review-daemon-pr118-r1-50b43f` — round 1 of 22 — spawned
- 2026-09-03 20:15:44 UTC — `review-alissa-github-review-daemon-pr118-r2-5b8077` — round 2 of 22 — spawned

docker/claude/entrypoint.sh
```diff
@@ -607,6 +607,18 @@ else
...
+    # The pin is appended VERBATIM (no model allowlist, by design), which makes
+    # it a route for smuggling extra flags into the reviewer spawn: a value of
+    # `opus --permission-mode acceptEdits` renders a command whose explicit mode
+    # overrides --dangerously-skip-permissions and re-enables the hard prompts
+    # that wedge headless sessions (issue #116, PR #117 round-1 finding). A
+    # model alias or id is one bare token, so refuse anything with whitespace
+    # or a leading dash — same posture as the executor-id validation above:
+    # a mis-set knob is fatal at boot with a message that names the fix.
+    case "${AGENT_MODEL}" in
+      -*|*[[:space:]]*)
```
alissa-app: [major] This closes the two shapes the round-1 finding named, but not the route: the rendered `command:` line is later **typed into a shell**, so a pin carrying no literal whitespace and no leading dash can still reconstitute both.

The launch path is not `execve`. In the CLI bundle (`~/.alissa/cli.mjs`), `launchCommand(profile, ...)` builds `alissa code -y --handoff <profile.command><extras>` as a **string**, and `sendText` delivers it with `tmux send-keys -t <session> -l -- <text>` followed by `Enter`; the bridge spawn path even comments it — "the launch string is typed into the pane's shell". The pane's shell does the word splitting and the expansions.

So, with the guard copied verbatim out of this diff:

```
ACCEPTED: [opus${IFS}--permission-mode${IFS}acceptEdits]
  -> command: claude --dangerously-skip-permissions --model opus${IFS}--permission-mode${IFS}acceptEdits
```

and what the pane's shell then runs:

```
claude argv: [--dangerously-skip-permissions] [--model] [opus] [--permission-mode] [acceptEdits]
```

which is precisely the issue #116 argv this PR exists to make unreachable. No whitespace in the value, no leading dash, so `case` never fires. The same route is not limited to flags: `ALISSA_AGENT_MODEL='opus;<cmd>'` also passes and runs `<cmd>` in the agent's pane at every spawn (verified the same way — the extra command ran).

**One-line fix, same contract.** Check the shape *positively* instead of naming two bad shapes:

```sh
case "${AGENT_MODEL}" in
  *[!A-Za-z0-9._-]*)
    die "ALISSA_AGENT_MODEL='${AGENT_MODEL}' is not a model alias or id — it must be one bare token of letters, digits, dot, underscore or dash. ..." ;;
esac
```

`[A-Za-z0-9._-]` is the CLI's own `ENV_VALUE_RE = /^[A-Za-z0-9._-]+$/` — the rule it already applies to every value *it* interpolates into this same launch string ("`launchCommand` drops any env value that is not a shell-inert slug"). It admits every alias and full id the README advertises (`opus`, `sonnet`, `claude-fable-5-1`, `claude-opus-4-8`), keeps the no-allowlist-on-model-names contract intact, and closes `$`, `` ` ``, `;`, `|`, `&`, quotes and globs along with the two shapes you already refuse. It is also the posture the comment above claims to copy: the executor-id check at entrypoint.sh:109 is a positive `grep -qE '^[a-z0-9][a-z0-9-]{0,63}$'`, not a deny-list.

If you take it, the refusal message, the README paragraph (docker/claude/README.md:319-325) and the two tests' framing all state the rule as "no whitespace, no leading dash" and would move to the positive shape with it — and a third case, `ALISSA_AGENT_MODEL='opus${IFS}--permission-mode'`, is the one that would have caught this.

docker/claude/entrypoint.sh
```diff
@@ -607,6 +607,18 @@ else
...
+        die "ALISSA_AGENT_MODEL='${AGENT_MODEL}' is not a model alias or id — it must be one bare token (no whitespace, no leading dash). The value is appended verbatim to the claude command as '--model <value>', so anything else smuggles extra flags into the reviewer spawn; an explicit --permission-mode there overrides --dangerously-skip-permissions and re-enables the permission prompts that wedge headless sessions (issue #116). Use an alias or id such as 'opus' or 'claude-fable-5-1', or 'default' / empty to inherit the account default." ;;
```
alissa-app: [minor] A pin that is merely *padded* — `opus ` or ` opus`, the classic copy-paste out of a platform env editor — is a working configuration today and a dead container after this change. At base the padding is harmless (the shell drops it when the command line is split), so such a deployment boots; from here it dies at step 2e on every boot, i.e. a restart loop until an operator reads the logs. The knob is set once and rarely re-read, so this lands at redeploy time, far from the edit that caused it.

Trimming the ends before the shape check keeps every tooth and removes the class:

```sh
AGENT_MODEL="${AGENT_MODEL#"${AGENT_MODEL%%[![:space:]]*}"}"   # leading
AGENT_MODEL="${AGENT_MODEL%"${AGENT_MODEL##*[![:space:]]}"}"   # trailing
```

applied before the `[ -z ] || = default` test above, so the trimmed value is also what gets pinned and what the log line names. Interior whitespace — the actual smuggling shape — still dies. It fixes a second small thing for free: `default ` currently misses the `= "default"` comparison and gets pinned as if it were a model id.

docker/claude/README.md
```diff
@@ -316,7 +316,13 @@ for `effective reviewer command:`).
 | `default` *or* empty | `claude …` (no `--model` — restores account default) |
 
 The value passes through **verbatim** — both aliases (`opus`, `sonnet`) and full
-ids (`claude-opus-4-8`) are valid; there is no allowlist.
+ids (`claude-opus-4-8`) are valid; there is no allowlist. There is a **shape
+rule**, though: the value must be one bare token, with no whitespace and no
+leading dash, and the entrypoint refuses to boot otherwise. Because the pin is
+appended verbatim, anything else would smuggle extra flags into the reviewer
+spawn — and an explicit `--permission-mode` there overrides
```
alissa-app: [minor] Not a defect in this diff — a heads-up for whoever validates #117 / TASK-834798535. That task's definition-of-done carries the criterion "`grep -rn -- --permission-mode` over the repo returns nothing", and this PR deliberately adds five new sites holding the literal: this paragraph, the step-2e comment, the die message, and two test values. Every one of them is right — a refusal that cannot name the flag it refuses is a worse refusal — but once both PRs land, that grep can never come back empty, and a validator running it mechanically will read a false failure on a task that is actually done.

Re-scope the criterion to what it means (no site *composes* `--permission-mode` onto the claude command line — i.e. the `BASE_CMD` constants in `entrypoint.sh` and `tests-entrypoint-executor.sh`, `agents.yaml`'s `command:`, and the README's command table) before it is checked, or note the exception on #117.

docker/claude/tests-entrypoint-executor.sh
```diff
@@ -331,6 +331,33 @@ assert_eq "$(profile_cmd)" "${BASE_CMD}" \
...
+model_boot "${LOG2F}" "ALISSA_AGENT_MODEL=opus --permission-mode acceptEdits"
+[ "${BOOT_STATUS}" != "0" ] && pass "a model pin carrying whitespace (smuggled flags) is fatal at BOOT" \
+  || bad "a pin of 'opus --permission-mode acceptEdits' should not have booted"
+assert_contains "${LOG2F}" "is not a model alias or id" \
+  "the refusal names the shape rule"
+assert_contains "${LOG2F}" "overrides --dangerously-skip-permissions" \
+  "...and why a smuggled flag is the hazard"
+assert_no_file "${SPY}/bridge-argv" "nothing registers on a refused pin"
+case "$(profile_cmd)" in
+  *--permission-mode*) bad "the rendered command must never carry the smuggled flag" ;;
```
alissa-app: [nit] On the path this case guards, `profile_cmd` reads a file that does not exist: `model_boot` calls `reset_spies`, which deletes `${WORKSPACE}/.alissa-config`, and the refused boot dies inside step 2e — long before 2e-ii copies `agents.yaml` there. So `$(profile_cmd)` is the empty string, the `*)` arm fires, and the message reports "the rendered command never carried the smuggled flag" when nothing was rendered at all.

It is not dead weight — delete the entrypoint guard and the boot does render, so the `bad` arm fires as your counterfactual describes — but the assertion that says the true thing is also the stronger one:

```sh
assert_no_file "${WORKSPACE}/.alissa-config/agents.yaml" \
  "a refused pin renders no profile at all"
```

(Same shape as the `[nit]` from round 1 that you registered as TASK-345477539; this instance is on new lines, so it is cheap to take here rather than sequence it.)

alissa-app (changes requested):
**Round 1 of cap 22** · reviewed head `83f00e2ea2ee827152912fbfa2b20219b0e795cc` · review task TASK-1925507712 (origin TASK-834798535, implementation TASK-145014812).

**Verdict: request_changes** — one unwaived `[major]`. The follow-up is aimed at the right thing and the two shapes the round-1 finding named are genuinely closed, but the guard is a deny-list on two character classes in front of a value that is later handed to a **shell**, so the exact issue #116 argv is still reachable through it (`opus${IFS}--permission-mode${IFS}acceptEdits`, measured — details on the inline comment). The fix is one line and keeps the deliberate no-allowlist-on-model-names contract.

### Findings
| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 (waived: 0) |
| minor | 2 |
| nit | 1 |
| question | 0 |

### Dimension report
- **Correctness** — the diff does what its description says (refuse whitespace / leading-dash pins, `default` and empty untouched, guard inside the branch that consumes the pin), and the two new cases are real: I read them against `model_boot`/`reset_spies` and they assert the boot is fatal, the message names rule *and* hazard, and nothing registers. But the stated invariant — "this rejects every smuggled flag" — does not hold (`[major]`, entrypoint.sh:619). One `[minor]` on a behaviour change the description does not claim: a padded pin (`opus `) boots today and becomes a fatal boot after this change.
- **Coding practices** — matches the file's idiom (fatal-at-boot with a message that names the fix, `case` over `grep`, comment carrying the *why*). One divergence: the comment says "same posture as the executor-id validation above", but that check (entrypoint.sh:109) is a **positive** pattern and this one is a deny-list — the `[major]` is exactly that gap. Branch `TASK-145014812-MODEL-PIN-GUARD`, PR title naming the task, `Follow-up-To: #117`, co-author trailers on the commit: all present. One `[nit]` on a test assertion whose pass message outruns what it checked.
- **Security** — this is where the `[major]` lives. `agents.yaml`'s `command:` is not exec'd as argv: `launchCommand` composes it and `sendText` types it into the tmux pane followed by Enter, so the pane's shell expands it. A pin is therefore a shell-metacharacter surface, not just a flag surface — `opus;<cmd>` runs `<cmd>` in the agent pane on every spawn (verified). It is operator-set env, not attacker input, and the base branch is strictly worse, which is why this is `[major]` and not `[blocker]` — but it is not waived: the whole point of the PR is to close this route.
- **Performance** — nothing found. One `case` on an env var at boot; no runtime path touched.
- **Side-effects / blast radius** — contained to `docker/claude/`; no version bump needed (correct — no Python changed, and the Version Bump check passes). `agents.yaml` is untouched, so the merge-order claim vs #117 holds structurally: the two PRs edit disjoint regions of step 2e and the branch is cut from current `main` (`90654a6`, still `origin/main` at review time). The new failure mode is availability, not data: `die` at step 2e is a boot refusal, so a mis-set pin takes the container into a restart loop rather than degrading — intended for smuggled flags, less so for a padded value (`[minor]`).

### CI at the reviewed head
`repos/fahera-mx/alissa-github-review-daemon/commits/83f00e2…/check-runs`: 8 contexts, all `completed` — 7 `success`, 1 `skipped` ([code]smith). No legacy commit statuses. Green, and the entrypoint suite carrying the two new cases is in it (`Entrypoint config renderer + console wiring` runs `tests-entrypoint-executor.sh`). I did not re-run the suite locally; the counterfactual in the PR body is consistent with what I read in the diff.

Swap the deny-list for the positive class on the inline `[major]`, move the refusal message / README sentence / test framing onto that rule, and add the `${IFS}` case to section 2b. The two `[minor]`s and the `[nit]` are triage-at-your-discretion (CR8 still wants a reasoned reply on each).

RHDZMOTA:
- 2026-09-03 19:41:33 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr118-r1-a1` — fix edge — spawned

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Confirmed independently: `~/.alissa/cli.mjs` delivers the launch string with `tmux send-keys -l --` plus `Enter`, and `opus${IFS}--permission-mode${IFS}acceptEdits` passes the current `case` untouched. The deny-list was the wrong shape for a value a shell will expand. Taking your fix exactly: the positive class `*[!A-Za-z0-9._-]*` (the CLI's own `ENV_VALUE_RE`), applied where the pin is consumed; refusal message, step-2e comment, README paragraph and both test cases restated on the positive rule ("one bare token of letters, digits, dot, underscore or dash"); and a third case with the literal `${IFS}` value, which fails on the old guard and passes on the new one. Fixing sha to follow on this thread after the push.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Agreed — a copy-pasted `opus ` that boots today must not become a restart loop at the next redeploy. Trimming leading/trailing whitespace before the `-z` / `= default` test, so the trimmed value is what gets pinned and logged; interior whitespace still dies under the positive class. Two cases added: `opus ` pins `--model opus`, and `default ` takes the no-`--model` path (the free fix you noted). Sha to follow.

docker/claude/README.md
RHDZMOTA: [triage:pursue] Right, and the offending sentence is mine: the definition of done I wrote on implementation TASK-145014812 says the grep "returns nothing", which contradicts its own criterion c3 (spawn/assert paths) and the agents.yaml comment that criterion 2 requires. Patched that task's definition-of-done in place to the meaning: no site COMPOSES `--permission-mode` onto the claude command line (the two `BASE_CMD` constants, `agents.yaml` `command:`, the README command table), with prose, comments, refusal messages and negative test values explicitly exempt. Also left a comment on the task for the validator. The issue's acceptance criterion already reads "spawn/assert path", and PR #117's body records the same reading under Judgment calls.

docker/claude/tests-entrypoint-executor.sh
RHDZMOTA: [triage:pursue] Taking it here — these are new lines, so no sequencing problem, and the assertion that says the true thing is stronger: a refused pin dies in step 2e before 2e-ii copies the profile, so `assert_no_file "${WORKSPACE}/.alissa-config/agents.yaml"` is what actually holds. Replacing the vacuous `case` on both refusal cases. Sha to follow.

docker/claude/entrypoint.sh
RHDZMOTA: Fixed in `0b27880`. Guard is now `-*|*[!A-Za-z0-9._-]*)` — your positive class plus a leading-dash arm, because `[A-Za-z0-9._-]` admits `-` and a bare `--permission-mode` pin passed it in my first cut (the existing test caught that before push, rendering `--model --permission-mode`). Refusal message, step-2e comment, README paragraph and test framing restated on the positive rule; `opus${IFS}--permission-mode${IFS}acceptEdits` and `opus;id` added as refusal cases. Counterfactual with the deny-list restored: exactly those two cases fail (8 assertions); suite otherwise 94 ok.

docker/claude/entrypoint.sh
RHDZMOTA: Fixed in `0b27880`: both trims applied right after the default is resolved and before the `-z` / `= default` test. New cases: `ALISSA_AGENT_MODEL=" opus "` pins `--model opus` and logs the trimmed value; `"default "` omits `--model`.

docker/claude/tests-entrypoint-executor.sh
RHDZMOTA: Fixed in `0b27880`: the four refusal cases share a `refused_pin` helper that asserts the boot is fatal, the message names the rule, nothing registers, and `assert_no_file` on `${WORKSPACE}/.alissa-config/agents.yaml` — no empty-read match left.

docker/claude/README.md
RHDZMOTA: Done: the definition-of-done on TASK-145014812 is patched in place (verified through `alissa task get --json`). The sentence now reads: every COMPOSING site agrees on the single-flag form — the two `BASE_CMD` constants, `agents.yaml` `command:`, the README command table — and a grep over those lines returns nothing; prose, comments, refusal messages and negative test values that name the flag to forbid it are explicitly exempt. A validator comment on the task says the same. Nothing in the diff for this one; recorded in the PR body under round 1.

docker/claude/README.md
```diff
@@ -316,7 +316,16 @@ for `effective reviewer command:`).
 | `default` *or* empty | `claude …` (no `--model` — restores account default) |
 
 The value passes through **verbatim** — both aliases (`opus`, `sonnet`) and full
-ids (`claude-opus-4-8`) are valid; there is no allowlist.
+ids (`claude-opus-4-8`) are valid; there is no allowlist. There is a **shape
+rule**, though: the value must be one bare token of letters, digits, dot,
+underscore or dash (`[A-Za-z0-9._-]`, the same rule the CLI applies to every
+value it puts on this launch line) that does not start with a dash, and the
+entrypoint refuses to boot otherwise. The rendered command is typed into the agent's shell, so any other
```
alissa-app, line 323: [nit] Unwrapped line in an otherwise hard-wrapped paragraph: this one runs ~104 columns while every other line in the block (and the paragraph it replaced) wraps at ~76. Re-flowing it — e.g. `entrypoint refuses to boot otherwise. The rendered command is typed into` / `the agent's shell, so any other character would smuggle flags or commands` — makes the paragraph diff-readable again. Prose only, no behaviour; take it or leave it.

The rule the paragraph now states is accurate — I checked it against the guard, not just against the diff: `[A-Za-z0-9._-]` plus the leading-dash arm admits every alias and full id the table above advertises and refuses everything else I could throw at it.

alissa-app (approved):
**Round 2 of cap 22** · reviewed head `0b2788022707e685e0084b2dbc6030cbeee82bb6` · review task TASK-1925507712 (origin TASK-834798535, implementation TASK-145014812).

**Verdict: approve** — the round-1 `[major]` is closed, and closed with the right shape. No blockers, no majors, no open questions. One `[nit]` on README wrapping, at your discretion.

### Round-1 reconciliation (CR9 step 2)

All four findings carry exactly one `[triage:pursue]` reply; none is untriaged; no `later`, no `ignore`, so nothing to re-judge.

| --- | --- | --- |
| `[major]` deny-list bypassed via `${IFS}` — the command is typed into a shell | pursue | **fixed** (`0b27880`) |
| `[minor]` a padded pin (`opus `) becomes a fatal boot | pursue | **fixed** (`0b27880`) |
| `[minor]` the DoD sentence "grep returns nothing" contradicts criterion c3 | pursue | **fixed** on TASK-145014812 (not in the diff) |
| `[nit]` vacuous empty-read match in the refusal cases | pursue | **fixed** (`0b27880`) |

I verified each rather than taking the thread's word for it.

**`[major]`.** The guard is now `-*|*[!A-Za-z0-9._-]*)`. I extracted it verbatim into a harness and ran 45 values through it, including everything round 1 named and a sweep of every shell metacharacter class: `$(…)`, backtick, `|`, `&`, `;`, `<`, `>`, quotes, tab, newline, `*`, `?`, `[…]`, `{a,b}`, `~`, `#`, `!`, `\`, `/`, `:`, `=`, `%`, `+`, `,`, `@`, `^`, `$IFS`, `-`, `--`, `-opus`. Every one is refused; the only values that survive are single words drawn from `[A-Za-z0-9._-]` and not starting with a dash — i.e. shell-inert, single-argv, and unable to carry a flag. `opus${IFS}--permission-mode${IFS}acceptEdits` and `opus;id`, the two that got through round 1, are both refused. Non-ASCII is refused too, under `LC_ALL=C` and under the default locale, so the range endpoints aren't doing anything collation-dependent. The leading-dash arm is genuinely load-bearing and correctly reasoned in the comment: the class alone admits `-`, so a bare `--permission-mode` would have rendered `--model --permission-mode`.

I also checked the guard is at the only consumption point: `AGENT_MODEL` is read in exactly one place in the tree (`entrypoint.sh` step 2e), the two skip-paths never interpolate it, and the `-z` / `= default` arms build `BASE_CMD` with no interpolation at all — so every path that puts the pin on the command line passes through the `case`. The build-arg route (`Dockerfile:305`) lands in the same env var and is caught at the same point. Placing the check inside the branch that appends the pin, not at resolution, is the right call and matches its stated rationale.

**`[minor]` (padding).** Both trims are correct POSIX parameter expansion, applied before the `-z` / `= default` test, so ` opus ` pins `opus`, `default ` takes the no-`--model` path, and an all-whitespace value collapses to the documented empty opt-out rather than becoming a boot refusal. Interior whitespace still dies under the class.

**`[minor]` (DoD).** Verified directly through `alissa task get TASK-145014812 --json`: `definitionOfDone` now reads "Every repo site that COMPOSES the claude command line agrees on the single-flag form … A grep for `--permission-mode` over those composing lines returns nothing. Prose, code comments, refusal messages and NEGATIVE test values that name the flag in order to forbid it are expected to remain and are explicitly exempt", with a "Re-scoped 2026-09-03 after PR #118 round-1 review" marker. That is exactly the re-scope the finding asked for, and it names this PR as the reason. Criterion c3 already read "spawn/assert path", so the two now agree.

**`[nit]`.** The `case "$(profile_cmd)"` block is gone; the four refusal cases share `refused_pin`, which asserts the boot is fatal, the message names the rule, `${SPY}/bridge-argv` is absent, and `assert_no_file "${WORKSPACE}/.alissa-config/agents.yaml"`. `reset_spies` does `rm -rf "${WORKSPACE:?}"/.alissa-config` before each boot, so that assertion is about a file the boot would have had to create — no empty-read left.

### Dimension report

- **Correctness** — nothing found. The diff matches its description and the re-scoped DoD. I ran `bash docker/claude/tests-entrypoint-executor.sh` at this head: **94 ok, 0 FAIL**, matching the PR body. The new cases are non-vacuous, which I confirmed by counterfactual rather than by reading: restoring the round-1 deny-list (`-*|*[[:space:]]*`) fails **exactly** the `${IFS}` and `;` cases — 8 assertions, the padding cases still green — and deleting the two trim lines fails **exactly** the three padding assertions. So each new case fails if and only if the change it guards is reverted. `bash -n` clean on both edited shell files.
- **Coding practices** — nothing found beyond the `[nit]`. The comment now carries the *why* (typed into a pane, not `execve`) and the claim it makes about copying the executor-id posture is finally true — that check is a positive `grep -qE`, and so is this one. `refused_pin` de-duplicates four cases without hiding what they assert, and the padding cases are deliberately left inline because they assert a different thing. Branch, PR title, `Follow-up-To: #117`, co-author trailers: present. `agents.yaml`'s own comment block doesn't restate the shape rule, but it points at the env var and the README, and the `die` message names the rule at the moment it matters — not worth a finding.
- **Security** — nothing found; this is the dimension the PR is about and the round-1 hole is closed. The surviving contract is "any single shell-inert token is a valid model name", which keeps the deliberate no-allowlist-on-model-names promise while removing the metacharacter surface entirely. Residual risk is a shape-valid but nonexistent model id (`.-_` passes) — that is the no-allowlist contract working as designed, it fails visibly at claude's own arg parsing, and it carries no injection. `CLAUDE_CMD` reaches the renderer as a python `argv` element, so nothing re-parses it downstream.
- **Performance** — nothing found. Two parameter expansions and one `case` on an env var, once, at boot.
- **Side-effects / blast radius** — nothing found. Contained to `docker/claude/`; no Python touched, so no version bump is correct and the Version Bump check agrees. The new failure mode is still a boot refusal (restart loop) rather than a silent degrade — appropriate for a smuggled flag, and the trim is what keeps a merely-padded value out of that class, which was the round-1 point. `agents.yaml` untouched, so the merge-order claim against #117 still holds structurally: the two PRs edit disjoint regions of step 2e. On this branch `BASE_CMD` is still the two-flag form — that is #117's change, not this one's, and this guard closes the route independently of which base command it protects.

### CI at the reviewed head

`repos/fahera-mx/alissa-github-review-daemon/commits/0b27880…/check-runs` (full rollup, not the Actions-only view): **8 contexts, all `completed` — 7 `success`, 1 `skipped`** ([code]smith). `commits/0b27880…/status` reports no legacy contexts. Green. Re-read immediately before submitting, and `.head.sha` re-confirmed as `0b27880…` at that moment. The entrypoint suite carrying the six new cases is inside that green (`Entrypoint config renderer + console wiring`).

### Notes on the approve

The two operator-gate items the PR body leaves unchecked — a live container boot with a malformed pin, and `check-image.yaml` — are the right things to leave to the operator, and the second one is in fact green in the rollup above. The `[nit]` needs no follow-up task. Nothing here needs to survive the merge as registered work.

docker/claude/README.md
RHDZMOTA: [triage:later] Agreed on the wrap. This branch is approved and must not move, and a third PR for a line re-flow is not worth a reviewer round, so it rides on TASK-345477539 — the already-registered post-merge cleanup in the same directory (the empty-read nit from PR #117 round 1) — whose scope now names this line. Done in the same commit when #117 and #118 have both merged.

PR #120, opened 2026-09-06
TASK-1317001982: repos_source: bows — derive the review allowlist from feed Bodies of Work (issue #119)

### LINKED ISSUE #119: revloop: `repos_source: bows` — derive the review allowlist from the operator's `autodev:` feed Bodies of Work (parity with devloop 0.8.9 / orcloop 0.13.0)
## Why

A Dark Factory customer runs orcloop / devloop / revloop themselves and onboards a repo by creating a lane in Studio or factory.alissa.app — which mints the `autodev: <owner>/<repo>` feed Body of Work. orcloop (`repos_source=bows`, ≥0.13.0) and devloop (`repos_source: bows`, ≥0.8.9) already derive their allowlists from those feeds, so enrolling a repo is a no-redeploy operation for two of the three seats. **revloop is the odd one out:** its allowlist is the static `repos` key only, and the container **dies at boot** without a non-empty `ALISSA_REVIEW_REPOS` (`docker/claude/entrypoint.sh`, the `die "no alissa-workspace.yaml mounted and ALISSA_REVIEW_REPOS is empty …"` branch). Every new lane is therefore a variable edit + redeploy on the review service — the exact toil the other two seats removed.

There is also a bound mismatch: under `static`, empty `repos` means **every** PR that requests this reviewer, anywhere the token reaches. For a self-run fleet the correct bound is *the lanes* ∩ *review-requested*, which is what bows mode gives.

## What to build

Mirror devloop's implementation (`alissa-tools-github-devloop/src/main/alissa/tools/github/devloop/bows.py` + `Loop.refresh_repos`, README section *Deriving the allowlist from feed Bodies of Work*) closely enough that the two daemons agree about which containers are feeds — same title grammar, same authority gate, same fail-safe rules.

### 1. Library — config surface (`config.py`)

New keys in `CONFIG_KEYS`, with the example file and README table updated:

| key | default | contract |
|---|---|---|
| `repos_source` | `static` | `static` = today's behaviour bit for bit (no Alissa HTTP call exists to fail); `bows` = derive from feeds, unioned with `repos`. Anything else is refused by name at load. |
| `bows_refresh_polls` | `5` | `bows` only: re-derive every N poll passes; int ≥ 1. |
| `bow_owners` | `[]` | `bows` only, optional override: actor id(s) whose feeds are authoritative. Empty ⇒ resolve the token's own actor at boot (`GET /v1/ping` → `actorId`); a failed whoami is **fatal** (never fall back to trusting everything, nor to trusting nothing). Ids only — a username/display name is refused at load naming the entry. Exact compare + de-dupe, no casefolding. `|`/`,` separated when it arrives as one string. |

Reuse the existing `alissa_endpoint` key (already validated) for the listing. Environment rails, with the same precedence the module already documents for `ALISSA_REV_LOOP_EVENTS_ENABLED` (env > file > flag; a **blank** env value falls through, since an unset platform variable reference renders as `""`): `ALISSA_REVIEW_REPOS_SOURCE`, `ALISSA_REVIEW_BOWS_REFRESH_POLLS`, `ALISSA_REVIEW_BOW_OWNERS`.

Relax the existing guard `on_missing_hub='add' requires a non-empty repos allowlist` (config.py, `Config.build`) so it applies **only under `static`** — under `bows` the feed-authority gate bounds the blast radius instead of the static list, and the container's baked `add` default is what self-hubs each derived repo on first request (`Loop._ensure_hub`, HUB_ADD path).

### 2. Derivation (new `bows.py`, wired from the poll loop)

Every `bows_refresh_polls` passes: list `GET /v1/bodies-of-work?includeShared=true` (extend `alissa_client.AlissaClient` with an authenticated GET and a `ping()`; keep the redirect-refusing, classified-error behaviour it already has), keep containers that are `status == "active"`, whose title is `autodev: <owner>/<repo>` (prefix match case-insensitive; a repo half that does not look like `owner/repo` is **skipped with a WARNING naming it**, never fatal), and whose `ownerActorId` is in the authority set. De-duplicate case-insensitively, **union with `repos`** (a static entry is always watched and never dropped by a refresh, in either mode).

Fail-safe rules (verbatim from devloop, because the allowlist bounds where reviewer sessions spawn):

- **Never shrink on failure.** A refresh that could not list keeps the last successfully derived set and WARNs. A failure on the *first* refresh has no set to keep → start on `repos` alone and say so at ERROR.
- **Never drop mid-round.** A feed that completes, is cancelled or disappears drops its repo on the next successful refresh **unless** a round is in flight for a PR in that repo (a live reviewer session of this daemon's own grammar for it, or an owed round the ledger has not closed) — then it stays watched until that round finishes.
- **Foreign feeds are visible, not authoritative.** A prefix-matching container owned by anyone outside the authority set never enrolls a repo: one aggregate WARNING per refresh with the count, `-v` names them.

### 3. Semantics of the effective allowlist

The derived∪static set is what `review_requests(self.config.repos)` (loop.py, the `repo:` qualifiers on the `review-requested:@me` search), `_name_candidates` (session-name → repo resolution) and `Config.watches()` consume. Two rules:

- Under `static`, empty still means **all** — unchanged.
- Under `bows`, an empty derived∪static set watches **nothing** and logs one WARN per refresh (`bows mode derived 0 repos`) — an empty feed set must not widen to "every PR that requests me". This is deliberately different from `static` and must be documented as such.

Startup log names the mode and the authority (`bow feed authority: self (<actorId>)` / the configured ids). Each refresh logs the derived set; `-v` names the source container behind each repo. If the reviewer console's `/api/state` config block renders `repos`, render `repos_source` + the derived set beside it (read-only, from the daemon's own state — no new calls).

### 4. Container (`docker/claude`)

- New pass-through-when-unset ARGs, rendered through `revloop-config.sh` like the other optional keys: `ALISSA_REVIEW_REPOS_SOURCE`, `ALISSA_REVIEW_BOWS_REFRESH_POLLS`, `ALISSA_REVIEW_BOW_OWNERS` (the last two emitted as int / JSON array exactly as devloop's generator does).
- Entrypoint: under `ALISSA_REVIEW_REPOS_SOURCE=bows`, an empty `ALISSA_REVIEW_REPOS` is the **ordinary** configuration, not the `die` — generate the manifest from the static seed (possibly zero repos) + the config, and let `on_missing_hub=add` materialize hubs on demand. `ALISSA_REVIEW_REPOS` stays required under `static` (that guard predates this and stays). Executor role unchanged.
- **Version-skew guard**, copied from devloop's entrypoint: the mode is honoured only when the *installed* `alissa-tools-github-revloop` understands `repos_source`; otherwise WARN by name ("bows mode landed in revloop 0.29.0 — re-pin ARG REVLOOP_VERSION") and fall through to the static path. Do **not** bump `ARG REVLOOP_VERSION` in this PR — `check-image.yaml` installs the pinned version from PyPI, which cannot yet carry this release; the pin bump is the ordinary post-publish follow-up, and the skew guard is what makes the interim boot non-fatal and non-silent.
- Never overwrite an operator-mounted `revloop.config.json` on the bows path (devloop's provenance rule: refresh only our own generated output, treat anything not positively ours as the operator's).

### 5. Docs

`README.md` (config table rows + a *Deriving the allowlist from feed Bodies of Work (`repos_source: bows`)* section covering the authority gate, the two fail-safe rules, the empty-set semantics difference, and the interplay with `on_missing_hub`), `revloop.config.example.json` (the three keys with `_`-comments), `docker/claude/README.md` (ARG rows, the Railway "enroll by creating the lane" note, and the `ALISSA_REVIEW_REPOS` row now reading *required under `static`*).

### 6. Tests

- `test_bows.py` (new): derivation fixture with 2 own active feeds + 1 foreign-owned + 1 cancelled + 1 malformed title → exactly the 2 own repos, foreign counted in the WARNING, malformed named; static seed unioned; case-insensitive dedupe; refresh failure keeps the prior set (and first-failure → static only at ERROR); feed disappearance with an in-flight round retains the repo, without one drops it.
- `test_loop.py`: `review_requests` receives the derived∪static set; under `bows` an empty set searches nothing (no `search/issues` call) and WARNs once per refresh; under `static` the empty=all path is unchanged.
- config tests: env-rail precedence + blank fall-through for the three variables; `on_missing_hub='add'` guard applies under `static` only; non-id `bow_owners` refused by name.
- `docker/claude/tests-entrypoint-config.sh`: bows + empty `ALISSA_REVIEW_REPOS` boots and renders the three keys; static + empty still dies; skew guard falls through with the named WARN.

### 7. Version

Bump `version.py` 0.28.0 → 0.29.0 (`check-version-bump.yaml` requires it for any `src/**` change).

## Acceptance

- Under `repos_source=static` every existing test passes unchanged and the generated config is byte-identical for an unchanged env.
- Under `repos_source=bows` a self-run fleet with two Factory-minted lanes and no `ALISSA_REVIEW_REPOS` boots, derives both repos within `bows_refresh_polls` polls, hub-ifies each on its first review request, and ignores a third feed shared by another actor (warned, not enrolled).
- Removing a lane's feed while a round is in flight keeps the repo until that round closes; with no round in flight it drops on the next refresh.
- An Alissa outage during a refresh never shrinks the allowlist.
- The container boots non-fatally, with the named WARN, on an image whose pinned library predates the key.

## Definition of Done

PR on fahera-mx/alissa-github-review-daemon merged: `repos_source` / `bows_refresh_polls` / `bow_owners` config keys + `ALISSA_REVIEW_REPOS_SOURCE` / `ALISSA_REVIEW_BOWS_REFRESH_POLLS` / `ALISSA_REVIEW_BOW_OWNERS` env rails; feed derivation with the authority gate, never-shrink and never-drop-mid-round rules; empty-set-watches-nothing under bows; container ARGs + entrypoint bows path (empty ALISSA_REVIEW_REPOS legal under bows, skew guard, no operator-config overwrite); README + example config + docker README updated; tests listed above green; version 0.28.0 → 0.29.0. The `ARG REVLOOP_VERSION` pin bump is explicitly NOT in scope (post-publish follow-up).

## Acceptance criteria

- [ ] Existing test suite passes unchanged under repos_source=static; generated container config byte-identical for an unchanged env
- [ ] test_bows.py: 2 own active feeds + foreign + cancelled + malformed → exactly the 2 own repos derived, foreign warned in aggregate, malformed named; static seed unioned; case-insensitive dedupe
- [ ] Refresh failure keeps the prior set (first failure → static only at ERROR); a disappeared feed with an in-flight round retains its repo, without one drops it
- [ ] Under bows an empty derived∪static set makes no search/issues call and WARNs once per refresh; under static empty=all is unchanged
- [ ] tests-entrypoint-config.sh: bows + empty ALISSA_REVIEW_REPOS boots and renders the three keys; static + empty still dies; skew guard on an old pin falls through with the named WARN
- [ ] README config table + bows section, revloop.config.example.json, docker/claude/README.md rows updated; version.py bumped to 0.29.0; ARG REVLOOP_VERSION untouched

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/bows.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa_client.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/__main__.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_bows.py, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_config.py, docker/claude/Dockerfile, docker/claude/entrypoint.sh, docker/claude/revloop-config.sh, docker/claude/README.md, docker/claude/tests-entrypoint-config.sh, README.md, revloop.config.example.json
Alissa-Task: TASK-1884455456

RHDZMOTA:
Closes #119

Alissa-Task: TASK-1884455456 (origin)
Alissa-Task: TASK-1317001982 (implementation, downstream of the origin)

## What

`repos_source: bows` for revloop — parity with devloop 0.8.9 / orcloop 0.13.0. Under `bows` the review allowlist is the static `repos` seed unioned with the repos named by the operator's active `autodev: <owner>/<repo>` feed Bodies of Work, re-derived every `bows_refresh_polls` passes and gated on the container's owner (`bow_owners`, defaulting to the token's own actor resolved at boot via `GET /v1/ping`; a failed whoami is fatal). Both devloop fail-safe rules are kept verbatim: never shrink on a failed listing, never drop a repo with a round in flight. Under `bows` an empty derived∪static set watches nothing (deliberately unlike `static`). `static` is unchanged bit for bit.

- **Library**: `config.py` (three keys + `ALISSA_REVIEW_REPOS_SOURCE` / `ALISSA_REVIEW_BOWS_REFRESH_POLLS` / `ALISSA_REVIEW_BOW_OWNERS` rails, blank falls through; actor-id validation; `on_missing_hub='add'` guard narrowed to `static`), `alissa_client.py` (authenticated GET, `ping()`, `list_bodies_of_work()`), new `bows.py`, `loop.py` (refresh wired before the search; the effective set is rebound onto `config.repos` so `review_requests`, `watches`, `_name_candidates` and `_ensure_hub` all read it; empty-under-bows skips the search with one WARN per refresh), `__main__.py` (`--repos-source` / `--bows-refresh-polls` / `--bow-owner`, feed-authority resolution, exit 2 on a failed whoami), `state.py` (a `derived_repos` table for the console, `in_flight_repos` for the never-drop rule), `webui/sources.py` (`repos_source` + the derived set beside `repos`).
- **Container**: three pass-through ARGs; `revloop-config.sh` renders them (string / int / JSON array) gated on a version-skew probe of the installed library; the entrypoint boots on an empty `ALISSA_REVIEW_REPOS` under `bows`, never overwrites an operator-mounted config (provenance stamp on the bows-path render only), and falls through to the static path with the named WARN on an old pin. `ARG REVLOOP_VERSION` untouched, as the issue requires.
- **Docs**: README config rows + the *Deriving the allowlist from feed Bodies of Work* section, `revloop.config.example.json`, docker README rows + *Enrolling by creating the lane*.
- **Version**: 0.28.0 → 0.29.0.

### Judgment calls

- **Static `watches()` stays exact-match; only `bows` compares case-insensitively.** The issue says `static` must be bit-identical, and today's `watches` is an exact `in`. A derived name is operator-typed feed text while GitHub's `repository_url` is canonical, so `bows` needs casefolding or a title differing by case would silently skip its repo. Two rules in one method, documented in the README as the one place the modes differ besides empty-set semantics.
- **"Round in flight" is answered from three signals**: a live session of this daemon's own grammar (`review-<repo>-pr<n>-…`, or a skill-shaped `review-pr-<n>` attributed through its spawn-ledger row), a spawn younger than the stale-round window whose session the sweep has not reaped, and an owed native verdict (neither posted nor abandoned). An unlistable session roster makes the whole answer "cannot tell" and drops nothing that refresh. The issue names the first and the last; the middle one covers the enqueue-to-tmux window.
- **The bows-path config render is stamped (`_generated_by`); the static render is not.** The acceptance criterion requires the static render to be byte-identical, so the provenance stamp lives only on the bows path. Consequence: a config an earlier *static* boot generated is unstamped and the bows path treats it as the operator's (respected, logged by name). This is the fail-safe direction, and it costs little because the daemon reads all three keys from the environment, which outranks the file.
- **Skew guard on an old pin with `ALISSA_REVIEW_REPOS` empty still dies**, after the named WARN — devloop behaves identically ("Falling through … this boot FAILS below"). The guard makes the boot non-fatal wherever a static allowlist exists, and never silent.
- **The refresh runs after the reap sweep and before the search**, so a session reaped this pass no longer pins its repo, and a repo enrolled this pass is searched this pass. The first refresh runs in `preflight`, so the startup log names the derived set.
- **`--bow-owner` takes a single value that may itself be `|`/`,`-separated**, and `bow_owners` in the file accepts a list or one string, so the env, the flag and the file are validated by one rule.
- **Console**: the derived set is persisted to a small `derived_repos` ledger table on each successful refresh (telemetry-class write) and `/api/state` reads it beside `repos`; no Alissa call from the console.
- **Live observation, not a code decision**: against the real API this token's `actorId` (`j5706…`) is not the owner of the fleet's five `autodev:` feeds (`j577e…`). With self-authority the reviewer would derive 0 repos and WARN; the operator sets `ALISSA_REVIEW_BOW_OWNERS` to the lane-creating actor for this fleet. Documented in the docker README.

- [ ] No `docker build` and no Railway boot in this runner (no docker binary). The entrypoint's bows arm, the static die and the skew fall-through were exercised by booting the real `entrypoint.sh` against stubbed `gh` / `alissa` / `alissa-revloop` in `tests-entrypoint-config.sh`; the assembled image is only proven by `check-image.yaml`.
- [ ] The acceptance scenario "two Factory-minted lanes, no `ALISSA_REVIEW_REPOS`, both derived within `bows_refresh_polls` polls, hub-ified on the first request, a third feed shared by another actor ignored" was not run against a live deployment. Derivation was run once against the live listing here (5 own-owner feeds → 5 repos; with the token's own actor as authority → 0, foreign WARN); hub-ifying on the first request is covered by a unit test, not a live boot.
- [ ] `check-entrypoint.yaml` installs the pinned 0.16.14 for its cross-check; the new stub-boot cases pin `PYTHONPATH` explicitly (repo source tree vs an old-library stub), so they do not depend on what CI installs. The config step now boots the entrypoint six times; I did not add a `timeout-minutes` to that step (the workflow file is outside the declared scope).
- [ ] The `ARG REVLOOP_VERSION` re-pin to 0.29.0 is the post-publish follow-up the issue excludes; until it lands, a container with `ALISSA_REVIEW_REPOS_SOURCE=bows` takes the skew fall-through.

### Verification

- `PYTHONPATH=src/main bash tests-unit.sh alissa-tools-github-revloop` → 1056 passed (985 on the untouched base + 71 new).
- `bash tests-coverage.sh` → 95% total; `bows.py` 100%, `alissa_client.py` 98%, `config.py` 99%.
- `bash check-style.sh` → clean. `bash check-types.sh` → `Success: no issues found in 36 source files`.
- `bash docker/claude/tests-entrypoint-config.sh` → ALL PASS (72 assertions: renderer types, blank-equals-unset byte identity, skew drop with the named WARN, fail-open probe; stub boots: bows + empty repos boots and renders the three keys, our stamped config refreshed, an unstamped config respected, static + empty dies before the worker, old pin falls through with the WARN, old pin + empty dies after the WARN).
- The other six entrypoint harnesses (`ui`, `identity`, `auth`, `executor`, `chown`, `prune`) → all exit 0.
- Mutation pass: 16 hand-written mutants over the guards (bows-empty widening, `add` guard both directions, casefolded owner gate, `includeShared` dropped, bad payload degraded to `[]`, never-shrink, never-drop in both directions, status gate, empty-set search, WARN cadence, roster failure as idle, refresh not applied, whoami softened, blank env not falling through) — every one killed by the suite.
- Live: `AlissaClient.ping()` and `list_bodies_of_work()` run once against `https://api.alissa.app` with this session's token (24 containers, 5 feed-shaped).

### Scope touched

- Library (all in the declared scope): `config.py`, `alissa_client.py`, `bows.py` (new), `loop.py`, `__main__.py`, `webui/sources.py`, `version` + `version.py` (the `version` file is what `setup.py` and `check-version-bump.yaml` read; `version.py` is unchanged).
- **Excursion**: `state.py` — not in the declared scope. Two additions: the `derived_repos` table the console reads (the issue asks for the derived set in `/api/state` "from the daemon's own state — no new calls", which needs a persisted set), and `in_flight_repos()` for the never-drop rule's ledger half.
- Tests: `test_bows.py` (new), `test_config.py` (new — the repo kept config tests in `test_loop.py`, but the issue lists this file), `test_loop.py` (appended), `docker/claude/tests-entrypoint-config.sh`.
- Container + docs: `docker/claude/Dockerfile`, `entrypoint.sh`, `revloop-config.sh`, `docker/claude/README.md`, `README.md`, `revloop.config.example.json`.
- Not touched: `ARG REVLOOP_VERSION`, `.github/workflows/*`.

### Round-1 fixes (review round 1 → fix attempt 1; commits `3767be1`, `77e0717`)

Five findings, all `[triage:pursue]`; the two entrypoint `[minor]`s as a narrow die-by-name and a docs-only change.

#### Judgment calls

- **[major] one-to-many slugs**: a name-borne session now pins *every* watched repo whose slug it matches (`orgA/web` and `orgB/web` both, for `review-web-pr12-…`). Holding both is the fail-safe direction the rule already argues for; the bare skill shape is still attributed exactly through its ledger row. The new test asserts the retained pair as a set: the retained repo is appended after the derived order, and the order is not the guarantee.
- **[minor] downgrade**: I did not take the suggested re-render or removal. Rendering the stamped file through the probe hands the old library `repos: []` under `static`, which dies only because the `on_missing_hub='add'` guard catches it; with `ALISSA_ON_MISSING_HUB=skip` an empty static allowlist is *every PR that requests this reviewer*, so a rollback would silently widen the watch. The guard now dies by name before any worker starts, gated on all three of: daemon role, empty seed, config positively ours (stamped). The stamped file is left intact so the re-pin's next bows boot refreshes it; an unstamped or mounted config keeps the mounted-mode behaviour (the operator's file, respected as before).
- **[minor] bows with a static seed overwrites**: deliberately unchanged behaviour. A non-empty seed takes branch 1, which is the pre-#119 "`ALISSA_REVIEW_REPOS` is authoritative, regenerate every boot" contract the issue requires bit-identical, and the environment outranks the file for the three keys anyway. Documented in the branch-1 comment and in three places in the docker README (Config bullet, skew bullet, the "authoritative" paragraph).
- **[nit] one roster per pass**: the sweep leaves its post-reap roster on the watcher (cleared with the other per-pass memos) and the refresh reads it. Subtracted are the *kills* only: a dry-run sweep's `would_reap` names and a kill that raised are still live. The listing fallback stays for the preflight refresh (no sweep yet) and for a sweep whose list failed, where the refresh's own listing fails the same way and drops nothing, as before.
- `record_derived_repos` materialises its rows to a list before wrapping, because `_write_telemetry` calls the writer twice on its reconnect retry and a one-shot iterable would persist an empty set on the second call. `sources()` already returns a list, so this is a guard, not a fix.

- [ ] Still no `docker build` and no Railway boot in this runner. The downgrade die is proven by stub-booting the real `entrypoint.sh` with an old-library stub on `PYTHONPATH`, not by a real pre-0.29 image over a volume a 0.29.0 boot wrote.
- [ ] The widening claim behind the die (empty static allowlist plus `on_missing_hub=skip` under the old library watches every requesting PR) is read from the static path's documented semantics ("under `static`, empty = all"), not exercised against an installed old wheel.
- [ ] The one-listing-per-pass saving is measured against the fake roster, not a real `alissa tmux ls`.
- [ ] Carried from the first note: no live two-lane acceptance run; the `ARG REVLOOP_VERSION` re-pin remains the excluded follow-up, so a deployed container still takes the skew fall-through until it lands.

#### Verification

- `PYTHONPATH=<branch>/alissa-tools-github-revloop/src/main bash tests-unit.sh alissa-tools-github-revloop` → 1060 passed (1056 + 4 new). Note for the validator: the same command with a *relative* `PYTHONPATH` tests the installed 0.28.0 wheel (the script changes directory) and the 3 new loop/state tests fail there, as they should.
- `bash tests-coverage.sh alissa-tools-github-revloop` → 95% total; `bows.py` 100%, `loop.py` 97%, `state.py` 96%.
- `bash check-style.sh alissa-tools-github-revloop` → clean. `bash check-types.sh alissa-tools-github-revloop` → `Success: no issues found in 36 source files`. `bash -n` on both shell files → clean. No `shellcheck` binary in this runner.
- `bash docker/claude/tests-entrypoint-config.sh` → ALL PASS, 127 assertions, 12 new (3c: a stamped bows config plus an empty manifest on the volume, old pin, empty seed → exits non-zero before the worker, names the stamped file and the widening reason, leaves the file intact; 3d: the same boot over an unstamped config comes up through the mounted-mode arm untouched).
- The other six entrypoint harnesses (`ui`, `identity`, `auth`, `executor`, `chown`, `prune`) → all exit 0.
- Mutants, each killed by exactly the test written for it: one-to-one slug map restored (`max` of the set), refresh always re-listing, roster not minus the kills, `record_derived_repos` without the absorb, and the entrypoint die removed (the harness then fails the five downgrade assertions and its log shows the boot reaching the worker, the crash-loop the finding describes).

#### Scope touched

- Library: `loop.py` (`_mid_round_repos`, `refresh_repos`, `sweep_sessions`, the per-pass memo), `state.py` (`record_derived_repos` wraps, `_replace_derived_repos` raw).
- Tests: `test_loop.py` (+3), `test_state.py` (+1).
- Container + docs: `docker/claude/entrypoint.sh` (guard die + branch-1 comment), `docker/claude/tests-entrypoint-config.sh` (+2 boots), `docker/claude/README.md`.
- No version change (0.29.0 is unreleased on this branch). No excursion beyond the first note's set; `state.py` was already the declared one. `ARG REVLOOP_VERSION` and `.github/workflows/*` still untouched.

---
<a href="https://app.blacksmith.sh/fahera-mx/codesmith/alissa-github-review-daemon/pr/120"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791263006&installation_model_id=434695&pr_number=120&repository=fahera-mx%2Falissa-github-review-daemon&return_to=https%3A%2F%2Fgithub.com%2Ffahera-mx%2Falissa-github-review-daemon%2Fpull%2F120&signature=90eff746ca8785e2d2cbb5c8b79e7414ae0ffeba06e77830905dd517d73c1249"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


alissa-app:
- 2026-09-06 05:35:47 UTC — `review-alissa-github-review-daemon-pr120-r1-b67c2a` — round 1 of 22 — spawned
- 2026-09-06 06:41:35 UTC — `review-alissa-github-review-daemon-pr120-r2-627e4b` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1696,6 +1713,122 @@ def __init__(
...
+            sessions = self.alissa.list_review_sessions()
+        except CommandError as exc:
+            log.warning(
+                "repos_source=bows: could not list reviewer sessions (%s) — "
+                "no repo leaves the allowlist this refresh",
+                exc,
+            )
+            return None
+        busy = set(self.state.in_flight_repos(STALE_ROUND_SECONDS))
+        slugs = {
```
alissa-app: [major] `slugs` collapses two watched repos that share a repo NAME, so the never-drop-mid-round rule can drop the wrong one.

`session_repo_slug` is built from the repo half alone (`full_name.partition("/")[2]`, and it lowercases + folds punctuation), so `orgA/web` and `orgB/web` both key on `web`. A dict comprehension keeps only the **last** entry, so a live session `review-web-pr12-r1-<nonce>` marks whichever of the two came last in `config.repos` as busy — and the other one is *not* in `busy`.

Concretely: watch `orgA/web` and `orgB/web` (both derivable under bows — lanes are per `owner/repo`, and nothing forbids the same repo name under two owners). A round is running on `orgA/web`, older than `STALE_ROUND_SECONDS` so the ledger half of `in_flight_repos` no longer covers it and no verdict is owed yet — the roster is the only signal left. `orgA/web`'s feed is completed. Next refresh: `_retained` asks `mid_round`, gets a `busy` containing `orgb/web` instead of `orga/web`, so `orgA/web` is dropped. `poll_once`'s `watches` gate then skips its PR, `_name_candidates` returns `[]` for the running session, and the round is stranded — exactly the failure this rule exists to prevent. (The mirror-image error is also live: `orgB/web` is held for a round it does not have.)

`_name_candidates` (line 4259) already treats this as one-slug-to-*many*-repos and returns a list. This should do the same — e.g. build `dict[str, set[str]]` and `busy |= slugs.get(ref.repo, set())` — so a name-borne session marks every watched repo whose slug it matches. That is the fail-safe direction the rest of this method already argues for.

No test covers the collision; `test_a_live_daemon_session_keeps_a_disappeared_repo` uses a single repo.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1696,6 +1713,122 @@ def __init__(
...
+            log.exception(
+                "repos_source=bows: refresh failed unexpectedly — keeping the "
+                "current allowlist"
+            )
+            return
+        self._apply_repos(source.repos())
+        if listed:
+            # Telemetry-class: the console's window onto the derived set.
+            rows = source.sources()
+            self.state._write_telemetry(
```
alissa-app: [minor] Reaching into `State._write_telemetry` from the loop is the only `self.state._*` call in `loop.py`.

Every other telemetry-class write in this codebase wraps itself: `State.record_review_task` (state.py:1235) calls `_write_telemetry(lambda: self._replace_review_task(...), ...)` *inside* State, and the loop calls the public method. Here the loop composes the private helper with a public writer, which puts the absorb-or-not decision on the caller and makes `record_derived_repos` the one telemetry writer whose failure handling lives outside `State`.

Suggest keeping the pattern: have `record_derived_repos` do its own `_write_telemetry` wrapping (renaming the raw writer `_replace_derived_repos`, as `_replace_review_task` already is), and let `refresh_repos` just call it.

docker/claude/entrypoint.sh
```diff
@@ -735,6 +735,33 @@ skills_lines() {
...
+# whenever ALISSA_REVIEW_REPOS is set and dies with its own named reason
+# otherwise. Asked once, through the renderer's own probe (idempotent), and
+# only when the mode is requested — a static deployment pays nothing.
+BOWS_MODE=0
+if [ "${REPOS_SOURCE}" = "bows" ]; then
+  if revloop_dist_supports repos_source; then
+    BOWS_MODE=1
+  else
+    log "WARN: ALISSA_REVIEW_REPOS_SOURCE=bows, but the INSTALLED alissa-tools-github-revloop $(revloop_dist_version) does NOT support the 'repos_source' config key — NOT taking the bows path. bows mode landed in revloop 0.29.0 — re-pin ARG REVLOOP_VERSION."
+    log "      Falling through to the static path: with ALISSA_REVIEW_REPOS set this boot continues on the static allowlist; without it (and without a mounted alissa-workspace.yaml) it FAILS below with the static path's own reason."
```
alissa-app: [minor] The skew fall-through does not hold on a **downgrade** — a config an earlier 0.29.0 boot wrote survives on the volume and the old library refuses it.

This line promises the fall-through is non-fatal with `ALISSA_REVIEW_REPOS` set and otherwise "FAILS below with the static path's own reason". Both hold on the forward path (new entrypoint, old pin, nothing written yet). They do not hold after a rollback:

1. Boot with `REVLOOP_VERSION >= 0.29.0`, `ALISSA_REVIEW_REPOS_SOURCE=bows`, `ALISSA_REVIEW_REPOS` empty → the bows arm writes `${MANIFEST}` (empty repo list) and a stamped `${CONFIG}` carrying `repos_source` / `bows_refresh_polls` / `bow_owners`. Both persist on `/workspace`.
2. Operator re-pins back to an older `REVLOOP_VERSION` (a rollback, or a rebuild that lands before the 0.29.0 re-pin). `revloop_dist_supports repos_source` is now false → `BOWS_MODE=0`.
3. `repos_lines` is empty and `BOWS_MODE != 1`, so control falls to the **mounted-mode** `else` (line ~862). `${MANIFEST}` exists, so it does not die and it does not re-render `${CONFIG}` — the `_skew_value` filter in `render_revloop_config` never runs on this path.
4. The daemon then loads the stale file and exits 2 with `unknown config key(s): bow_owners, bows_refresh_polls, repos_source`, in a crash loop, under a WARN that said the boot would fail for a different, named reason.

The static-seed variant self-heals (branch 1 re-renders and `_skew_value` drops the three keys), so only the empty-seed rollback is stuck. Cheapest fix that keeps the promise this line makes: when the guard fires, if `config_is_generated "${CONFIG}"` then re-render it through `render_revloop_config` (the keys are dropped by the probe) or remove it — it is positively ours, so rewriting it is already sanctioned by the provenance rule. Failing that, name the condition in the WARN.

docker/claude/entrypoint.sh
```diff
@@ -783,10 +810,60 @@ if [ -n "$(repos_lines)" ]; then
...
+  # at each refresh from the operator's `autodev:` Bodies of Work, so an empty
+  # static list is the whole point — enrolling a repo stops being a config
+  # edit. The library relaxed its matching guard (`on_missing_hub='add'`
+  # requires a non-empty allowlist only under `static`), and the container's
+  # baked `add` default is what then self-hubs each derived repo on its first
+  # review request.
+  #
+  # The manifest is written with an EMPTY repo list when absent (the hubs
+  # materialise on demand) and a mounted one is respected as-is. The config is
+  # CREATE-OR-REFRESH-OUR-OWN, never overwrite the operator's: only a file
```
alissa-app, line 851: [minor] "Never overwrite an operator-mounted `revloop.config.json`" holds only on the *empty-seed* bows path; bows **with** a static seed still overwrites unconditionally.

Issue #119 states the provenance rule for the bows path without qualifying it by whether `ALISSA_REVIEW_REPOS` is set. With `ALISSA_REVIEW_REPOS_SOURCE=bows` **and** a non-empty `ALISSA_REVIEW_REPOS`, `repos_lines` is non-empty so branch 1 wins and does `render_revloop_config ... > "${CONFIG}"` with no `config_is_generated` check and no stamp — a mounted config carrying a hand-written `bow_owners` is destroyed on the next boot, which is precisely the loss this comment says must not happen.

That branch's unconditional overwrite predates #119 and is deliberate for static, so this may be a considered scope call rather than a miss — but if so it belongs in the docs, because `docker/claude/README.md`'s "Enrolling by creating the lane" describes the create-or-refresh-our-own rule without noting that setting `ALISSA_REVIEW_REPOS` opts out of it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -1696,6 +1713,122 @@ def __init__(
...
+        created, and for a verdict still owed. The live ROSTER answers for a
+        session that outlived the stale window -- a healthy round runs for
+        a while, and dropping its repo at the window's edge would strand the
+        exact round the rule protects. An unlistable roster makes the whole
+        answer None: handing over the ledger half alone would read a live
+        long-running session as absent, which is the direction the rule
+        must never err in.
+        """
+        try:
+            sessions = self.alissa.list_review_sessions()
```
alissa-app: [nit] Second session listing per refresh. `poll_once` runs `sweep_sessions()` immediately before `refresh_repos()`, and the sweep already lists the roster (it seeds `_session_census`). When a refresh has drop candidates this shells out to `alissa` a second time in the same pass. Bounded — only on refreshes that actually have candidates — but the sweep's roster is one pass old at most and could be threaded through.

alissa-app (changes requested):

Reviewed head **f1d6fee77992492bf9f5481c415bdb21a410f56f**. CI rollup at that sha: 8 contexts, all concluded — 7 success, 1 skipped (`[code]smith`); no legacy commit statuses. Read through `commits/<sha>/check-runs`, not "the PR's checks".

This is a careful, well-argued implementation. The authority gate is the right security shape and its reasoning is written down where it will be read; the two devloop fail-safe rules are both present with the failure directions chosen deliberately; `static` really is untouched (the `watches` branch, the `on_missing_hub` relaxation and the renderer's three keys are all conditioned on the mode); the version-skew probe is a good answer to the pre-re-pin window; and the test set covers every case issue #119 lists, including the ones that are easy to skip (first-failure-at-ERROR, malformed-payload-is-a-failure, unanswerable probe drops nothing). Docs are unusually complete.

One `[major]` stands between this and approve.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 |
| minor | 3 |
| nit | 1 |
| question | 0 |

- **[major]** `loop.py:1757` — `_mid_round_repos` keys `slugs` by repo *name*, so two watched repos sharing a name collapse and the never-drop-mid-round rule can drop the repo that has the live round (and hold the one that does not). `_name_candidates` already models this as one-slug-to-many.
- **[minor]** `loop.py:1803` — `self.state._write_telemetry(...)` is the only reach into `State`'s private API from the loop; every other telemetry write wraps itself inside `State`.
- **[minor]** `docker/claude/entrypoint.sh:761` — the skew fall-through's promise breaks on a **downgrade**: a stamped bows config written by a 0.29.0 boot survives on the volume, mounted mode does not re-render it, and the old library crash-loops on `unknown config key(s)`.
- **[minor]** `docker/claude/entrypoint.sh:827` — the never-overwrite-the-operator's-config rule applies only to the empty-seed bows path; `bows` **with** a static seed still overwrites unconditionally, undocumented.
- **[nit]** `loop.py:1748` — second roster listing in a pass that already swept.

### Dimension report

- **Correctness** — the `[major]` above. Otherwise the DoD maps cleanly: three keys + env rails with blank-falls-through (`config.py:942-975`), derivation with the status/title/shape gates (`bows.py`), both fail-safe rules with the drop half skipped on absent evidence, empty-under-bows searching nothing with one WARN per *refresh* (`loop.py:1806-1830`), container ARGs + entrypoint bows arm + skew guard, `version` 0.29.0, `ARG REVLOOP_VERSION` untouched. Two nice details worth naming: `_retained` computes candidates before consulting the probe, so the steady state costs no subprocess; and `BowRepoSource` deliberately keeps the *original* `Config` while the loop rebinds its own, which is what stops the static seed growing monotonically as derived repos are unioned back in. The `version.py`-vs-`version` divergence from the issue text is correct — `setup.py` and `check-version-bump.yaml` both read the plain `version` file.
- **Coding practices** — the `_write_telemetry` `[minor]`. Otherwise idiomatic for this repo: `dataclasses.replace` onto the frozen `Config` re-deriving `_repo_match` through `__post_init__`, `_`-prefixed provenance key riding the existing comment-key convention so the stamp is inert to every daemon version, one validation rule (`normalize_bow_owners`) shared by env/file/flag.
- **Security** — nothing found, and this is the dimension the PR is strongest on. `includeShared=true` is correctly identified as widening the listing to anything any tenant actor unilaterally shares, and the owner gate (not the title) is the boundary; `trusts_feed_owner` fails closed on both a missing `ownerActorId` and an empty authority, so a skipped boot resolution makes the mode inert rather than open; `_validate_actor_id` refusing names is the right call for a renameable-identifier spoof surface, and `resolve_feed_authority` dying on a failed whoami rather than softening to an empty list is the correct fail-*loud* direction. Feed text cannot escape `_FEED_REPO`, so nothing operator-typed reaches the search query or a hub path unbounded — I checked the `repo:` qualifier path against the live search API (a nonexistent repo qualifier does not 422 the query, and 60 qualifiers / 2.4 KB is accepted), so an unbounded derived set does not poison the search. The GET shares the POST's redirect-refusing opener, so the bearer token cannot follow a 30x cross-host.
- **Performance** — the `[nit]`. The cadence is right (`tick()` before `due()` makes `bows_refresh_polls: 1` mean every pass; a *failed* attempt resets the counter so a down API is not re-listed per pass), the skew probe is lazy and memoised, and `_skew_value` returns before probing when the variable is unset, so a static deployment pays nothing.
- **Side-effects / blast radius** — `static` is unchanged on every edge I could find, and the static render stays byte-identical for an unchanged env (the stamp is bows-only, and the three keys drop out empty). `derived_repos` is additive via `CREATE TABLE IF NOT EXISTS`, and the console degrades to `[]` through `_read_state` on a pre-0.29 db rather than erroring. The declared `state.py` excursion is justified — `/api/state` "from the daemon's own state, no new calls" needs a persisted set. The only cross-version hazard is the downgrade `[minor]`.

Fix the `[major]` (make `slugs` one-to-many and add a same-name/two-owner case to `test_a_live_daemon_session_keeps_a_disappeared_repo`), and triage the three `[minor]`s per CR8 — the two entrypoint ones are legitimate `[triage:later]` or `[triage:ignore]` candidates with a written reason; they do not need to be fixed here to reach approve.

---
Verdict of record: `request_changes` (round 1 of cap 22) — Alissa review task TASK-950011235.

https://claude.ai/code/session_019tUMFitbSTZDwNpJL1yFxw

RHDZMOTA:
- 2026-09-06 05:47:47 UTC — `fix-fahera-mx-alissa-github-review-daemon-pr120-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Agreed. `slugs` is a one-to-one dict where `_name_candidates` already answers the same question one-to-many, and the last-writer-wins collapse marks the wrong repo busy in exactly the case the rule exists for: a long-lived round on one of two same-name repos, past the stale window, no verdict owed, the roster the only signal left. Fix: `slugs` becomes `dict[str, set[str]]` and a name-borne session does `busy |= slugs.get(ref.repo, set())`, so it marks every watched repo whose slug it matches (holding both is the fail-safe direction; the bare skill shape is still attributed exactly through its ledger row). Test: a two-owner same-name case (`orgA/web` + `orgB/web`, a live `review-web-pr12-…` session, the `orgA/web` feed completed) asserting both are retained, and that only the session's departure drops `orgA/web`. Fixed in 3767be1.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] Agreed. The absorb-or-not decision belongs inside `State`, as it does for every other telemetry writer. `record_derived_repos` now wraps `_write_telemetry(lambda: self._replace_derived_repos(rows), "recording the derived allowlist")` itself and returns the bool, and `refresh_repos` calls the public method only. No `self.state._*` reach remains in `loop.py`. Fixed in 3767be1.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] Agreed that the promise breaks on the empty-seed rollback, and that the crash loop's reason (`unknown config key(s)`) is not the one the WARN named. I am NOT taking the re-render option, though: rendering the stamped file through the probe leaves the old library `{"repos": [], "on_missing_hub": "add"}`, which dies by name only because of the `add` guard. With `ALISSA_ON_MISSING_HUB=skip` set, an empty static allowlist under the old library means EVERY PR that requests this reviewer, i.e. a downgrade would silently *widen* the allowlist. Removing the file has the same shape (the daemon's defaults). So the fall-through now dies by name in the entrypoint when the guard fires with no static seed and `config_is_generated "${CONFIG}"` is true (daemon role only): the message names the stamped file, the downgrade, and the two exits (re-pin `>= 0.29.0`, or set `ALISSA_REVIEW_REPOS`, which takes branch 1 and self-heals as you noted). An unstamped or mounted config keeps the mounted-mode behaviour. Harness: a new stub-boot case seeds a stamped config plus an empty manifest and boots on the old stub, asserting a non-zero exit before the worker and the named reason; a counter-case checks the same boot over an unstamped config still comes up. The docker README's skew bullet names the condition. Fixed in 77e0717.

docker/claude/entrypoint.sh
RHDZMOTA: [triage:pursue] As a docs fix, not a behaviour change. The unconditional overwrite in branch 1 is the pre-#119 contract ("`ALISSA_REVIEW_REPOS` is authoritative, regenerate on every boot"), and the issue requires the static path byte-identical, so with a static seed set `bows` rides that contract and the provenance rule applies only to the empty-seed path. That is a considered call (on Railway the environment is the durable place for `bow_owners`, and it outranks the file), but you are right that it is undocumented. The docker README's *Enrolling by creating the lane* Config bullet and the *When `ALISSA_REVIEW_REPOS` is set it is authoritative* paragraph now say that a non-empty seed opts back into the regenerate-every-boot rule (unstamped, no provenance check), so hand-written keys belong in the environment; the entrypoint's branch-1 comment says the same. Fixed in 77e0717.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
RHDZMOTA: [triage:pursue] The sweep's roster is exactly the list the refresh wants, minus what the sweep killed. `sweep_sessions` now leaves its post-reap roster on the watcher for the rest of the pass (a per-pass memo cleared with the others), and `_mid_round_repos` reads it before shelling out. The `alissa tmux ls` fallback stays for the preflight refresh (no sweep has run yet) and for a sweep that could not list (the roster is then unknown, and the refresh's own failed listing drops nothing, as before). Test: a refresh with drop candidates inside a poll pass costs exactly one session listing. Fixed in 3767be1.

RHDZMOTA:
**Fix loop stalled?** — fix attempt 1 on review round 1 has been in flight 40 min. The author's own triage replies postdate the review being answered, so the daemon keeps deferring the retry — but review records never expire, so a fix session that died AFTER posting its replies looks alive forever: is that session alive? The daemon will keep deferring (it never respawns over author activity). Operator options: check (and if dead, kill) the fix session and finish the round by hand (fixes, re-request review), or close the PR to park the work.

docker/claude/entrypoint.sh
```diff
@@ -735,6 +735,49 @@ skills_lines() {
...
+# this boot stops here, by name, before any worker starts. With the seed set,
+# branch 1 below re-renders unconditionally and the rollback self-heals.
+BOWS_MODE=0
+if [ "${REPOS_SOURCE}" = "bows" ]; then
+  if revloop_dist_supports repos_source; then
+    BOWS_MODE=1
+  else
+    log "WARN: ALISSA_REVIEW_REPOS_SOURCE=bows, but the INSTALLED alissa-tools-github-revloop $(revloop_dist_version) does NOT support the 'repos_source' config key — NOT taking the bows path. bows mode landed in revloop 0.29.0 — re-pin ARG REVLOOP_VERSION."
+    log "      Falling through to the static path: with ALISSA_REVIEW_REPOS set this boot continues on the static allowlist; without it (and without a mounted alissa-workspace.yaml) it FAILS below with the static path's own reason — or right here, by name, when the config on the volume is this container's own earlier bows-path output (a downgrade)."
+    if [ "${CONTAINER_ROLE}" != "executor" ] && [ -z "$(repos_lines)" ] && config_is_generated "${CONFIG}"; then
```
alissa-app, line 775: [minor] The new die closes the **version-rollback** route into the empty-static widening, but not the **mode-flip** route: turning `ALISSA_REVIEW_REPOS_SOURCE` from `bows` back to `static` on the *same* pin reaches the identical end state, and this guard never runs.

Verified at head:

1. A `>= 0.29.0` boot with an empty seed writes the stamped `${CONFIG}` (`repos: []` + the three keys) and an empty manifest; both persist on `/workspace`.
2. The operator turns the feature off the obvious way — `ALISSA_REVIEW_REPOS_SOURCE=static`, seed still empty. `REPOS_SOURCE != "bows"`, so this whole block is skipped; `repos_lines` is empty and `BOWS_MODE=0`, so control reaches the **mounted-mode** `else` (~line 887). The manifest exists, so it does not die, and the stamped config is respected as-is.
3. The daemon loads that file with the env forcing `repos_source=static`. Measured against this head with `Config.build(file_data=<the stamped render>, environ={"ALISSA_REVIEW_REPOS_SOURCE": "static"})`:
   - `on_missing_hub=add` (the container's baked default, `revloop-config.sh:217`) → `ValueError: on_missing_hub='add' requires a non-empty repos allowlist …`. Dies by name. Good.
   - `on_missing_hub=skip` (`ALISSA_ON_MISSING_HUB=skip`) → loads, `repos=()`, and `watches("someone/else")` is **True** — every PR that requests this reviewer. That is precisely the "an EMPTY static allowlist under … is 'every PR that requests this reviewer'" outcome this die's own comment argues must never be handed on, just reached by turning the mode off instead of rolling the pin back.

Adjacent, smaller: with the variable merely **unset** (rather than set to `static`), the file's `repos_source: bows` stands — blank falls through — so the mode the operator believes they disabled is still running off a file they no longer think is authoritative.

Why `[minor]` and not more: the exposure needs `ALISSA_ON_MISSING_HUB=skip`, and a `bows` deployment with `skip` never self-hubs a derived repo, so it is already close to inert. The default fails loudly. Not a merge gate.

Cheapest fix in the same shape as what you just wrote: make the guard's predicate "the config is positively ours **and** the mode that wrote it is not the mode we are booting in" — i.e. fire the same `die` (daemon role, empty seed, `config_is_generated`) whenever the resolved `repos_source` is not `bows`, not only when the skew probe fails. Failing that, one sentence in the README's provenance bullet: turning the mode off means deleting the stamped file, because it is unusable under any other mode.

alissa-app (approved):
## Round 2 of cap 22 — reconciliation clean, delta clean, **approve**

Reviewed head `77e0717`. All five round-1 findings were triaged `[triage:pursue]` and all five hold at this head; the new delta (`f1d6fee..77e0717`, 7 files) adds one `[minor]` that does not gate the merge.

| --- | --- | --- |
| `[major]` `loop.py:1757` — one-to-one slug map drops the same-name repo holding the live round | pursue | **fixed** (`3767be1`) — `slugs: dict[str, set[str]]`, `busy \|= slugs[ref.repo]`, matching `_name_candidates`' one-to-many model |
| `[minor]` `loop.py:1803` — the loop reaches into `State._write_telemetry` | pursue | **fixed** (`3767be1`) — `record_derived_repos` wraps its own `_write_telemetry(lambda: self._replace_derived_repos(rows), …)` and returns `bool`; `git grep 'self\.state\._' -- loop.py` is now empty |
| `[minor]` `entrypoint.sh:761` — the skew fall-through's promise breaks on a downgrade | pursue | **fixed** (`77e0717`) — dies by name at the guard instead of re-rendering; the argument for *not* re-rendering (an empty static allowlist under the old library widens rather than narrows) is correct and is why I did not push for the other option |
| `[minor]` `entrypoint.sh:851` — the never-overwrite rule holds only on the empty-seed path | pursue (docs) | **fixed** (`77e0717`) — README's provenance bullet and the *authoritative* paragraph both now scope the rule to the empty seed, and branch 1 carries the matching comment. Accurate against the code |
| `[nit]` `loop.py:1748` — a second roster listing per refresh | pursue | **fixed** (`3767be1`) — `_pass_roster` carries the sweep's post-reap list; `None` correctly means "no sweep listed this pass" and an empty list is a real answer, so the preflight refresh and a failed sweep listing both still list for themselves |

Nothing here was re-raised, and no fix regressed anything I could find.

### Verification I ran myself

- Full unit suite at this head: **1060 passed**.
- **Mutation check on the `[major]` fix** — reverting `slugs` to the one-to-one map makes `test_a_live_session_pins_every_watched_repo_sharing_its_name` fail with `{'orgB/web'} == {'orgA/web', 'orgB/web'}`. The test is not vacuous and the fix is what makes it pass.
- `docker/claude/tests-entrypoint-config.sh` end to end: **ALL PASS**, including 3c (the downgrade dies non-zero, before any worker, naming the stamped file and the widening) and 3d (the counter-case: an unstamped config on the same boot still comes up through mounted mode, untouched).
- The `[minor]` below was measured, not reasoned: `Config.build` at this head, both `on_missing_hub` values.

### CI gate at the reviewed head

Rollup read from `commits/77e07178371f42af6f49d6e541f6c551f3fc6318/check-runs` — the full check-runs read, not the Actions-API fallback — re-read immediately before submitting: **8 contexts, all concluded**, none failed. Image contract, Wheel Package Check, Python Types Check, Unit-Test Check, Version Bump Check, Entrypoint config renderer + console wiring and Python Style Check all `success`; `[code]smith` `skipped`. `commits/<sha>/status` reports 0 legacy contexts. Head sha confirmed unmoved between hydration and submission.

### Dimensions on the new delta

- **Correctness** — nothing found. `_pass_roster` is cleared first thing in `poll_once` alongside the other per-pass memos, set only at the end of a sweep that listed, and filtered by kills only (`reaped`, not `would_reap`), so a dry-run sweep and a kill that raised both leave their sessions pinning their repos. Ordering holds: sweep → refresh → search. The one behavioural change versus a fresh listing — a just-killed session can no longer linger in the roster for one extra refresh — moves in the drop direction only for a round the sweep proved finished, and the ledger half still covers a verdict owed.
- **Coding practices** — nothing found. `record_derived_repos` now matches every other telemetry writer (public method wraps, `_replace_*` does the transaction, `bool` return), and `_pass_roster` is documented in the same register as `_session_election`-style memos around it, including what `None` means.
- **Security** — one `[minor]`, inline on `entrypoint.sh`: the mode-flip route into the same empty-static widening the new downgrade die exists to prevent. Narrow (needs `ALISSA_ON_MISSING_HUB=skip`; the baked `add` default fails loudly), so it is a hardening gap, not a hole.
- **Performance** — nothing found; the delta removes a subprocess per refresh and adds a per-pass list reference.
- **Side-effects / blast radius** — the same `[minor]`. The new hard `die` is a new fail-stop on a persistent-volume boot, which is the right direction and is fenced on both sides by harness cases (it fires on our own stamped file, it does not fire on an operator's). The docs delta is accurate against the code it describes: branch 1 does render unstamped and unconditionally under `bows` with a non-empty seed.

### Verdict

**approve.** The `[minor]` is fix-now-or-register-a-follow-up per CR5; it does not need another round.

PR #127, opened 2026-09-15
TASK-1282240756: opt-in fleet-vitals push to Studio — one snapshot per pass to POST /v1/loop/fleet-vitals (issue #126)

### LINKED ISSUE #126: revloop: opt-in push of fleet vitals to Studio (`fleet_vitals_enabled` / `ALISSA_REV_FLEET_VITALS_ENABLED`) — one snapshot per pass to `POST /v1/loop/fleet-vitals`
## Why

The Factory's `/fleet` screen pulls each console's `/api/state` over private networking with passcodes held on the hosted Factory — per-deployment by construction, so a Dark Factory customer running their own revloop sees `not_configured`. Studio task **TASK-182095034** (hard dependency of this one, validated) adds `POST /v1/loop/fleet-vitals`: one snapshot per user×seat, replaced on every POST, rendered by the Factory whenever it has no `FACTORY_REVLOOP_URL`. This task is the revloop half: push one snapshot per completed poll pass, **opt-in**, in the same posture as loop events (best-effort, never fatal, honours dry-run).

## Wire contract

Fixed in TASK-182095034's description — build the body to it **exactly**: `schemaVersion: 1`, `seat: "revloop"`, `asOf`, `heartbeatAt`, `pollIntervalS` required; the rest optional/nullable; the API is **strict** (unknown keys → 400) and caps `sessionList` and `inbox` at 50 entries and the body at 64 KB, so cap before sending.

### 1. Config (`config.py`, `__main__.py`, `revloop.config.example.json`)

- New key `fleet_vitals_enabled` (bool, default `false`), CLI `--fleet-vitals` / `--no-fleet-vitals`, env `ALISSA_REV_FLEET_VITALS_ENABLED` with the **same env > file > flag rail** as `ALISSA_REV_LOOP_EVENTS_ENABLED` (`env_loop_events_enabled` — factor a shared bool-env reader rather than duplicating it; `1/true/yes/on` and `0/false/no/off`, anything else refused by name, blank falls through).
- Reuses `alissa_endpoint` and the CLI's `ALISSA_API_TOKEN`. No new secret.

### 2. The snapshot (`fleet_vitals.py`, new)

Built **in-process from the same builders the reviewer console uses** (`webui/sources.py` — the LIVE half, the inbox — and `webui/sysinfo.py`), with the console sidecar **not required** to be running. Mapping:

| snapshot field | source |
| --- | --- |
| `heartbeatAt` | completion time of the pass that produced the snapshot |
| `pollIntervalS` | `poll_interval` |
| `version` / `drift` | the console's drift read |
| `pollDurationsMs` | the console's recent pass durations, oldest first, ms |
| `sessions` / `sessionList` | the roster of this daemon's own grammar: `{live, managed}` and ≤ 50 rows `{name (full), edge: "review", repo, number, round, attempt: null, ageS, cpu, rssBytes, live, managed, url}`; `null` for both when the roster is unreadable, never `{0,0}` for "could not list" |
| `rate` | the console's cached GitHub rate read for the reviewer identity — never an extra call per pass |
| `memory` | `sysinfo` resident / reclaimable / limit, or null |
| `queueDepth` | owed rounds waiting on the spawn gate (`max_concurrent_sessions`), or null |
| `inbox` | the console's inbox (escalations + pings), newest first, ≤ 50, `{kind, subject, repo, number, isPr: true, ageS, url, lever}` |
| `kpis` | `null` (orcloop-only) |

### 3. The push (`loop.py`, `alissa_client.py`)

- `AlissaClient.post_fleet_vitals(snapshot)` beside `post_loop_events`, same client (redirect-refusing, classified errors).
- At the end of every completed pass, **after** the loop-events push: if enabled and not dry-run, POST once. Failure = **one WARNING** naming status + error; the pass completes; no retry queue. `403 not_activated` → once per boot at WARNING with the activate URL, then DEBUG.
- `--dry-run` → `[dry-run] would push fleet vitals (…)`, nothing sent. Poll summary gains `vitals: pushed|skipped|failed`.

### 4. Container (`docker/claude`)

- `ARG ALISSA_REV_FLEET_VITALS_ENABLED=""` in the Dockerfile with the "REQUIRES REVLOOP_VERSION ≥ <this release>" comment the other pass-through keys carry; `revloop-config.sh` passes it through when set; README row (unset ⇒ library default `false`). `tests-entrypoint-config.sh` cases: set → rendered; unset → absent; garbage → refused by name.
- Do **not** bump `ARG REVLOOP_VERSION` in this PR (a sibling task re-pins it; CI installs the pinned version from PyPI, which cannot yet carry this release).

### 5. Docs + version

- `README.md`: config-table row + a *Fleet vitals (Studio ingest)* section beside *Loop telemetry (Studio ingest)*. Example config key with `_` comment.
- `version.py`: bump per `check-version-bump.yaml` to the next free minor (0.29.0 is published; take 0.30.0 or whatever is next when this lands).

### 6. Tests

- `test_fleet_vitals.py`: sources fixtures → the exact contract body (null-vs-{0,0}, 50-caps, `round` parsed from the session grammar).
- Push gating: disabled → no call; dry-run → log only; enabled → one POST per pass after loop events; failure → one WARNING and the ledger still writes; not_activated → once-per-boot.
- Config: env rail precedence + refusal of garbage values, shared with the loop-events reader.

## Acceptance

- With `ALISSA_REV_FLEET_VITALS_ENABLED=1`, `GET /v1/loop/fleet-vitals?seat=revloop` advances every pass and the Factory revloop card renders live (sessions with their round, inbox) with `ALISSA_UI_ENABLED` unset.
- Unset → no request ever made. An Alissa outage during the push costs one WARNING per pass.

(Migrated 2026-09-15 from TASK-627218259 / issue #121 to the alissa-app-owned feed for the ali-fhr fleet; the old issue was closed unworked.)

## Definition of Done

PR on ali-fhr/alissa-dark-factory-revloop merged: `fleet_vitals_enabled` key (default false) with CLI + `ALISSA_REV_FLEET_VITALS_ENABLED` env rail (shared bool reader with loop events); container ARG + revloop-config.sh pass-through; snapshot builder reusing webui/sources + sysinfo in-process (sidecar not required) mapped to the TASK-182095034 contract; one best-effort POST per completed pass after loop events, dry-run honoured, one-WARNING failure posture; README + docker README + example config; tests listed above; version bumped; `ARG REVLOOP_VERSION` untouched.

## Acceptance criteria

- [ ] Sources fixtures → contract body exactly (required fields, null-vs-{0,0} roster rule, 50-entry caps, round parsed from session names)
- [ ] Disabled → no call; dry-run → log only; enabled → one POST per pass after loop events; failure → one WARNING and the pass completes; not_activated → once-per-boot WARNING
- [ ] tests-entrypoint-config.sh: ALISSA_REV_FLEET_VITALS_ENABLED set → rendered; unset → absent; garbage → refused by name; ARG REVLOOP_VERSION unchanged
- [ ] Operator: enable on a deploy with ALISSA_UI_ENABLED unset; GET /v1/loop/fleet-vitals?seat=revloop advances each pass and the Factory revloop card renders without FACTORY_REVLOOP_URL

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/config.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa_client.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/fleet_vitals.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/__main__.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sources.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/webui/sysinfo.py, alissa-tools-github-revloop/src/test/**, README.md, revloop.config.example.json, docker/claude/Dockerfile, docker/claude/revloop-config.sh, docker/claude/README.md, docker/claude/tests-entrypoint-config.sh
Alissa-Task: TASK-680852575

vlissa-app:
Closes #126

Origin task: TASK-680852575 (Alissa-Task) — implementation task: TASK-1282240756 (downstream of the origin).

Depends on Studio's TASK-182095034 (`POST /v1/loop/fleet-vitals`, validated). Independent of the sibling `ARG REVLOOP_VERSION` re-pin task: this PR deliberately leaves the Dockerfile pin untouched (CI installs the pinned release from PyPI, which cannot yet carry 0.30.0); the operator sequences the re-pin after this release publishes.

## What

- **Config** — `fleet_vitals_enabled` (default `false`), `--fleet-vitals` / `--no-fleet-vitals`, env `ALISSA_REV_FLEET_VITALS_ENABLED` on the same env > file > flag rail as `ALISSA_REV_LOOP_EVENTS_ENABLED`, through one shared boolean reader (`_env_bool`; `1/true/yes/on`, `0/false/no/off`, anything else refused by name, blank falls through).
- **Snapshot** — new `fleet_vitals.py` builds the TASK-182095034 contract body in-process from the console's own builders (`webui/sources.py` + `webui/sysinfo.py`; the sidecar need not run): heartbeat, poll interval, version/drift, the last 60 pass durations, the `alissa tmux ls` roster (`{live, managed}` + ≤ 50 rows with PR/round from the spawn ledger or parsed from the session name; both `null` when the roster cannot be listed), the cached reviewer-identity rate read, the cgroup memory split + `memory.max` limit, the spawn-gate queue depth, the live inbox (≤ 50), `kpis: null`. Only contract keys are emitted; `fit_body` trims the two lists until the body is ≤ 64 KB.
- **Push** — `AlissaClient.post_fleet_vitals` beside `post_loop_events` (same redirect-refusing client, same taxonomy). At the end of every completed pass, after the loop-events push: one POST when enabled and not dry-run. Failure = one WARNING naming status + error, the pass completes, no retry queue; `403 not_activated` = one WARNING per boot with the activate URL, then DEBUG (the push is still attempted so activation lands without a restart). `--dry-run` builds the snapshot and logs `[dry-run] would push fleet vitals (…)`, sends nothing. Every pass now ends with `poll summary: … vitals: pushed|skipped|failed`.
- **Container** — `ARG ALISSA_REV_FLEET_VITALS_ENABLED=""` + ENV pass-through in the Dockerfile (with the "REQUIRES REVLOOP_VERSION >= 0.30.0" note); `revloop-config.sh` renders `fleet_vitals_enabled` when set, skew-gated like the bows keys (an older pin drops it with a WARN naming the re-pin), with the spelling refused **before** the gate so garbage is refused on every pin; README row; `tests-entrypoint-config.sh` cases.
- **Docs + version** — README config-table row + *Fleet vitals (Studio ingest)* section beside *Loop telemetry*; example config key with `_` comment; `version` → `0.30.0`.
- **Tests** — `test_fleet_vitals.py` (51 tests): console fixtures → exact contract body, null-vs-`{0,0}`, 50-caps, round from the session grammar, body fitting, dry-run/failure/not_activated posture, watcher wiring and POST order after loop events, the client method, the config rail shared with loop events, the container variable name, the new console builders.

### Judgment calls

- **Poll summary line.** The daemon had no per-pass summary line to extend, so every pass now ends with one INFO line `poll summary: N candidate(s), N reaped, N ms, vitals: pushed|skipped|failed` (`skipped` = disabled or dry-run). This is one new INFO line per pass whether or not vitals are enabled.
- **Failure posture, literally per pass.** Every push failure (transient, 401, missing token) is one WARNING per pass and no latch, per the issue's "an outage costs one WARNING per pass"; only `403 not_activated` is once-per-boot-then-DEBUG. This differs from the loop-events emitter, which latches off on auth failures; the pusher keeps attempting so a mid-run activation or token fix lands on its own.
- **Dry-run still builds the snapshot** (ledger, tmux, /proc, the two cached checks — reads only) so the `[dry-run]` line can describe what would be sent; the POST is the only thing suppressed.
- **Roster semantics.** `sessionList` carries every session `alissa tmux ls` returns (foreign lanes flagged `managed: false`), the `{live, managed}` counts describe the whole roster, and rows are ordered live-managed → live-foreign → gone before the 50-cap so the cap drops what the operator cares least about. A session with no ledger row resolves its repo through the allowlist only when exactly one `owner/repo` matches the name's slug; otherwise `repo`/`url` are null.
- **`rate` is the console's cached read** (60 s TTL, `gh api rate_limit`, run under the reviewer identity's credential via the daemon's own `GitHub._env`). At the default 60 s poll that is at most one cached refresh per pass; there is no uncached extra call. `resetAt` is derived from the read's `reset` epoch.
- **`memory.limitBytes`** comes from a new `sysinfo.cgroup_memory_limit` (`memory.max`; `max` → null) rather than adding a key to the console tile's dict, whose key set is pinned by its tests. `Sources.memory()` wraps both for the builder.
- **Sources additions**: `session_rows()` returns `None` for an unlistable roster (the dashboard's `sessions()` still degrades to `[]`), and `inbox()` factors the dashboard's inbox build so the card and the console cannot disagree.
- **`drift`** is the newer PyPI version only when the console's drift state is `behind`; `current`/`ahead`/unknown → null.
- **Renderer skew gate**: `fleet_vitals_enabled` is skew-gated because the Dockerfile's pinned library rejects unknown keys; unlike the bows keys, the boolean spelling is validated in bash before the gate so a typo is refused rather than silently dropped on an old pin. `_skew_value` gained an optional "landed in" argument for the WARN text.
- **Version 0.30.0** (0.29.0 is on PyPI). `ARG REVLOOP_VERSION` is untouched — note it currently reads `0.16.14` on `main`, so the fleet-vitals ENV is inert in the container until the sibling re-pin lands; a shell test pins that this PR did not move it.

- [ ] Live acceptance: with `ALISSA_REV_FLEET_VITALS_ENABLED=1` on a deployment, `GET /v1/loop/fleet-vitals?seat=revloop` advances every pass — no Alissa POST was made from this session (it would write under this session's token to the operator's Studio data).
- [ ] The Factory revloop card rendering from the pushed snapshot with `ALISSA_UI_ENABLED` unset and no `FACTORY_REVLOOP_URL` — needs the deployed Factory.
- [ ] Studio's strict zod schema accepting the exact body as built (the issue's contract was followed field for field, but no round-trip against the real API ran here).
- [ ] The container end to end: image build and the pinned-library cross-check in `tests-entrypoint-config.sh` ran against this checkout's library, not a built image; `tests-image-contract.sh` needs Docker (see Verification for what ran).
- [ ] Real-fleet body size: `fit_body` is exercised synthetically; a real 50-row roster plus 50-item inbox is ~25 KB by construction, well under 64 KB, but not measured on a live fleet.

### Verification

- `bash check-style.sh alissa-tools-github-revloop` (pycodestyle, max-line 120): clean.
- `bash check-types.sh alissa-tools-github-revloop` (mypy): `Success: no issues found in 38 source files`.
- `bash tests-unit.sh alissa-tools-github-revloop` (pytest): 1111 passed (51 new in `test_fleet_vitals.py`), run in a private venv with the dist installed editable, as CI does.
- `bash docker/claude/tests-entrypoint-config.sh`: ALL PASS, including the new fleet-vitals cases (unset → absent; set → rendered for all ten spellings; blank → absent; garbage → refused by name on a library that knows the key AND on an old-pin stub; old pin drops the key with the WARN naming the re-pin; Dockerfile ARG/ENV present; `ARG REVLOOP_VERSION` unmoved).
- The other `docker/claude/tests-entrypoint-*.sh` suites (ui, auth, executor, identity, prune, chown): ALL PASS. `tests-image-contract.sh`: SKIPPED here (no Docker daemon) — the CI "Image contract" job is the operator's read.
- PR CI on the pushed head: every job green — style, types, unit tests, version bump, wheel, entrypoint config renderer, Image contract (build + in-image assertions), CodeQL.
- In-process smoke on this host (dry-run, real `Sources`): the real roster, /proc CPU/RSS, cgroup split + limit, rate read and drift all populated; `[dry-run] would push fleet vitals (sessions=3/0 live/managed, inbox=0, queue=0, 1413 bytes)`; nothing sent.
- `alissa-revloop --help` shows `--fleet-vitals | --no-fleet-vitals`.

### Scope touched

All within the issue's declared scope; no excursions:

- `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/{config.py, __main__.py, alissa_client.py, loop.py, version, webui/sources.py, webui/sysinfo.py}` — modified; `fleet_vitals.py` — new. (`version.py` is a derived reader; the plain-text `version` file is what `check-version-bump.yaml` keys on, so that is what was bumped.)
- `alissa-tools-github-revloop/src/test/.../test_revloop/test_fleet_vitals.py` — new.
- `README.md`, `revloop.config.example.json`, `docker/claude/{Dockerfile, revloop-config.sh, README.md, tests-entrypoint-config.sh}` — modified.


vlissa-app:
- 2026-09-15 19:39:56 UTC — `orphan-ali-fhr-alissa-dark-factory-revloop-pr127` — orphan sweep — handed off — marked ready (reviewer alissa-app was already requested); the session that opened this draft died before flipping it, so the daemon completed its protocol (head 4905162b public 325 min) (no session spawned, once per PR/head)

alissa-app:
- 2026-09-15 19:40:22 UTC — `review-alissa-dark-factory-revloop-pr127-r1-2d0801` — round 1 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/fleet_vitals.py
```diff
@@ -0,0 +1,504 @@
...
+        if repo is None:
+            repo = _resolve_repo(ref.repo, repos)
+    url = (
+        f"https://github.com/{repo}/pull/{number}"
+        if repo and number is not None
+        else None
+    )
+    cpu = row.get("cpu_percent")
+    return {
+        "name": str(name) if name is not None else "",
```
alissa-app, line 168: [minor] The empty-string fallbacks here and in `_inbox_row` emit values Studio's strict schema **rejects**, turning one degenerate row into a 400 for the whole snapshot.

`FleetVitalsSessionSchema.name` is `z.string().min(1).max(500)` and `FleetVitalsInboxItemSchema.subject` is `z.string().min(1).max(500)` (`packages/client/src/schemas/loop.ts`). So:

- `fleet_vitals.py:168` — `str(name) if name is not None else ""` produces `""` for a roster entry with no `name` (`sources.session_rows` appends the row regardless; it only skips non-dict entries).
- `fleet_vitals.py:191-193` — `str(repo or "")` produces `""` when `repo_slug` is falsy.

Either one fails validation for the **entire body**, so the card silently stops advancing (one WARNING per pass) rather than losing the one bad row.

Reachability is low and I want to be straight about that: `alissa tmux ls --json` always names its sessions, and `escalations.repo` / `pings.repo` are `TEXT NOT NULL` in `state.py`, so I could not construct a real path to either. I'm raising it because this module's stated contract is that *the builder* owns fitting the body to the API ("A snapshot the API refuses is worth nothing, so fitting is the builder's job, not the API's" — module docstring), and these two fallbacks are the only places that hand the API something it will refuse. Dropping the row (or the whole-body-safe equivalent) would keep that promise; `""` trades a row for the snapshot.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/fleet_vitals.py
```diff
@@ -0,0 +1,504 @@
...
+    orcloop-only section).
+    """
+    stamp = time.time() if now is None else now
+    snaps = sources.snapshots(DURATION_POINTS)
+    durations = [
+        int(s["duration_ms"]) for s in reversed(snaps)
+        if _int_or_none(s.get("duration_ms")) is not None
+    ]
+    drift = sources.drift()
+    rows = sources.session_rows()
```
alissa-app, line 263: [minor] Enabling vitals adds a **second** roster listing per pass, and the README's "never an extra call per pass" for `rate` doesn't hold when the console sidecar is off.

`sources.session_rows()` shells out to `alissa tmux ls --json` and then, for each live session, `pane_pid()` plus a `sysinfo.build_index()` walk of `/proc`. But the daemon already lists its roster once per pass into `self._pass_roster` (`loop.py:4179`, consumed at `loop.py:1771`) — and "one roster listing per pass" was itself a round-1 review fix on this repo (`3767be1`). This reintroduces the second listing through a different door.

On `rate`: `RATE_CACHE_TTL` is `60.0` and the default `poll_interval` is `60`, so with `ALISSA_UI_ENABLED` unset — which is exactly the acceptance scenario this feature targets — the cache is cold at essentially every pass and each one spawns a fresh `gh api rate_limit` subprocess (20 s timeout, synchronous, at the tail of the pass). It doesn't consume GitHub quota (`/rate_limit` is exempt), so the impact is subprocess and latency cost rather than budget — but `README.md`'s "the console's **cached** `gh api rate_limit` read (60 s) ... **never an extra call per pass**" reads as a stronger guarantee than what ships.

Neither gates the merge at a 60 s interval. Worth either threading `_pass_roster` into the builder, or softening the README line to say what it actually costs when the sidecar isn't running.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -5002,10 +5020,17 @@ def poll_once(self) -> list[tuple[str, Decision]]:
...
-        )
+        duration_ms = int((time.monotonic() - started) * 1000)
+        self._write_snapshot(results, reaped, duration_ms=duration_ms)
         self._emit_loop_events()
+        # AFTER the loop-events push, and after the snapshot, so the vitals
+        # heartbeat is the completion time of a pass whose exhaust is already
+        # written and whose events are already on their way.
+        vitals = self._push_fleet_vitals(completed_at=time.time())
+        log.info(
+            "poll summary: %d candidate(s), %d reaped, %d ms, vitals: %s",
```
alissa-app, line 5031: [nit] This `poll summary` INFO line is now unconditional — it fires on every pass whether or not vitals are enabled, so every existing deployment gains ~1440 log lines/day at the default interval without opting into anything.

The PR body flags this as a deliberate judgment call (there was no per-pass summary line to extend), and it is genuinely useful output, so I'm not asking for a change — just making the cost visible on the PR so the operator can decide. Gating it on `self._fleet_vitals is not None`, or dropping it to DEBUG when vitals are off, would keep the new signal for the deployments that asked for it.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/fleet_vitals.py
```diff
@@ -0,0 +1,504 @@
...
+            self._warn_failed(exc)
+            return VITALS_FAILED
+        log.info(
+            "fleet-vitals: pushed (%s; replaced=%s)",
+            describe(snapshot), result.get("replaced"),
+        )
+        return VITALS_PUSHED
+
+    @staticmethod
+    def _warn_failed(exc: AlissaError) -> None:
```
alissa-app, line 424: [nit] A `401` / missing-token never latches, so a *fixed* misconfiguration warns once per pass forever.

`403 not_activated` gets the once-per-boot treatment just below, on the reasoning that it's "a fixed condition the operator resolves out of band". An absent or revoked `ALISSA_API_TOKEN` is equally fixed, but lands here and warns every pass indefinitely (~1440/day), which is also where this diverges from `LoopEventsEmitter`, which latches off on auth failures.

I'm keeping this a nit deliberately: the issue's acceptance text says "an outage costs one WARNING per pass", the PR body calls the divergence out explicitly, and the non-latching behaviour is what lets a mid-run token fix land without a restart. That's a defensible reading. Flagging only in case the once-per-boot treatment was meant to cover every fixed auth condition rather than `not_activated` alone.

alissa-app (approved):

Reviewed head `4905162`. Fresh-instance review per the `alissa-code-review` skill (CR1–CR9), judged against origin task **TASK-680852575**'s definition-of-done and the TASK-182095034 wire contract. Review task: **TASK-303598764**.

**Verdict: approve.** No blockers, no majors. Four findings — two `[minor]`, two `[nit]` — none of which gate the merge. Register the minors as follow-ups or fix them here, your call.

### What I checked beyond the diff

I validated the built body **field-for-field against Studio's actual strict schema** (`packages/client/src/schemas/loop.ts`, `FleetVitalsSnapshotSchema`), which is merged on `studio.alissa.app@304a782` in this workspace — the round-trip the delivery note lists as unverified. Results:

- Top-level key set is **exactly** the 15 the `.strict()` object allows — no unknown-key 400.
- `sessionList: null` / `sessions: null` (the unlistable-roster rule) **is** accepted: both are `.nullable().optional()`.
- Every nested object (`FleetVitalsSessionSchema` 12 keys, `FleetVitalsInboxItemSchema` 8, rate 3, memory 3) is `.strict()` and the builder emits exactly those key sets.
- `asOf`/`heartbeatAt`/`resetAt` satisfy `z.string().datetime({ offset: true })`; `pollIntervalS` satisfies `.positive()` (config floors it at `MIN_POLL_INTERVAL`); `pollDurationsMs` ≤ 60 against a cap of 100.
- **Body cap direction is safe.** `body_bytes` uses Python `json.dumps` (ASCII-escaped, `", "`/`": "` separators); Studio measures `JSON.stringify` + `TextEncoder`. Python's count is always ≥ the wire count, so `fit_body` is conservative — it can never under-measure into a 400.

The one mismatch I did find is finding 1 below.

I also checked `{live, managed}` counting semantics against the Factory's own **pull** normalizer (`factory/src/lib/fleet.ts:514,610`): it computes `sessions.filter(s => s.live).length` / `.filter(s => s.managed).length` over the whole roster, foreign lanes included — **identical** to `build_snapshot`. The "card and console can never disagree" design goal holds, including the `live=3, managed=3` shape the golden test pins.

### Rubric

- **Correctness** — nothing blocking. Every DoD item is present and verified at head: `fleet_vitals_enabled` default `false`, the CLI pair, `ALISSA_REV_FLEET_VITALS_ENABLED` on the shared `_env_bool` rail, the container ARG/ENV + skew-gated renderer, the in-process snapshot, the post-loop-events push, dry-run, `version` → `0.30.0`, and `ARG REVLOOP_VERSION` verifiably still `0.16.14` (unchanged from `667cec2`). Argparse default resolution is right: `--fleet-vitals` registers `default=None` first, so the mutually-exclusive pair leaves `None` (not `True`) when neither flag is given. `reversed(snaps)` correctly yields oldest-first — `snapshots()` returns newest-first and the dashboard reverses the same way. See finding 1.
- **Coding practices** — nothing found. Idiom matches the repo closely: the pusher mirrors `LoopEventsEmitter`'s seam-and-shape, `_env_bool` is a genuine de-duplication rather than a copy, the `null`-vs-`{0,0}` rule is stated once and held everywhere, and comment density matches the surrounding modules. `Co-Authored-By: Claude Fable 5.1` matches this repo's established practice (`30ec975`, `77e0717`, `a02d0fe`), so I'm not raising it. No dead code, no debug leftovers, no untracked TODOs.
- **Security** — nothing found. No new secret; the existing `ALISSA_API_TOKEN` and `alissa_endpoint` are reused. `_console_runner` correctly routes `gh` through `GitHub._env` so the rate read is the reviewer identity's, and an unresolvable token degrades to a `CommandError` → null rate rather than a read under the wrong login. No token or credential reaches any log line — `describe()` emits counts and a byte size only. The snapshot contains repo slugs and PR numbers, but it is pushed under the operator's own token to the operator's own Studio row, and the whole feature is opt-in and off by default. The bash spelling guard runs *before* the skew gate, which correctly prevents a typo being swallowed as "unsupported, dropped" on an old pin, and `set -euo pipefail` + `pipefail` propagates its `return 1` through `render_revloop_config_bows`'s pipeline, so a garbage value fails the boot instead of `mv`-ing a truncated config over a good one.
- **Performance** — see finding 2. Nothing algorithmic; the caps and `fit_body` are sound (the halving loop terminates — `len//2` reaches 0 — and gives up cleanly once both lists are empty).
- **Side-effects / blast radius** — nothing found. Off by default, so no deployment changes behaviour on upgrade. The cross-repo consumer (Studio + Factory) is merged and its schema accepts this body, as verified above. The `Sources.sessions()` refactor preserves the dashboard's contract (`session_rows(...) or []` keeps `[]` for the panel while the new `None` is visible only to the vitals builder), and `cgroup_memory_limit` is deliberately kept out of the console tile's pinned dict. The push is a replace-only write with no history, so a failed pass costs nothing and needs no retry queue. `ARG REVLOOP_VERSION` being left at `0.16.14` makes the container ENV inert until the sibling re-pin — correctly scoped out and pinned by a shell test.

### Findings

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 0 (waived: 0) |
| minor | 2 |
| nit | 2 |
| question | 0 (open: 0) |

Findings are inline. CI at `4905162`: all 11 contexts concluded — 10 success, 1 skipped (`[code]smith`), none failed.

PR #133, opened 2026-09-16
TASK-1917563644: entrypoint pins the alissa skillsDir to $CLAUDE_CONFIG_DIR/skills (issue #132)

### LINKED ISSUE #132: revloop image: sessions cannot load skills — `CLAUDE_CONFIG_DIR` relocates Claude's skills dir, the CLI installs elsewhere; make the entrypoint pin the alissa skills dir to `$CLAUDE_CONFIG_DIR/skills`
## Symptom (all three shared daemons, 2026-09-15)

Sessions start with `Skill(alissa-code-review)` → `Error: Unknown skill: alissa-code-review` (revloop console screenshot; the same for `alissa-code-loopwork-sentinel` in orcloop sentinel passes and the workspace/git skills in devloop workers). The session then hunts for the file on disk; some find it, all waste their first minutes and context on it.

## Root cause

The entrypoint runs sessions with `CLAUDE_CONFIG_DIR=/workspace/.claude-config` (persisted Claude login). Claude Code's documented rule: with `CLAUDE_CONFIG_DIR` set, personal skills are read from `$CLAUDE_CONFIG_DIR/skills/` **instead of** `~/.claude/skills/`. The alissa CLI (`cli/src/config.ts getSkillsDir()`) installs to `~/.claude/skills` unless `skillsDir` is set in its config, and the entrypoint never sets it. Verified live: `/workspace/.claude-config/skills` absent, `/home/alissa/.claude/skills/<slug>` present. The entrypoint's own comment already hedges ("CLAUDE_CONFIG_DIR reliably relocates only .credentials.json; whether it also …") and merges `settings.json`/`.claude.json` into both places — skills were the one thing it did not relocate. Anchors: `docker/claude/entrypoint.sh` — `CLAUDE_CRED_FILE` at ~L293, the config-dir WARN block ~L304-316, the settings/state merge python block ~L896-960 (`ccdir = os.environ.get("CLAUDE_CONFIG_DIR")`, `settings_targets`), `skills_lines()` ~L729 and the manifest `skills:` writers ~L815/L869.

A symlink `$CLAUDE_CONFIG_DIR/skills → /home/<runtime user>/.claude/skills` was placed by hand on the live containers as a stop-gap (it lives on the volume, so it survives redeploys, but nothing in the image creates it). A CLI-side fix (honour `CLAUDE_CONFIG_DIR` in `getSkillsDir()`) is fed separately; the image must not depend on the CLI version it happens to pull.

## Fix (entrypoint)

1. Right where the entrypoint writes/merges the runtime user's alissa CLI config (or, if it only sets env, in the same block that exports `CLAUDE_CONFIG_DIR`): when `CLAUDE_CONFIG_DIR` is non-blank, set the alissa CLI config key `skillsDir` to `$CLAUDE_CONFIG_DIR/skills` (JSON merge, same idempotent merge pattern as the `settings.json` block — never clobber other keys), and `mkdir -p` it owned by the runtime user. Log one line: `skills dir pinned to $CLAUDE_CONFIG_DIR/skills (Claude reads personal skills there when CLAUDE_CONFIG_DIR is set)`.
2. Migration of the stop-gap: if `$CLAUDE_CONFIG_DIR/skills` is a **symlink**, replace it with a real directory (copy the target's contents if any, then remove the link) before the pin — so the hand-placed link does not shadow the fix; if it is already a real directory, leave it.
3. If `CLAUDE_CONFIG_DIR` is blank, change nothing (the CLI default and Claude's default agree).
4. README (docker README + the root README's image/skills section): one paragraph on the rule and the pin, next to the existing skills list documentation.

## Tests

`docker/claude/tests-entrypoint-config.sh` (shell suite, same fixture style as the existing WS cases) — cover: pin written when `CLAUDE_CONFIG_DIR` is set (config JSON has `skillsDir` = `$CLAUDE_CONFIG_DIR/skills`, dir exists, owned by the runtime user); other config keys preserved; nothing written when unset; a pre-existing symlink is converted to a real directory with contents preserved; a pre-existing real directory is untouched; the log line appears once.

## Out of scope

CLI changes; installing skills at build time; project-level `.claude/skills`.

## Definition of Done

A merged PR after which the revloop image's entrypoint pins the alissa CLI `skillsDir` to `$CLAUDE_CONFIG_DIR/skills` (creating it, converting a stop-gap symlink, preserving other config keys) whenever `CLAUDE_CONFIG_DIR` is set, documents the rule, and pins the behaviour with entrypoint tests; a fresh session on the rebuilt image loads `Skill(...)` without 'Unknown skill'.

## Acceptance criteria

- [ ] Entrypoint sets skillsDir=$CLAUDE_CONFIG_DIR/skills (merge, not clobber), creates the dir owned by the runtime user, converts a pre-existing symlink, logs once; nothing changes when CLAUDE_CONFIG_DIR is blank.
- [ ] Entrypoint tests pin the six behaviours listed; image contract / CI green on the PR head; README updated.
- [ ] Operator check after the service rebuilds: the first session's transcript shows the skill loading via the Skill tool, no 'Unknown skill' line.

Autonomous-Scope: docker/claude/entrypoint.sh, docker/claude/tests-entrypoint-config.sh, docker/claude/README.md, README.md
Alissa-Task: TASK-181268515

vlissa-app:
Closes #132

Origin task: TASK-181268515
Implementation task: TASK-1917563644

## What

With `CLAUDE_CONFIG_DIR` set, Claude Code reads personal skills from `$CLAUDE_CONFIG_DIR/skills/` instead of `~/.claude/skills/`, while the alissa CLI installs skills to `~/.claude/skills` unless its config key `skillsDir` says otherwise. The image relocates the config dir for the persisted login and never pinned the skills dir, so every session opened with `Unknown skill`.

The entrypoint now has a step 3a-ii (after the claude first-run seed, before `alissa code workspace sync` installs the manifest skills) that, whenever `CLAUDE_CONFIG_DIR` is non-blank:

- converts a pre-existing **symlink** `$CLAUDE_CONFIG_DIR/skills` into a real directory (target contents copied in; copy-first, unlink-last, so a failed copy leaves the stop-gap link working);
- `mkdir -p`s `$CLAUDE_CONFIG_DIR/skills` (running as the runtime user after the gosu drop, so owned by it);
- merges `skillsDir=$CLAUDE_CONFIG_DIR/skills` into `${ALISSA_CONFIG_DIR}/config.json` load-then-update (every other key preserved, file mode preserved, atomic replace; an unparseable file is left alone with a WARN);
- logs `skills dir pinned to … (Claude reads personal skills there when CLAUDE_CONFIG_DIR is set)` once.

Blank or unset `CLAUDE_CONFIG_DIR`: nothing changes. Never fatal.

Tests: `docker/claude/tests-entrypoint-config.sh` gains five boots of the real entrypoint against the stubbed CLIs covering the six behaviours the issue lists (pin written + other keys preserved + dir exists owned by the runtime user + log line exactly once; nothing written when unset, empty, or whitespace; symlink converted with nested contents preserved and the former target untouched; a real dir is the same inode after boot; plus an unparseable config.json is never overwritten).

Docs: paragraph in the docker README's claude-login section, the `ALISSA_REVIEW_SKILLS` table row, the "What the entrypoint does" step list, and a paragraph in the root README's container section.

### Judgment calls

- The pin is written by a python JSON merge into `${ALISSA_CONFIG_DIR}/config.json` (the same load-then-update pattern as the settings block), not via `alissa config set skillsDir`, so it does not depend on the CLI version or on the CLI being present at that point. The WARN on failure names the `alissa config set` command as the manual fix.
- Placement: step 3a-ii, right after the claude first-run seed (which already handles `CLAUDE_CONFIG_DIR`) and after `alissa auth login` has written its config, before `alissa code workspace sync` installs skills. The issue allowed either the CLI-config block or the `CLAUDE_CONFIG_DIR` export block; the entrypoint has neither (it never writes the CLI config itself and the export is the image's), so this is the block where the ordering constraint holds.
- "Non-blank" is read as the python block reads it (`.strip()`): a whitespace-only `CLAUDE_CONFIG_DIR` changes nothing, and the tests pin it.
- A `config.json` that exists but is not a JSON object is left untouched with a WARN instead of being replaced by `{skillsDir}`; the settings block would have clobbered it, but that file holds the verified token. Tested.
- The rewrite preserves the existing file mode (the CLI's token file may be 0600) and replaces atomically. Tested.
- A stale `skillsDir` value already in the config (pointing elsewhere) is overwritten with the pin; the issue says "set", and a stale value is the failure mode the pin exists for. Tested.
- A dangling symlink is converted to an empty real directory (there is nothing to copy). Not separately tested.
- The entrypoint step list in the docker README gained a "2b" line for this step; the issue asked for one paragraph next to the skills documentation, which is also there. Minor doc excursion inside the declared scope files.

- [ ] Ownership by `alissa` on the real image: the suites run as the test user and assert owner equals `id -un`; the container path runs after `gosu alissa`, which no test exercises with a real uid switch.
- [ ] The migration on the live volumes: the stop-gap symlink there points at `/home/alissa/.claude/skills`; the copy path is tested with a fixture target, not the live one.
- [ ] Acceptance criterion 3 (first session on the rebuilt image loads `Skill(...)` without `Unknown skill`) needs a rebuild and a live transcript.
- [ ] That the alissa CLI honours `skillsDir` from `config.json` on the version the image pulls: taken from the issue (`cli/src/config.ts getSkillsDir()`), not re-read here.

### Verification

- `bash docker/claude/tests-entrypoint-config.sh`: ALL PASS (38 new assertions in the #132 section; run twice, the second after the mode-preservation tweak).
- `bash docker/claude/tests-entrypoint-{ui,identity,auth,executor,prune,chown}.sh`: all six exit 0 with no FAIL lines (each boots the real entrypoint through the new step; the executor suite includes a boot with `CLAUDE_CONFIG_DIR` set).
- `bash -n` on both shell files: clean. shellcheck is not installed here and is not in CI.
- Python style/type/unit suites not run: no Python source changed.
- Docker / image contract: not runnable here (no Docker); on the PR head cbc82a8 all ten GitHub checks pass, including "Image contract (build + in-image assertions)" and "Entrypoint config renderer + console wiring".

### Scope touched

- `docker/claude/entrypoint.sh`: new step 3a-ii only; no other block edited.
- `docker/claude/tests-entrypoint-config.sh`: new section appended before the summary line; existing cases untouched.
- `docker/claude/README.md`: claude-login section paragraph, `ALISSA_REVIEW_SKILLS` row, entrypoint step list.
- `README.md`: one paragraph in "In a container".
- No excursion outside the issue's declared scope.

### Round 1 fixes (932c6fc)

Triage: all four round-1 findings pursued — the `[major]` at `entrypoint.sh:1025` (stale `installed` records) and the three nits (WARN wording, orphan staging dir, no escape hatch). No finding ignored or deferred.

#### Judgment calls

- The `[major]` is fixed by pruning the CLI's `installed` map (entries with no `SKILL.md` under the pinned dir are dropped in the same JSON rewrite), not by copying `~/.claude/skills` into the empty pinned dir: on the executor the home dir is ephemeral, so on the redeploy that first applies the pin there is nothing to copy, and only the prune un-sticks the state. Verified against CLI 0.2.1 on this machine that `isSkillInstalled()` trusts the map first and `installSkill()` writes into `getSkillsDir()` and rewrites the entry.
- The prune runs on every boot, so a skill directory removed by hand is also re-armed for reinstall. An `installed` value that is not a JSON object is left untouched.
- The orphan `skills.migrate.<pid>` sweep sits before the symlink test, not at the top of the symlink branch as the reviewer suggested: after an interrupted conversion the path is no longer a symlink, so a sweep inside that branch would never run for the orphan it targets.
- The escape-hatch nit is answered with a README sentence (the pin is unconditional and wins over a hand-set `skillsDir`; no opt-out knob), not an `ALISSA_SKILLS_DIR_PIN=0` variable: with `CLAUDE_CONFIG_DIR` set any other value is a place Claude does not read, so an opt-out would only reproduce the issue's failure.
- The conversion WARN is split into three per-failure lines (copy failed / rm failed / mv failed), each stating what is on disk at that point.

- [ ] The three conversion-failure log lines are not exercised by a test: none of the failures can be provoked against the stubbed boot without a foreign-owned fixture.
- [ ] The prune on a live executor volume: tested with a fixture `installed` map, not the `${WORKSPACE_ROOT}/.alissa-config/config.json` of a deployed executor. Expected first-boot log: `dropped N installed-skill record(s) …` followed by the CLI reinstalling in step 3c.
- [ ] Everything under the round-0 checklist above still stands (ownership after the real gosu drop, live-volume symlink migration, acceptance criterion 3 on a rebuilt image, CLI honouring `skillsDir`).

#### Verification

- `bash docker/claude/tests-entrypoint-config.sh`: ALL PASS, exit 0 — the `#132` section now carries 50 assertions (11 new: no-drop on a clean config, orphan sweep, prune with present/absent/dir-without-SKILL.md slugs, non-object `installed` preserved).
- `bash docker/claude/tests-entrypoint-{ui,identity,auth,executor,prune,chown}.sh`: all six exit 0, no FAIL line.
- `bash -n` on `entrypoint.sh` and `tests-entrypoint-config.sh`: clean.
- Python style/type/unit suites not run: no Python source changed (the python is a heredoc inside the shell entrypoint). Docker image contract not runnable here (no Docker); left to CI on the new head.

#### Scope touched

- `docker/claude/entrypoint.sh`: step 3a-ii only (header comment, sweep line, split WARN branches, prune in the python block).
- `docker/claude/tests-entrypoint-config.sh`: `#132` section only (three assertions added to existing cases, one new case with two boots).
- `docker/claude/README.md`: the claude-login paragraph and the `2b.` step-list line. `README.md`: one clause in the container paragraph.
- No excursion outside the issue's declared scope.


alissa-app:
- 2026-09-16 02:22:28 UTC — `review-alissa-dark-factory-revloop-pr133-r1-68ed42` — round 1 of 22 — spawned
- 2026-09-16 02:45:15 UTC — `review-alissa-dark-factory-revloop-pr133-r2-fe16ec` — round 2 of 22 — spawned

docker/claude/entrypoint.sh
```diff
@@ -954,6 +954,91 @@ for t in settings_targets:
...
+if os.path.exists(path):
+    try:
+        d = json.load(open(path))
+    except Exception as e:
+        print(f"[entrypoint] WARN: {path} is not readable JSON ({e}) — not pinning skillsDir over it", file=sys.stderr)
+        sys.exit(1)
+    if not isinstance(d, dict):
+        print(f"[entrypoint] WARN: {path} is not a JSON object — not pinning skillsDir over it", file=sys.stderr)
+        sys.exit(1)
+d["skillsDir"] = skills_dir
```
alissa-app, line 1044: [major] **Pinning `skillsDir` does not make the skills appear in the new directory.**

The CLI gates installation on this (CLI 0.2.1, the version this image pulls — `~/.alissa/cli.mjs:21896`):

```js
function isSkillInstalled(slug) {
  if (slug in installedSkills()) return true;              // config `installed` map wins
  return existsSync2(join2(getSkillsDir(), slug, "SKILL.md"));
}
```

`installedSkills()` is `loadConfig().installed`, and `runPrechecks()` (`cli.mjs:31903-31904`) filters the required skills **plus the workspace manifest's `skills:`** through it:

```js
const required = [...new Set([...loadConfig().requiredSkills ?? DEFAULT_REQUIRED_SKILLS, ...workspaceSkills])];
const missing = required.filter((s) => !isSkillInstalled(s));
```

So a slug recorded in `installed` is treated as present **without the filesystem ever being consulted**. Moving `skillsDir` to a fresh empty directory does not re-arm the install.

**Where this bites.** In the executor role `ALISSA_CONFIG_DIR` is `${WORKSPACE_ROOT}/.alissa-config` — on the persistent volume, deliberately (`entrypoint.sh:130`, "Point the whole config dir at the persisted volume"). Before this PR those skills installed into the ephemeral `~/.claude/skills` while the `installed` map was written to the persisted `config.json`. So on the first boot that applies this pin:

1. `${WORKSPACE_ROOT}/.alissa-config/config.json` survives the redeploy still listing `alissa-code-review`, `alissa-code-workspace`, … under `installed`;
2. this step creates `$CLAUDE_CONFIG_DIR/skills` **empty** (`:1010`) — the symlink branch above only helps a container that actually carries the hand-placed stop-gap link;
3. `alissa code` computes `missing = []` and installs nothing;
4. the session opens with `Unknown skill` — the exact symptom the issue is about.

It does not self-heal on any later boot either: nothing ever clears `installed`, so the state is sticky.

The daemon role is fine — its config dir is the ephemeral `${HOME}/.config/alissa`, so `installed` is empty at pin time — and the live daemon volumes are covered because the migration copies the link target's contents in. It is the executor service, on this same shared code path, that is left behind, and acceptance criterion 3 ("first session on the rebuilt image loads `Skill(...)` without 'Unknown skill'") is what is at risk.

**Suggested fix**, in scope and still not a CLI change — a few lines in this same python block, right after `d["skillsDir"] = skills_dir`:

```python
# A slug recorded as installed but absent from the newly pinned dir would
# never be reinstalled: isSkillInstalled() trusts this map over the disk.
installed = d.get("installed")
if isinstance(installed, dict):
    d["installed"] = {
        slug: meta for slug, meta in installed.items()
        if os.path.exists(os.path.join(skills_dir, slug, "SKILL.md"))
    }
```

Equivalent alternative: when the pinned directory is created empty and `~/.claude/skills` is populated, copy it in — the same move the symlink branch already makes, generalized to the no-symlink case. Either way it deserves a test case alongside the five already added: `installed` carries a slug, the pinned dir is empty, and the boot leaves the CLI able to see it as missing.

docker/claude/entrypoint.sh
```diff
@@ -954,6 +954,91 @@ for t in settings_targets:
...
+      # would hide every installed skill.
+      SKILLS_COPY_OK=1
+      if [ -n "${SKILLS_LINK_TARGET}" ] && [ -d "${SKILLS_LINK_TARGET}" ]; then
+        cp -a "${SKILLS_LINK_TARGET}/." "${SKILLS_MIGRATE_TMP}/" || SKILLS_COPY_OK=0
+      fi
+      if [ "${SKILLS_COPY_OK}" = "1" ] && rm -f "${CLAUDE_SKILLS_DIR}" && mv "${SKILLS_MIGRATE_TMP}" "${CLAUDE_SKILLS_DIR}"; then
+        log "converted the stop-gap symlink ${CLAUDE_SKILLS_DIR} -> ${SKILLS_LINK_TARGET:-<dangling>} into a real directory (contents copied)"
+      else
+        rm -rf "${SKILLS_MIGRATE_TMP}"
+        log "WARN: could not convert the symlink ${CLAUDE_SKILLS_DIR} -> ${SKILLS_LINK_TARGET:-<dangling>} into a real directory — leaving the link in place; the alissa CLI will install through it"
```
alissa-app: [nit] This WARN can state the opposite of what happened. The guard is `[ "${SKILLS_COPY_OK}" = "1" ] && rm -f "${CLAUDE_SKILLS_DIR}" && mv …`, so it is also reached when the `rm -f` succeeded and only the `mv` failed — the link is gone at that point, yet the line says "leaving the link in place; the alissa CLI will install through it".

Consequences are mild: the staged copy is discarded, the `mkdir -p` two lines down recreates the path as an empty real directory, and the original link target was never touched — so nothing is lost, the log is just wrong about it. The window is also narrow, since both operations act on the same directory. Splitting the condition so the "link removed but not replaced" case logs its own line would make a 3am reading of this log trustworthy.

docker/claude/entrypoint.sh
```diff
@@ -954,6 +954,91 @@ for t in settings_targets:
...
+# Runs after the privilege drop, so the directory is created by (and owned by)
+# ${RUNTIME_USER}. Never fatal: a skills dir that cannot be pinned is a slower
+# first session, not a dead daemon.
+# -----------------------------------------------------------------------------
+CLAUDE_SKILLS_DIR=""
+if [ -n "${CLAUDE_CONFIG_DIR:-}" ] && [ -n "${CLAUDE_CONFIG_DIR//[[:space:]]/}" ]; then
+  CLAUDE_SKILLS_DIR="${CLAUDE_CONFIG_DIR}/skills"
+  if [ -L "${CLAUDE_SKILLS_DIR}" ]; then
+    SKILLS_LINK_TARGET="$(readlink -f "${CLAUDE_SKILLS_DIR}" 2>/dev/null || true)"
+    SKILLS_MIGRATE_TMP="${CLAUDE_SKILLS_DIR}.migrate.$$"
```
alissa-app, line 1002: [nit] A boot killed between the `rm -f` of the link and the `mv` leaves `${CLAUDE_CONFIG_DIR}/skills.migrate.<pid>` behind, and nothing ever sweeps it — this path is the persistent volume, so the orphan is permanent and a future interrupted boot adds another under a different pid. The next boot does recover correctly (the path is no longer a symlink, so `mkdir -p` just recreates it), which is what keeps this a nit rather than a real leak.

A `rm -rf "${CLAUDE_SKILLS_DIR}".migrate.* 2>/dev/null || true` at the top of the symlink branch would cost nothing and keep the volume tidy.

docker/claude/entrypoint.sh
```diff
@@ -954,6 +954,91 @@ for t in settings_targets:
...
+tmp = path + ".tmp"
+with open(tmp, "w") as f:
+    json.dump(d, f, indent=2)
+    f.write("\n")
+if os.path.exists(path):  # keep whatever mode the CLI chose for its token file
+    os.chmod(tmp, os.stat(path).st_mode & 0o777)
+os.replace(tmp, path)
+PY
+  then
+    log "skills dir pinned to ${CLAUDE_SKILLS_DIR} (Claude reads personal skills there when CLAUDE_CONFIG_DIR is set)"
```
alissa-app, line 1065: [nit] No escape hatch: a `skillsDir` an operator set deliberately (`alissa config set skillsDir …`) is silently overwritten on every boot, since the pin is unconditional whenever `CLAUDE_CONFIG_DIR` is non-blank.

This is in spec — the issue asks for "set", and the PR body argues the case for overwriting a stale value, which I agree with as the default. It is worth flagging only because every other knob in this entrypoint is a pass-through env variable with an operator override, so this one quietly breaks the pattern. Either a one-line note in the README's claude-login paragraph saying the pin always wins over a hand-set value, or an `ALISSA_SKILLS_DIR_PIN=0` bail-out, would close the gap. Your call.

alissa-app (changes requested):

Reviewed head **cbc82a8**. CI at that sha: **11 contexts, all concluded, none failed** (`[code]smith` skipped; the legacy `/status` endpoint reports `total_count: 0`, i.e. nothing to judge there). I also ran all seven entrypoint suites locally on the PR head — `config`, `ui`, `identity`, `auth`, `executor`, `prune`, `chown` — every one exits 0 with no FAIL line, the `#132` section contributing 39 assertions.

The pin itself is well built: it lands at step 3a-ii (`entrypoint.sh:958`), after the gosu drop so the directory is created by the runtime user, after `alissa auth login` has written the config, and before both `alissa code workspace sync` (`:1088`) and the worker start — the ordering the issue requires. I verified independently that `${ALISSA_CONFIG_DIR}` (`:419`, default `${HOME}/.config/alissa`) is genuinely the file the CLI reads (`alissa config path` → `/home/alissa/.config/alissa/config.json`, and the CLI does honour an exported `ALISSA_CONFIG_DIR`), and that Claude really does read `$CLAUDE_CONFIG_DIR/skills` — this very reviewer session loaded its skills through the hand-placed stop-gap symlink. Refusing to overwrite an unparseable `config.json` is a genuine improvement over the 3a block above it, which would clobber it via `except: d = {}`.

One **[major]** keeps this from an approve: pinning `skillsDir` at a new empty directory does not, by itself, get the skills installed into it. Detail on the thread at `entrypoint.sh:1025`.

### Findings by severity

| severity | count |
| --- | --- |
| blocker | 0 |
| major | 1 |
| minor | 0 |
| nit | 3 |
| question | 0 |

### Dimension report

- **Correctness:** one `[major]` — the CLI's `isSkillInstalled()` short-circuits on the config's `installed` map before it checks the filesystem, so re-pointing `skillsDir` does not repopulate the new directory on the executor role, whose config dir is persistent. Everything else in the step checks out: the no-op path is exact (unset / empty / whitespace all tested), `readlink -f` + `[ -d ]` + `cp -a … /.` + copy-first/unlink-last is the right shape, and a dangling link degrades to an empty real directory.
- **Coding practices:** nothing above a nit. The step matches the file's idiom — numbered sub-step with a why-comment, load-then-update python heredoc, WARN-never-die — `bash -n` is clean on both shell files, and the `2b.` marker in the README step list is consistent with the `0b.` / `3c-ii.` / `3d.` / `4b.` markers already there. No CHANGELOG or version bump is owed: `check-version-bump.yaml` explicitly exempts `docker/` and repo-level `README.md`, and the four most recent entrypoint-only commits touch neither.
- **Security:** nothing found. The write targets the token-bearing `config.json` but preserves its mode when the file exists and otherwise creates it at the same default permissions the CLI itself uses (`docker/claude/README.md:975` already documents the token as cleartext with default permissions, so no regression); the replace is atomic via `os.replace`; no secret reaches a log line; and `cp -a` run as the runtime user fails closed on foreign-owned content, which correctly leaves the stop-gap link in place.
- **Performance:** nothing found. One `cp -a` of a small skills tree, and only while the path is still a symlink — after the first boot it is a real directory and the branch is skipped. No recursive walk of the volume, so the dentry/slab hazard that step 0 is guarded against is not reintroduced.
- **Side-effects / blast radius:** no schema, no migrations, no dependency, no deletion, no CI/deploy/workflow file touched. Two things an operator should know: the symlink→directory conversion is a one-way change to operator-placed state on a persistent volume, and the pin overwrites any deliberately operator-set `skillsDir` on every boot (nit at `:1036`). The step sits on the path **both** container roles share, so it changes the executor service too — which is exactly where the `[major]` bites.

Address the `[major]` at `entrypoint.sh:1025` (a few lines inside the same python block), or waive it with a written justification — e.g. that the executor service is known to carry the stop-gap symlink too, so its pinned directory is never created empty. The nits are yours to take or leave.

Per CR8, every finding above `[nit]` needs exactly one `[triage:*]` reply on its thread before re-requesting review.

vlissa-app:
- 2026-09-16 02:34:31 UTC — `fix-ali-fhr-alissa-dark-factory-revloop-pr133-r1-a1` — fix edge — spawned

docker/claude/entrypoint.sh
vlissa-app: [triage:pursue] Confirmed against the CLI the image pulls (0.2.1): `isSkillInstalled()` returns true on the `installed` map before touching the disk, and `installSkill()` writes into `getSkillsDir()` and rewrites the entry — so pruning the map is enough to re-arm the install in step 3c. Took the prune, not the copy-from-`~/.claude/skills` alternative: on the executor the home dir is ephemeral, so on the redeploy that first applies the pin there is nothing to copy — the prune is the only variant that fixes the sticky state. In the same python rewrite, right after `skillsDir` is set, every `installed` entry with no `SKILL.md` under the pinned dir is dropped and the dropped slugs are logged by name; a present entry keeps every field; an `installed` value that is not an object is left untouched. Two new boots in `tests-entrypoint-config.sh` cover this (three slugs: present / absent / dir-without-SKILL.md; and the non-object case), plus a no-drop assertion on the first boot. README paragraph and step list updated.
Fixed in 932c6fc.

docker/claude/entrypoint.sh
vlissa-app: [triage:pursue] Agreed, the combined condition made that WARN lie in the rm-succeeded-mv-failed case. Split into three branches, each logging what is actually on disk at that point: copy failed → link still in place; `rm -f` failed → link still in place; `mv` failed → "removed the symlink … but could not move the staged copy into place — recreating it as an empty directory; the alissa CLI will reinstall the skills there". Not separately tested: none of the three failures can be provoked against the stubbed boot without a foreign-owned fixture.
Fixed in 932c6fc.

docker/claude/entrypoint.sh
vlissa-app: [triage:pursue] Took the sweep, but not at the spot suggested: after a boot dies between the `rm -f` and the `mv`, the next boot no longer sees a symlink (the path is missing, then `mkdir -p` recreates it), so a sweep at the top of the symlink branch would never run for exactly the orphan it is meant for. It sits right after `CLAUDE_SKILLS_DIR` is computed, before the `-L` test, as `rm -rf "${CLAUDE_SKILLS_DIR}".migrate.* 2>/dev/null || true`, so every boot with `CLAUDE_CONFIG_DIR` set clears leftovers. The real-directory test case now plants a `skills.migrate.99999` and asserts it is gone after boot.
Fixed in 932c6fc.

docker/claude/entrypoint.sh
vlissa-app: [triage:pursue] Took the README note rather than an `ALISSA_SKILLS_DIR_PIN=0` knob: with `CLAUDE_CONFIG_DIR` set, any other `skillsDir` is by definition a place Claude does not look, so an opt-out would only re-create the `Unknown skill` failure the issue is about, and it would add a code path and a test for a value nobody should want. The claude-login paragraph in `docker/claude/README.md` now says the pin is unconditional, rewritten on every boot, wins over a hand-set `alissa config set skillsDir …`, and that there is no opt-out env knob — so the break from the pass-through pattern is at least documented.
Fixed in 932c6fc.

docker/claude/entrypoint.sh
```diff
@@ -954,6 +954,120 @@ for t in settings_targets:
...
+# every other key (the verified token, apiBase, ...) is preserved. A config file
+# that exists but does not parse is left alone (WARN), never overwritten.
+#
+# One more key IS touched: the CLI's `installed` map. isSkillInstalled() trusts
+# that map before it looks at the disk, so a slug recorded there but absent from
+# the newly pinned directory would never be reinstalled — the executor role
+# keeps its config dir on the persistent volume, so after a redeploy the map
+# still lists skills that were installed into the ephemeral ~/.claude/skills.
+# Entries whose SKILL.md is not under the pinned dir are dropped; the CLI then
+# sees them as missing and installs them in 3c, into the pinned dir.
```
alissa-app, line 979: [minor] **"installs them in 3c" / "reinstalled … on the same boot" is not what happens — nothing in this entrypoint installs skills at boot.**

Step `3c` in this file is `alissa code workspace sync` (`:1104`, and the docker README's step list spells it out: *"**`alissa code workspace sync`** — materialize the worktree hubs the manifest declares"*). I read its implementation in the CLI the image pulls (0.2.1, `cli.mjs:31667`): it resolves the workspace root, `hubify()`s each manifest repo, optionally refreshes READMEs, and runs the drift doctor. It never touches `installSkill`, `isSkillInstalled`, or `getSkillsDir` — `workspace sync` installs **no** skills, and neither does any other boot step.

The only thing that installs a skill is `runPrechecks()` (`cli.mjs:31897-31916`), reached from `alissa code` — i.e. at the **session handoff**, not at boot:

```js
const required = [...new Set([...loadConfig().requiredSkills ?? DEFAULT_REQUIRED_SKILLS, ...workspaceSkills])];
const missing = required.filter((s) => !isSkillInstalled(s));
if (missing.length > 0) { /* … installSkill(client, slug) for each … */ }
return startSession(...);
```

**This does not break the fix.** `alissa code -y --handoff claude` (the wrapper this image's `agents.yaml` deliberately keeps, `:674`) auto-installs without prompting, and prechecks complete *before* claude is launched, so the skill is on disk under the pinned dir before the session reads it — acceptance criterion 3 still holds. That is why this is `[minor]` and not above: the behaviour is right, the description of it is wrong.

What it costs is the operator, in exactly the situation this PR exists for. Four places say the reinstall happens at boot:

- `entrypoint.sh:979` — "the CLI then sees them as missing and installs them in 3c, into the pinned dir"
- `entrypoint.sh:967` — "the pin is written here, before 3c installs anything"
- `docker/claude/README.md:302` — "once — before `alissa code workspace sync` installs the manifest's `skills:`"
- `docker/claude/README.md:311` — "dropped entries are logged by slug and reinstalled by the CLI on the same boot"

(the step-list line `2b.` — "so the skills step 3 installs land where Claude reads them" — and the PR body's own operator-gate bullet, "followed by the CLI reinstalling in step 3c", carry it too.)

An operator who redeploys, watches the boot log print `dropped 4 installed-skill record(s) …`, then finds `$CLAUDE_CONFIG_DIR/skills` still empty when the boot finishes has been told by all four that something failed. It didn't — the directory fills on the first `alissa code`, which on the daemon is the first review spawn and on the executor the first job. That is a materially different thing to wait for, and the gap is the window in which someone reaches for `alissa skill install` by hand.

Suggested wording, keeping it to the two load-bearing spots: at `:978-979`, *"Entries whose SKILL.md is not under the pinned dir are dropped; the CLI's prechecks then see them as missing and install them into the pinned dir at the first `alissa code` handoff (not at boot — no boot step installs skills)."* And in the README, *"reinstalled by the CLI at the first session handoff"* instead of *"on the same boot"*, with the same correction to the `before alissa code workspace sync installs the manifest's skills:` clause.

Fine as a follow-up task rather than another round — it is comment and README prose only, and the shipped behaviour is correct as it stands.

alissa-app (approved):
## Round 2 — approve

All four round-1 findings were pursued, and all four fixes hold at `932c6fc`. One new `[minor]` below, on comment/README prose only; nothing shipped is wrong. Verdict: **approve**.

### Round-1 reconciliation

| --- | --- | --- |
| `[major]` pinning `skillsDir` does not make the skills appear in the new dir | pursue | **fixed** (`932c6fc`) |
| `[nit]` conversion WARN can state the opposite of what happened | pursue | **fixed** (`932c6fc`) |
| `[nit]` orphaned `skills.migrate.<pid>` is never swept | pursue | **fixed** (`932c6fc`) |
| `[nit]` no escape hatch for a hand-set `skillsDir` | pursue | **fixed** (`932c6fc`, doc route) |

**The `[major]`.** The prune is the right variant and I agree with the reasoning for choosing it over copying `~/.claude/skills` in — on the executor the home dir is ephemeral, so on the redeploy that first applies the pin there is nothing to copy, and only the prune un-sticks the map. Re-verified against the CLI the image pulls (0.2.1): `isSkillInstalled()` returns true on `slug in installedSkills()` before `existsSync2()` is reached (`cli.mjs:21896`), `installSkill()` writes into `getSkillsDir()` and rewrites `installed[slug]` (`:21869-21882`), and `getSkillsDir()` reads `loadConfig().skillsDir` (`:21762`). So dropping the record does re-arm the install, and the reinstall lands in the pinned dir.

Two properties of the placement I checked rather than assumed, both good:

- The prune runs **after** the symlink conversion has copied the link target in, so a container carrying the hand-placed stop-gap keeps its records instead of pointlessly re-downloading.
- If the conversion *failed* and the link is still in place, `os.path.exists(skills_dir/slug/SKILL.md)` follows it into the home dir, so the records are kept and the CLI keeps installing through the stop-gap. The prune degrades in the same direction as the rest of the step.

I also checked the one regression this could have introduced: pruning makes `missing` non-empty on the first executor boot, and `runPrechecks()` prompts `Install them now?` when it is. It cannot hang — this image's `agents.yaml` deliberately routes through `alissa code -y --handoff claude` (`entrypoint.sh:674`), and `-y` is `autoYes`. The daemon role is unaffected either way (ephemeral config dir, empty map at pin time).

**The three nits.** The WARN split is correct per branch: copy-failed and `rm`-failed both leave the link on disk and say so; the `mv`-failed line now says the link was removed and the path is being recreated empty, which is what `mkdir -p` at `:1029` then does. The sweep placement is better than what I suggested — after an interrupted conversion the path is no longer a symlink, so a sweep inside the `-L` branch would never fire for the orphan it targets; at `:999` it does, and it runs before `SKILLS_MIGRATE_TMP` is created, so it cannot eat its own staging dir. The escape-hatch answer is reasoned and I accept it: with `CLAUDE_CONFIG_DIR` set, any other `skillsDir` is a place Claude does not read, so an opt-out would only re-create the failure the issue is about — and it is now documented as unconditional.

### Rubric

- **Correctness:** one `[minor]` (`entrypoint.sh:979`) — the boot sequence is described as installing the dropped skills "in 3c" / "on the same boot"; `workspace sync` installs no skills, `runPrechecks()` does, at the first `alissa code`. The behaviour is right; only the description is wrong. Everything else in the delta checks out — prune keys and placement, the three WARN branches against actual disk state, the sweep, and the `isinstance(installed, dict)` guard that leaves a non-object value alone.
- **Coding practices:** nothing found — the additions stay inside step 3a-ii, match the file's idiom (numbered sub-step, load-then-update python heredoc, WARN-never-die), and `bash -n` is clean on both shell files. The new test case follows the existing fixture style and asserts the log line by name, not by count alone.
- **Security:** nothing found — the config rewrite still preserves the file mode, still refuses an unparseable file rather than clobbering it, and the new log line prints slugs, never values from the config. The `rm -rf` sweep is anchored to `"${CLAUDE_SKILLS_DIR}".migrate.*` inside the non-blank `CLAUDE_CONFIG_DIR` guard, so the variable cannot be empty when it runs; a non-matching glob is a literal path that `rm -f` exits 0 on.
- **Performance:** nothing found — one extra glob and one `stat` per `installed` entry, on a map of single digits.
- **Side-effects / blast radius:** no schema, migrations, dependencies, deletions, or CI/workflow files; the whole diff is the container entrypoint, its test suite, and two READMEs. Three operator-visible effects, all on the persistent volume and all at every boot: the one-way symlink→directory conversion (round 1), the new unconditional `rm -rf` of `skills.migrate.*`, and the new deletion of keys from the persisted `installed` map. All three are self-healing — the CLI reinstalls what the prune drops — but they are deletions on operator state, which is what the merge-readiness line below reflects. One thing I looked at and am *not* raising: the sweep is a wildcard, so two containers sharing this volume and booting simultaneously could have one delete the other's in-flight staging dir; the loser logs the `mv`-failed WARN and recovers to an empty directory the CLI refills, and no deployment in this repo defines that shape.

### CI

Read at the reviewed head `932c6fc`, not at the PR: 11 contexts, all `completed`, 10 `success` and `[code]smith` `skipped`, none failed. The legacy `/status` endpoint answers `state: pending` with `total_count: 0` — no legacy contexts exist for this sha, judged on `check-runs`. I also ran `bash docker/claude/tests-entrypoint-config.sh` against `932c6fc` in a throwaway worktree: **ALL PASS**, including all six new assertions (orphan sweep, the no-drop case, the three-slug prune, the non-object `installed`).

PR #135, opened 2026-09-16
TASK-1105449672: session-posted verdicts carry the Merge-Readiness trailer — directives require it, daemon warns on a trailer-less approve (issue #134)

### LINKED ISSUE #134: revloop: session-posted verdicts carry the Merge-Readiness trailer too — directives require the native review body to end with the bare `Merge-Readiness:` line equal to the envelope's, the daemon warns (log + activity row) when a reviewer-identity APPROVE on the current head lacks a parseable trailer, README trailer section names both paths — next free patch after 0.31.0
## Why

PR #131 (0.31.0, issue #130) made `_close_round_natively` carry the envelope's Merge-Readiness judgment onto the native review as a line-anchored trailer (`NATIVE_VERDICT_BODY` `{readiness}` slot, `READINESS_TRAILER_LABEL`, `READINESS_MISSING_REASON`; `loop.py:392-417`). That path runs only when the envelope count is ahead of the native count (`loop.py` ~`:2170-2200`, `owed > native`) — i.e. when the session did NOT submit its own review. The module's own comment beside `_gate_on_checks` says it: the session posting the verdict "is the normal path: the daemon's native post exists precisely for the rounds where the session did not submit one." On the normal path the review body is the session's and passes through nothing, so the trailer never lands. Observed 2026-09-16 on ali-fhr/studio.alissa.app#1258 (round 2 approve 03:36 UTC) and ali-fhr/alissa-dark-factory-orcloop#111 (approve 03:18 UTC), both after 0.31.0 was live: neither approve body has a parseable trailer (the studio one has the line inside backticks mid-sentence). The orcloop merge edge (orcloop PR #111) holds such approves as `no-trailer`, so with the edge on, the normal path never auto-merges.

The source fix is the skill (sibling task on ali-fhr/skills.alissa.app: the session writes the trailer as the last line of its own review). This task makes the daemon **say so** and **see it**: the directives carry the rule (the session reads the directive even when its skill copy is stale), and a session approve that lacks the trailer is reported instead of silently holding downstream.

## Changes

1. **Directives** — `_MERGE_READINESS_LINE` (`loop.py:651`), spliced into `ROUND_1_DIRECTIVE` (`:931`) and `ROUND_K_DIRECTIVE` (`:949`): extend to say the session's OWN native review (every `gh pr review` form and the reviews-API POST) must END with the bare line `Merge-Readiness: auto` or `Merge-Readiness: operator — <one-line reason>` as its last non-empty line, byte-equal in value and reason to the envelope's line; `auto` only on an APPROVE of the reviewed head; the daemon copies the envelope's line only when it posts the verdict itself. Keep the grammar sentence identical to the README's. `test_directives.py`: both directives contain `Merge-Readiness:` and the phrase "last non-empty line"; existing invariants (no stray braces, credential routing, head+checks gate) hold.
2. **Observe** — where the daemon stamps every reviewer-identity review it sees on the current head into the `verdicts` ledger (the 0.30.1 round-admission read, `README.md:~584`): when such a review is an APPROVE by the reviewer login on the current head and its body has no line matching the pinned grammar (`alissa.py`'s `_READINESS_RE`'s bare-line sibling — add `_TRAILER_RE` next to it, one regex shared with `_close_round_natively`'s emitter so the two cannot drift), log ONE `WARNING` per (PR, head): `<repo>#<n> approve at <head7> by <login> carries no Merge-Readiness trailer — the merge edge will hold it; the session must end its review body with the line (see directive)`, and add one row to the activity comment (`ACTIVITY_HEADER` renderer) reading `readiness=missing` for that round. Never post a review, never edit the session's review, never re-run the round: this is observation only. A parseable trailer logs the existing `readiness=auto|operator` term. Once per (PR, head) — a ledger flag beside the `verdicts` row, not a new table.
3. **README** — the Merge-Readiness section (`README.md:~350-365`): state the two paths explicitly (session-posted: the session writes the trailer per the skill, the daemon verifies and warns; daemon-posted: the daemon copies the envelope's line), the WARNING's wording, and that a missing trailer is a hold on the merge edge, not a review failure. `CHANGELOG.md`: a new top entry.
4. **Version** — `alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version` → the next free patch above whatever `main` and PyPI hold when you branch (0.31.1 at the time of writing; the `Check Version Bump` workflow is the judge).

## Tests (`test_loop.py`, existing fakes `FakeGitHub.submit_review`, `make_pr`, `review`, `watcher`; `test_directives.py`)
- A session-posted APPROVE on the current head whose body ends with `Merge-Readiness: auto` → no warning, `readiness=auto` in the round-close/observe log line and the activity row.
- Same with `Merge-Readiness: operator — touches convex/schema.ts` → `readiness=operator`.
- Same with the line inside backticks mid-sentence (the studio#1258 shape) → WARNING once, activity row `readiness=missing`; a second poll on the same head logs nothing new; a new head re-arms.
- A REQUEST_CHANGES review without a trailer → no warning (approve events only).
- The daemon-posted path (`_close_round_natively`) is unchanged: existing 0.31.0 tests green, and the emitter and `_TRAILER_RE` agree (a test renders the emitter's output for `auto` and for a reason and asserts the regex matches it, and rejects the backticked and bold variants).
- `test_directives.py` as in change 1.

## Out of scope
The skill text (sibling task on skills.alissa.app); the orcloop consumer (PR #111 / a follow-up there may tolerate the backticked form); posting or editing reviews on the session's behalf; any config key or env var.

## Definition of Done

PR ready-for-review and converged through the review loop, PR URL attached as evidence, version bumped to the next free patch: both round directives require the session's own native review to end with the bare `Merge-Readiness:` line equal to the envelope's; a reviewer-identity APPROVE on the current head with no parseable trailer produces one WARNING per (PR, head) and a `readiness=missing` activity row, with one shared regex between the emitter and the check; README and CHANGELOG updated; the 0.31.0 daemon-posted path unchanged; unit tests, style, types, wheel and version-bump checks green.

## Acceptance criteria

- [ ] ROUND_1_DIRECTIVE and ROUND_K_DIRECTIVE tell the session to end its own native review body with the bare `Merge-Readiness:` line as the last non-empty line, equal to the envelope's; test_directives pins it and its existing invariants hold.
- [ ] A reviewer-identity APPROVE on the current head without a parseable trailer logs one WARNING per (PR, head) and an activity row `readiness=missing`; a trailer-bearing approve logs readiness=auto|operator; REQUEST_CHANGES never warns; the emitter's output matches the shared regex and the backticked/bold variants do not; the daemon-posted path's 0.31.0 tests are unchanged and green.
- [ ] Version is the next free patch, CI green on the PR head (unit, style, types, wheel, version-bump, image contract); README Merge-Readiness section names both paths and the WARNING; CHANGELOG entry present. Operator check after the fleet re-pins: a session approve on any ali-fhr PR shows readiness=auto|operator (or the WARNING) in the revloop log.

Autonomous-Scope: alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/alissa.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py, alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/version, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_loop.py, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_directives.py, alissa-tools-github-revloop/src/test/test_alissa/test_tools/test_github/test_revloop/test_state.py, README.md, CHANGELOG.md
Alissa-Task: TASK-1672903146

vlissa-app:
Closes #134

Origin task: TASK-1672903146
Implementation task: TASK-1105449672

## Summary

- **Directives** (`ROUND_1_DIRECTIVE`, `ROUND_K_DIRECTIVE`): the session's OWN native review (every `gh pr review` form and the reviews-API POST) must END with the bare `Merge-Readiness: auto` or `Merge-Readiness: operator — <one-line reason>` line as its last non-empty line, byte-equal to the envelope's; `auto` only on an APPROVE of the reviewed head; the daemon copies the envelope's line only when it posts the verdict itself.
- **Observe**: the first poll that sees a reviewer-identity APPROVE on the current head (session-posted, not the daemon's own) reads the body with the shared trailer regex. Parseable → `readiness=auto|operator` log line + activity row. Missing → one `WARNING` per (PR, head) (`<repo>#<n> approve at <head7> by <login> carries no Merge-Readiness trailer — the merge edge will hold it; the session must end its review body with the line (see directive)`) + a `readiness=missing` activity row. Never posts, never edits, never re-runs. `REQUEST_CHANGES` never warns. Once-per-head guard is a nullable `readiness` column on the `verdicts` ledger row.
- **One regex**: `_TRAILER_RE` / `parse_trailer` in `alissa.py`, built from `READINESS_TRAILER_LABEL`, which the 0.31.0 emitter also builds from; a test renders the emitter's output for every envelope shape and asserts the regex reads it back, and rejects the backticked/bold/bulleted forms.
- README Merge-Readiness section names both paths and the WARNING; CHANGELOG 0.31.1; version 0.31.1.

### Judgment calls
- **Where the observation hooks in.** The issue points at the 0.30.1 round-admission read (`_last_verdict_at`), but that read only runs from `_admit_round`, which sits downstream of convergence — and a session APPROVE on the current head converges the PR on that same pass, so the read is never reached for exactly the review this task is about. The check (`_observe_session_readiness`) runs in `evaluate` after the round counts are known and before the native-post/convergence branches; it decides nothing and reports only.
- **Any matching line, not strictly the last one.** Change 2 of the issue says "its body has no line matching the pinned grammar", so the check is `_TRAILER_RE.search` (first line-anchored match wins, like the consumer's grammar). The directive tells the session to make it the last non-empty line; the daemon does not fail a body whose bare line sits elsewhere.
- **The daemon's own native approve is skipped by the observer.** A review with a verdict marker (`verdict_round is not None`) is emitter-built and already logged `readiness=` plus an activity row at post time; observing it again on the next poll would duplicate the row. So the observer scopes to session-posted reviews only.
- **A parseable trailer also gets an activity row** (`readiness=auto|operator`), not only a log line — the issue's test list asks for "the round-close/observe log line and the activity row" for both the auto and operator cases.
- **Row wording.** The session-posted rows read `readiness=missing` / `readiness=auto` / `readiness=operator — <reason>` literally (the issue's wording), whereas the 0.31.0 round-close row reads ``merge-readiness: `…` ``. Left the 0.31.0 row untouched.
- **Ledger flag = nullable `readiness` column on `verdicts`**, migrated via `_ADDED_COLUMNS`; `note_readiness` inserts the verdict row if GitHub's list showed the review before the ledger did. Telemetry-absorbed like `note_observed_verdict`: an unwritable ledger costs a repeated WARNING on the next poll, not a decision.
- **Constants moved.** `READINESS_TRAILER_LABEL` now lives in `alissa.py` next to `_TRAILER_RE`/`parse_trailer` (the regex is built from the label) and is re-imported by `loop.py`; `READINESS_MISSING = "missing"` added beside `READINESS_AUTO`/`READINESS_OPERATOR`.
- **CRLF bodies** (a review typed into GitHub's web form) are normalised to LF before the regex runs; the pinned grammar itself is unchanged.
- **Version 0.31.1**: `main` and PyPI both hold 0.31.0 at branch time.

- [ ] `tests-image-contract.sh` / `check-image.yaml` — needs Docker, which this hub machine does not have; left to CI.
- [ ] `check-version-bump.yaml` and `check-wheel.yaml` on the PR head — only CI runs them; the wheel builds locally as `alissa_tools_github_revloop-0.31.1`.
- [ ] Post-fleet-re-pin operator check from the acceptance criteria: a session approve on any ali-fhr PR shows `readiness=auto|operator` (or the WARNING) in the revloop log. Not reproducible here.
- [ ] The skill-side fix (sibling task on skills.alissa.app) is out of scope and not touched; until it lands, session approves will keep warning.

### Verification
- `pytest src/test` in a private venv (editable install): **1200 passed** (1167 pre-existing + 33 new), all 0.31.0 daemon-posted-path tests unchanged and green.
- `pycodestyle . --max-line-length 120`: clean.
- `mypy . --ignore-missing-imports`: `Success: no issues found in 38 source files`.
- `python -m build --wheel`: built `0.31.1` wheel.
- Not run: entrypoint suites (`docker/claude/tests-entrypoint-*.sh`, no entrypoint files touched) and the image contract (no Docker).

### Scope touched
- `alissa.py` — `READINESS_TRAILER_LABEL`, `READINESS_MISSING`, `_TRAILER_RE`, `parse_trailer`.
- `loop.py` — `_MERGE_READINESS_LINE` (both directives), `_observe_session_readiness` + its call in `evaluate`, label import.
- `state.py` — `verdicts.readiness` column (schema + `_ADDED_COLUMNS`), `readiness_observed`, `note_readiness`, `read_verdicts` includes the column.
- `version` → 0.31.1; `README.md` (Merge-Readiness section: two paths, WARNING wording, hold-not-failure; one sentence in the round-admission section); `CHANGELOG.md` 0.31.1 entry.
- Tests: `test_loop.py`, `test_directives.py`, `test_state.py`.
- No excursions outside the issue's declared scope. `docker/claude/Dockerfile`'s `REVLOOP_VERSION` pin deliberately untouched.

#### Judgment calls
- **`[major]` ledger-flag ordering / dry-run — pursued.** `_observe_session_readiness` now emits first and records second: `note_readiness` runs only after `_append_activity` returned True, and never when `config.dry_run` is set (explicit guard, even though `_append_activity` already returns False in dry-run — the guard states the rule where the next reader looks). Under dry-run the WARNING therefore repeats on every poll; that is the same behaviour `_note_checks_hold` has in dry-run and is what the `_warn_identity_drift` docstring asks for ("a fresh dry-run always announces"). I did not add an in-memory dry-run gate for it.
- **`[minor]` newest-review scoping — pursued.** The observer now picks the newest `APPROVED` review on the current head rather than `my_reviews[-1]`, which is what the merge edge reads; a later `COMMENTED` write-up by the same login no longer hides a trailer-less approve. The README sentence was widened to say "the newest such approve" rather than narrowed.
- **`[minor]` synthetic `time.time()` stamp — pursued, one step further than suggested.** When `_epoch(submitted_at)` is None the observer skips the ledger write, as suggested — and also skips the activity row, because with no flag ever settable the row would be re-appended on every poll and the activity comment would grow without bound. The log line still fires each poll. The `note_readiness` docstring now says the row is a verdict row `last_verdict_at` reads and must carry the review's own stamp.
- **`[nit]` first-match vs last-line — answered, no change.** Agreed with the reviewer's own reading; `TASK-660326447` carries the last-match move for both seats.
- **Order of the two guards.** The stamp check sits before the activity append so the "no stamp" case posts nothing; the dry-run check sits after the append (which is a no-op in dry-run) so the shape matches the three sibling call sites the reviewer cited.

- [ ] Same as round 0: `tests-image-contract.sh` / `check-image.yaml` need Docker, absent on this hub machine; left to CI.
- [ ] `check-version-bump.yaml` / `check-wheel.yaml` on the new head — only CI runs them; version stays 0.31.1 (unreleased on `main` and PyPI at branch time).
- [ ] A real `--dry-run` pass over a live PR carrying a trailer-less session approve prints the WARNING and leaves `state.db` untouched — covered by a unit test, not reproduced against a live ledger here.

#### Verification
- `pytest src/test` in a private venv (editable install): **1203 passed** (1200 at round 0 + 3 net new; one existing test rewritten), all 0.31.0 daemon-posted-path tests unchanged and green.
- `pycodestyle . --max-line-length 120`: clean.
- `mypy . --ignore-missing-imports`: `Success: no issues found in 38 source files`.
- Not run: entrypoint suites (no entrypoint files touched) and the image contract (no Docker).

#### Scope touched
- `loop.py` — `_observe_session_readiness` only (scoping, emit-then-record, dry-run and stamp guards, docstring).
- `state.py` — `note_readiness` docstring only; no schema or query change.
- Tests: `test_loop.py` — the failed-post test rewritten to pin retry-next-poll; three new tests (dry-run, later comment review, unreadable stamp).
- `README.md` (Merge-Readiness section: newest-approve wording, flag-after-row and dry-run sentence), `CHANGELOG.md` (0.31.1 entry extended). No version change, no excursions.

---
<a href="https://app.blacksmith.sh/ali-fhr/codesmith/alissa-dark-factory-revloop/pr/135"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1792129209&installation_model_id=436981&pr_number=135&repository=ali-fhr%2Falissa-dark-factory-revloop&return_to=https%3A%2F%2Fgithub.com%2Fali-fhr%2Falissa-dark-factory-revloop%2Fpull%2F135&signature=5c31e0e10fbe821d529b220acf4047e3c2b6f6c4b5bb5cfe4312e591cae574a7"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>


alissa-app:
- 2026-09-16 05:43:53 UTC — `review-alissa-dark-factory-revloop-pr135-r1-511f26` — round 1 of 22 — spawned
- 2026-09-16 06:03:10 UTC — `review-alissa-dark-factory-revloop-pr135-r2-cbf12a` — round 2 of 22 — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -4042,6 +4070,77 @@ def _defer_stale_round(
...
+            or newest.verdict_round is not None
+        ):
+            return
+        if self.state.readiness_observed(pr.full_name, pr.number, pr.head_sha) is not None:
+            return
+
+        value, reason = parse_trailer(newest.body)
+        readiness = value if value is not None else READINESS_MISSING
+        posted_at = _epoch(newest.submitted_at)
+        self.state.note_readiness(
```
alissa-app: [major] The once-per-(PR, head) flag is written **before** the activity append, and unconditionally under `--dry-run`. Both halves are settled elsewhere in this file, against this ordering.

`_note_checks_hold` (`loop.py:3633`): *"Ledger row after the append, like every other activity note here, so a transient comment failure retries next poll instead of losing the line."* `_append_activity` catches broadly and returns `False` on a rate limit or a 5xx — and here that return is discarded, with the flag already set, so the `readiness=missing` row criterion `observe` asks for is lost permanently for that head.

The dry-run case is worse and fully deterministic. `_append_activity` opens with `if self.config.dry_run: ... return False`, so under dry-run the row can never be appended — but `note_readiness` still writes the flag, into the same production `state.db`. `_drift_gated` (`loop.py:~4370`) names exactly this failure: *"In `--dry-run` the SAME two gates apply, held in memory instead of in the ping ledger — which the run must not touch in either direction. Writing would let a diagnostic pass durably silence the daemon it was run to diagnose (`state_db` has no dry-run branch, so the rows land in the same `state.db` production reads...)."* One diagnostic pass over a PR whose session approve lacks the trailer permanently silences both the WARNING and the row for that head once the daemon goes live — the precise outcome issue #134 exists to prevent.

`test_a_missing_trailer_is_still_reported_when_the_activity_post_fails` pins the current order on the grounds that the WARNING survives; that rationale does not reach dry-run, where nothing survives but the flag.

Suggested shape, following the three sibling call sites (`:3639`, `:3962`, `:4052`):

```python
if not self._append_activity(pr, line):
    return          # retried next poll, like every other activity note
if not self.config.dry_run:
    self.state.note_readiness(...)
```

which also keeps the emit-then-record ordering for the parseable branch.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -4042,6 +4070,77 @@ def _defer_stale_round(
...
+
+        Observation only, by contract: never a review posted, never the
+        session's review edited, never a round re-run. Scoped to the newest
+        reviewer-identity review, APPROVED, on the current head, and NOT the
+        daemon's own post -- that one is emitter-built and already reported
+        `readiness=` at post time (a second row for it would be noise). Once
+        per (PR, head) via the ledger flag beside the verdicts row: a new
+        head re-arms, a re-poll of the same head is silent.
+        """
+        newest = my_reviews[-1] if my_reviews else None
```
alissa-app: [minor] Scoped to `my_reviews[-1]`, but the README this PR writes says *"the first poll that sees a reviewer-identity `APPROVE` on the current head"* — those differ. `my_reviews` is every **substantive** submitted review by the reviewer login, oldest first, so a later substantive non-approve record by the same login (a round-k reviewer that fell back to `--comment`, a follow-up write-up) makes `newest.state != "APPROVED"` and the approve is never observed — while the merge edge, which looks for the APPROVE rather than the newest review, still finds it and still holds it, silently. That is the exact silence this feature was built to end.

Zero-body inline-comment artifacts are already dropped by `my_reviews`, so the window is narrow, and `[-1]` is this module's convention (`_last_verdict_at`, `_convergence_reason`) — which is why this is a `[minor]` and not more. Either scope to the newest `APPROVED` review on the head, or narrow the README sentence to say "the newest reviewer-identity review, when it is an `APPROVE` on the current head". Fix here or carry it as a follow-up task.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -4042,6 +4070,77 @@ def _defer_stale_round(
...
+            return
+        if self.state.readiness_observed(pr.full_name, pr.number, pr.head_sha) is not None:
+            return
+
+        value, reason = parse_trailer(newest.body)
+        readiness = value if value is not None else READINESS_MISSING
+        posted_at = _epoch(newest.submitted_at)
+        self.state.note_readiness(
+            pr.full_name, pr.number, pr.head_sha,
+            int(posted_at if posted_at is not None else time.time()),
```
alissa-app: [minor] The `time.time()` fallback writes a synthetic timestamp into a column another decision reads. When `_epoch(newest.submitted_at)` is `None`, this row is inserted at wall-clock *now* rather than at the review's own time — and on a converging pass `_observe_session_readiness` is the **first** writer of that row, because `_last_verdict_at` (the other writer, which skips its write entirely when `_epoch` returns `None`) is reached only from `_admit_round`, downstream of convergence.

`last_verdict_at` then reads that row as "the newest verdict of record on this head landed *now*", which is what the post-verdict cooldown in `_admit_round` compares a fresh re-request against. So `note_readiness`'s docstring claim — *"nothing downstream reads it but the dedupe"* — is true of the `readiness` **column** but not of the **row** it creates.

An unparseable GitHub timestamp is close to unreachable in practice, hence `[minor]`. Cleanest fix is to not invent one: skip the ledger write when `posted_at is None` (the WARNING still fires, and the next poll retries), or say in the docstring that the row is a verdict row and the timestamp must therefore be the review's.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -644,15 +651,28 @@ def readiness_trailer(envelope: "VerdictEnvelope | None") -> str:
...
+# README's ("The `Merge-Readiness` trailer on a native approve"), verbatim.
 _MERGE_READINESS_LINE = (
     "Your verdict envelope MUST carry the skill's `- **Merge-Readiness:** auto | "
-    "operator — <reason>` line — the daemon copies it onto the native review as "
-    "the `Merge-Readiness:` trailer the merge edge reads, and an envelope "
-    "without the line posts as `operator`. "
+    "operator — <reason>` line, AND your OWN native review (every `gh pr review` "
+    "form and the reviews-API POST alike) MUST END with the bare line "
+    "`Merge-Readiness: auto` or `Merge-Readiness: operator — <one-line reason>` "
+    "as its last non-empty line — plain text at the start of the line, not in "
```
alissa-app, line 669: [nit] The directive now tells the session the line must be its *last* non-empty line, while `parse_trailer` takes the *first* line-anchored match. That is the right choice today — it matches the grammar the README publishes and the orcloop consumer's own first-match read, so the daemon reports what the merge edge will actually see — and the PR body calls it out explicitly. Noting it only so the next reader does not re-derive it: a session that quotes a bare `Merge-Readiness:` line earlier in its body would be reported on the quoted line. `TASK-660326447` already carries moving both seats to last-match, so nothing to do here.

alissa-app (changes requested):
**Review round 1 — `request_changes`** · reviewed head `58a0fbb` · origin `TASK-1672903146` · review task `TASK-1995696122`

The shape of this change is right and the argument in the PR body for *where* the observation hooks in (`evaluate`, above the native-post and convergence branches, because a session approve converges the PR on the same pass) is correct — I traced it and the 0.30.1 `_last_verdict_at` read the issue points at really is unreachable for exactly this review. The shared `_TRAILER_RE` / `parse_trailer` is the right call and the emitter↔regex agreement test is the right test. CI is green at `58a0fbb` (11 contexts: 10 success, 1 skipped).

One `[major]` holds the round, and it is a five-line fix.

`_observe_session_readiness` writes its once-per-(PR, head) ledger flag **before** the activity append, and with no `dry_run` branch. This module already answered both halves of that question, in its own words, a few hundred lines away — `_note_checks_hold`: *"Ledger row after the append, like every other activity note here, so a transient comment failure retries next poll instead of losing the line"*; and `_drift_gated`: *"In `--dry-run` the SAME two gates apply, held in memory instead of in the ping ledger — which the run must not touch in either direction. Writing would let a diagnostic pass durably silence the daemon it was run to diagnose."* A dry-run pass over a trailer-less approve now does precisely that: the flag lands in the production `state.db`, and the WARNING this feature exists to emit never fires for that head again.

`test_a_missing_trailer_is_still_reported_when_the_activity_post_fails` pins the current ordering deliberately, so I read this as a judgment call rather than an oversight — but it is a judgment call against a convention the module states verbatim three times, and the dry-run consequence is not covered by that test's rationale ("the WARNING does not depend on it") because under dry-run the row can never land at all.

Two `[minor]`s and one `[nit]` follow; none of them gate.

## Rubric

- **Correctness:** the directive change satisfies criterion `directives` and the observer satisfies `observe` on the paths the tests cover; `[major]` on the ledger-flag ordering / dry-run write, `[minor]` on the newest-review scoping vs. what the README promises. The daemon-posted 0.31.0 path is genuinely untouched and its tests are unchanged.
- **Coding practices:** repo idiom is otherwise matched closely (docstrings carry the reasoning, telemetry absorbed via `_write_telemetry`, migration through `_ADDED_COLUMNS`). The one divergence is the `[major]`, and it is a divergence from a rule this file writes down.
- **Security:** nothing found. Review bodies are filtered to the reviewer login before parsing, and the reason string passes through `clean_readiness_reason` (flattened, backticks stripped, 200 chars) before it reaches a log line or a GitHub comment, so it cannot fence the activity comment.
- **Performance:** nothing found. One indexed `SELECT` per qualifying poll and one activity-comment fetch once per head; the guard short-circuits before the DB read on every non-qualifying review.
- **Side-effects / blast radius:** the `verdicts.readiness` migration is additive and nullable (`ALTER TABLE ... ADD COLUMN`, no rewrite, old rows read as "not yet observed"); no config key, no contract change for the orcloop consumer, and `parse_trailer` matches the grammar the README already publishes byte for byte. `read_verdicts` gains a column — harmless for a positional consumer since it is read as a row mapping. The new writes into `verdicts` are the one thing that reaches beyond telemetry; see the second `[minor]`.

Shortest path back to approve: move the `note_readiness` call below a successful `_append_activity` (and skip it under `dry_run`), then re-request review.

Merge-Readiness: operator — verdict is request_changes

vlissa-app:
- 2026-09-16 05:53:24 UTC — `fix-ali-fhr-alissa-dark-factory-revloop-pr135-r1-a1` — fix edge — spawned

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
vlissa-app: [triage:pursue] Agreed on both halves, and the module precedent you cite is the right one. Fixed in 9e26050: `_observe_session_readiness` now emits first and records second — `note_readiness` runs only after `_append_activity` returned True, and is skipped outright under `config.dry_run` (explicit guard, following the three sibling call sites, even though `_append_activity` already returns False in dry-run). The pinning test is rewritten: a failed activity post keeps the WARNING and leaves the flag unset, the next poll lands the row and then flags, the poll after that is silent. A new dry-run test asserts two polls warn twice, no row, no flag, no verdict row, and that a live daemon on the same ledger still observes once. One consequence worth naming: under dry-run the WARNING repeats every poll, which matches `_note_checks_hold` in dry-run and the "a fresh dry-run always announces" rule in `_warn_identity_drift`; I did not add an in-memory dry-run gate.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
vlissa-app: [triage:pursue] Fixed in 9e26050. The observer now selects the newest `APPROVED` review on the current head (`approves[-1]`) instead of `my_reviews[-1]`, which is what the merge edge reads; the `verdict_round is not None` skip for the daemon's own post still applies to that pick. A new test puts a later substantive `COMMENTED` review by the same login after a trailer-less approve and asserts the approve is still observed. The README sentence was widened rather than narrowed ("reads the body of the newest such approve — the review the merge edge reads, whatever the identity posted after it"), and the docstring says the same.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
vlissa-app: [triage:pursue] Fixed in 9e26050, one step further than the suggested shape. When `_epoch(submitted_at)` is None the observer skips the ledger write, as you suggest — and also skips the activity row, because with no flag ever settable that row would be re-appended on every poll and the activity comment would grow without bound. The log line (WARNING or `readiness=` INFO) still fires each poll, so nothing is hidden. `note_readiness`'s docstring now says what you pointed out: the `readiness` column is dedupe-only, but the row is a verdict row that `last_verdict_at` reads for the cooldown, so `posted_at` must be the review's own GitHub stamp and the caller skips the write without one. A new test feeds an unreadable stamp and asserts two WARNINGs, no activity row, no readiness flag and no verdict row.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
vlissa-app: [triage:answer] Agreed, and thanks for writing the reasoning down where the next reader will find it: first-match is the consumer's grammar today, so the daemon reports what the merge edge will actually see, and `TASK-660326447` carries the last-match move for both seats. No change in this round.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/loop.py
```diff
@@ -4042,6 +4070,95 @@ def _defer_stale_round(
...
+            # No stamp to key a verdict row on; the log line above is the
+            # whole report for this poll, and the next poll says it again.
+            log.debug(
+                "%s: approve at %s has no readable submitted_at; readiness row "
+                "and flag skipped", pr.slug, head7,
+            )
+            return
+        if not self._append_activity(pr, line):
+            return  # retried next poll, like every other activity note
+        if self.config.dry_run:
```
alissa-app, line 4156: [nit] This guard is unreachable: `_append_activity` opens with `if self.config.dry_run: ... return False` (`loop.py:5199`), so the line above has already returned in every dry-run pass. Harmless as belt-and-braces, and I would not spend a round on it — but the rationale given for it in the PR body ("the shape matches the three sibling call sites the reviewer cited") is the one part that does not hold: none of `:3639`, `:3962` or `:4052` carries an explicit `dry_run` check; all three rely on exactly the `_append_activity` return that the line above already consumes.

So the choice is between deleting it and keeping it as a deliberate seatbelt against a future `_append_activity` that appends in dry-run. If you keep it, the honest comment is "not reachable today; states the rule where the next reader looks" rather than a claim of sibling precedent. Your call.

alissa-tools-github-revloop/src/main/alissa/tools/github/revloop/state.py
```diff
@@ -777,6 +788,51 @@ def note_observed_verdict(
...
+
+    def note_readiness(
+        self, repo: str, number: int, head_sha: str, posted_at: int,
+        readiness: str, review_url: str = "",
+    ) -> bool:
+        """TELEMETRY: flag the verdict row for `head_sha` at `posted_at` with
+        the readiness the loop observed on it, creating the row if GitHub's
+        list showed the review before the ledger did (issue #134). Absorbed
+        like note_observed_verdict: a flag the ledger cannot take costs one
+        repeated WARNING on the next poll, not a wrong decision -- nothing
```
alissa-app, line 812: [nit] This sentence is now slightly understated by the round-1 fix. With `note_readiness` moved after the append, a flag the ledger cannot take costs a repeated WARNING **and** a repeated activity row on the next poll — the row lands, the flag does not, and the next poll appends another one.

The conclusion the docstring draws is still correct (telemetry, not a wrong decision), and the shape is the same as `_note_checks_hold`, so there is nothing to fix in the code. It is only worth noting because the observer's *other* new branch — the unreadable-stamp skip at `loop.py:4148` — was added precisely to avoid an activity row re-appended behind an unsettable flag, and this path has that shape too, just gated behind a persistently failing ledger rather than a missing timestamp. One clause ("and a repeated activity row") would keep the two branches' reasoning consistent for the next reader.

alissa-app (approved):
**Round 2 of cap 22** — reviewed head `9e26050b0c6291c0d8e208dbd00954162cfd80e2`. Verdict: **approve**.

## Prior-finding reconciliation (round 1, head `58a0fbb`)

All four findings carry exactly one `[triage:*]` reply; the three `pursue` items are fixed in `9e26050` and I verified each against the shipped code, not the claim.

- **`[major]` ledger flag before the append, and unconditional under `--dry-run`** — *fixed*. `_observe_session_readiness` now logs, then `_append_activity`, then `note_readiness` (`loop.py:4154-4159`). A failed post returns before the flag, so the row retries next poll, matching `_note_checks_hold` at `:3639`. `--dry-run` writes nothing durable. `test_a_missing_trailer_is_still_reported_when_the_activity_post_fails` was rewritten to pin retry-next-poll rather than the old order, and `test_a_dry_run_pass_reports_the_missing_trailer_but_never_touches_the_ledger` asserts two polls warn twice with no row, no flag and no verdict row, then that a live daemon on the same ledger still observes once. That last assertion is the one that actually closes the finding — the silencing I described was cross-run, and the test reproduces it cross-run.
- **`[minor]` scoped to `my_reviews[-1]`** — *fixed*. The observer now filters to `APPROVED` reviews whose `commit_id` is the head and takes the newest (`loop.py:4126-4131`), which is what the merge edge reads; the `verdict_round is not None` skip still applies to that pick. `test_a_later_comment_review_by_the_reviewer_does_not_hide_the_approve` covers the case I named. The README was widened to "the newest such approve — the review the merge edge reads, whatever the identity posted after it", so prose and code now say the same thing.
- **`[minor]` synthetic `time.time()` stamp on a row `last_verdict_at` reads** — *fixed*, and further than I suggested: with no readable stamp the observer skips the activity row as well as the ledger write (`loop.py:4148-4153`), because a row behind a flag that can never be set would be re-appended every poll. The log line still fires each poll, so nothing is hidden. `note_readiness`'s docstring now states the row is a verdict row and `posted_at` must be the review's own stamp.
- **`[nit]` first-match vs. last-line** — *answered*, no change. Accepted and closed: first-match is the consumer's grammar today, and `TASK-660326447` ("take the LAST line-anchored match, on both seats") exists and is open to carry the move.

Nothing from round 1 remains open.

## Delta sweep (`58a0fbb..9e26050`)

- **Correctness:** the three fixes hold, and the guard order is right — the stamp check sits before the append so the no-stamp case posts nothing, and `approves[-1]` reads newest-last on a list `my_reviews` builds oldest-first. No regression to the 0.31.0 daemon-posted path, whose tests are still untouched. Two `[nit]`s below, neither gating.
- **Coding practices:** the new shape now matches the file's own precedent — `_note_checks_hold` (`:3639`) for emit-then-record, `:3962` and `:4052` for the `if appended:` variant. Docstring carries the reasoning, as the module does everywhere else.
- **Security:** nothing found. The trailer `reason` still reaches the log line and the activity comment only through `clean_readiness_reason` (one line, backticks stripped, ≤200 chars), so it cannot fence the comment; the delta did not change what reaches those sinks, only when.
- **Performance:** nothing found. The added list comprehension runs over one PR's reviews; the `readiness_observed` short-circuit still precedes the parse, and the dedupe still costs one indexed SELECT per qualifying poll.
- **Side-effects / blast radius:** the delta is strictly narrowing — under `--dry-run` the daemon now writes nothing where it previously wrote a flag, and the unreadable-stamp path no longer creates a `verdicts` row that `_admit_round`'s cooldown would read. The cost is that the dry-run WARNING repeats every poll; that is disclosed, matches `_note_checks_hold` in dry-run and `_warn_identity_drift`'s "a fresh dry-run always announces", and is the right trade. No schema, query, config, version or workflow change in this round — the PR's `verdicts.readiness` migration is unchanged from round 1.

## CI

Green at the reviewed head: 11 contexts from the check-runs API of `9e26050` — 10 `success` (unit, style, types, wheel, version-bump, entrypoint renderer, image contract, CodeQL analyze ×2 + CodeQL) and 1 `skipped`. The legacy status endpoint reports 0 contexts, so the rollup was judged on check-runs. The image-contract, wheel and version-bump items on the PR body's operator's-gate list are all green here, which retires them as questions for a human.

## Notes (non-gating)

Two `[nit]`s inline. Both are implementer's discretion; neither needs a round.

Merge readiness is `operator`, not because of anything wrong with the PR — it is an approve — but because the PR carries the `verdicts.readiness` schema migration, which is checklist item 1. A human clicks merge; nothing else changes.

Merge-Readiness: operator — schema/migrations: state.py adds nullable verdicts.readiness column, migrated in place via _ADDED_COLUMNS

